PR #117, opened 2019-01-29
Create REST API server service

CodeSandwich:
This merge adds rest-server module, which contains a server framework for REST API server. It's a wrapper around Actix-web allowing it to be run as a background service.


src/rest/server_service/error.rs
```diff
@@ -0,0 +1,9 @@
+use std::io::Error as IoError;
+
+#[derive(Debug)]
+pub enum Error {
+    BindFailed(IoError),
+    ServerAlreadyStopped,
+    ServerStopTimeout,
+    ServerStopFailed,
+}
```
NicolasDP: can you also implement the `std::error::Error` trait ?

PR #121, opened 2019-02-20
Rest service

NicolasDP:


.gitignore
```diff
@@ -1,2 +1,5 @@
 /target
 **/*.rs.bk
+
+# IntelliJ project directory
+.idea
```
mzabaluev: I prefer to add my IDE-specific project files to `.git/info/ignore` rather than the version controlled `.gitignore`.

src/main.rs
```diff
@@ -183,20 +201,31 @@ fn run() -> Result<(), Error> {
...
         tasks.task_create_with_inputs("block", move |r| {
-            block_task(blockchain, selection, clock, r, broadcast_sender)
+            block_task(
+                blockchain,
+                selection,
+                clock,
+                r,
+                broadcast_sender,
+                shared_stats,
+            )
```
mzabaluev: It may improve maintainablility to group some or all of these these parameters into a struct at this point.

src/rest/server_service/error.rs
```diff
@@ -0,0 +1,43 @@
...
+    Pkcs12Invalid(TlsError),
+    BindFailed(IoError),
+    ServerAlreadyStopped,
+    ServerStopTimeout,
+    ServerStopFailed,
+}
+
+impl Display for Error {
+    fn fmt(&self, f: &mut Formatter) -> FmtResult {
+        let error_name = match *self {
```
mzabaluev: It's more conventional to use human-readable error messages in `Display`. Variant names are better suited for the `Debug` impl, which is macro-derived.

src/rest/server_service/error.rs
```diff
@@ -0,0 +1,43 @@
...
+            Error::Pkcs12LoadFailed(_) => "Pkcs12LoadFailed",
+            Error::Pkcs12Invalid(_) => "Pkcs12Invalid",
+            Error::BindFailed(_) => "BindFailed",
+            Error::ServerAlreadyStopped => "ServerAlreadyStopped",
+            Error::ServerStopTimeout => "ServerStopTimeout ",
+            Error::ServerStopFailed => "ServerStopFailed ",
+        };
+        write!(f, "Server service error: {}", error_name)?;
```
mzabaluev: The choice of a prefix message should be left to a higher-level error handling code which knows it has this error type to represent. Just use a human-readable string describing the problem.

src/rest/server_service/error.rs
```diff
@@ -0,0 +1,43 @@
...
+        if let Some(cause) = self.source() {
+            write!(f, "caused by {}", cause)?
```
mzabaluev: Don't do this here either, it's up to a top level error handler to walk the error chain if needed.

.gitignore
CodeSandwich: Will fix in next MR

src/main.rs
CodeSandwich: I agree, 5 parameters are hard to follow. Should I refactor it now or wait with it until the demo? We will have to do quite some clean up then and developing fresh code is slightly easier with unpacked args.

src/rest/server_service/error.rs
CodeSandwich: Display must be implemented for Error, it's a direct replacement for `Error::description()`. I don't think it must be actually displayed and understandable by end-user, at least not at this point.

src/rest/server_service/error.rs
CodeSandwich: Will fix in next PR

src/main.rs
mzabaluev: Whichever works for you. I'd write even new code with easy to read structures that can be autocompleted.

src/rest/server_service/error.rs
mzabaluev: Yes, but it should be human readable, no need to stuff things like enum variant names there.

CodeSandwich: You're right, that's the point of a description. I'll fix it in next PR.

src/rest/server_service/error.rs
CodeSandwich: Will fix it in next PR

src/rest/server_service/error.rs
mzabaluev: This PR is not getting anywhere without being rebased to resolve the conflicts, so you might as well fix it here.

CodeSandwich: This PR was already rebased and merged as a part of a different branch with additional functionalities. I would close or reuse this branch and this PR if I knew you were going to review it.

mzabaluev: OK, feel free to close this PR then.

PR #123, opened 2019-02-20
setup a development environment via shell.nix

sjmackenzie:


edolstra:
Looks good to me. A small tweak would be to pin the Nixpkgs version:
```
nixpkgs ? fetchTarball https://github.com/NixOS/nixpkgs/archive/7fff567ee99c1f343ecdd82fef2e35fb6f50e423.tar.gz
 ```
I've done that in `shell.nix` in rust-cardano (not pushed yet) because we quite often depend on a very recent rustc version that might not be in the channel yet.

**Comment by sjmackenzie** [comment]
Eelco, note I'm preferring `rustup` and not `rustc`, this gives the user easy access to bleeding edge nightlies. I'll pin at 7fff567ee99c1f343ecdd82fef2e35fb6f50e423 nonetheless.

**Comment by sjmackenzie** [comment]
@edolstra `rustup` doesn't compile on your pinned version. It's better to keep `nixpkgs` on `channel:nixos-unstable` and use `rustup` to get the version of `rustc` you want. Then do things like:
```
$ rustup install nightly
$ rustup run nightly cargo build --examples
$ rustup run nightly cargo run --example generate_genesis_data boot 3 3
```
To this effect, you can swap out to whichever compiler version you want, or even target different platforms if you so wish.

PR #124, opened 2019-02-21
Update setting command

NicolasDP:
This is a little rework of the settings command


src/settings/command_arguments.rs
```diff
@@ -63,13 +55,66 @@ pub struct CommandArguments {
...
+
+#[derive(StructOpt, Debug)]
+pub enum Command {
+    /// start jormungandr service and start participating to the network
+    #[structopt(name = "start")]
+    Start(StartArguments),
+
+    /// initialize a configuration
+    #[structopt(name = "init")]
+    Init(InitArguments),
```
mzabaluev: Maybe use `InitConfig` for clarity?

src/settings/command_arguments.rs
```diff
@@ -1,20 +1,12 @@
...
-)]
-pub struct CommandArguments {
-    /// activate the verbosity, the more occurrences the more verbose.
-    /// (-v, -vv, -vvv)
-    #[structopt(short = "v", long = "verbose", parse(from_occurrences))]
-    pub verbose: u8,
+use crate::settings::logging::LogFormat;
 
+#[derive(StructOpt, Debug)]
+pub struct StartArguments {
```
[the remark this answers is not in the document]
NicolasDP: The purpose will come in the following PR, but mainly it is to initialize a new jormungandr blockchain. I'll elaborate when I have a better idea of what is actually needed.

src/settings/command_arguments.rs
NicolasDP: I'd rather like to remain consistent with the other option. Especially that we are still in the command line arguments module here.

PR #131, opened 2019-02-25
Update obft input

### LINKED ISSUE #128: Move bft configution to the genesis.yaml
We should move `bft` configuration to the genesis, as this that configuration is related to the blockchain itself, not to the node itself. Additionaly we should remove constants.t and use `slots.secs` instead.

NicolasDP:
This update the OBFT configuration setting. Instead of coming from the config.yaml of the Node it comes with the Genesis file. This is because it is expected to be shared securely between node and a genesis file with the wrong initial leader may lead to an undesired state of the blockchain.

fix #128

### DIFF
--- a/src/blockcfg/genesis_data.rs
+++ b/src/blockcfg/genesis_data.rs
@@ -1,6 +1,7 @@
 //! Generic Genesis data
+use cardano::util::hex;
 use chain_addr::AddressReadable;
-use chain_impl_mockchain::transaction::Value;
+use chain_impl_mockchain::{key, transaction::Value};
 
 use serde;
 use serde_yaml;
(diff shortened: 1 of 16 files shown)


src/blockcfg/genesis_data.rs
```diff
@@ -32,12 +37,38 @@ impl fmt::Display for ParseError {
     }
 }
 
+impl std::str::FromStr for PublicKey {
+    type Err = String;
```
mzabaluev: It should be considered technical debt to return non-`Error` values as `Err`.

src/network/grpc/bootstrap.rs
```diff
@@ -13,8 +13,10 @@ use std::fmt::Debug;
 pub fn bootstrap_from_target<P, B>(peer: P, blockchain: BlockchainR<B>)
 where
     B: BlockConfig,
-    P: tokio_connect::Connect<Error = io::Error> + 'static,
-    P::Connected: Send,
+    P: tower_service::Service<(), Error = std::io::Error> + 'static,
+    <P as tower_service::Service<()>>::Response:
+        tokio::io::AsyncWrite + tokio::io::AsyncRead + 'static + Send,
+    // P::Connected: Send,
```
mzabaluev: Just clean this up, the line above is the replacement bound now.

src/blockcfg/genesis_data.rs
NicolasDP: agree

PR #133, opened 2019-02-26
fix circle-ci

NicolasDP:


NicolasDP:
@mzabaluev could you have a look on the error here ? looks related to the network.

.circleci/config.yml
```diff
@@ -35,6 +36,7 @@ jobs:
       - run:
           name: Print version information
           command: rustc --version; cargo --version
+      - run: git submodule update --init --remote
```
mzabaluev: It's better to put this right after the checkout step.

.circleci/config.yml
```diff
@@ -6,6 +6,7 @@ jobs:
     working_directory: /mnt/crate
     steps:
       - checkout
+      - run: git submodule update --init --remote
```
mzabaluev: See also the [recommended way to do it](https://discuss.circleci.com/t/can-not-get-submodules-to-clone-properly/22120) accordingly to CircleCI support staff.

mzabaluev:
@NicolasDP a local checkout of jormungandr master builds for me if submodules are checked out accordingly to the [documentation](https://circleci.com/docs/2.0/configuration-reference/#checkout).

.circleci/config.yml
mzabaluev: Or a slightly different [recipe](https://circleci.com/docs/2.0/configuration-reference/#checkout) in the reference doc.

mzabaluev:
I think the `--remote` option breaks it, as it checks out the tip of the submodule's branch, which is often too advanced for jormungandr and we want to keep it pinned to a particular commit in rust-cardano.

mzabaluev:
Works now.

PR #136, opened 2019-02-26
Sync with rust-cardano master

edolstra:
The main change is that the leadership task now knows its leader ID, while Leadership is leader-agnostic.


src/leadership/process.rs
```diff
@@ -14,6 +15,8 @@ pub fn leadership_task<B>(
     B: BlockConfig,
     <B as BlockConfig>::TransactionId: Eq + std::hash::Hash,
     <B as BlockConfig>::Settings: Settings,
+    // FIXME: LeaderId should always require PartialEq.
```
mzabaluev: Please add a PR on rust-cardano with the necessary bound (also `Eq` and perhaps `Hash`) added.

PR #143, opened 2019-02-28
New testing framework

qnikst:


testing/thor/LICENSE-APACHE
```diff
@@ -0,0 +1,201 @@
+                              Apache License
```
mzabaluev: Hmm, this subfolder now claims to have an Apache/MIT license, but jormungandr as a whole has no declared license. We should give some care about this.

testing/thor/index.js
```diff
@@ -0,0 +1,155 @@
...
+}
+
+window.document.getElementById("sign_btn").onclick = function(self) {
+   console.log("Creating a builder.");
+   var txbuilder = mjolnir.TransactionBuilder.new();
+   console.log("Adding inputs:");
+   inputs.forEach(function (input, i) {
+      console.log("  input: ", input);
+      var tx = mjolnir.TransactionId.from_hex(input.in_txid);
+      var utxo = mjolnir.UtxoPointer.new(tx, input.in_idx, BigInt(input.out_value));
```
mzabaluev: This is not idiomatic JS, can we bind to the JS class constructor?

testing/thor/index.js
```diff
@@ -0,0 +1,155 @@
...
+        console.log(output);
+        var li = document.createElement("li");
+        var a = document.createElement("a")
+        var t = document.createTextNode("" + output.value + " " + output.address);
+        total += parseInt(output.value);
+        li.appendChild(a);
+        a.appendChild(t);
+        a.setAttribute("class","clickable");
+        a.onclick = function() { alert('hi!') };
+        output_block.appendChild(li);
```
mzabaluev: A less tedious, more maintainable way to go about it is to have a markup template for a list item somewhere in the HTML document, e.g. in an invisible `<div/>`. When a new item is created, clone it into a document fragment and populate the fragment with values, then append the fragment as a child to the list.

testing/thor/LICENSE-APACHE
```diff
@@ -0,0 +1,201 @@
...
+   To apply the Apache License to your work, attach the following
+   boilerplate notice, with the fields enclosed by brackets "[]"
+   replaced with your own identifying information. (Don't include
+   the brackets!)  The text should be enclosed in the appropriate
+   comment syntax for the file format. We also recommend that a
+   file or class name and description of purpose be included on the
+   same "printed page" as the copyright notice for easier
+   identification within third-party archives.
+
+Copyright [yyyy] [name of copyright owner]
```
NicolasDP:
```suggestion
Copyright 2019 Input Output HK Ltd
```

testing/thor/index.js
qnikst: not sure if I can do that easily, the value is not a normal JS clas.

qnikst:
I've updated PR.

testing/thor/index.js
qnikst: Here were are clearing multiple `<li>` so I expect that amount of code and complexities would be pretty much similar, but thanks, for the idea, I'll use that if representation will be more complex.

Issue #149, opened 2019-03-01
Implement the network topology state module:


NicolasDP:
* utilise poldercast crate;
* add simple configuration for the BFT nodes in this module;

PR #154, opened 2019-03-01
update the settings to reflect the peer 2 peer settings

NicolasDP:
the network setting now looks like:

```yml
storage: "/tmp/storage"
logger:
  verbosity: 7
  format: json
peer_2_peer:
  trusted_peers: []
  public_access: "/ip4/127.0.0.1/tcp/8080"
  topics_of_interests:
    transactions: low
    blocks: normal
```

The new settings reflects the desired static properties of the Node for the network.

relates to #149

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,8 +107,6 @@ Configuration fields meaning:
 Example of node config:
 
 ```
-grpc_listen:
-       - "127.0.0.1:8081"
 storage: "/tmp/storage"
 logger:
   verbosity: 1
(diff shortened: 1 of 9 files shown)


README.md
```diff
@@ -117,14 +115,18 @@ rest:
   listen: "127.0.0.1:8443"
   pkcs12: "example.p12"
   prefix: "api"
+peer_2_peer:
+  trusted_peers: []
+  public_access: "/ip4/127.0.0.1/tcp/8080"
```
mzabaluev: Would it be better to use a property map with sensible defaults for some of the fields (IPv4/TCP) instead of this microformat?

PR #156, opened 2019-03-01
Introduce get_transactions call in the TransactionsService.

qnikst:


qnikst:
I'll update rustfmt and add some additional logic that is needed for the PR

qnikst:
Included in #158

PR #157, opened 2019-03-01
Post transactions support

qnikst:


src/rest/v0/transaction/mod.rs
```diff
@@ -0,0 +1,40 @@
...
+use actix_web::{App, Error as ActixError, HttpMessage, HttpRequest, Responder};
+use bytes::IntoBuf;
+use chain_core::property::Deserialize;
+use chain_impl_mockchain::transaction::SignedTransaction;
+use futures::Future;
+use std::sync::{Arc, Mutex};
+
+pub type Task = Arc<Mutex<TaskMessageBox<TransactionMsg<Mockchain>>>>;
+
+pub fn crate_handler(
```
mzabaluev: Why is this named `crate_handler`?

CodeSandwich: Because it creates item, which then can be passed to `add_handler`

mzabaluev: But what is a "crate", in this context?

CodeSandwich: :D I'll fix it

PR #158, opened 2019-03-01
Introduce get_transaction functionality 

### LINKED ISSUE #146: Add GRPC getters:
* [x] TipRequest
* [x] getBlocks
* [ ] getTransactions

They need to be as per the spec in jormungandr's network spec

qnikst:
Closes #146


src/transaction/process.rs
```diff
@@ -2,6 +2,7 @@ use std::sync::{mpsc::Receiver, Arc, RwLock};
 
 use crate::blockcfg::{BlockConfig, Ledger, Transaction};
 use crate::blockchain::BlockchainR;
+use crate::client::do_stream_reply;
```
mzabaluev: I think `do_stream_reply` should be moved to `intercom`, now that many tasks use this helper function.

PR #166, opened 2019-03-05
Persistent block storage

edolstra:
This makes jormungandr store blocks persistently in `<storage>/blocks.sqlite` (where `<storage>` is specified in the configuration file), ensuring that the chain is maintained across restarts.


src/blockchain/chain.rs
```diff
@@ -69,11 +106,31 @@ impl State<Mockchain> {
...
+
+        let storage = SQLiteBlockStore::new(sqlite.to_str().unwrap());
+
+        if let Some(tip_hash) = storage.get_tag(LOCAL_BLOCKCHAIN_TIP_TAG).unwrap() {
+            debug!("restoring state at tip {}", tip_hash);
+
+            // FIXME: should restore from serialized chain state once we have it.
+            for info in storage
+                .iterate_range(&<Mockchain as BlockConfig>::BlockHash::zero(), &tip_hash)
+                .unwrap()
```
mzabaluev: We might want to start introducing proper error handling instead of an `unwrap` bonanza everywhere. Will this go away as the FIXME comment above suggests, or should the initialization code change to returning `Result`?

src/client.rs
```diff
@@ -71,7 +72,7 @@ where
     let mut checkpoints = checkpoints
         .iter()
         .filter_map(
-            |checkpoint| match blockchain.storage.get_block(&checkpoint) {
```
mzabaluev: If the lock's `read()` failure here always means something fatal (another thread has panicked while holding the lock?), this looks OK. Otherwise, I'd rather invest some time in converting these errors into domain errors.

src/client.rs
```diff
@@ -71,7 +72,7 @@ where
...
+            |checkpoint| match blockchain.storage.read().unwrap().get_block(&checkpoint) {
```
mzabaluev: If the lock's `read()` failure here always means something fatal (another thread has panicked while holding the lock?), this looks OK. Otherwise, I'd rather invest some time in converting these errors into domain errors.

src/blockchain/chain.rs
edolstra: Yeah, we should introduce proper error handling, but that probably should be a separate PR. Also as you say this code will probably end up getting rewritten/dropped anyway when we have proper chain state restoration.

doc/jormungandr_keys.md
```diff
@@ -20,6 +20,6 @@ ed25519_secret1sx3vjuc733mmlfm86xc2ufrjx6hltas57yj6pklra3s2n03yfv2svckr67
 and to extract the associated public key:
 
 ```
-$ jormungandr generate-pub-key ed25519_secret1sx3vjuc733mmlfm86xc2ufrjx6hltas57yj6pklra3s2n03yfv2svckr67
+$ echo ed25519_secret1sx3vjuc733mmlfm86xc2ufrjx6hltas57yj6pklra3s2n03yfv2svckr67 | jormungandr generate-pub-key
```
NicolasDP: thanks

NicolasDP:
```
error: Server does not allow request for unadvertised object cd47f1f4afa7c5c5a05e96b732dcb64e1a6e876d
Fetched in submodule path 'cardano-deps', but it did not contain cd47f1f4afa7c5c5a05e96b732dcb64e1a6e876d. Direct fetching of that commit failed.
```

PR #168, opened 2019-03-06
REST API GET tip, GET block and GET block/next_id

CodeSandwich:
This PR finished REST API V0 for Mjolnir demo. It allows tracking block tip and getting appended blocks data.

### DIFF
--- a/src/rest/mod.rs
+++ b/src/rest/mod.rs
@@ -25,8 +25,9 @@ pub fn start_rest_server(config: &Rest, context: Context) -> Result<ServerServic
         .unwrap_or("/");
     ServerService::builder(config.pkcs12.clone(), config.listen.clone(), prefix)
         .add_handler(v0::node::stats::create_handler(context.stats_counter))
-        .add_handler(v0::utxo::create_handler(context.blockchain))
+        .add_handler(v0::utxo::create_handler(context.blockchain.clone()))
         .add_handler(v0::transaction::create_handler(context.transaction_task))
+        .add_handler(v0::tip::create_handler(context.blockchain))
         .build()
         .map_err(|e| SettingsError::Start(ConfigError::InvalidRest(e)))
 }
--- a/src/rest/mod.rs
+++ b/src/rest/mod.rs
@@ -32,6 +32,9 @@ pub fn start_rest_server(config: &Rest, context: Context) -> Result<ServerServic
         .clone();
     ServerService::builder(config.pkcs12.clone(), config.listen.clone(), prefix)
         .add_handler(v0::block::create_handler(context.blockchain.clone()))
+        .add_handler(v0::block::next_id::create_handler(
+            context.blockchain.clone(),
+        ))
         .add_handler(v0::node::stats::create_handler(context.stats_counter))
         .add_handler(v0::tip::create_handler(settings))
         .add_handler(v0::transaction::create_handler(context.transaction_task))
--- a/src/rest/mod.rs
+++ b/src/rest/mod.rs
@@ -22,7 +22,7 @@ pub fn start_rest_server(config: &Rest, context: Context) -> Result<ServerServic
         .prefix
         .as_ref()
         .map(|prefix| prefix.as_str())
-        .unwrap_or("/");
+        .unwrap_or("");
     let settings = context.blockchain.read().unwrap().state.settings.clone();
     ServerService::builder(config.pkcs12.clone(), config.listen.clone(), prefix)
         .add_handler(v0::block::create_handler(context.blockchain.clone()))
--- a/src/rest/mod.rs
+++ b/src/rest/mod.rs
@@ -23,11 +23,19 @@ pub fn start_rest_server(config: &Rest, context: Context) -> Result<ServerServic
         .as_ref()
         .map(|prefix| prefix.as_str())
         .unwrap_or("/");
+    let settings = context
+        .blockchain
+        .read()
+        .unwrap_or_else(|e| e.into_inner())
+        .state
+        .settings
+        .clone();
     ServerService::builder(config.pkcs12.clone(), config.listen.clone(), prefix)
+        .add_handler(v0::block::create_handler(context.blockchain.clone()))
         .add_handler(v0::node::stats::create_handler(context.stats_counter))
-        .add_handler(v0::utxo::create_handler(context.blockchain.clone()))
+        .add_handler(v0::tip::create_handler(settings))
         .add_handler(v0::transaction::create_handler(context.transaction_task))
-        .add_handler(v0::tip::create_handler(context.blockchain))
+        .add_handler(v0::utxo::create_handler(context.blockchain))
         .build()
         .map_err(|e| SettingsError::Start(ConfigError::InvalidRest(e)))
 }
--- a/src/rest/mod.rs
+++ b/src/rest/mod.rs
@@ -23,13 +23,7 @@ pub fn start_rest_server(config: &Rest, context: Context) -> Result<ServerServic
         .as_ref()
         .map(|prefix| prefix.as_str())
         .unwrap_or("/");
-    let settings = context
-        .blockchain
-        .read()
-        .unwrap_or_else(|e| e.into_inner())
-        .state
-        .settings
-        .clone();
+    let settings = context.blockchain.read().unwrap().state.settings.clone();
     ServerService::builder(config.pkcs12.clone(), config.listen.clone(), prefix)
         .add_handler(v0::block::create_handler(context.blockchain.clone()))
         .add_handler(v0::block::next_id::create_handler(
--- a/src/rest/server_service/path_predicate.rs
+++ b/src/rest/server_service/path_predicate.rs
@@ -7,20 +7,73 @@ pub struct PathPredicate {
 }
 
 impl PathPredicate {
-    pub fn new(prefix: &str, pattern: &str) -> Self {
-        let full_pattern = format!("^/{}/{}$", regex::escape(prefix), pattern);
-        let regex = Regex::new(&full_pattern).unwrap_or_else(|e| {
-            panic!(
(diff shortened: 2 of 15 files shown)


src/rest/mod.rs
```diff
@@ -23,10 +23,22 @@ pub fn start_rest_server(config: &Rest, context: Context) -> Result<ServerServic
         .as_ref()
         .map(|prefix| prefix.as_str())
         .unwrap_or("/");
+    let settings = context
+        .blockchain
+        .read()
+        .unwrap_or_else(|e| e.into_inner())
```
mzabaluev: What does this `unwrap_or_else` do?

src/rest/v0/block/mod.rs
```diff
@@ -0,0 +1,41 @@
...
+use chain_storage::store::BlockStore;
+
+pub fn create_handler(
+    blockchain: BlockchainR<Mockchain>,
+) -> impl Fn(&str) -> App<BlockchainR<Mockchain>> + Send + Sync + Clone + 'static {
+    move |prefix: &str| {
+        let full_prefix = format!("{}/v0/block", prefix);
+
+        App::with_state(blockchain.clone())
+            .filter(PathPredicate::new(&full_prefix, "[^/]+"))
```
mzabaluev: It's better to use [raw string literals][raw-string-literal] for regexp literals.

[raw-string-literal]: https://doc.rust-lang.org/reference/tokens.html#raw-string-literals

src/rest/v0/block/next_id/mod.rs
```diff
@@ -0,0 +1,62 @@
...
+use chain_core::property::Settings;
+use chain_storage::store::BlockStore;
+
+pub fn create_handler(
+    blockchain: BlockchainR<Mockchain>,
+) -> impl Fn(&str) -> App<BlockchainR<Mockchain>> + Send + Sync + Clone + 'static {
+    move |prefix: &str| {
+        let full_prefix = format!("{}/v0/block", prefix);
+        App::with_state(blockchain.clone())
+            .filter(PathPredicate::new(&full_prefix, "[^/]+/next_id"))
```
mzabaluev: Same here, you might slip in a backslash and it would get interpreted as a string literal escape.

src/rest/server_service/path_predicate.rs
```diff
@@ -0,0 +1,26 @@
...
+use actix_web::Request;
+use regex::{self, Regex};
+
+pub struct PathPredicate {
+    regex: Regex,
+}
+
+impl PathPredicate {
+    pub fn new(prefix: &str, pattern: &str) -> Self {
+        let full_pattern = format!("^/{}/{}$", regex::escape(prefix), pattern);
```
mzabaluev: I would rather demand the leading slash to be part of the prefix pattern, for usability reasons.

src/rest/server_service/path_predicate.rs
```diff
@@ -0,0 +1,26 @@
+use actix_web::pred::Predicate;
+use actix_web::Request;
+use regex::{self, Regex};
+
+pub struct PathPredicate {
+    regex: Regex,
+}
+
+impl PathPredicate {
+    pub fn new(prefix: &str, pattern: &str) -> Self {
```
mzabaluev: It seems that all uses of this predicate try to parse three components: the path prefix, the path segment without slashes in the middle, and the possibly empty expected path selector at the end. If it's the only thing this predicate is expected to do, I would just have the prefix and the end selector as parameters and not require the type user to supply regex syntax.

src/rest/server_service/path_predicate.rs
```diff
@@ -0,0 +1,29 @@
...
+impl PathPredicate {
+    pub fn new(pattern: &str) -> Self {
+        let pattern = regex::escape(&pattern);
+        let segment_regex = Regex::new(r#"\\\{[^}]*\}"#).unwrap();
+        let pattern = segment_regex.replace_all(&pattern, r#"[^/]*"#);
+        let prefix = match pattern.starts_with('/') {
+            true => "",
+            false => "/",
+        };
+        let pattern = format!("^{}{}$", prefix, pattern);
```
mzabaluev: I don't understand this. `PathPredicate::new` needs documentation to tell what the pattern is supposed to be, and then the code feels like it could do without much of this replacement stuff.

mzabaluev: Perhaps adding a few unit tests for `PathPredicate` could help clarify it, and also test that it does what it's supposed to?

CodeSandwich: Sure, docs will help, tests as well. I'm under impression that we're not supposed to write good code yet, just release proof of concept as soon as possible.

The replacement stuff is introduced to be compatible with Actix route definition. The idea is to keep path definition DRY and not force to write it multiple times with slight formatting changes, which is prone to mistakes.

mzabaluev: So please put some minimum viable documentation on what syntax the pattern should have, or an `# #Examples` section, or a reference to an Actix API providing the same, even just for the purposes of this review. I refuse to analyse this code without an idea of what it's supposed to do :)

src/rest/server_service/path_predicate.rs
```diff
@@ -0,0 +1,79 @@
...
+
+pub struct PathPredicate {
+    regex: Regex,
+}
+
+impl PathPredicate {
+    // The pattern should be a full URI, with or without slash on the beginning.
+    // Wildcards in form of `{some_name}` are allowed, they accept everything except `/`.
+    // The predicate will accept only requests with URIs fully matching the pattern.
+    pub fn for_pattern(pattern: &str) -> Self {
```
mzabaluev: Seems legit now. My last question about this is: Doesn't Actix have built-in path matchers for this?

src/rest/server_service/path_predicate.rs
```diff
@@ -0,0 +1,79 @@
...
+        let pattern = regex::escape(&pattern);
+        let segment_regex = Regex::new(r#"\\\{[^}]*\}"#).unwrap();
```
mzabaluev: What should the escaped backslash in the beginning match? The docs do not mention any backslashes in the pattern.

The raw literal could have less line noise. As the pattern does not have any inner double quotes, the `#` guards can be dropped.

src/rest/server_service/path_predicate.rs
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    fn is_uri_accepted(pattern: &str, uri: &str) -> bool {
+        let predicate = PathPredicate::for_pattern(pattern);
+        let request = TestRequest::with_uri(uri).finish();
+        predicate.check(&request, &())
+    }
+
+    #[test]
+    fn path_predicate_accepts_only_matching_uris() {
```
mzabaluev: :raised_hands:

src/rest/server_service/path_predicate.rs
CodeSandwich: Yes and no. It has excellent path matcher for routes inside `App`. Unfortunately we need to have multiple `App`s and we must decide which one of them should receive the request and it's an irreversible routing decision. We must have multiple `App`s, because each of them captures different context and we also want to keep them composable and `App`s are good building blocks.

src/rest/server_service/path_predicate.rs
CodeSandwich: `{` and `}` are backslash-escaped by `regex::escape

This IS raw literal. I think you've asked me to put all regex patterns into raw literals and I agree that it's a safe habit.

src/rest/server_service/path_predicate.rs
CodeSandwich: I hope it's a good thing :thinking:

src/rest/server_service/path_predicate.rs
mzabaluev: Yeah, but it's legal and less noisy to have a `r"foo"` literal if there are no inner double quotes.

CodeSandwich: That's great, I've been always using at least one `#`!

src/rest/server_service/path_predicate.rs
mzabaluev: Maybe each of the apps could have a simple predicate that matches just the prefix as a static string and we can do away with regex completely?

src/rest/server_service/path_predicate.rs
```diff
@@ -0,0 +1,79 @@
...
+}
+
+impl PathPredicate {
+    // The pattern should be a full URI, with or without slash on the beginning.
+    // Wildcards in form of `{some_name}` are allowed, they accept everything except `/`.
+    // The predicate will accept only requests with URIs fully matching the pattern.
+    pub fn for_pattern(pattern: &str) -> Self {
+        let pattern = regex::escape(&pattern);
+        let segment_regex = Regex::new(r"\\\{[^}]*\}").unwrap();
+        let pattern = segment_regex.replace_all(&pattern, r"[^/]*");
```
mzabaluev: Ah, I get it now. So all this complexity and regex-tweaking-regex is just to have some `{foo}` constructs in your pattern that you then replace with simple matches? Why have them and the pattern syntax in the first place?

src/rest/server_service/path_predicate.rs
CodeSandwich: It worked that way before introduction of `block/{block_id}` and `block/{block_id}/next_id`.

src/rest/server_service/path_predicate.rs
CodeSandwich: I think this question was answered in other thread: https://github.com/input-output-hk/jormungandr/pull/168#discussion_r263232925

mzabaluev:
Please rebase this and make sure it works, I have build failures.

mzabaluev:
Approved with actions: #172

Please fix the post-rebase failure before merging.

PR #170, opened 2019-03-06
Introduce gossip service.

qnikst:
This is a  draft PR as code references cardano-deps that are not in master yet.

### DIFF
--- a/src/network/p2p_topology.rs
+++ b/src/network/p2p_topology.rs
@@ -1,20 +1,127 @@
 //! module defining the p2p topology management objects
 //!
 
+use bincode;
+use chain_core::property;
+use network_core::gossip;
 use poldercast::topology::{Cyclon, Module, Rings, Topology, Vicinity};
 use poldercast::Subscription;
 pub use poldercast::{Address, Id, InterestLevel, Node};
-use std::collections::BTreeMap;
-
-use std::sync::{Arc, RwLock};
+use std::{
+    collections::BTreeMap,
+    error, fmt, io,
+    sync::{Arc, RwLock},
+    vec,
+};
 
 pub const NEW_TRANSACTIONS_TOPIC: u32 = 0u32;
 pub const NEW_BLOCKS_TOPIC: u32 = 1u32;
+
+#[derive(Debug, Clone)]
+pub struct Gossip(pub Vec<Node>);
+
+#[derive(Debug)]
+pub enum Error {
+    Encoding(Box<bincode::ErrorKind>),
+    IO(io::Error),
+}
+
+/*
+impl Error {
+    pub fn new<E>(source: E) -> Self
+    where
+        E: Into<Box<dyn error::Error + Send + Sync>>,
+    {
+        Error {
+            source: source.into(),
+        }
+    }
+}
+*/
+
+impl error::Error for Error {
+    fn source(&self) -> Option<&(dyn error::Error + 'static)> {
+        match self {
+            Error::Encoding(source) => Some(source),
+            Error::IO(source) => Some(source),
+        }
+    }
+}
+
+impl fmt::Display for Error {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match self {
+            Error::Encoding(_) => write!(f, "serialization error"),
+            Error::IO(_) => write!(f, "io error"),
+        }
+    }
+}
+
+impl From<io::Error> for Error {
+    fn from(e: io::Error) -> Self {
+        Error::IO(e)
+    }
+}
+
+impl From<Box<bincode::ErrorKind>> for Error {
+    fn from(e: Box<bincode::ErrorKind>) -> Self {
+        Error::Encoding(e)
+    }
+}
+
+pub fn from_node_id(node_id: &gossip::NodeId) -> Id {
+    let bytes = node_id.to_bytes();
+    let mut buf = [0; 16];
+    buf[0..16].clone_from_slice(&bytes);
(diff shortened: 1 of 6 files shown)


src/network/p2p_topology.rs
```diff
@@ -1,22 +1,74 @@
...
+    //}
+
+    fn from_nodes(iter: Box<Iterator<Item = Self::Node>>) -> Self {
+        Gossip(iter.collect())
+    }
+}
+
+impl IntoIterator for Gossip {
+    type Item = Node;
+    type IntoIter = ::std::vec::IntoIter<Node>;
```
mzabaluev: I usually import `std::vec` at the top and then use `vec::` for disambiguation.

src/network/p2p_topology.rs
```diff
@@ -1,22 +1,74 @@
...
+}
+
 /// object holding the P2pTopology of the Node
 #[derive(Clone)]
 pub struct P2pTopology {
     topology: Arc<RwLock<Topology>>,
 }
 
+impl property::Serialize for Gossip {
+    type Error = std::boxed::Box<bincode::ErrorKind>;
```
mzabaluev: Let's create a nice static error type wrapping the bincode stuff, and have it implement `std::error::Error`.

src/network/service.rs
```diff
@@ -5,10 +5,12 @@ use crate::intercom::{
     ReplyStream, SubscriptionFuture, SubscriptionStream, TransactionMsg,
 };
 use crate::utils::task::TaskMessageBox;
+use std::collections::BTreeMap;
 use std::marker::PhantomData;
 
+use network::p2p_topology::{Gossip, Id, P2pTopology};
 use network_core::{
-    gossip::NodeId,
+    gossip::{self, Gossip as G},
```
mzabaluev: Could better do `Gossip as GossipTrait`.
Or do `<Gossip as gossip::Gossip>` where trait disambiguation is needed.
Or name our concrete gossip type `PoldercastGossip` to avoid ambiguity.

src/network/service.rs
```diff
@@ -206,16 +207,47 @@ impl<B: BlockConfig> TransactionService for ConnectionTransactionService<B> {
...
     /// Record and process gossip event.
-    fn record_gossip(&mut self, _node_id: NodeId, _gossip: &Self::Message) -> Self::MessageFuture {
-        unimplemented!()
+    fn record_gossip(
+        &mut self,
+        node_id: gossip::Id,
+        gossip: &Self::Message,
+    ) -> Self::MessageFuture {
+        let nodes: BTreeMap<_, _> = gossip
+            .clone()
```
mzabaluev: No way to avoid cloning, even though it's passed by reference?

src/network/service.rs
```diff
@@ -206,16 +207,47 @@ impl<B: BlockConfig> TransactionService for ConnectionTransactionService<B> {
...
+            .0
+            .into_iter()
+            .map(|node| (*node.id(), node))
+            .collect();
+        let node_id: Id = <Gossip as G>::from_node_id(&node_id);
```
mzabaluev: `Gossip::from_node_id` should work once you have imported the trait.

src/network/p2p_topology.rs
mzabaluev: This one is not resolved, would you like this to be the module's `Error` type?

qnikst: ha you are right, fixing!

qnikst:
Fixed!

src/network/p2p_topology.rs
```diff
@@ -1,22 +1,118 @@
...
+
+impl From<io::Error> for Error {
+    fn from(e:  io::Error) -> Self {
+        Error {
+            source: Box::new(e),
+        }
+    }
+}
+
+pub fn from_node_id(node_id: &gossip::NodeId) -> Id {
```
mzabaluev: Please remind me, why do we need these conversions rather than make `NodeId` an abstract associated type of `Gossip` and instantiating it here with `poldercast::Id`?

qnikst: the problem with keeping this abstract is that then we will need to abstract all the commands inside gossip message handler so we can work with that.

The other solution is to change rust-cardano Gossip trait and make NodeId generic there.
But I think that can be done in the separate commit if needed.

src/network/p2p_topology.rs
```diff
@@ -1,22 +1,118 @@
...
 #[derive(Clone)]
 pub struct P2pTopology {
     topology: Arc<RwLock<Topology>>,
 }
 
+impl property::Serialize for Gossip {
+    type Error = Error;
+
+    fn serialize<W: std::io::Write>(&self, writer: W) -> Result<(), Self::Error> {
+        bincode::serialize_into(writer, &self.0).map_err(|err| Error::new(err))
```
mzabaluev: Note that `Error` could just encapsulate the bincode error with no need for the box.

src/network/p2p_topology.rs
```diff
@@ -1,22 +1,129 @@
...
 
 pub const NEW_TRANSACTIONS_TOPIC: u32 = 0u32;
 pub const NEW_BLOCKS_TOPIC: u32 = 1u32;
 
+#[derive(Debug, Clone)]
+pub struct Gossip(pub Vec<Node>);
+
+#[derive(Debug)]
+pub enum Error {
+    BincodeError(Box<bincode::ErrorKind>),
```
mzabaluev: I'd name the variants `Encoding` and `Io`, as we do elsewhere.

PR #171, opened 2019-03-07
update cardano-deps to utilise the latest change in the mockchain

NicolasDP:


src/secure/mod.rs
```diff
@@ -30,15 +30,15 @@ impl NodeSecret {
         fs.read_to_end(&mut vec)?;
         let v = hex::decode(String::from_utf8(vec).unwrap().as_ref()).unwrap();
         // TODO propagate error properly
-        if v.len() != redeem::PRIVATEKEY_SIZE {
+        if v.len() != 32 {
```
mzabaluev: Make a private constant for this magic number here and below.

NicolasDP: will be fixed when we have @CodeSandwich 's missing serialisation/deserialisation tool

PR #173, opened 2019-03-11
Drop hex dependency, use FromStr impl of Hash

mzabaluev:


CodeSandwich:
This exact changes are pending in my next PR, we may merge them now and I will remove them from my branch or we can merge them later.

mzabaluev:
@CodeSandwich Cool, let's go with your branch then!

PR #174, opened 2019-03-11
Simplification of REST resource routing

### LINKED ISSUE #172: Simplify REST dispatch
The REST API dispatch is complicated. It's built around App-based composition, where each route is a separate App, which may not be necessary. What also may not be necessary is routing based on path patterns with wildcards if REST API is redesigned.

mzabaluev:
Get rid of `PathPredicate`, do what Actix-web wants its apps to do rather than work around it.
More concise naming and module structure.
Grouped and cleaned up module imports.

Fixes #172


src/rest/v0/block/mod.rs
```diff
@@ -16,10 +15,12 @@ pub fn create_handler(
     blockchain: BlockchainR<Mockchain>,
 ) -> impl Fn(&str) -> App<BlockchainR<Mockchain>> + Send + Sync + Clone + 'static {
     move |prefix: &str| {
-        let path = format!("{}/v0/block/{{block_id}}", prefix);
         App::with_state(blockchain.clone())
-            .filter(PathPredicate::for_pattern(&path))
-            .resource(&path, |r| r.get().with(handle_request))
+            .prefix(prefix.to_string())
+            .resource("/v0/block/{{block_id}}", |r| r.get().with(handle_request))
+            .resource("/v0/block/{{block_id}}/next_id", |r| {
```
CodeSandwich: The double brackets in `{{block_id}}` causes crash on runtime. They are needed for `format` macro, where single ones are substitutions and double ones are just brackets, but this is a literal.

Please run code before merging to master. It will get better when we have automated integration tests, but now we must try not to die.

src/rest/v0/block/mod.rs
```diff
@@ -16,10 +15,12 @@ pub fn create_handler(
     blockchain: BlockchainR<Mockchain>,
 ) -> impl Fn(&str) -> App<BlockchainR<Mockchain>> + Send + Sync + Clone + 'static {
     move |prefix: &str| {
-        let path = format!("{}/v0/block/{{block_id}}", prefix);
         App::with_state(blockchain.clone())
-            .filter(PathPredicate::for_pattern(&path))
-            .resource(&path, |r| r.get().with(handle_request))
+            .prefix(prefix.to_string())
```
CodeSandwich: The prefix should be full `format!("{}/v0/block", prefix)` like in other routes. Now this app consumes everything starting with configured prefix, for example `api/v0/node/stats`, which stopped working and now returns 404.

src/rest/v0/block/mod.rs
mzabaluev: Sorry. It was only a little change, what could possibly go wrong? ;)

PR #175, opened 2019-03-12
Pass origin to the grpc-client.

qnikst:
grpc client required the origin to be passed to the
connection so the proper connection may be established.
here we generate origin based on the connection type
passed.


src/network/grpc/client.rs
```diff
@@ -16,10 +16,12 @@ where
     B: NetworkBlockConfig,
 {
     info!("connecting to subscription peer {}", peer.connection);
+    info!("address: {}", peer.address());
     let peer = grpc_peer::TcpPeer::new(*peer.address());
     let addr = peer.addr().clone();
+    let origin: http::Uri = format!("http://{}", addr).parse().unwrap();
```
mzabaluev: Better do it with `uri::Builder`, we have all the parts, no need to format and parse.

src/network/grpc/mod.rs
```diff
@@ -19,6 +20,7 @@ where
     <B::Leader as property::LeaderSelection>::Update: Clone,
 {
     info!("connecting to bootstrap peer {}", peer.connection);
+    let origin : http::Uri = format!("http://{}",peer.address()).parse().unwrap();
```
mzabaluev: Same as above.

qnikst:
I've addressed all the comments PTAL.

src/network/grpc/mod.rs
```diff
@@ -19,6 +21,18 @@ where
...
+        peer.address().ip(),
+        peer.address().port()
+    )))
+    .unwrap();
+    let origin = http::uri::Builder::new()
+        .scheme("http")
+        .authority(authority)
+        .path_and_query("/")
+        .build()
+        .unwrap();
```
mzabaluev: Gah. It's almost like there should be a helper function for this.

PR #176, opened 2019-03-12
Update README: add instructions to build

NicolasDP:
add instructions to build jormungandr from sources.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,6 +56,11 @@ to utilise the most recent stable version of the rust compiler.
 2. make sure you have cloned the submodule too: `git submodule update`
 3. install: `cargo install`
 
+Note:
+
+* on windows, you'll need to add the `/userProfile/.cargo/bin` into the Path;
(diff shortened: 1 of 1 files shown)


dorin100:
@NicolasDP - I was able to do the first 2 steps but I cannot install 'cargo'

D:\iohk\otherProjects\jormungandr>cargo install
'cargo' is not recognized as an internal or external command,
operable program or batch file.

dorin100:
ok. so on windows it seems that I need to add the /userProfile/.cargo/bin into the Path and then the 'cargo install' will work if executed from the jormungandr root

NicolasDP:
thanks, I'll add this in the README

dorin100:
I installed Cargo and I also created the genessis.yaml file. Now I am at the Node Configuration part and it is not clear where this node config is stored... Where is this file located? Or should I create it - if yes, what kind of file/extension should it have?

PR #179, opened 2019-03-13
Gen key

CodeSandwich:
CLI tooling for public and private key generation

### DIFF
--- a/src/settings/command_arguments.rs
+++ b/src/settings/command_arguments.rs
@@ -152,6 +152,11 @@ pub enum Command {
     /// generate a random private key and print it to stdout encoded in bech32
     #[structopt(name = "generate-priv-key")]
     GeneratePrivKey(GeneratePrivKeyArguments),
+
+    /// generates a public key corresponding to a private key,
+    /// reads private from stdin and prints its public to stdout, both encoded in bech32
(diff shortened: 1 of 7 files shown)


src/settings/command_arguments.rs
```diff
@@ -96,6 +97,24 @@ pub struct InitArguments {
...
 }
 
+#[derive(StructOpt, Debug)]
+pub struct GeneratePrivKeyArguments {
+    /// Type of a private key
+    #[structopt(long = "type")]
+    pub key_type: GenPrivKeyType,
+}
+
+arg_enum! {
```
NicolasDP: neat. didn't see that last time <3

NicolasDP:
what about removing the old command then ? It is deprecated and not really usable.

also please, can you update the README to provide the necessary information regarding how to generate the keys ?

CodeSandwich:
@NicolasDP I'm not sure if we should already remove the old command and update README. The old key generator produces hex-encoded keys while the new one bech32-encoded. Genesis and xprv files all work on hex-encoded keys, so until we accept them in bech32 the new key generator is pretty much useless.

vincenthz:
it's ok to do another PR

PR #184, opened 2019-03-18
Update latest cardano-deps

NicolasDP:
this PR brings the new State model of the chain-impl (and all the new perks that came around).

1. block 0 from the genesis configuration;
2. state management;
3. Less traits;


src/blockchain/chain.rs
```diff
@@ -107,41 +94,28 @@ where
...
+
+                // TODO: update with the multiverse here
+                self.state = state;
             }
+            Err(error) => error!("Error with incoming block: {}", error),
         }
     }
+
+    /// return the current tip hash and date
+    pub fn get_tip(&self) -> B::BlockHash {
```
mzabaluev: How about gradually removing `get_` from method names so that the APIs are more conventional for Rust?

src/blockchain/chain.rs
```diff
@@ -1,36 +1,22 @@
-use std::collections::{BTreeMap, HashMap, HashSet};
+use std::collections::BTreeMap;
 use std::sync::{Arc, RwLock};
 
-use chain_core::property::{self, HasTransaction};
-use chain_impl_mockchain::{leadership, ledger, setting};
+use chain_core::property::{
+    Block as _, HasHeader as _, HasMessages as _, Settings as _, State as _,
```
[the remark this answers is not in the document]
NicolasDP: is it a problem if we start moving to 1.33 as a minimal stable?

src/blockchain/chain.rs
NicolasDP: agree. This PR is a bit massive and there are lot of component that are a bit hacky. I'd like to merge this ASAP and then we can start the surgery and do the proper cleanup of jormungandr.

PR #190, opened 2019-03-20
Tx cli

CodeSandwich:
Initial version of transaction generator in CLI. Currently does not connect with the node, so both txids and values of inputs must be manually provided.

@NicolasDP Because of future HTTP communication with node it's extension of sender and not a new debug cli tool.

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,6 +91,7 @@ jormungandr start --genesis-config genesis.yaml \
 # Extra tooling
 
 * [CLI sender](./doc/cli_sender.md) for HTTP management
+* [CLI transaction builder](./doc/cli_tx_builder.md) for HTTP management
 * [CLI address](./doc/cli_address.md) for debugging addresses
 
 # License
--- a/doc/cli_sender.md
+++ b/doc/cli_sender.md
@@ -20,8 +20,13 @@ Many CLI commands have common arguments:
 Fetches node stats
 
 ```
-sender v0 node stats get -h <node_addr>
+sender v0 node stats get <options>
 ```
+
+The options are
+
+- -h <node_addr> - see [conventions](#conventions)
+
 
 JSON printed on success
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -417,6 +417,7 @@ dependencies = [
 name = "chain-crypto"
 version = "0.1.0"
 dependencies = [
+ "bech32 0.6.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "cryptoxide 0.1.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "curve25519-dalek 1.1.3 (registry+https://github.com/rust-lang/crates.io-index)",
  "digest 0.8.0 (registry+https://github.com/rust-lang/crates.io-index)",
--- a/doc/cli_tx_builder.md
+++ b/doc/cli_tx_builder.md
@@ -1,0 +1,33 @@
+# Transaction builder
+
+Builds a signed transaction
+
+## Building
+
+```
+cargo build --bin tx_builder
+```
+
+## Usage
+
+```
+tx_builder <options>
+```
+
+The options are
+
+- -f <value> or --fee-base <value> - fee which will be always added to the transaction
+- -a <value> or --fee-per-addr <value> - fee which will be added to the transaction for every
+input and output
+- -i <input> or --input <input> - transaction input. Must have format
+`<hex-encoded-transaction-id>:<output-index>:<value>`. E.g. `1234567890abcdef:2:535`.
+At least 1 value required.
+- -o <output> or --output <output> -transaction output. Must have format
+`<hex-encoded-address>:<value>`. E.g. `abcdef1234567890:501`. At least 1 value required.
+- -c <address> or --change <address> - change address. Value taken from inputs and not spent on
+outputs or fees will be returned to this address. If not provided, the change will go to treasury.
+Must be hex-encoded.
+- -s <key> or --spending-key <key> - transaction spending keys. Must be hex-encoded.
+Required as many as provided inputs.
+
+Value outputted to stdout on success is binary blob with transaction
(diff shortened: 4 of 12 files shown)


doc/cli_sender.md
```diff
@@ -20,15 +20,46 @@ Many CLI commands have common arguments:
...
+
+- -h <node_addr> - see [conventions](#conventions)
+
+
 JSON printed on success
 
 ```
 {
-  "blockRecvCnt": 7,  // Blocks received by node
+  "blockRecvCnt": 7,  // Blocks received by node
```
NicolasDP:
```suggestion
  "blockRecvCnt": 7,  // Blocks received by node
```

CodeSandwich:
Tx_create is now a separate binary

README.md
```diff
@@ -91,6 +91,7 @@ jormungandr start --genesis-config genesis.yaml \
 # Extra tooling
 
 * [CLI sender](./doc/cli_sender.md) for HTTP management
+* [CLI TX creator](./doc/cli_tx_create.md) for HTTP management
```
NicolasDP:
```suggestion
* [CLI Transaction builder](./doc/cli_tx_create.md) for HTTP management
```

doc/cli_tx_create.md
```diff
@@ -0,0 +1,33 @@
+# TX creator
```
NicolasDP:
```suggestion
# Transaction Builder
```

doc/cli_tx_create.md
```diff
@@ -0,0 +1,33 @@
...
+
+Creates a signed transaction
+
+## Building
+
+```
+cargo build --bin tx_create
```
NicolasDP: `transaction` would be a more fitted name. We will use this to:

* build transaction;
* decode/debug transactions (display the content and all)

src/bin/tx_create.rs
```diff
@@ -0,0 +1,127 @@
...
+    /// E.g. `1234567890abcdef:2:535`. At least 1 value required.
+    #[structopt(name = "input", short, long, parse(try_from_str = "parse_input"))]
+    inputs: Vec<Input>,
+    /// transaction output. Must have format `<hex-encoded-address>:<value>`.
+    /// E.g. `abcdef1234567890:501`. At least 1 value required.
+    #[structopt(name = "output", short, long, parse(try_from_str = "parse_output"))]
+    outputs: Vec<(Address, Value)>,
+    /// return address. Value taken from inputs and not spent on outputs or fees will be returned
+    /// to this address. If not provided, the excess will go to treasury. Must be hex-encoded.
+    #[structopt(name = "return", short, long, parse(try_from_str = "parse_address"))]
```
NicolasDP: this option is more commonly known as `change`

src/bin/tx_create.rs
```diff
@@ -0,0 +1,127 @@
...
+#[structopt(rename_all = "kebab-case")]
+pub struct TxCreate {
+    /// transaction input. Must have format `<hex-encoded-transaction-id>:<output-index>:<value>`.
+    /// E.g. `1234567890abcdef:2:535`. At least 1 value required.
+    #[structopt(name = "input", short, long, parse(try_from_str = "parse_input"))]
+    inputs: Vec<Input>,
+    /// transaction output. Must have format `<hex-encoded-address>:<value>`.
+    /// E.g. `abcdef1234567890:501`. At least 1 value required.
+    #[structopt(name = "output", short, long, parse(try_from_str = "parse_output"))]
+    outputs: Vec<(Address, Value)>,
```
NicolasDP: don't use hexadecimal for the address, use bech32

src/bin/tx_create.rs
```diff
@@ -0,0 +1,127 @@
...
+    /// return address. Value taken from inputs and not spent on outputs or fees will be returned
+    /// to this address. If not provided, the excess will go to treasury. Must be hex-encoded.
```
NicolasDP: don't use hexadecimal for the address, use bech32

src/bin/tx_create.rs
```diff
@@ -0,0 +1,127 @@
...
+    /// fee base which will be always added to the transaction
+    #[structopt(short = "f")]
+    fee_base: Option<u64>,
+    /// fee which will be added to the transaction for every input and output
+    #[structopt(short = "a")]
+    fee_per_addr: Option<u64>,
+    /// fee which will be added to the transaction for certificate
+    #[structopt(short = "c")]
+    fee_per_cert: Option<u64>,
+    /// transaction spending keys. Must be hex-encoded. Required as many as provided inputs.
```
NicolasDP: use the standard bech32 encoding with the prefix

doc/cli_tx_create.md
CodeSandwich: Don't we want another binary for decoding transaction in UNIX spirit? Docoding will be a separate subcommand with no common arguments which will act like a completely independent app anyway.

src/bin/tx_create.rs
CodeSandwich: Hex in output is to keep consistency with hex in input which is to keep consistency with REST API. We probably should keep APIs consistent. I prefer bech32 as well, so maybe we should change REST API to return bech32s everywhere instead of hexes?

NicolasDP: yes we need to change that too. We need bech32 here as this is _user_ facing.

NicolasDP: and input are hex decimal if they are from a utxo pointer. Otherwise they need to be bech32 if it is an account.

doc/cli_tx_create.md
NicolasDP: that's more or less the spirit. But let be consistent too.

* we need one executable for the REST API glue (query/post);
* we need one for the address (already present) (create/debug);
* we need one for the transaction (this one) (create/sign/debug);
* we need one for the certificate (create/debug)

src/bin/tx_create.rs
CodeSandwich: OTOH the REST API is not supposed to be used by humans anymore, they should get information via CLI and it can convert hexes to bech32 (or push it to REST API itself with `Accept` header)

PR #193, opened 2019-03-20
add block creation/display

NicolasDP:
Add command to create/display block via command line. Simplest example is:

```
cargo run --bin block -- create | \
    cargo run --bin block -- info
```

The first command create the block 0 (with no content) and pire the raw serialiased block to the standard input of the second command (which parse it from the standard input) and display (Debug format) the block to the standard output

### DIFF
--- a/src/bin/block.rs
+++ b/src/bin/block.rs
@@ -1,0 +1,118 @@
+extern crate chain_core;
+extern crate chain_impl_mockchain;
+extern crate structopt;
+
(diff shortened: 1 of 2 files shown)


NicolasDP:
@CodeSandwich I have added a change in the tx_builder: I am now using the Address's standard bech32 format (So the kind/discrimination is human readable).

Have a look at the `cargo run --bin address` for more details on what are the different options of the address and the different outputs expected.

PR #205, opened 2019-03-22
Make CLI TX builder resumable

CodeSandwich:
The CLI needs to be able to build TX incrementally with state saved in between.

Also the private key is now passed as path to file storing it to avoid touching it in plain text.


CodeSandwich:
@vincenthz Simple rebase was enough, no fixes required, everything just works

PR #210, opened 2019-03-25
Full support for fees and accounts in mjolnir.

qnikst:


testing/mjolnir/src/lib.rs
```diff
@@ -177,6 +187,16 @@ impl TransactionId {
         tx::TransactionId::from_str(input)
             .map_err(|e| JsValue::from_str(&format!("{:?}", e)))
             .map(TransactionId)
+        /*
```
vincenthz: left over comment ?

PR #229, opened 2019-03-29
Enhance circle ci tests

NicolasDP:


.circleci/config.yml
```diff
@@ -8,71 +8,106 @@ jobs:
       - checkout
       - run: git submodule sync
       - run: git submodule update --init
-      - restore_cache:
```
mzabaluev: Not restoring cache from past builds, even if those don't completely match this build's `Cargo.lock`, will make every build job re-fetch every crate.

.circleci/config.yml
```diff
@@ -8,71 +8,106 @@ jobs:
...
-          keys:
-            - cargo-v2-{{ checksum "Cargo.toml" }}-
-            - cargo-v2-
       - run: cargo fetch
-      - persist_to_workspace:
-          root: "."
-          paths:
-            - Cargo.lock
```
mzabaluev: OK since `Cargo.lock` is version-controlled.

.circleci/config.yml
```diff
@@ -8,71 +8,106 @@ jobs:
...
+      - run:
+          name: Print version information
+          command: rustc --version; cargo --version
+      - run:
+          name: Build
+          command: cargo build --frozen
+      - persist_to_workspace:
+          root: target
+          paths:
+            - debug/*
```
mzabaluev: Consider pruning it to only persisting the cached sources; the test will rebuild the binaries anyway.

.circleci/config.yml
```diff
@@ -8,71 +8,106 @@ jobs:
...
-  test:
+  rustfmt:
+    docker:
+      - image: rust:latest
+    working_directory: /mnt/crate
+    steps:
+      - checkout
+      - run:
+          name: Install rustfmt
+          command: rustup component add rustfmt
```
mzabaluev: Ideally, we'd want a separate container image with rustfmt pre-installed to save build time. This will do for now.

.circleci/config.yml
NicolasDP: thanks, I've put back the restore cache here too now

.circleci/config.yml
```diff
@@ -6,73 +6,125 @@ jobs:
     working_directory: /mnt/crate
     steps:
       - checkout
-      - run: git submodule sync
-      - run: git submodule update --init
       - restore_cache:
           keys:
-            - cargo-v2-{{ checksum "Cargo.toml" }}-
-            - cargo-v2-
+            - cargo-cache-{{ checksum "Cargo.toml" }}-{{ checksum "Cargo.lock" }}
```
mzabaluev: Good idea to add this one.

.circleci/config.yml
```diff
@@ -6,73 +6,125 @@ jobs:
...
+            - cargo-cache-{{ checksum "Cargo.toml" }}
+            - cargo-cache-
```
mzabaluev: Why did you rename the prefix? The old `cargo-v2-*` entries are still valid for this cache, and the versioning is a useful practice for forcing a fresh cache when we decide we don't need some previously cached directories.

NicolasDP: re-renamed

.circleci/config.yml
mzabaluev: The binaries might be reused as well, at least accordingly to my quick comparison of the build directory between the steps.

PR #234, opened 2019-03-31
Migrate genesis.yml from block_version to consensus

CodeSandwich:
Finish block and consensus versioning split. Switch genesis.yml from pretending that it accepts block version to accepting consensus version without changing meaning of accepted values.


doc/genesis_file.md
```diff
@@ -23,7 +23,7 @@ initial_setting:
   bootstrap_key_slots_percentage: ~
   slot_duration: 15
   epoch_stability_depth: 10
-  block_version: 1
+  consensus: 1
```
NicolasDP: actually since you are there, can you put a string here?

`bft` or `genesis`

vincenthz: here and also in the block0 state

NicolasDP: I think this should be `bft` and  no longer `1`right ?
```suggestion
  consensus: bft
```

CodeSandwich: @vincenthz I may be looking at the wrong place, but AFAIK block0 is `chain_impl_mockchain::block::Block`, which is either stored as binary or generated from genesis YAML in CLI. Using string representation in binary is not a good idea and genesis.yml is already switched to using string. What am I missing?

vincenthz: I'm talking about the Message `Initial` and its content related to ConfigParam in mockchain config.rs

CodeSandwich: `Message::Initial` contains tag-value pairs and the only tags we support for now are `Block0Date` (1) and `Discriminant` (2). I don't see state, block version or consensus version. Should I add another tag for consensus version with string value?

vincenthz: yes, we are indeed missing consensus as one parameter. this is crucial to know what consensus the blockchain is starting with (which is what the `Initial` message is about)

CodeSandwich: Ok, I'll add it. Should it go in this PR or the next one?

vincenthz: it can be the next one

Issue #191, opened 2019-03-20
transaction management CLI


NicolasDP:
- [x] create a transaction from input/output (staging mode, no signature);
- [x] sign a transaction with signing keys (finalise mode);
- [x] display the transaction info (from hexa decimal dump binary dump);

for creating transaction there are multiple options:

1. in cardano-cli, we utilise a staging file with a staging ID, but the issue with this is that we are storing file in the user's environment.
2. I think we would be better adding a command line `--staging FILE` to add inputs/outputs (to create if not yet on the disk.

PR #238, opened 2019-04-02
Add to CLI message debug print command

CodeSandwich:
It allows printing debug representation of the message.

This closes the third point of https://github.com/input-output-hk/jormungandr/issues/191.

### DIFF
--- a/src/bin/jcli_app/debug/message.rs
+++ b/src/bin/jcli_app/debug/message.rs
@@ -1,0 +1,27 @@
+use cardano::util::hex;
+use chain_core::property::Deserialize as _;
+use chain_impl_mockchain::message::Message as MockMessage;
+use jcli_app::utils;
+use std::fs::File;
+use std::io::{self, Read};
+use std::path::PathBuf;
+use structopt::StructOpt;
+
+#[derive(StructOpt)]
+pub struct Message {
+    /// file containing hex-encoded message. If not provided, it will be read from stdin.
+    #[structopt(short, long)]
+    input: Option<PathBuf>,
+}
+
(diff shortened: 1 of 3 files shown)


src/bin/jcli_app/debug/message.rs
```diff
@@ -0,0 +1,27 @@
...
+    input: Option<PathBuf>,
+}
+
+impl Message {
+    pub fn exec(self) {
+        let mut hex = String::new();
+        match self.input {
+            Some(path) => File::open(path).unwrap().read_to_string(&mut hex).unwrap(),
+            None => io::stdin().read_line(&mut hex).unwrap(),
+        };
```
NicolasDP: There is a function in `jcli_app::utils::io` to help you read from the input or the given file

src/bin/jcli_app/mod.rs
```diff
@@ -21,6 +22,8 @@ pub enum JCli {
     Rest(rest::Rest),
     /// Build and view offline transaction
     Transaction(transaction::Transaction),
+    /// Debug tools for developers
+    Debug(debug::Debug),
```
NicolasDP: It seems to me the `message` command should be merged with the transaction CLI.

`jcli message new ...`
`jcli message decode < message.bin`

CodeSandwich: I was considering this, but then noticed that the output of this command is unstable, developer-only-readable, inconsistent with the rest of CLI (not YAML) and probably not really usable to the end user. That's why I decided to follow jormungandr-cli and put it in debug subcommand.

PR #241, opened 2019-04-02
Update cardano-deps; ground work for managing propagation state 

mzabaluev:
Requires input-output-hk/rust-cardano#599.


src/blockchain/process.rs
```diff
@@ -1,14 +1,15 @@
...
-use crate::utils::task::TaskBroadcastBox;
+
 use chain_core::property::Header as _;
 
+use futures::sync::mpsc;
+
 pub fn process(
     blockchain: &chain::BlockchainR,
     bquery: BlockMsg,
-    network_broadcast: &mut TaskBroadcastBox<Header>,
```
NicolasDP: can you hid this behind the TaskBroadcastBox instead then ? It is the only object utilsign the TaskboardcastBox and we introduced abstraction to not have to change the whole code everytime we decide to change how it is working internally

NicolasDP:
this PR is going to conflict with #240

src/blockchain/process.rs
mzabaluev: `TaskMessageBox` then, as it won't be a broadcast message any more.

mzabaluev: Trouble is, `TaskMessageBox` can only be constructed by `Services::spawn_with_inputs`. But we have a chicken and egg problem here, as the network task is constructed last and the block task would need its message box to send broadcasts. Previously, there was also an `UnboundedSender` there which I restored from history; I guess it was done with the abstraction-breaking workaround for the same reason.

mzabaluev: I should think of adding another abstraction type to `network::propagate` that will encapsulate the sender.

mzabaluev: commit dc6ec7cbdd8cfa77510e157ae5c5d4cd025fdd43 provides an abstraction.

PR #242, opened 2019-04-02
Update service management:

NicolasDP:
1. remove hoops to allow for mutable inputs;
2. add support to tokio services
3. move the blockchain service within tokio

### DIFF
--- a/src/network/grpc/client.rs
+++ b/src/network/grpc/client.rs
@@ -46,7 +46,8 @@ pub fn run_connect_socket(peer: Peer, state: GlobalState) -> impl Future<Item = 
                     state
                         .channels
                         .block_box
-                        .send_to(BlockMsg::AnnouncedBlock(header));
+                        .clone()
+                        .send(BlockMsg::AnnouncedBlock(header));
                     future::ok(())
                 })
                 .map_err(|err| {
--- a/src/network/grpc/client.rs
+++ b/src/network/grpc/client.rs
@@ -10,7 +10,7 @@ use futures::prelude::*;
 use tokio::{executor::DefaultExecutor, sync::mpsc};
 
 pub fn run_connect_socket(peer: Peer, state: GlobalState) -> impl Future<Item = (), Error = ()> {
-    let state = ConnectionState::new_peer(&state, &peer);
+    let mut state = ConnectionState::new_peer(&state, &peer);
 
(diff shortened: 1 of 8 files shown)


src/utils/task.rs
```diff
@@ -175,6 +273,34 @@ impl ThreadServiceInfo {
     }
 }
 
+impl TokioServiceInfo {
+    /// get the time this service has been running since
+    #[inline]
+    pub fn up_time(&self) -> Duration {
```
mzabaluev: The common methods of `*ServiceInfo` could be a trait, if it ever needs to be used in a polymorphic way.

src/utils/task.rs
```diff
@@ -143,13 +158,96 @@ impl Services {
...
+            executor: executor,
+        };
+
+        let (msg_box, msg_queue) = async_msg::channel(1000);
+        let future = msg_queue.for_each(move |input| f(&future_service_info, Input::Input(input)));
+
+        runtime.spawn(future);
+
+        let task = Service::new_runtime(name, runtime, now);
+        self.services.push(task);
```
mzabaluev: Lots of code duplication here; could this method be done in terms of `spawn_future`?

src/network/grpc/client.rs
```diff
@@ -46,7 +46,8 @@ pub fn run_connect_socket(peer: Peer, state: GlobalState) -> impl Future<Item =
                     state
                         .channels
                         .block_box
-                        .send_to(BlockMsg::AnnouncedBlock(header));
+                        .clone()
```
mzabaluev: It's enough to make `state` mutable, no need for cloning.

src/utils/task.rs
NicolasDP: I'll think about it, since there is only 2 objects I am not sure it is worth adding abstraction here.

src/utils/task.rs
NicolasDP: that could be a solution, though there are requirements we still need to comply to regarding some of our services. The leadership task needs to run in a thread and we need to make sure it is well enough isolated from anything else.

src/network/grpc/client.rs
NicolasDP: ha yes. I've been lazy there. I missed that point

PR #244, opened 2019-04-03
Readme fixes

dkijania:
Fixing docs:

- conventions section is not anchored
- IMHO in cli_address.md there is no need to add cargo run in each command
- rest transaction -> message since` jcli rest v0 transaction post <options>` does not exists
- starting bft: prameters should be underscored not dashed

### DIFF
--- a/doc/cli_transaction.md
+++ b/doc/cli_transaction.md
@@ -40,3 +40,11 @@ ed25519extended_public key.
 bech32-encoded ed25519extended_secret. Required one for every input.
 
 Value outputted to stdout on success is transaction message encoded as hex.
+
+
+### Example
+
+```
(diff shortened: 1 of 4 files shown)


doc/cli_transaction.md
```diff
@@ -40,3 +40,11 @@ ed25519extended_public key.
 bech32-encoded ed25519extended_secret. Required one for every input.
 
 Value outputted to stdout on success is transaction message encoded as hex.
+
+
+### Example
+
+```
+jcli transaction build --input f7f1b60d6033bf72409f5fca14fea21f657a0cee19729146351a698d2b6a853e:0:100  \
```
NicolasDP: message, then?

PR #254, opened 2019-04-04
First integration tests case for genesis block encode

dkijania:
Signed-off-by: dkijania <dariusz.kijania@iohk.io>

1) Assumption

a) i use JCLI which is already installed on env. Maybe in the future we would need to have an option to use jcli which is built (for debugging purposes).
b) I created genesis.yaml in project under resources, because it didn't want to create it each test. This required to write custom build script (copy-resources.rs) with one workaround (which causes that build script is run at each build). Please comment if this is good approach. 
c) after cargo test command in debug profile command there is an executable with test_{Hash}.exe which then can execute tests without cargo. 

2) TODO's:

a) maybe we should move output file path to temp


testing/integration-tests/Cargo.toml
```diff
@@ -0,0 +1,16 @@
+[package]
+name = "integration_tests"
```
mzabaluev: Package names are supposed to be global, so I'd name it "jormungandr-integration-tests". It's more conventional to use dashes in package names.

testing/integration-tests/Cargo.toml
```diff
@@ -0,0 +1,16 @@
...
+version = "0.1.0"
+authors = ["dkijania <dariusz.kijania@iohk.io>"]
```
mzabaluev: Consider putting the full name here.

testing/integration-tests/copy_resources.rs
```diff
@@ -0,0 +1,27 @@
+extern crate fs_extra;
+use fs_extra::dir::copy;
+use fs_extra::dir::*;
```
mzabaluev: Please avoid wildcard imports except from preludes and the like.
This code is not "productoin", so this is more guidance on good style.

testing/integration-tests/src/file_assert.rs
```diff
@@ -0,0 +1,44 @@
...
+///
+/// use integration_tests::file_assert::assert_file_exists;
+/// let file_name = "./test.txt";
+/// assert_file_exists(&file_name);
+///
+pub fn assert_file_exists(file_name: &str) {
+    let metadata = metadata(&file_name).unwrap();
+    assert!(
+        metadata.is_file(),
+        "file '{}' does not exists, while it should",
```
mzabaluev: Spelling: "does not exist"

testing/integration-tests/src/file_assert.rs
```diff
@@ -0,0 +1,44 @@
...
+///
+/// use integration_tests::file_assert::assert_file_not_empty;
+/// let file_name = "./test.txt";
+/// assert_file_not_empty(&file_name);
+///
+pub fn assert_file_not_empty(file_name: &str) {
+    let metadata = metadata(&file_name).unwrap();
+    assert!(
+        metadata.len() > 0,
+        "file '{}' is empty (len = {}), while it shouldn't",
```
mzabaluev: "shouldn't be"? :)

mzabaluev: This will always print "(len = 0)" unless the type is signed (meaning the API is strange), why print it?

testing/integration-tests/src/lib.rs
```diff
@@ -0,0 +1,18 @@
...
+///
+/// * `file_name` - A string slice that holds the name of the file
+///
+/// # Example
+///
+/// use integration_tests::assert_file_exists_and_not_empty;
+/// let file_name = "./test.txt";
+/// assert_file_exists_and_not_empty(&file_name);
+///
+pub fn assert_file_exists_and_not_empty(file_name: &str) {
```
mzabaluev: Why is this function not in the same module with the others?

testing/integration-tests/src/file_assert.rs
```diff
@@ -0,0 +1,44 @@
...
+/// # Example
+///
+/// use integration_tests::file_assert::assert_file_not_empty;
+/// let file_name = "./test.txt";
+/// assert_file_not_empty(&file_name);
+///
+pub fn assert_file_not_empty(file_name: &str) {
+    let metadata = metadata(&file_name).unwrap();
+    assert!(
+        metadata.len() > 0,
```
NicolasDP:
```suggestion
        ! metadata.empty(),
```

testing/integration-tests/src/lib.rs
NicolasDP: yes, move it to the `file_assert` like the other. that's where it belongs

testing/integration-tests/Cargo.toml
```diff
@@ -0,0 +1,16 @@
...
+edition = "2018"
+build = "copy_files.rs"
```
NicolasDP:
```suggestion
build = "copy_resources.rs"
```

NicolasDP: otherwise it does not compile at all

NicolasDP:
something else, since this is only testing, actually why not adding it to jormungandr itself ? I am thining if you put them in the `https://doc.rust-lang.org/cargo/guide/project-layout.html` it should make things easier, no?

dkijania:
> something else, since this is only testing, actually why not adding it to jormungandr itself ? I am thining if you put them in the [tests directory](https://doc.rust-lang.org/cargo/guide/project-layout.html) it should make things easier, no?
> 
> there is some doc about it [here](https://doc.rust-lang.org/book/ch11-03-test-organization.html#the-tests-directory)

Yes it is possible.  we can add attribute to test `#[cfg(feature = "integration_tests")]` and then if we want to run only integration tests we will fire up: `cargo test --features integration_tests` then cargo test will only launch unit. Otherwise integration tests will be launched in each cargo test phase, and i believe this can cause troubles, like some devs just want to build project and run unit tests only or integration tests could required some setup (already launched node?). What you think ?

NicolasDP:
I'm happy with that

testing/integration-tests/copy_resources.rs
dkijania: Thanks. Every code should have a production quality, even test code. So don't hesitate to put more remarks about its quality

testing/integration-tests/src/file_assert.rs
dkijania: Actually after this change. I've got compilation issue with message that there is no empty method for metadata

NicolasDP: sorry, it is `is_empty`: https://doc.rust-lang.org/std/vec/struct.Vec.html#method.is_empty

PR #260, opened 2019-04-08
Add integration test job to circle ci

dkijania:


.circleci/config.yml
```diff
@@ -145,6 +145,29 @@ jobs:
           name: Build and test with nightly Rust
           command: cargo test --frozen --verbose
 
+  test_integration_nightly:
+    docker:
+      - image: rustlang/rust:nightly
```
NicolasDP: why on nightly only ?

also why not keeping the build outputs from the build job?

PR #262, opened 2019-04-08
Update node secret

NicolasDP:
Merging this PR will generate the following change in the format of the node's secret file:

```yaml
bft:
    signing_key: ed25519extended_secretbech32...
```

There is an _undocumented part_ (for now) about starting the node in genesis (stake pool):

```yaml
genesis:
    vrf_key: curve25519_2hashdh_secretbech32...
    sig_key: fakemmm_secretbech32...
    node_id: aa7215b6bb9d0b85f77bb0452c0b2e5543eb57981f1d99cc1ea62927dcba1bc2
```

### DIFF
--- a/src/secure/mod.rs
+++ b/src/secure/mod.rs
@@ -1,12 +1,57 @@
-use bech32::{Bech32, FromBase32};
-use chain_crypto::{AsymmetricKey, Ed25519Extended, PublicKey, SecretKey};
-use std::fs;
+use chain_crypto::{
+    Blake2b256, Curve25519_2HashDH, Ed25519Extended, FakeMMM, PublicKey, SecretKey,
+};
+use chain_impl_mockchain::leadership::{bft, BftLeader, GenesisLeader};
+use serde::Deserialize;
 use std::path::Path;
 
+/// hold the node's bft secret setting
+#[derive(Clone, Deserialize)]
+pub struct Bft {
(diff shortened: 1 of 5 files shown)


src/secure/mod.rs
```diff
@@ -15,26 +60,28 @@ pub struct NodePublic {
     pub block_publickey: PublicKey<Ed25519Extended>,
 }
 
+custom_error! {pub NodeSecretFromFileError
+    Io { source: std::io::Error } = "Cannot read node's secrets: {source}",
+    Format { source: serde_yaml::Error } = "Invalid Node secret file: {source}",
```
mzabaluev: This custom error stuff inlines the source error in `Display`...

src/start_up/error.rs
```diff
@@ -15,6 +15,7 @@ custom_error! {pub Error
     StorageError { source: StorageError } = "Storage error",
     Blockchain { source: blockchain::LoadError } = "Error while loading the blockchain state",
     Block0 { source: blockcfg::Block0Error } = "Error in the genesis-block",
+    NodeSecrets { source: secure::NodeSecretFromFileError} = "Error while loading the node's secrets."
```
mzabaluev: ... while this one does not.

NicolasDP: Yes, they are not displayed the same in the main.rs in case of error. The startup::Error is the one being catch at the main level. If it fails it display the error message and then the source. But does not recurse through every sources (that would make UX difficult IMHO).

PR #264, opened 2019-04-09
First node integration test

dkijania:


dkijania:
@NicolasDP  after your changes (get jcli and jormungandr from build directory not from installed one) I think cargo install before test is necessary, am i right?

tests/__galvanic_test/jcli_wrapper.rs
```diff
@@ -0,0 +1,64 @@
+use std::process::Command;
+use std::path::PathBuf;
+
+use crate::__galvanic_test::configuration;
```
mzabaluev: What with the double underscore name?

tests/__galvanic_test/configuration.rs
```diff
@@ -0,0 +1,40 @@
...
+    let path = PathBuf::from("./tests/resources/genesis/genesis.yaml");
+    path
+}
+pub fn get_node_config_path() -> PathBuf {
+    let path = PathBuf::from("./tests/resources/jormungandr/node.config");
+    path
+} 
+
+/// Get jormungandr executable defined as system environment
+pub fn get_jormungandr_app_variable_as_path_from_os() -> PathBuf {
```
mzabaluev: These names could be shorter, no need to tell the whole story in the name. If you find yourself using `{common_part}_{specific_part}` names a lot, it probably means in Rust you could do with a module: `{common_part}::{specific_part}`.

tests/__galvanic_test/configuration.rs
```diff
@@ -0,0 +1,40 @@
...
+    let mut path : PathBuf = env!("JCLI").into();	
+    if cfg!(windows) {
+        path.set_extension("exe");
+    }
+    assert!(path.is_file(), "File does not exist: {:?}, pwd: {:?}", path, list_dir());
+    path	
+}
+
+
+/// list directory depenging on os
```
mzabaluev: This is unclear. A list, of what? The function actually obtains the CWD from the environment, are you sure you don't just want `std::env::current_dir()` or at least use it under the hood?
Typo: "depending"

tests/__galvanic_test/jcli_wrapper.rs
```diff
@@ -0,0 +1,64 @@
...
+    let mut command =  Command::new(get_jcli_app_path().as_os_str());
+    command.arg("rest")
+        .arg("v0")
+        .arg("node")
+        .arg("stats")
+        .arg("get")
+        .arg("-h")
+        .arg(&configuration::JORMUNGANDR_ADDRESS);
+    command   
+}
```
mzabaluev: Please apply rustfmt and/or set up your editor to fix this automatically, these unterminates lines look annoying.

tests/jcli_genesis.rs
```diff
@@ -1,57 +1,39 @@
...
 
     fixture genesis_fixture() -> PathBuf {
         setup(&mut self) {
-             let temp_dir = Temp::new_dir().unwrap();
-             let mut path = temp_dir.to_path_buf();
-             path.push("block-0.bin");
-             println!("Setup: location for output block file: {}",&path.to_str().unwrap());
-             temp_dir.release();
+             let mut path = file_utils::get_path_in_temp("block-0.bin");
+             println!("Setup: location for output block file: {}",path.as_os_str().to_str().unwrap());
```
NicolasDP:
```suggestion
             println!("Setup: location for output block file: {:?}", path);
```

tests/resources/jormungandr/node.config
```diff
@@ -0,0 +1,13 @@
...
+  format: json
+rest:
+  listen: "127.0.0.1:8443"
+  prefix: "api"
+peer_2_peer:
+  trusted_peers: []
+  public_access: "/ip4/127.0.0.1/tcp/8080"
+  topics_of_interests:
+    messages: low
+    blocks: normal
```
NicolasDP: We need to find a way to generate this document depending on the needs.

One easy way we can do that is by preparing the default configuration file in a _format ready way_.

NODE_CONFIG.yaml

```yaml
storage: {storage}
logger:
  verbosity: 1
  format: json
rest:
  listen: "{rest_listen}"
  prefix: "{rest_api}"
peer_2_peer:
  trusted_peers: {p2p_trusted_peers}
  public_access: "{p2p_public_access}"
  topics_of_interests:
    messages: {topic_message_interest_level}
    blocks: {topic_blocks_interest_level}
```

Then we can generate the configuration file for every nodes we need to setup:

```rust
    writeln!(output_file, include_str!("NODE_CONFIG.yaml"),
        storage = config.storage.unwrap_or("~"),
        rest_listen = config.rest.listen,
        rest_prefix = config.rest.prefix,
        p2p_trusted_peers = config.p2p.peers,
        p2p_public_access = config.p2p.public,
        topic_messages_interest = config.p2p.message_interest,
        topic_blocks_interest = config.p2p.blocks_interest,
   )
```

This will allow to call the function above with pre-defined parameters, to setup more easily the node(s) and how the topology will look like.

Ideally, we will need to do the same with the `genesis.yaml` file for the blockchain. and the `node_secret.yaml`  file too for the nodes we need to be able to write blocks in the blockchain.

dkijania: Absolutely agree. At this moment i hardcoded a few things (like node.config and gensis.yaml), but i will generate them on the fly in the nearest future as you proposed

tests/__galvanic_test/jcli_wrapper.rs
dkijania: I thought i did it couple of times. It's funny but there is a bug: 

https://github.com/rust-lang/rustfmt/issues/1820

which prevents cargo fmt from formatting subdirectories in tests folder

tests/__galvanic_test/configuration.rs
dkijania: Removed method

tests/__galvanic_test/jcli_wrapper.rs
mzabaluev: I still don't like the name. What's the purpose of the underscore?

Besides, this use statement could just use `super::` if I'm not mistaken.

NicolasDP: It's cleaner to use the `crate::path::to::feature`. It gives a real indication where it is in the crate. Simpler to find out where the data came from.

PR #266, opened 2019-04-09
Add propagation among subscribed peers

mzabaluev:
Propagate blocks and messages announced by their respective service tasks.

Currently propagation is only expected to work among the peers who have subscribed. On-demand client propagation toward peers discovered via gossip is not yet implemented.

### DIFF
--- a/src/network/propagate.rs
+++ b/src/network/propagate.rs
@@ -6,7 +6,10 @@ use network_core::{error::Error, gossip::Gossip};
 use futures::prelude::*;
 use futures::sync::mpsc;
 
-use std::{collections::HashMap, sync::RwLock};
+use std::{
+    collections::{hash_map, HashMap},
+    sync::Mutex,
+};
 
 // Buffer size determines the number of stream items pending processing that
 // can be buffered before back pressure is applied to the inbound half of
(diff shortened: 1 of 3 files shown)


src/network/propagate.rs
```diff
@@ -122,25 +125,56 @@ fn ensure_propagation_peer<'a>(
...
+
+    fn propagate_with<F>(&self, ids: &[p2p::NodeId], f: F)
+    where
+        F: Fn(&mut PeerHandles) -> Result<(), PropagateError>,
+    {
+        let mut map = self.mutex.lock().unwrap();
+        for id in ids {
+            if let hash_map::Entry::Occupied(mut entry) = map.entry(id.clone()) {
+                f(entry.get_mut()).unwrap_or_else(|e| {
+                    info!("propagation stream error: {:?}", e);
```
NicolasDP: a `warn` would make more sense, no?

mzabaluev: It might happen for benign reasons, like loss of network connectivity or the peer being slow. The mitigation is to simply unsubscribe the peer.

NicolasDP: the keyword error in the log is confusing.

PR #270, opened 2019-04-10
Propagate on demand

mzabaluev:
When there are peer nodes selected for propagation that are not currently
subscribed, connect to them, follow through with propagating the item,
and add them to the map.


src/network/grpc/client.rs
```diff
@@ -1,5 +1,8 @@
 use super::{
-    super::{propagate, subscription, BlockConfig, Channels, ConnectionState, GlobalStateR},
+    super::{
```
NicolasDP: utilising `crate::network::{` would be easier really...

PR #271, opened 2019-04-11
Removed 'install crate' task from integration jobs. Removed galvanic …

dkijania:
…test usage in tests.

Added Model for genesis yaml.

There is some duplication in tests. My plan is to remove it in next PR. Since there are bugs in transaction build and transaction post, I've limited tests only to verifify in utxos are correct based on genesis yaml.


tests/common/file_utils.rs
```diff
@@ -15,9 +18,18 @@ use std::path::PathBuf;
 /// get_path_in_temp(&path_in_temp);
 ///
 pub fn get_path_in_temp(file_path: &str) -> PathBuf {
-    let temp_dir = Temp::new_dir().unwrap();
+    let temp_dir = mktemp::Temp::new_dir().unwrap();
     let mut path = temp_dir.to_path_buf();
     path.push(&file_path);
     temp_dir.release();
```
mzabaluev: I'd rather use crate `tempfile` as it is more mature. This is for tests, so it does not matter much as long as the test can clean up for itself.

tests/common/file_utils.rs
```diff
@@ -15,9 +18,18 @@ use std::path::PathBuf;
...
     path
 }
+
+/// Creates file in temporary folder
```
mzabaluev: The docs should decide if they want to use "folder" or "directory". I'd prefer the latter, "folder" is a UI term :)

tests/common/process_utils/mod.rs
```diff
@@ -1,7 +1,65 @@
...
+
+pub fn run_process_and_get_output_line(command: Command) -> String {
+    let mut content = run_process_and_get_output(command);
+    if content.ends_with("\n") {
+        let len = content.len();
+        content.truncate(len - 1);
+    }
+    content
+}
+
```
mzabaluev: This family of functions could be more composable: `run_and_collect_output` could return a value of type `Output` that has methods `to_yaml_node`, `to_yaml_nodes`, etc.

tests/common/file_utils.rs
NicolasDP: arf. I think my bad I suggested mktemp last time. And now looking at it, tempfile seems a better long term bet.

dkijania:
Ok. I'll prepare different PR with circle Ci change

PR #282, opened 2019-04-17
Extend config.yaml init params

CodeSandwich:
Finishes https://github.com/input-output-hk/rust-cardano/issues/610

### DIFF
--- a/src/blockcfg/mod.rs
+++ b/src/blockcfg/mod.rs
@@ -14,49 +14,58 @@ pub use chain_impl_mockchain::{
 };
 use std::time::{Duration, SystemTime};
 
-custom_error! {pub Block0Malformed
-    NoInitialSettings = "Missing its initial settings",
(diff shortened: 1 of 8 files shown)


NicolasDP:
Seems like the block0 is malformed. Did you try running jormungandr with your update?

src/blockcfg/mod.rs
```diff
@@ -50,6 +50,21 @@ pub fn block_0_get_slot_duration(block: &Block) -> Result<std::time::Duration, B
     }
 }
 
+pub fn block_0_get_slots_per_epoch(block: &Block) -> Result<Option<usize>, Block0Error> {
```
mzabaluev: Why does this API feel like Pascal to me? :)
Can we have a `GenesisBlock` newtype with these methods?

CodeSandwich:
@NicolasDP That's right, it was broken, I've fixed it, but that was dangerous :(

PR #287, opened 2019-04-24
network: function to fetch the genesis block

mzabaluev:
Add a blocking function to connect to the first of the trusted
peers and fetch the block specified by hash.

### DIFF
--- a/src/network/grpc/client.rs
+++ b/src/network/grpc/client.rs
@@ -1,7 +1,11 @@
 use super::origin_authority;
-use crate::network::{
-    p2p_topology as p2p, propagate, subscription, BlockConfig, Channels, ConnectionState,
-    GlobalStateR,
+use crate::{
+    blockcfg::{Block, HeaderHash},
+    network::{
+        p2p_topology as p2p, propagate, subscription, BlockConfig, Channels, ConnectionState,
+        GlobalStateR,
+    },
+    settings::start::network::Peer,
 };
 
 use network_core::{
@@ -18,7 +22,7 @@ use http::uri;
 use tokio::{executor::DefaultExecutor, net::TcpStream};
 use tower_service::Service as _;
 
-use std::net::SocketAddr;
+use std::{net::SocketAddr, slice};
 
 pub fn connect(
     addr: SocketAddr,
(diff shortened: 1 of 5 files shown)


src/network/mod.rs
```diff
@@ -196,9 +197,14 @@ fn handle_propagation(
...
     input.for_each(move |msg| {
+        debug!("to propagate: {:?}", &msg);
         let channels = channels.clone();
         let state = state.clone();
-        let nodes = state.topology.view().collect();
+        let nodes = state.topology.view().collect::<Vec<_>>();
+        debug!(
+            "will propagate to: {:?}",
+            nodes.iter().map(|node| node.id()).collect::<Vec<_>>()
+        );
```
mzabaluev: These debug lines got swept in with the other changes.

src/network/mod.rs
```diff
@@ -307,3 +320,23 @@ pub fn bootstrap(config: &Configuration, blockchain: BlockchainR) {
...
+/// The calling thread is blocked until the block is retrieved.
+/// This function is called during blockchain initialization
+/// to retrieve the genesis block.
+pub fn fetch_block(config: &Configuration, hash: &HeaderHash) -> Result<Block, ()> {
+    if config.protocol != Protocol::Grpc {
+        unimplemented!()
+    }
+    match first_trusted_peer_address(config) {
+        None => {
+            error!("no gRPC peers specified, cannot retrieve block {}", hash);
```
NicolasDP: can we have an error report instead ?

src/network/mod.rs
```diff
@@ -307,3 +320,23 @@ pub fn bootstrap(config: &Configuration, blockchain: BlockchainR) {
...
+
+/// Queries the trusted peers for a block identified with the hash.
+/// The calling thread is blocked until the block is retrieved.
+/// This function is called during blockchain initialization
+/// to retrieve the genesis block.
+pub fn fetch_block(config: &Configuration, hash: &HeaderHash) -> Result<Block, ()> {
+    if config.protocol != Protocol::Grpc {
+        unimplemented!()
+    }
+    match first_trusted_peer_address(config) {
```
NicolasDP: This will do for now. But we need to have concurrent fetch instead in the long run.

src/network/grpc/client.rs
```diff
@@ -67,3 +71,38 @@ fn subscribe(
             Ok((node_id, prop_handles))
         })
 }
+
+// Fetches a block from a network peer in a one-off, blocking call.
+// This function is used during node bootstrap to fetch the genesis block.
+pub fn fetch_block(peer: Peer, hash: &HeaderHash) -> Result<Block, ()> {
```
NicolasDP: I'd rather like this function to return a proper error code.

You can now create new error type fairly easily:
```rust
custom_error! {pub FetchBlockError
    // ...
}

PR #300, opened 2019-04-26
Removing build script from Cargo.toml

dkijania:
Removing build script from Cargo.toml which caused constant build even if there is no change in crate. Moved logic to configuration.


tests/common/configuration/mod.rs
```diff
@@ -34,3 +36,14 @@ pub fn get_jcli_app() -> PathBuf {
     );
     path
 }
+
+/// Gets working directory
+fn get_working_directory() -> PathBuf {
+    let mut output_directory: PathBuf = std::env::current_exe().unwrap().into();
+
+    /// current exe directory is ./target/{profile}/deps/{app_name}.exe
+    /// We would like to navigate to ./target/{profile}
```
mzabaluev: I wonder if cargo passes down some environment variables that could help locate the target dir.

Cargo.toml
```diff
@@ -14,7 +14,6 @@ documentation = "https://github.com/input-output-hk/jormungandr#USAGE.md"
 description = """
 Midgard Serpent
 """
-build = "tests/build/deploy_tests_resources.rs"
```
mzabaluev: :raised_hands:

tests/common/configuration/mod.rs
dkijania: After reading below article:

https://doc.rust-lang.org/cargo/reference/environment-variables.html

and printing all env variables available in test. so far i couldn't find a better way

PR #301, opened 2019-04-26
Implement block 0 fetching from network on startup

### LINKED ISSUE #294: fetch block0 from the network

mzabaluev:
Resolves #294


mzabaluev:
Even better now, it actually works.

PR #303, opened 2019-04-26
new integration tests for keys

dkijania:


tests/common/file_utils.rs
```diff
@@ -58,3 +63,10 @@ pub fn create_file_with_content(path: &PathBuf, content: &str) -> () {
     file.write_all(content.as_bytes())
         .expect(&format!("cannot write to file {:?}", path));
 }
+
+pub fn read_file(path: &PathBuf) -> String {
+    let mut file = File::create(&path).unwrap();
```
NicolasDP: calling [`File::create`](https://doc.rust-lang.org/stable/std/fs/struct.File.html#method.create) will create the file if it does not exist or will truncate the file if it already exists, is it the original intent ?

tests/common/file_utils.rs
```diff
@@ -58,3 +63,10 @@ pub fn create_file_with_content(path: &PathBuf, content: &str) -> () {
...
+    let mut buffer = String::new();
+    std::fs::read_to_string(&mut buffer);
+    buffer
```
NicolasDP: I don't understand what this function is doing. `file` is not used at all. What are you reading?

The function [`std::fs::read_to_string`](https://doc.rust-lang.org/stable/std/fs/fn.read_to_string.html) does not read in the buffer here... please read the doc.

tests/common/file_utils.rs
```diff
@@ -58,3 +63,10 @@ pub fn create_file_with_content(path: &PathBuf, content: &str) -> () {
...
+}
```
NicolasDP: It seems to me, this function `read_file` is not needed and `std::fs::read_to_string` is sufficient.

tests/common/file_utils.rs
dkijania: Yes this is mistake, removing method

tests/common/file_utils.rs
dkijania: Yes this is mistake, removing method

tests/common/file_utils.rs
dkijania: Yes this is mistake, removing method

PR #304, opened 2019-04-28
add block0 hash when signing the witness

### LINKED ISSUE #292: make sure we sign with the block0 hash
Sign the transactions and the certificates with the block0 hash. This will prevent utilising messages in multiple blockchains.

relates to input-output-hk/rust-cardano#644

NicolasDP:
fix #292

### DIFF
--- a/src/bin/jcli_app/transaction/mk_witness.rs
+++ b/src/bin/jcli_app/transaction/mk_witness.rs
@@ -4,6 +4,7 @@ use chain_crypto::bech32::{self as bech32_crypto, Bech32 as _};
 use chain_crypto::{AsymmetricKey, SecretKey, SecretKeyError};
 use chain_impl_mockchain::{
     account::SpendingCounter,
+    block::HeaderHash,
     transaction::{TransactionId, Witness},
 };
 use jcli_app::{transaction::common, utils::io};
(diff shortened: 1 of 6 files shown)


src/bin/jcli_app/transaction/mk_witness.rs
```diff
@@ -36,9 +37,14 @@ pub struct MkWitness {
     /// If omitted it will be printed to the standard output.
     pub output: Option<PathBuf>,
 
+    /// the type of witness to build: account, UTxO or Legacy UtxO
     #[structopt(long = "type", parse(try_from_str))]
     pub witness_type: WitnessType,
 
+    /// the hash of the block0, the first block of the blockchain
+    #[structopt(long = "block0-hash", parse(try_from_str))]
```
mzabaluev: There is a similar option `--genesis-block` in jormungandr which has been bothering me, because it can be a file path or a hash, but it's only interpreted as a path when it fails to parse as a hash.

Can we harmonize these options in both executables so that: 1) they use the same names; 2) clearly disambiguate between file and hash parameters?

NicolasDP: I am not going to change the parameters in jormungandr. But I agree, this was not a great choice in there. I'll rename it jcli for now.

PR #305, opened 2019-04-29
Integration tests for jcli address

dkijania:


tests/jcli_address_test.rs
```diff
@@ -0,0 +1,163 @@
...
+
+    let mut public_key = jcli_wrapper::assert_key_to_public_default(&private_key);
+    println!("public key: {}", &public_key);
+
+    public_key.push('A');
+
+    // Assertion changed due to issue #306. After fix please change it to correct one
+    process_assert::assert_process_failed_and_contains_message(
+        jcli_wrapper::jcli_commands::get_address_single_command_default(&public_key),
+        "thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: MixedCase",
```
mzabaluev: It's fragile to depend on precise formatting of panic messages by the standard library and `fmt` impls. Would it be better to match the essential substring or a regex?

dkijania: fixed

PR #315, opened 2019-05-02
Automation report 3MAY2019

dkijania:


dkijania:
Ger Moroney asked us to provide such data (coverage + passrate). After discussion with nicolas we decided to keep it in github. Keeping it in readme file + images is just for convenience to have some presentation of  CI artifacs which is in a very raw form. Then i can point him to page or paste it to mail.

If you mean that we should generate also those metrics with presentation form, then i agree but i haven't got fully automated process yet.  Will add also Nicolas to speak his mind about it. because he also got some idea how it should be stored

vincenthz:
possibly that might make sense to have a separate repository to commit reports ?

dkijania:
@vincenthz  I don't mind having it on separate repo if this is what you prefer

dkijania:
But i don't think i have rights to do this. so can anyone create it for me?

mzabaluev:
I think doing something similar to what #318 did for documentation is more sensible. I wouldn't like automation noise in our commit history.

dkijania:
I do understand benefits from using mdbook and this is a good idea for presenting markdown files but i don't understand how #318 solves problem with clean commits history (not polluted by automation noise). Is this comment refers to idea of keeeing automation reports commit in different repo ?

mzabaluev:
I did not mean using mdbook specifically, but the general approach of deploying artefacts to input-output-hk.github.io.

NicolasDP:
The automation report and the documentation is different. @vincenthz will create a repo for these

vincenthz:
it's created at https://github.com/input-output-hk/jormungandr-qa

PR #327, opened 2019-05-06
Sync with the rust-cardano unify-proposals-and-initial-config branch

edolstra:
This means block 0 no longer contains a (malformed) `UpdateProposal` message, only an `Initial` message containing configuration parameters.

This also means that `initial_setting` in the YAML genesis configuration file is gone - all settings are now part of `blockchain_configuration`. (Originally the former was supposed to be the immutable parameters, but it was questionable whether some of them are really immutable, e.g. slot duration, BFT leaders,  consensus version etc.)

(This PR depends on https://github.com/input-output-hk/rust-cardano/pull/669.)

### DIFF
--- a/src/bin/jcli_app/block/genesis/yaml.rs
+++ b/src/bin/jcli_app/block/genesis/yaml.rs
@@ -1,16 +1,13 @@
 use chain_addr::{Address, Discrimination};
 use chain_core::property::HasMessages;
-use chain_crypto::bech32::Bech32;
-use chain_crypto::{Ed25519Extended, PublicKey};
 use chain_impl_mockchain::{
     block::{Block, BlockBuilder, ConsensusVersion},
-    certificate::{Certificate, SignatureRaw},
+    certificate::Certificate,
     config::{Block0Date, ConfigParam},
     fee::LinearFee,
     legacy::{self, OldAddress},
-    message::{InitialEnts, Message},
+    message::{ConfigParams, Message},
     milli::Milli,
-    setting::{SignedUpdateProposal, UpdateProposal, UpdateProposalWithProposer},
     transaction,
     value::Value,
 };
@@ -29,7 +26,6 @@ pub struct Genesis {
     /// All that is static and does not need to have any update
     /// mechanism.
     blockchain_configuration: BlockchainConfiguration,
-    initial_setting: Update,
     #[serde(default)]
     initial_funds: Vec<InitialUTxO>,
     #[serde(default)]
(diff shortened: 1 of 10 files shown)


src/bin/jcli_app/block/genesis/yaml.rs
```diff
@@ -308,10 +294,13 @@ impl BlockchainConfiguration {
                 ConfigParam::EpochStabilityDepth(param) => epoch_stability_depth
                     .replace(*param)
                     .map(|_| "EpochStabilityDepth"),
-                ConfigParam::ConsensusLeaderId(param) => {
+                ConfigParam::AddBftLeader(param) => {
                     consensus_leader_ids.push(SerdeLeaderId(param.clone()));
                     None
                 }
+                ConfigParam::RemoveBftLeader(_) => {
```
mzabaluev: Does this mean that config is now not declarative, but procedural?

sjmackenzie: BFT nodes aren't immutable.

edolstra: Yeah that's right. Also, `ConfigParams` is now best seen as a diff to the "current" configuration (since that's what an update proposal describes).

src/start_up/mod.rs
```diff
@@ -102,7 +102,7 @@ pub fn prepare_clock(block0: &Block) -> Result<Clock, Error> {
 
     let initial_epoch = ClockEpochConfiguration {
         slot_duration,
-        slots_per_epoch: slots_per_epoch.unwrap_or(10 * 10),
+        slots_per_epoch: slots_per_epoch.unwrap_or(100), // FIXME
```
joevegan: whats wrong with the default value?

edolstra: This should actually be `date::EPOCH_DURATION`, but rust-cardano doesn't export that constant at the moment. (Also it's not a very good default probably.)

edolstra: Hm, I see the default got removed from rust-cardano, so maybe we should just panic here if `slots_per_epoch` is not set? @vincenthz

src/bin/jcli_app/block/genesis/yaml.rs
```diff
@@ -46,27 +42,19 @@ struct BlockchainConfiguration {
...
-
-/// the initial configuration of the blockchain
-///
-/// This is the data tha may be updated but which needs
-/// to have an initial value in the blockchain (or not)
-#[derive(Clone, Serialize, Deserialize)]
-struct Update {
     max_number_of_transactions_per_block: Option<u32>,
-    bootstrap_key_slots_percentage: Option<u8>,
+    bft_slots_ratio: Option<SerdeAsString<Milli>>,
```
NicolasDP: Is this really an optional parameter ? It seems to me that if we start the blockchain without this parameter, the node will fail to start.

edolstra: The default in `Settings::new()` is `Milli::ONE`, so if it's not set in the genesis block, it will get that value. Also, `bft_slots_ratio` is not actually used anywhere at the moment AFAIK.

src/start_up/mod.rs
vincenthz: yes, slots_per_epoch & slot_duration should always be present now

PR #329, opened 2019-05-07
integration test for the REST API: fix the expected error message

NicolasDP:
expect `Status(400)` as this is simple to pattern match and less
likely to change

### DIFF
--- a/tests/jcli_rest_test.rs
+++ b/tests/jcli_rest_test.rs
@@ -103,7 +103,7 @@ pub fn test_correct_error_is_returned_for_incorrect_block_id() {
             &incorrect_block_id,
             &jormungandr_rest_address,
         ),
-        "value: Error { kind: ClientError(400)",
+        "Status(400)",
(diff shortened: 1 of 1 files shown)


tests/jcli_rest_test.rs
```diff
@@ -103,7 +103,7 @@ pub fn test_correct_error_is_returned_for_incorrect_block_id() {
             &incorrect_block_id,
             &jormungandr_rest_address,
         ),
-        "value: Error { kind: ClientError(400)",
+        "Status(400)",
         "This assertion is incorrect on purpose to aviod failing build when running test,
         after #298 is fixed it need to be changed to correct one",
```
dkijania: Since you fixed assertion, this message can be removed

NicolasDP: the error message is still not correct here. I'll leave this message. We still need proper report.
There have been a change in the down stream and the previous error message has been changed. But still not correct.

PR #331, opened 2019-05-07
Fixed git submodule init missing instruction in README

robcohen:
Current README instructions omit the following instruction:
```
$ git submodule init
```

This PR adds this command to the README.


robcohen:
Changed the command to:

git submodule update --init --recursive

Based on recommendation from @NicolasDP

Issue #298, opened 2019-04-26
[JCLI_REST] Panic instead of well-formed error message


dkijania:
**Describe the bug**
There are few places where jcli throws internal exception on bad input while it should produce error message like in the other parts. 

**To Reproduce**
1. Start jormungandr with any settings
2.  run command with bad data like below:
  `jcli rest v0 block e1049ea45726f0b1fc473af54f706546b3331765abf89ae9e6a8333e49621641aa get --host http://127.0.0.1:8443/api `
3. See panic is produced:
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Error { kind: ClientError(400), url: `Some("http://127.0.0.1:8443/api/v0/block/e1049ea45726f0b1fc473af54f706546b3331765abf89ae9e6a8333e49621641aa") }', src\libcore\result.rs:997:5
note: Run with `RUST_BACKTRACE=1` environment variable to display a backtrace.`

**Expected behavior**
Similar output should be presented:
Cannot get block
  |-> unknown id

**Additional context**

There are others commands in jcli rest subcommand that have the same behavior:

1) ` jcli rest v0 block e1049ea45726f0b1fc473af54f706546b3331765abf89ae9e6a8333e49621a641a next-id get -c "1001" --host "http://127.0.0.1:8443/api"`

2) ` jcli rest v0 tip get -h tcp://127.0.0.1:1/apixxxxxxxx`
3) `jcli rest v0 tip get -h http://127.0.0.100:1/api`

CodeSandwich:
This is a tip of a JCLI-wide problem of inconsistent error handling. The final goal is to make it fully panic-free  with components returning errors and a single point of handling them.

CodeSandwich:
@vincenthz Milestone week 1 is for removal of transaction panics

PR #335, opened 2019-05-08
Jcli no panic utils

CodeSandwich:
Prepare JCLI tooling for proper error handling. This is needed as part of https://github.com/input-output-hk/jormungandr/issues/298


src/bin/jcli.rs
```diff
@@ -19,8 +19,21 @@ extern crate strfmt;
...
+}
+
+fn report_error(error: Box<Error>) {
+    eprintln!("{}", error);
+    let mut source = error.source();
+    while let Some(sub_error) = source {
+        eprintln!("  |-> {}", sub_error);
+        source = sub_error.source();
+    }
+    std::process::exit(1)
```
NicolasDP: Excellent having this reported to the most top level.

src/bin/jcli_app/utils/io.rs
```diff
@@ -1,37 +1,48 @@
...
-            std::fs::OpenOptions::new()
-                .create(false)
-                .write(false)
-                .read(true)
-                .append(false)
-                .open(path)
-                .unwrap(),
-        ))
-    } else {
-        Box::new(std::io::BufReader::new(std::io::stdin()))
```
NicolasDP: This Diff seems like you changed the order of the functions... this is not necessary

src/bin/jcli_app/utils/io.rs
```diff
@@ -1,37 +1,48 @@
...
+pub fn open_file_write<P: AsRef<Path>>(path: &Option<P>) -> Result<impl Write, Error> {
+    match path {
+        Some(path) => {
+            let writer = File::create(path).map_err(|e| Error::new_open_file_write(path, e))?;
```
NicolasDP: I prefer the `OpenOptions` method. we see explicitly the options when opening the file. What is the rational for the change of function ?

CodeSandwich: There was a subtle bug in open_file_write: the truncate flag was not set. If we try to overwrite a file containing 20 bytes with 10 bytes of JCLI output, the file will still be 20 bytes, 10 new and 10 old. This bug would not be possible if we used the methods provided by std lib, which do what we want, but are guaranteed to be correct. Also the old version was unnecessarily noisy and verbose.

NicolasDP: 1. indeed the `truncate` option was not set, this was the bug, it should have been the only line changing here.
2. verbose is fine, especially if it allows to be precise, this was the case here. It was precise and it was easy to identify the error.
3. OpenOptions is the method provided by rust `std`.

Please revert change and fix the missing truncate and the error report as 2 different commit:

1. for the missing truncate;
2. one to add the Error

src/bin/jcli_app/utils/io.rs
```diff
@@ -1,37 +1,48 @@
...
+            Ok(Box::new(writer) as Box<Write>)
+        }
+        None => Ok(Box::new(stdout()) as Box<Write>),
     }
 }
+
+custom_error! { pub Error
+    OpenFileRead { path: String, source: IoError } = "Failed to open file '{path}' to read",
+    OpenFileWrite { path: String, source: IoError } = "Failed to open file '{path}' to write",
+}
```
mzabaluev: This looks like an unnecessarily unified type, because you'd hardly ever have a situation where both variants can be returned, yet it would not be served better with a more case-specific error to tell which file was not opened (e.g. when generating block 0 you'd have a `GenesisError::FailedToOpenSourceFile` and `GenesisError::FailedToOpenOutputFile`).

So throwing the specific case analysis to the upper level, you could return `std::io::Error` from both of these basic utility functions and write the path into the error message.

src/bin/jcli_app/utils/io.rs
```diff
@@ -1,37 +1,37 @@
...
                 .write(true)
                 .read(false)
                 .append(false)
-                .open(path)
-                .unwrap(),
-        )
-    } else {
-        Box::new(std::io::stdout())
+                .truncate(true)
+                .open(path)?;
```
mzabaluev: Good. You may want to check that the I/O error message captures the information so no `map_err` is needed to augment it with the file path.

CodeSandwich: I've checked it previously and no, it returns error which prints vague message that file does not exist. The error also has no source. Correct me if I'm wrong, but I think it could be slightly dangerous to depend on undocumented formatting of the message anyway, even if it's nothing serious.

mzabaluev: Yes, that's why there could be a `map_err` which will reuse the original error's code, but prepend the file path information to the message of the returned error.

CodeSandwich: That would make sense. Unfortunately the path is not captured, but that's not a problem, we'll just add path in higher-level errors.

mzabaluev: That's a solution to, no real need to encode it in errors here.

PR #337, opened 2019-05-09
unit tests for new & staging commands

dkijania:


dkijania:
Only few test cases added. feel free to comment if the general idea is ok. Or i should use some other approach

PR #346, opened 2019-05-14
Transaction witness module tests

dkijania:


tests/common/configuration/node_config_model.rs
```diff
@@ -89,7 +89,7 @@ impl NodeConfig {
 
 fn get_available_port() -> u16 {
     let available_port = loop {
-        let port = rand::thread_rng().gen_range(8000, 9999);
+        let port = rand::thread_rng().gen_range(6000, 9999);
```
CodeSandwich: I don't think that this change will fix the test unstability regarding port binding. This loop never fails and always returns a free port. If lack of ports was problem, it would hang forever and not panic.

I suspect that the real cause is that 2 threads generated the same port, validated it at different times, both succesfuly, then one started using it and the other got rejected. The time gap between validation and usage creates a race condition.

CodeSandwich: What I think we could do here is use the fact that every test binary is run sequentially in separate process and only tests functions are run in parallell threads. If we have a static (atomic!) counter with last used port, we could eliminate conflicts. Every process will then create it from a scratch with value 8000 (no conflict here, processes are run sequentially). During search for a free port it could be incremented on every try in every thread. We probably won't run to 9999 and if we do, we probably have a problem anyway and should panic anyway.

tests/jcli_transaction/e2e_transaction_tests.rs
```diff
@@ -1,43 +1,11 @@
-#![allow(dead_code)]
+#![cfg(feature = "integration-test")]
```
CodeSandwich: :heart:

tests/jcli_transaction/e2e_transaction_tests.rs
```diff
@@ -156,3 +121,52 @@ pub fn test_transaction_from_account_to_account_is_accepted_by_node() {
...
+
+    let transaction_message = JCLITransactionWrapper::new_transaction(&block0_hash)
+        .assert_add_input_from_utxo(&utxo)
+        .assert_add_output(&reciever.address, &transfer_amount)
+        .assert_finalize()
+        .seal_with_witness_deafult(&sender.private_key, "utxo")
+        .assert_transaction_to_message();
+
+    jcli_wrapper::assert_transaction_post_accepted(&transaction_message, &jormungandr_rest_address);
+}
```
CodeSandwich: Do we need delegation-to-delegation test?

tests/common/configuration/node_config_model.rs
mzabaluev: Could port number ranges be assigned by the test runner, so that they could not conflict between processes?

CodeSandwich: I'm not aware of such tool. Isn't it enough that processes don't run in parallel?

dkijania:
it looks like set readonly permission to a file in linux does not work as it should. See this:
https://doc.rust-lang.org/std/fs/fn.set_permissions.html

I'm still developing on windows (where it works ok), once i've setup vm with linux will try to debug a problem, because testing code on circleCI is not fast solution.

dkijania:
No issue with it but Isn't it enough to squash them at merge?

Issue #328, opened 2019-05-07
REST: Add an endpoint to gather the account state


vincenthz:
**Is your feature request related to a problem/context ? Please describe if applicable.**

it's useful to get the state of a specific account from the node, specifically the spending counter.

**Describe the solution you'd like**

the REST endpoint takes an account id, then try to find it in the current ledger, then return some of the content (at the moment everything) of the mockchain::account::State structure in JSON

CodeSandwich:
`mockchain::account::State` contains 2 fields: value and spending counter. Value is obviously a valuable piece of information, but are users interested in getting the counter?

PR #352, opened 2019-05-15
Add account state getter to REST API

CodeSandwich:
This is core functionality of https://github.com/input-output-hk/jormungandr/issues/328


CodeSandwich:
This PR may not be valid, it makes REST API accept account address encoded as bech32 `ed25519e_pk`, it probably should be `ta` or `ca`

PR #356, opened 2019-05-16
Added retry for jormungandr startup if port is not available

dkijania:


tests/common/configuration/node_config_model.rs
```diff
@@ -72,25 +72,33 @@ impl NodeConfig {
...
+        };
+        available_port
+    }
 
-fn port_is_available(port: u16) -> bool {
-    match TcpListener::bind(("127.0.0.1", port)) {
-        Ok(_) => true,
-        Err(_) => false,
+    fn port_is_available(port: u16) -> bool {
+        match TcpListener::bind(("127.0.0.1", port)) {
```
mzabaluev: If the listener is dropped immediately, the port is released, so there is a race condition.

dkijania: I changed it to bind ot 0 port which binds to random unoccupied port.

tests/common/configuration/node_config_model.rs
```diff
@@ -86,19 +86,7 @@ impl NodeConfig {
...
-        available_port
-    }
-
-    fn port_is_available(port: u16) -> bool {
-        match TcpListener::bind(("127.0.0.1", port)) {
-            Ok(_) => true,
-            Err(_) => false,
-        }
+        let tcp_listener = TcpListener::bind(("127.0.0.1", 0)).unwrap();
+        tcp_listener.local_addr().unwrap().port()
```
mzabaluev: I'm afraid this is even worse than random: the listener is immediately dropped, meaning that the ephemeral port is up for grabs again and the OS might very well recycle it for a concurrent test.

tests/common/configuration/node_config_model.rs
mzabaluev: What I meant is that we can change jormungandr to bind gRPC to a random port. The problem is how to discover that port in a test after starting jormungandr.

Another way is to hand an open, pre-bound socket descriptor down to jormungandr with a particular option. IIRC this is also how systemd loves to start daemons on demand. But I only know how to do this on Unix.

tests/common/configuration/node_config_model.rs
```diff
@@ -72,25 +77,20 @@ impl NodeConfig {
...
-    };
-    available_port
-}
 
-fn port_is_available(port: u16) -> bool {
-    match TcpListener::bind(("127.0.0.1", port)) {
-        Ok(_) => true,
-        Err(_) => false,
+    fn get_available_port() -> u16 {
+        NEXT_AVAILABLE_PORT_NUMBER.fetch_add(1, Ordering::SeqCst)
```
mzabaluev: Much better. Assuming that there is only one test process at a time handing out port numbers. Otherwise there is a running race for ports starting predictably from 8000.

tests/common/configuration/secret_model.rs
```diff
@@ -23,6 +23,10 @@ impl SecretModel {
         node_config_file_path
     }
 
+    pub fn empty() -> SecretModel {
```
mzabaluev: OK, though it's more conventional to mark constructor-like methods with `Self` in the return type.

tests/common/startup/jormungandr_starter.rs
```diff
@@ -0,0 +1,102 @@
...
+use common::jormungandr_wrapper;
+use common::process_utils;
+use common::process_utils::{
+    output_extensions::ProcessOutput, process_guard::ProcessKillGuard, ProcessError,
+};
+
+use std::process::{Command, Output};
+
+custom_error! {pub StartupError
+    JormungandrNotLaunched{ source: ProcessError } = "could not start jormungandr",
```
mzabaluev: Solid, with a room to add more causes later on.

tests/common/startup/jormungandr_starter.rs
```diff
@@ -0,0 +1,102 @@
...
+}
+
+fn try_to_start_jormungandr_node(
+    rest_address: &str,
+    command: &mut Command,
+) -> Result<ProcessKillGuard, StartupError> {
+    println!("Starting jormungandr node...");
+    let process = command
+        .spawn()
+        .expect("failed to execute 'start jormungandr node'");
```
mzabaluev: This catches something totally unexpected like failure of exec call or missing the executable, right?

tests/common/startup/jormungandr_starter.rs
```diff
@@ -0,0 +1,102 @@
...
+    let guard = ProcessKillGuard::new(process, String::from("Jormungandr node"));
+
+    let proces_start_result = process_utils::run_process_until_response_matches(
+        jcli_wrapper::jcli_commands::get_rest_stats_command(&rest_address),
+        &is_node_up,
+        2,
+        5,
+        "get stats from jormungandr node",
+        "jormungandr node is not up",
+    );
```
mzabaluev: The signature of this function is problematic: six parameters with no idea of what most of them are from the invoking code. Consider making some parameter structs or a builder API to improve readability, out of scope of this PR though.

tests/common/startup/jormungandr_starter.rs
```diff
@@ -0,0 +1,102 @@
...
+
+    println!("Starting node with configuration : {:?}", &config);
+    let process = start_jormungandr_node_sync_with_retry(&rest_address, &mut command, &mut config);
+    process
+}
+
+pub fn start_jormungandr_node_as_leader(mut config: &mut JormungandrConfig) -> ProcessKillGuard {
+    let rest_address = &config.node_config.get_node_address();
+
+    let mut command = jormungandr_wrapper::get_start_jormungandr_as_leader_node_command(
```
mzabaluev: Some of these utility functions might even take the whole config struct by reference instead of receiving its contents piecemeal.

tests/common/configuration/node_config_model.rs
mzabaluev: Nitpick: it better remain a freestanding function, as there is nothing actually connecting it to `NodeConfig`

tests/common/startup/jormungandr_starter.rs
```diff
@@ -0,0 +1,102 @@
...
+
+fn start_jormungandr_node_sync_with_retry(
+    rest_address: &str,
+    command: &mut Command,
+    config: &mut JormungandrConfig,
+) -> ProcessKillGuard {
+    let first_attempt = try_to_start_jormungandr_node(rest_address, command);
+    match first_attempt {
+        Ok(guard) => return guard,
+        _ => (),
```
mzabaluev: Might want to report the failure?

mzabaluev:
Please rebase and make sure your `Cargo.lock` is up to date.

tests/common/process_utils/mod.rs
```diff
@@ -146,10 +153,13 @@ pub fn run_process_until_response_matches<F: Fn(Output) -> bool>(
...
-            &error_description, &max_attempts, &timeout
-        );
+        return Err(ProcessError::ProcessExited {
+            message: format!(
+                "{} (tried to connect {} times with {} s interval)",
+                &error_description, &max_attempts, &timeout
+            ),
+        });
     }
     println!("Success: {}", &command_description);
```
mzabaluev: `command_description` is already a reference, as is `error_description` above. Does not matter, it's transparent to the `Display` impl.

tests/common/startup/configuration_builder.rs
```diff
@@ -0,0 +1,67 @@
...
+impl ConfigurationBuilder {
+    pub fn new() -> Self {
+        ConfigurationBuilder {
+            funds: vec![],
+            trusted_peers: None,
+            block0_hash: None,
+        }
+    }
+
+    pub fn with_funds<'a>(&'a mut self, funds: Vec<Fund>) -> &'a mut Self {
```
mzabaluev: Good. Conventionally, there's no need for `with_` prefix in builder-style methods, just the names like members.

mzabaluev: You can omit the lifetime parameter in this case, it is inferred.

tests/common/startup/mod.rs
```diff
@@ -1,109 +1,32 @@
...
-    );
+use std::path::PathBuf;
 
-    config.genesis_block_path = path_to_output_block.clone();
-    println!("Starting node with configuration : {:?}", &config);
+pub use self::configuration_builder::ConfigurationBuilder;
 
-    let process = start_jormungandr_node_and_wait(&rest_address, command);
-    process
+pub fn start_jormungandr_node(mut config: &mut JormungandrConfig) -> ProcessKillGuard {
```
mzabaluev: Why does the parameter have to be mutable?

tests/jcli_rest_test.rs
```diff
@@ -57,7 +60,7 @@ pub fn test_correct_utxos_are_read_from_node() {
 #[test]
 #[cfg(feature = "integration-test")]
 pub fn test_correct_id_is_returned_for_block_tip_if_only_genesis_block_exists() {
-    let mut config = JormungandrConfig::new();
+    let mut config = startup::ConfigurationBuilder::new().build();
```
mzabaluev: This could still be `JormungandrConfig::new` as a shorthand.

PR #357, opened 2019-05-16
Improved block solicitation

mzabaluev:
Streamlined the code a great deal.

### DIFF
--- a/src/network/grpc/client.rs
+++ b/src/network/grpc/client.rs
@@ -103,7 +103,7 @@ fn process_block_solicitations(
                 })
             })
             .map_err(|e| {
-                info!("block solicitation failed: {:?}", e);
+                warn!("block solicitation failed: {:?}", e);
             }),
     );
 }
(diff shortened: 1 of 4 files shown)


src/network/grpc/client.rs
```diff
@@ -80,12 +76,38 @@ fn subscribe(
...
+                client.get_blocks(&block_ids).and_then(move |blocks| {
+                    let mut block_box = block_box.clone();
+                    blocks.for_each(move |block| {
+                        block_box.send(BlockMsg::NetworkBlock(block));
+                        Ok(())
+                    })
+                })
+            })
+            .map_err(|e| {
+                info!("block solicitation failed: {:?}", e);
```
NicolasDP: a `warn` instead maybe? This is not really an error, i agree. But not being able to solicit a block might be problematic, and this is a warning to something that may become a problem down the line (not being able to sync the blockchain with the rest of the world).
Better having this already well sorted.

Issue #349, opened 2019-05-14
Logging improvements


NicolasDP:
need to improve the logging usage in `jormungandr`. Utilise `slog` with the `slog::Logger` instead of the per thread global variables;

PR #366, opened 2019-05-22
Slog macro

CodeSandwich:
Migrates away from global logger to local, contextualizable loggers which are directly used with slog macros. Fixes https://github.com/input-output-hk/jormungandr/issues/349


src/network/p2p/comm.rs
```diff
@@ -319,18 +324,29 @@ impl PeerMap {
         }
     }
 
-    pub fn solicit_blocks(&self, node_id: topology::NodeId, hashes: Vec<HeaderHash>) {
+    pub fn solicit_blocks(
+        &self,
+        node_id: topology::NodeId,
+        hashes: Vec<HeaderHash>,
+        logger: &Logger,
```
mzabaluev: Likewise, no logger in methods, put it in the host object.

src/network/p2p/topology.rs
```diff
@@ -138,35 +138,43 @@ impl P2pTopology {
 
     /// Returns a list of neighbors selected in this turn
     /// to contact for event dissemination.
-    pub fn view(&self) -> impl Iterator<Item = Node> {
+    pub fn view(&self, logger: &Logger) -> impl Iterator<Item = Node> {
```
mzabaluev: Already commented on this above, there should be `self.logger` for this.

src/network/service.rs
```diff
@@ -21,23 +21,27 @@ use network_core::{
...
-
-use futures::future::{self, FutureResult};
-use futures::prelude::*;
+use slog::Logger;
 
 #[derive(Clone)]
 pub struct NodeService {
     channels: Channels,
     global_state: GlobalStateR,
+    logger: Logger,
```
mzabaluev: Yeah, that's more like it

src/network/service.rs
```diff
@@ -21,23 +21,27 @@ use network_core::{
...
 }
 
 impl NodeService {
-    pub fn new(channels: Channels, global_state: GlobalStateR) -> Self {
+    pub fn new(channels: Channels, global_state: GlobalStateR, logger: Logger) -> Self {
```
mzabaluev: If `GlobalState` has a logger member, the constructor could make its own clone and add a "server" tag, no need for an extra parameter.

src/network/service.rs
```diff
@@ -21,23 +21,27 @@ use network_core::{
...
         NodeService {
             channels,
             global_state,
+            logger,
         }
     }
+
+    pub fn logger(&self) -> &Logger {
```
mzabaluev: :+1:

src/network/service.rs
```diff
@@ -193,7 +198,7 @@ impl ContentService for NodeService {
 
     fn message_subscription<S>(
         &mut self,
-        subscriber: Self::NodeId,
+        _subscriber: Self::NodeId,
```
mzabaluev: Eh, I've been keeping this so the warning reminds me that we have to add an implementation at some point. Please remove this completely unrelated change :)

src/network/service.rs
```diff
@@ -203,6 +208,8 @@ impl ContentService for NodeService {
     }
 }
 
+// NodeService implements GossipService, which has blanket impl for every Node
```
mzabaluev: I don't get this comment. What blanket impl is this about?

src/settings/logging.rs
```diff
@@ -44,25 +33,19 @@ impl FromStr for LogFormat {
...
                 let drain = slog_json::Json::default(std::io::stderr()).fuse();
-                let drain = slog::LevelFilter::new(drain, self.verbosity).fuse();
-                let drain = slog_async::Async::new(drain).build().fuse();
-                slog::Logger::root(drain, o!())
+                slog_async::Async::new(drain).build().fuse()
             }
         };
-        log_wrapper::logger::set_global_logger(log);
+        let drain = slog::LevelFilter::new(drain, self.verbosity).fuse();
+        slog::Logger::root(drain, o!())
```
mzabaluev: I like this code. `o!()`

src/network/client.rs
```diff
@@ -54,19 +55,23 @@ where
         let mut peer_comms = PeerComms::new();
         let block_req = service.block_subscription(peer_comms.subscribe_to_block_announcements());
         let gossip_req = service.gossip_subscription(peer_comms.subscribe_to_gossip());
+        let err_logger = global_state.logger().clone();
```
mzabaluev: This and the instance in the `and_then` closure below could be augmented with the connection address. At the end the logger that goes inside the `Client` would further clone one of these instances to add the node ID, so we'll get both the remote address and the node ID in the context of the client logger.

src/network/mod.rs
```diff
@@ -151,40 +159,53 @@ pub fn run(config: Configuration, input: MessageQueue<NetworkMsg>, channels: Cha
 
     let addrs = global_state
         .topology
-        .view()
+        .view(&logger)
         .filter_map(|paddr| paddr.address())
         .collect::<Vec<_>>();
     let state = global_state.clone();
     let conn_channels = channels.clone();
+    let conn_logger = logger.clone();
```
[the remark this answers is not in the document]
mzabaluev: I thought we'd put this one inside `ConnectionState`, and now there is some redundancy?

src/network/mod.rs
```diff
@@ -156,20 +172,25 @@ pub fn run(config: Configuration, input: MessageQueue<NetworkMsg>, channels: Cha
...
     let connections = stream::iter_ok(addrs).for_each(move |addr| {
-        info!("connecting to initial gossip peer at {}", addr);
+        info!(conn_logger, "connecting to initial gossip peer at {}", addr);
         let peer = Peer::new(addr, Protocol::Grpc);
         let conn_state = ConnectionState::new(state.clone(), &peer);
         let state = state.clone();
+        let client_logger = conn_logger.clone();
```
mzabaluev: Likewise, there is a logger in `Client` now, so this one is not needed?

src/network/mod.rs
```diff
@@ -283,16 +309,18 @@ fn connect_and_propagate_with<F>(
                 Ok(()) => (),
                 Err(e) => {
                     info!(
+                        state.logger(),
```
mzabaluev: This more properly belongs to the `conn_state` logger so that we'd have the address context?

src/network/mod.rs
```diff
@@ -283,16 +309,18 @@ fn connect_and_propagate_with<F>(
...
                     );
                     return;
                 }
             }
         } else {
             info!(
-                "peer at {} responded with different node id: {}",
-                addr, connected_node_id
+                state.logger(),
+                "peer at {} responded with different node id: {}", addr, connected_node_id
```
mzabaluev: Similarly here, use the per-connection logger.

src/network/mod.rs
```diff
@@ -266,12 +289,15 @@ fn connect_and_propagate_with<F>(
...
+            info!(
+                state.logger(),
+                "ignoring P2P node without an IP address: {:?}", node
+            );
             return;
         }
     };
     let node_id = node.id();
-    debug!("connecting to node {} at {}", node_id, addr);
+    debug!(state.logger(), "connecting to node {} at {}", node_id, addr);
```
mzabaluev: A `ConnectionState` is created two lines below, so maybe it should go under per-connection logger, and the "at {}" part of the message is no longer needed.

src/network/mod.rs
```diff
@@ -309,17 +337,17 @@ fn first_trusted_peer_address(config: &Configuration) -> Option<SocketAddr> {
...
-pub fn bootstrap(config: &Configuration, blockchain: BlockchainR) {
+pub fn bootstrap(config: &Configuration, blockchain: BlockchainR, logger: &Logger) {
     if config.protocol != Protocol::Grpc {
         unimplemented!()
     }
     match first_trusted_peer_address(config) {
         Some(address) => {
             let peer = Peer::new(address, Protocol::Grpc);
-            grpc::bootstrap_from_peer(peer, blockchain)
+            grpc::bootstrap_from_peer(peer, blockchain, logger)
```
mzabaluev: Make a clone with the peer address and a "bootstrap" task tag?

src/network/mod.rs
```diff
@@ -328,15 +356,19 @@ pub fn bootstrap(config: &Configuration, blockchain: BlockchainR) {
...
+) -> Result<Block, FetchBlockError> {
     if config.protocol != Protocol::Grpc {
         unimplemented!()
     }
     match first_trusted_peer_address(config) {
         None => Err(FetchBlockError::NoTrustedPeers),
         Some(address) => {
             let peer = Peer::new(address, Protocol::Grpc);
-            grpc::fetch_block(peer, hash)
+            grpc::fetch_block(peer, hash, logger)
```
mzabaluev: This should be a cloned and augmented instance, similarly to the bootstrap task above.

src/network/subscription.rs
```diff
@@ -32,15 +32,16 @@ pub fn process_gossip<S>(inbound: S, state: GlobalStateR) -> tokio::executor::Sp
 where
     S: Stream<Item = Gossip<Node>, Error = core_error::Error> + Send + 'static,
 {
+    let err_logger = state.logger().clone();
```
mzabaluev: Now this function could receive the logger from the caller, similarly to `process_block_announcements`. Both server and client tasks can use this utility function, so we'd be interested to see which connection the activity is related to.

src/network/client.rs
CodeSandwich: Good point. I've solved it slightly differently, instead of accepting `GlobalState` now it accepts `ConnectionState`, which has logger with address already there.

src/network/mod.rs
CodeSandwich: It's already done in `main::bootstrap`, line 224. Thanks to that all actions in bootstrapping phase have this tag. It's insanely hard to follow, we have to think about some way to make it clearer.

mzabaluev: Right, thanks.

src/network/mod.rs
CodeSandwich: Will fix it like bootstrap, in `main::initialize_node`

src/network/client.rs
```diff
@@ -48,29 +49,31 @@ where
 {
     fn subscribe(
         mut service: S,
-        global_state: GlobalStateR,
+        state: ConnectionState,
```
mzabaluev: :+1:

PR #367, opened 2019-05-22
Split testes into smaller pieces

dkijania:
Since there are more and more integration tests in jormungandr node, this PR is to reorganize tests and split them into smaller parts


dkijania:
I realize that this PR is not easy to review, so my recommendation is to open code i some IDE and analyze test folder structure

Issue #65, opened 2018-12-11
Configuration: logging options


NicolasDP:
We need to extend the logging option in the configuration file:

* [x] where to dump the logs;
* [x] the verbosity level of the logs;

I can see cases where we might want to have the logs redirected to syslog but also the warning and the error ones to be on the standard error output too.

might need to sync with #55

NicolasDP:
we need to finish the _where to output the logs_ part

CodeSandwich:
@NicolasDP Here's my propositon:
# Current design
## User input
- CMD provides verbosity in form of `-vvv` (default 0 a.k.a. `info`) and optional log format
- Config provides optional verbosity and optional log format
## Logger setup
- If config provides verbosity, it overwrites CMD verbosity
- If CMD does not provide log format, it's set to `plain`
- Logger uses CMD verbosity and log format
## Consequences
- Only 1 logging output
- Config log format is always ignored

# Design proposition
## User input
- CMD provides verbosity in form of `-vvv` (default 0 a.k.a. info) and optional log format
- Config provides **list** of structures with mandatory verbosity and log format, may be empty
## Logger setup
- Convert config list to list of loggers
- If CMD provides log format, create logger and append to list
- If list is still empty, append logger with CMD verbosity and `plain` format
- Create logger which dispatches logs to each logger using slog::Duplicate
## Consequences
- User can log to multiple outputs, some may have different logging levels from others
- Config log format is always respected
- Sane default - same as in current solution
## ALSO log format
Log format should be upgraded from `LogFormat`
- `plain` (implicit written to stderr)
- `json` (implicitly written to stderr)

to `LogOutput`
- `stderr` (same as `plain`)
- `stderr_json` (same as `json`)
- `syslog`

gufmar:
From a SysOps point of view and considering multi proxy-node setups it would be very useful to have syslog as an option.

As all the GRPC communication is stateless and happening to multiple neighbors in parallel I wonder if and how the logfiles can be written in order to improve readability. 
If there is a node-internal "session ID" valid for a certain task (who is a sequence of multiple stateless messages) it would be useful to have it in the logs. If there is no such internal ID, not even a sequential number counted up by the logger process, I have seen loggers in the past who sum up all messages for one process in the RAM and write it to the logfile at once (one start time, + x seconds since ...)

Also please consider to log all startup parameters only once (at startup) and don't repeat redundant information in many lines. Eg `allow_accounts: true, consensus-version: bft`

[...]

PR #369, opened 2019-05-24
Add syslog and journald logging outputs

CodeSandwich:
This is first part of solution from https://github.com/input-output-hk/jormungandr/issues/65. It renames and extends output formats.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,8 +25,9 @@ serde_json = "1.0.38"
 serde_yaml = "0.8"
 slog = "^2.4"
 slog-async = "2.3.0"
-slog-term = "2.4.0"
+slog-journald = { version = "2.0.0", optional = true }
 slog-json = "2.3.0"
+slog-term = "2.4.0"
 humantime = "1.2"
 bincode = "1.0.1"
 cbor_event = "2.1.2"
@@ -70,6 +71,11 @@ version = "2.32"
 default-features = false
 features = [ "suggestions", "color", "wrap_help" ]
 
+[target.'cfg(unix)'.dependencies]
+# Crates.io version does not work on NetBsd due to dependency on outdated `nix` crate
+# TODO switch to crates.io when fixed version is released
+slog-syslog = { git = "https://github.com/slog-rs/syslog", rev = "336da50ed86058bfa87d9b497f5921fb7f95ea42" }
+
 [dev-dependencies]
 quickcheck = "0.8"
 galvanic-test = "0.2.0"
@@ -82,3 +88,4 @@ custom_error = "1.6"
 [features]
 with-bench = []
 integration-test = []
+systemd = ["slog-journald"]
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -128,6 +128,33 @@ jobs:
           paths:
             - release/*
 
+  build_release_all_features:
+    docker:
+      - image: rust:latest
+    working_directory: /mnt/crate
+    steps:
+      - run:
+          name: Install systemd dev tools
+          command: apt update && apt install -y libsystemd-dev
+      - checkout
+      - run: git submodule sync
+      - run: git submodule update --init
+      - restore_cache:
+          keys:
+            - cargo-v3-{{ checksum "Cargo.toml" }}-{{ checksum "Cargo.lock" }}
+            - cargo-v3-{{ checksum "Cargo.toml" }}
+            - cargo-v3-
+      - run:
+          name: Print version information
+          command: rustc --version; cargo --version
+      - run:
+          name: Build in release profile
+          command: cargo build --release --features systemd --frozen
+      - persist_to_workspace:
+          root: target
+          paths:
+            - release/*
+
   test_release:
     docker:
       - image: rust:latest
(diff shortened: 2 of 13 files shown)


.travis.yml
```diff
@@ -1,7 +1,7 @@
 # Based on the "trust" template v0.1.2
 # https://github.com/japaric/trust/tree/v0.1.2
 
-dist: trusty
```
mzabaluev: Ooh, travis has rolled out a newer environment at last.

Cargo.toml
```diff
@@ -70,6 +70,12 @@ version = "2.32"
 default-features = false
 features = [ "suggestions", "color", "wrap_help" ]
 
+[target.'cfg(unix)'.dependencies]
+slog-syslog = "0.11.0"
```
mzabaluev: Can you check this works on Mac OS X?

ci/install.sh
```diff
@@ -42,6 +45,17 @@ main() {
            --git japaric/cross \
            --tag $tag \
            --target $target
+
+    if [ $TRAVIS_OS_NAME = linux ]; then
+        echo "AM INSTALLING"
+        sudo apt update
+        sudo apt install -y libsystemd-dev
```
mzabaluev: Interesting how on linux we target musl and then link to libsystemd, which is installed as a shared library? Have you tested that it works in CI?

ci/install.sh
```diff
@@ -42,6 +45,17 @@ main() {
            --git japaric/cross \
            --tag $tag \
            --target $target
+
+    if [ $TRAVIS_OS_NAME = linux ]; then
+        echo "AM INSTALLING"
```
mzabaluev: Could these messages be more human readable: no all caps, full English sentences?

.travis.yml
CodeSandwich: They had to, 14.04 officially lost support almost a month ago :D

ci/install.sh
CodeSandwich: This is only temporary and for me so I know what's going in logs

ci/install.sh
CodeSandwich: Good point, I'll write a few smoke tests which will try to write anything to each of supported outputs

CodeSandwich:
@mzabaluev @NicolasDP @vincenthz  This PR is still a WIP. I had to open it so CI would cross-run the tests. It's not passing yet, sorry for the noise.

mzabaluev:
OK, please squash it into better annotated comments when done.

vincenthz:
hmm I'm not sure this is wise to fence the systemd support behind os=linux. it seems to me it should be a independent feature (only available on os=linux) that someone can turn on and off. And also that would prevent this massive test change, as only 1 extra target would be required (the one that activated systemd feature)

CodeSandwich:
@vincenthz You're right, I should've thought about it earlier, not every Linux has Systemd. We can keep tests with journald on CircleCI and stop fighting with Cross on Travis.

Do you think we need to double CircleCI tests and have versions with and without journald? I think that it will be enough to have Travis run tests without journald.

CodeSandwich:
The `systemd` feature will be available on all targets. Right now it's not possible to restrict features to specific targets: https://github.com/rust-lang/cargo/issues/1197.

NicolasDP:
and this is fine. it will be the user's problem. If they want to activate `systemd` and it is not available on their machine: then this is their problem. Make it a crate feature and remove the `os = linux` guards.

NicolasDP:
also, once this is done, remove your commit with the changes in the CI scripts. we don't need `systemd` enabled in these for now.

CodeSandwich:
From Travis - of course, they didn't work anyway. But for CircleCI we probably could keep them? It's already working and the change is simple. It's good to have a feature tested on CI.

NicolasDP:
yes, and that would be in a new circle ci test, depending on `build_release`, called `build_release_all_features` and would do the necessary tests with all the features enabled. But not on by default.

ci/install.sh
mzabaluev: Check carefully. As far as I understand, the dynamic libraries packaged for Debian are built with the glibc toolchain, so they require ldd and depend on a dynamically loaded libc.so. There may even be a subtle dependency on runtime state managed in the single dynamically linked runtime library. Adding a statically linked musl libc alongside that would break it, even if it is possible.

The [musl FAQ](https://www.musl-libc.org/faq.html) has this to say:

> At present, some glibc-linked shared libraries can be loaded with musl, but all but the simplest glibc-linked applications will fail if musl is dropped-in in place of `/lib/ld-linux.so.2`.

mzabaluev: Also static libraries built for glibc may require stuff from libc that is not standard and is not present in musl.

CodeSandwich:
I'm more and more thinking about completely removing different logger kinds smoke tests. The CI systems don't have syslog and it's not easy to install it. It's arous inside Docker inside CircleCI, but seemingly impossible inside VM managed by cross inside Travis. Are tests checking if logger can log without panicking very precious?

vincenthz:
Ok, I think this PR has run its course, let's restart with a simpler path:

1) move away the start_up error refactoring into its own PR. that could go in straight away; note, if it was a separated commit, you could just cherry-pick it out of this branch,
2) let's *keep* `LogFormat` and introduce a `LogOutput`. I think it's perfectly valid to have JSON going to stdout, or plain text going to systemd, or whichever other product of configuration. Since both syslog and systemd are not straightforward to support, let's put it behind feature flag for now
3) let's pun the test for later

CodeSandwich:
1. The `start_up::Error` isn't refactored. Its `LoggingInitializationError` is extended with `logging::Error` field. The same commit creates said `logging::Error`, which is used only during syslog logger creation. I can move it to separate commit including the creation of never-type-ish `logging::Error` if that's going to make it easier to review.
2. The output and output format are inseparable. We can't push JSON or plain text to syslog, it has its own custom format. We can't output syslog format to stderr either.
3. I think that logger tests should not be run on the CI, because the configuration is killing us. I proposed complete removal of logger tests, because they don't make much sense outside of CI either. If we agree to do that, this PR will become much cleaner and ready to be merged almost right away.

mzabaluev:
> The output and output format are inseparable. We can't push JSON or plain text to syslog, it has its own custom format. We can't output syslog format to stderr either.

I agree, to an extent. JSON is only workable with plain file descriptor based outputs. If you want to be really flexible, there can be a formatter that can optionally augment a file-based log sink, but it will be useless on syslog, and journald would actually make good use of the key-value pairs from the messages.

vincenthz:
I'm not particularly concerned whether or not this is a good idea to combine certain things, but there's always a way to flatten JSON into a string, which syslog can accept without a problem. With a bit of flexibility in the interface between Output and Format, you can make sure structure construction get passed as is for best effect, or indeed flatten into a string, if no such thing is possible

mzabaluev:
In the command line `--log-format=json --log-output=stdout` certainly looks better than the special case `stdout_json`. Maybe orthogonality is OK and if you use strange combinations of options, you get strange results.

PR #370, opened 2019-05-27
Allow starting the node with multiple leaders

edolstra:
This allows the `--secret` flag to be specified multiple times in order to start a node with multiple BFT leaders, which is primarily useful for testing.

### DIFF
--- a/doc/advanced/02_starting_bft_blockchain.md
+++ b/doc/advanced/02_starting_bft_blockchain.md
@@ -80,3 +80,6 @@ $ jormungandr --genesis-block block-0.bin \
     --config example.config \
     --secret node_secret.yaml
 ```
+
+It's possible to use the flag `--secret` multiple times to run a node
+with multiple leaders.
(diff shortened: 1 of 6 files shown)


doc/advanced/02_starting_bft_blockchain.md
```diff
@@ -80,3 +80,6 @@ $ jormungandr --genesis-block block-0.bin \
     --config example.config \
     --secret node_secret.yaml
 ```
+
+It's possible to use the flag `--secret` multiple times to run a node
+with multiple leaders.
```
NicolasDP: I think we ought to balance pro and cons a little bit here. Especially that in production this is not recommended to have all your _stakepools_' secret key in the same machine, in the same process. This is mainly interesting for testing the blockchain behaviour with multiple stakepool without having to setup a whole network of nodes.

vincenthz: This is not going to be the final mechanism, so we can skip putting too much details here for now

PR #371, opened 2019-05-27
Added jormungandr log handler

dkijania:


tests/common/jormungandr/commands.rs
```diff
@@ -65,7 +68,22 @@ pub fn get_start_jormungandr_as_passive_node_command(
...
 }
+
+#[cfg(target_os = "windows")]
+fn get_stdio_from_log_file(log_file_path: &PathBuf) -> std::process::Stdio {
+    use std::os::windows::io::{FromRawHandle, IntoRawHandle};
+    let file = File::create(log_file_path).expect("couldn't create log file for jormungandr");
+    unsafe { Stdio::from_raw_handle(file.into_raw_handle()) }
+}
+
+#[cfg(not(target_os = "windows"))]
```
mzabaluev: That should better be `#[cfg(unix)]`. If we come across a platform that is neither Windows nor has `std::os::unix`, we should know we need to create a `get_stdio_from_log_file` for it :)

tests/common/jormungandr/logger.rs
```diff
@@ -0,0 +1,34 @@
...
+    pub fn get_lines_with_error(&self) -> Vec<String> {
+        let lines = self.get_lines_from_log();
+        lines
+            .iter()
+            .filter(|n| n.to_uppercase().contains("ERROR"))
+            .cloned()
+            .collect::<Vec<String>>()
+    }
+
+    fn get_lines_from_log(&self) -> Vec<String> {
```
mzabaluev: This could return an `impl Iterator<Item = String>` instead of buffering everything internally, only to filter it later into another vector in `get_lines_with_error`.

tests/common/jormungandr/logger.rs
```diff
@@ -0,0 +1,34 @@
...
+pub struct JormungandrLogger {
+    pub log_file_path: PathBuf,
+}
+
+impl JormungandrLogger {
+    pub fn new(log_file_path: PathBuf) -> Self {
+        JormungandrLogger { log_file_path }
+    }
+
+    pub fn get_lines_with_error(&self) -> Vec<String> {
```
mzabaluev: Similarly, this could return `impl Iter<String>` and let the caller decide whether to buffer the data or process it iteratively.

tests/common/jormungandr/logger.rs
```diff
@@ -0,0 +1,34 @@
...
+        let lines = self.get_lines_from_log();
+        lines
+            .iter()
+            .filter(|n| n.to_uppercase().contains("ERROR"))
```
mzabaluev: It would be better and faster to use a regex, which 1) can be case-insensitive without you needing to fiddle with the string; 2) can match a particular location in the line, like `^ERROR`.

mzabaluev: In fact, we have a machine-readable output format, `stderr_json`. Why not use that and parse the severity level as a JSON field?

tests/common/jormungandr/process.rs
```diff
@@ -19,7 +19,7 @@ impl JormungandrProcess {
     }
 
     pub fn assert_no_erros_in_log(&self) {
```
mzabaluev: Just noticed typo: errors

tests/common/jormungandr/logger.rs
```diff
@@ -0,0 +1,51 @@
...
+        ));
+        entry.level == "ERROR"
+    }
+
+    fn get_lines_from_log(&self) -> impl Iterator<Item = String> {
+        let file = File::open(self.log_file_path.clone()).unwrap();
+        let mut data: Vec<String> = Vec::new();
+        let reader = BufReader::new(file);
+
+        for (_index, line) in reader.lines().enumerate() {
```
mzabaluev: Why `enumerate` if the index is not used?

tests/common/jormungandr/logger.rs
```diff
@@ -0,0 +1,51 @@
...
+            data.push(line.unwrap());
+        }
+        data.into_iter()
```
mzabaluev: Try just returning `reader.lines().map(|line| line.unwrap())`, no need for an intermediate vector.

tests/common/startup/mod.rs
```diff
@@ -11,37 +10,34 @@ use common::data::{
...
-        &mut config,
-        expected_message_part,
-    )
+    starter::assert_start_jormungandr_node_as_passive_fail(&mut config, expected_message_part)
 }
 
-pub fn start_jormungandr_node_as_passive(mut config: &mut JormungandrConfig) -> ProcessKillGuard {
-    jormungandr_starter::start_jormungandr_node_as_passive(&mut config)
+pub fn start_jormungandr_node_as_passive(mut config: &mut JormungandrConfig) -> JormungandrProcess {
+    starter::start_jormungandr_node_as_passive(&mut config)
```
mzabaluev: All these functions just delegate to the other module. Why not export those functions directly? You can use `pub use starter::start_jormungandr_node` to re-export.

mzabaluev: It's weird to get a mutable reference to a config structure and then pass it under another layer of mutable reference. What's that for?

tests/common/jormungandr/starter.rs
```diff
@@ -76,53 +81,57 @@ fn is_node_up(output: Output) -> bool {
     }
 }
 
-pub fn start_jormungandr_node(mut config: &mut JormungandrConfig) -> ProcessKillGuard {
+pub fn start_jormungandr_node(mut config: &mut JormungandrConfig) -> JormungandrProcess {
```
mzabaluev: Why this mutability fest for `JormungandrConfig`? Should the function modify the configuration for later use?

PR #374, opened 2019-05-28
Corner cases for jcli tests

dkijania:
Signed-off-by: dkijania <dariusz.kijania@iohk.io>


tests/jcli/address/info.rs
```diff
@@ -0,0 +1,64 @@
...
+
+#[test]
+pub fn test_info_delegation_address() {
+    let private_key = jcli_wrapper::assert_key_generate("ed25519Extended");
+    let public_key = jcli_wrapper::assert_key_to_public_default(&private_key);
+
+    let private_key = jcli_wrapper::assert_key_generate("ed25519Extended");
+    let delegation_key = jcli_wrapper::assert_key_to_public_default(&private_key);
+    let account_address =
+        jcli_wrapper::assert_address_delegation_for_testing(&public_key, &delegation_key);
```
mzabaluev: Could have used some profile data structures to select between testing and production, rather than functions with duplicate code.

tests/jcli/address/single.rs
```diff
@@ -85,9 +132,10 @@ pub fn test_delegation_address_made_of_random_string() {
 
     // Assertion changed due to issue #306. After fix please change it to correct one
     process_assert::assert_process_failed_and_contains_message(
-        jcli_wrapper::jcli_commands::get_address_delegation_command_default(
+        jcli_wrapper::jcli_commands::get_address_delegation_command(
             &public_key,
             &delegation_key,
+            true,
```
mzabaluev: The boolean parameter is not informative. Invent an enum with descriptive values?

tests/common/jcli_wrapper/mod.rs
```diff
@@ -15,6 +15,12 @@ use super::process_utils::output_extensions::ProcessOutput;
 use std::collections::BTreeMap;
 use std::path::PathBuf;
 
+#[derive(PartialEq)]
+pub enum Discrimination {
```
mzabaluev: I'd call it `Profile` or maybe `BlockchainSelection` for clarity, but it's a little matter.

tests/common/jcli_wrapper/mod.rs
```diff
@@ -15,6 +15,12 @@ use super::process_utils::output_extensions::ProcessOutput;
...
+    Production,
+    Test,
+}
+
```
mzabaluev: Tip: if you'd like to have a clearly designated default value for this, make `impl Default for Discrimination { ... }`, then you can use `Default::default()` as a constructor.

tests/common/jcli_wrapper/mod.rs
mzabaluev: Unless "address discrimination" is an established term that I've missed.

tests/common/jcli_wrapper/mod.rs
dkijania: Ok. will implement this in next PR

PR #378, opened 2019-05-29
Add error for logger init

CodeSandwich:
This is first step of #65. Some loggers will be able to fail on initialization and we want to be prepared


vincenthz:
looking good :+1:

PR #380, opened 2019-05-29
Add enclave in the leadership

vincenthz:


src/secure/enclave.rs
```diff
@@ -0,0 +1,156 @@
...
+
+    pub fn from_vec(leaders: Vec<Leader>) -> Self {
+        let mut e = Self::new();
+        for leader in leaders {
+            e.add_leader(leader);
+        }
+        e
+    }
+
+    pub fn get_leaderids(&self) -> Vec<LeaderId> {
```
mzabaluev: Nitpick: name `get_leader_ids` for snake case consistency.

src/secure/enclave.rs
```diff
@@ -0,0 +1,156 @@
...
+    pub fn new() -> Self {
+        Enclave {
+            leaders: Arc::new(RwLock::new(BTreeMap::new())),
+        }
+    }
+
+    pub fn from_vec(leaders: Vec<Leader>) -> Self {
+        let mut e = Self::new();
+        for leader in leaders {
+            e.add_leader(leader);
```
mzabaluev: It's better to construct the whole map and then put it under the lock. Shouldn't be much of a difference with a small number of leaders, though.

src/secure/enclave.rs
vincenthz: this will go away once this is used as an abstract box

src/secure/enclave.rs
vincenthz: I'ld agree but it's also to load 0 to 2 leaders probably, and in very soon future will be done by a control-api REST endpoint

PR #381, opened 2019-05-29
Updated network protocol doc

mzabaluev:
Brought the description of the methods in sync with the current definitions in the .proto file in network-grpc.


mzabaluev:
Note that I did not update the sync algorithm pseudocode. It uses some old methods that we discarded for the simpler current protocol.

PR #390, opened 2019-05-30
Slog output

CodeSandwich:
Adds Systemd and Jounald log outputs. This is another step for https://github.com/input-output-hk/jormungandr/issues/65

### DIFF
--- a/src/settings/command_arguments.rs
+++ b/src/settings/command_arguments.rs
@@ -76,7 +76,8 @@ pub struct CommandLine {
     #[structopt(long = "log-format", parse(try_from_str), default_value = "plain")]
     pub log_format: LogFormat,
 
-    /// Set format of the log emitted. Can be "stderr"
+    /// Set format of the log emitted. Can be "stderr", "syslog" (unix only) or "journald"
+    /// (linux with systemd only, must be enabled during compilation)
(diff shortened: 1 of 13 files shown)


src/settings/command_arguments.rs
```diff
@@ -69,10 +72,15 @@ pub struct CommandLine {
     #[structopt(short = "v", long = "verbose", parse(from_occurrences))]
     pub verbose: u8,
 
-    /// Set format of the log emitted. Can be "json" or "plain"
+    /// Set format of the log emitted. Can be "json" or "plain", default "plain"
```
NicolasDP: this is redundant with what clap already gives:

```
        --log-format <log_format>              Set format of the log emitted. Can be "json" or "plain" [default: plain]
```

src/settings/command_arguments.rs
```diff
@@ -69,10 +72,15 @@ pub struct CommandLine {
...
     #[structopt(long = "log-format", parse(try_from_str), default_value = "plain")]
     pub log_format: LogFormat,
 
+    /// Set format of the log emitted. Can be "stderr", "syslog" (unix only) or "journald"
+    /// (linux with systemd only, must be enabled during compilation), default "stderr"
```
NicolasDP: no need to add `default "stderr"`. This is already given by `clap`

Issue #383, opened 2019-05-30
JCLI Autocompletion Panic on Linux & Windows


dkijania:
**Describe the bug**
Following guide -> https://input-output-hk.github.io/jormungandr/quickstart/01_command_line.html#help-and-auto-completion
Command:
`jcli auto-completion bash ${HOME}/.bash_completion.d`
produces error:
(Ubuntu 18.04)
```thread 'main' panicked at 'couldn't create completion file: other os error', /home/piotr/.cargo/registry/src/github.com-1ecc6299db9ec823/clap-2.33.0/src/app/parser.rs:131:25
note: Run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
```
(Windows 10 x64)
```
$ bash --version
GNU bash, version 4.3.42(5)-release (x86_64-pc-msys)
```
error:
```
thread 'main' panicked at 'couldn't create completion file: entity not found', C:\Users\Nobody\.cargo\registry\src\github.com-1ecc6299db9ec823\clap-2.33.0\src\app\parser.rs:131:25
note: Run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
```

**To Reproduce**
1. jcli auto-completion bash ${HOME}/.bash_completion.d

**Expected behavior**
no error

piotr-iohk:
For the record - exact error on Ubuntu:
```
thread 'main' panicked at 'couldn't create completion file: entity not found', /home/user/.cargo/registry/src/github.com-1ecc6299db9ec823/clap-2.33.0/src/app/parser.rs:131:25
note: Run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
```

CodeSandwich:
I have error from the same place but with other message, Ubuntu 19.04:
```
thread 'main' panicked at 'couldn't create completion file: other os error', /home/zuczek/.cargo/registry/src/github.com-1ecc6299db9ec823/clap-2.33.0/src/app/parser.rs:131:25
```

CodeSandwich:
This is and isn't a bug. For case when output does not exist here's a PR with a fix. For case when output is an existing file, this is not a bug, JCLI explicitly asks for an output **directory**.

PR #393, opened 2019-05-31
JCLI autocompletion creates output directory if it doesn't exist

CodeSandwich:
Fixes https://github.com/input-output-hk/jormungandr/issues/383

### DIFF
--- a/src/bin/jcli_app/auto_completion.rs
+++ b/src/bin/jcli_app/auto_completion.rs
@@ -12,7 +12,24 @@ pub struct AutoCompletion {
 }
 
 impl AutoCompletion {
-    pub fn exec<S: StructOpt>(self) {
-        S::clap().gen_completions("jcli", self.shell, self.output)
+    pub fn exec<S: StructOpt>(self) -> Result<(), Error> {
+        validate_output(&self.output)?;
+        S::clap().gen_completions("jcli", self.shell, self.output);
+        Ok(())
     }
+}
+
+fn validate_output(output: &PathBuf) -> Result<(), Error> {
+    if !output.exists() {
+        return Err(Error::OutputNotExist);
+    }
+    if !output.is_dir() {
+        return Err(Error::OutputNotDir);
+    }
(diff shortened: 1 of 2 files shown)


CodeSandwich:
Last minute change: added error if output dir exists and is not a dir

src/bin/jcli_app/auto_completion.rs
```diff
@@ -12,7 +12,18 @@ pub struct AutoCompletion {
 }
 
 impl AutoCompletion {
-    pub fn exec<S: StructOpt>(self) {
-        S::clap().gen_completions("jcli", self.shell, self.output)
+    pub fn exec<S: StructOpt>(self) -> Result<(), Error> {
+        if !self.output.exists() {
+            std::fs::create_dir_all(&self.output)?;
```
mzabaluev: It's _very_ surprising if autocompletion makes any changes to the filesystem.
Could it just silently produce nothing if the directory does not exist?

CodeSandwich: But I think that it's exactly the behavior that users were expecting when they were filing the bug

NicolasDP: ha no I think I prefer to fail if the directory does not exit. I don't want we create a directory in the user's environment. Especially they they will expect to send it in their already specific environment (for bash, in `/etc/completion.d`, zsh: `~/.zfunc`)

NicolasDP: yes add an error to not make it silent, but don't create the directory.

also a little update on the doc would be welcome regarding this command. It seems users are confused so would be great to make sure we have given them enough information.

NicolasDP:
LGTM now

src/bin/jcli_app/auto_completion.rs
CodeSandwich: The documentation explicitly asks for directory in output parameter, that probably should be enough

PR #406, opened 2019-06-03
Small doc update on Starting BFT node and jcli auto-completion

piotr-iohk:
I have slightly updated the docs adding:
 - step by step on how to start BFT node
 - more details on setting up `jcli` autocompletion


NicolasDP:
Thanks for the update

PR #415, opened 2019-06-04
Add genesis documentation and passive config note

enzoc4:
# Summary
- Add documentation of jcli *genesis* command.
- Add a few notes in the *starting a passive node* section.


NicolasDP:
looking good, thanks

PR #417, opened 2019-06-05
Pass in top level variants to customize build at deploy time

sjmackenzie:


ci/script.sh
```diff
@@ -6,13 +6,16 @@ set -ex
 main() {
     cross build --target $TARGET --release
 
-    if [ ! -z $DISABLE_TESTS ]; then
-        return
+    if [ -z $DISABLE_TESTS ]; then
+        cross test --target $TARGET --features=integration-test --release
     fi
-
-    cross test --target $TARGET --features=integration-test --release
```
NicolasDP: why changing this part of the script?

sjmackenzie:
```
if [ $DISABLE_TESTS is zero length ]; then
   execute tests
fi
```
is the same as 
```
if [ $DISABLE_TESTS is *not* zero length]; then 
    return
fi
Now execute the tests
```
The former is clearer, easier to read and does exactly the same thing as the latter.

sjmackenzie: Let me add a comment to clarify.

sjmackenzie:
Closing this as it's better I make these changes on my own release branch.

Issue #395, opened 2019-05-31
Add different output formats for JCLI


CodeSandwich:
Currently JCLI outputs structured data as YAML.

The users might want to get it as JSON as well or even something else.

JCLI should provide special option for such commands, which will enable different formats.

NicolasDP:
What about using the same as in `jcli transaction info` ? Let the user decide what is being printed by providing the format in the CLI?

CodeSandwich:
I think you're referring to a `format` parameter. Example:
```
"Transaction `{id}' ({status})\n  Input:   {input}\n  Output:  {output}\n  Fees:    {fee}\n  Balance: {balance}\n"
```
This could be tricky when processing non-flat structures like JSON, which may have nested lists and structures.

I think that this could be an addition to JSON/YAML/etc. as yet another output format not replacement for them.

carl-shepherd:
Hello CodeSandwich,

We'd love to hear what you think of our customer service. Please take a moment to tell us. How would you rate the support you received? Please one of these in your reply that will help us to report on our efforts to the team.

Good, I'm satisfied
Bad, I'm unsatisfied

Thanks,
The IOHK Technical Support Desk Team

PR #418, opened 2019-06-05
Jcli json

CodeSandwich:
Add output formatting in JCLI REST commands. This fixes https://github.com/input-output-hk/jormungandr/issues/395.

This PR proposes small change in existing output format in UTxO getter and in logs getter. This is caused by a bug, design issue or lacking docs in TinyTemplate: https://github.com/bheisler/TinyTemplate/issues/3.

### DIFF
--- a/src/bin/jcli_app/rest/v0/node/stats.rs
+++ b/src/bin/jcli_app/rest/v0/node/stats.rs
@@ -1,4 +1,4 @@
-use jcli_app::utils::{DebugFlag, HostAddr, RestApiSender};
+use jcli_app::utils::{DebugFlag, HostAddr, OutputFormat, RestApiSender};
 use structopt::StructOpt;
 
 #[derive(StructOpt)]
@@ -10,14 +10,18 @@ pub enum Stats {
         addr: HostAddr,
         #[structopt(flatten)]
         debug: DebugFlag,
+        #[structopt(flatten)]
+        output_format: OutputFormat,
     },
 }
 
 impl Stats {
     pub fn exec(self) {
-        let (addr, debug) = match self {
-            Stats::Get { addr, debug } => (addr, debug),
-        };
+        let Stats::Get {
+            addr,
+            debug,
+            output_format,
+        } = self;
         let url = addr
             .with_segments(&["v0", "node", "stats"])
             .unwrap()
@@ -25,8 +29,8 @@ impl Stats {
         let builder = reqwest::Client::new().get(url);
         let response = RestApiSender::new(builder, &debug).send().unwrap();
         response.response().error_for_status_ref().unwrap();
-        let status: serde_json::Value = response.body().json().unwrap();
-        let status_yaml = serde_yaml::to_string(&status).unwrap();
-        println!("{}", status_yaml);
+        let status = response.body().json_value().unwrap();
+        let formatted = output_format.format_json(status).unwrap();
+        println!("{}", formatted);
     }
 }
--- a/src/bin/jcli_app/rest/v0/utxo/mod.rs
+++ b/src/bin/jcli_app/rest/v0/utxo/mod.rs
@@ -1,4 +1,4 @@
-use jcli_app::utils::{DebugFlag, HostAddr, RestApiSender};
+use jcli_app::utils::{DebugFlag, HostAddr, OutputFormat, RestApiSender};
 use structopt::StructOpt;
 
 #[derive(StructOpt)]
@@ -10,20 +10,24 @@ pub enum Utxo {
         addr: HostAddr,
         #[structopt(flatten)]
         debug: DebugFlag,
+        #[structopt(flatten)]
+        output_format: OutputFormat,
     },
 }
 
 impl Utxo {
     pub fn exec(self) {
-        let (addr, debug) = match self {
-            Utxo::Get { addr, debug } => (addr, debug),
-        };
+        let Utxo::Get {
+            addr,
+            debug,
+            output_format,
+        } = self;
         let url = addr.with_segments(&["v0", "utxo"]).unwrap().into_url();
         let builder = reqwest::Client::new().get(url);
         let response = RestApiSender::new(builder, &debug).send().unwrap();
         response.response().error_for_status_ref().unwrap();
-        let utxos: serde_json::Value = response.body().json().unwrap();
-        let utxos_yaml = serde_yaml::to_string(&utxos).unwrap();
-        println!("{}", utxos_yaml);
+        let status = response.body().json_value().unwrap();
+        let formatted = output_format.format_json(status).unwrap();
+        println!("{}", formatted);
     }
 }
--- a/src/bin/jcli_app/utils/output_format.rs
+++ b/src/bin/jcli_app/utils/output_format.rs
@@ -1,0 +1,90 @@
+use gtmpl::Value as GtmplValue;
+use serde_json::{Map as JsonMap, Number as JsonNumber, Value as JsonValue};
+use std::fmt::{self, Display, Formatter};
+use structopt::StructOpt;
+
+#[derive(StructOpt)]
+pub struct OutputFormat {
+    /// Format of output data. Possible values: json, yaml.
+    /// Any other value is treated as a custom format using values from output data structure.
+    /// Syntax is Go text template: https://golang.org/pkg/text/template/.
+    #[structopt(long = "output-format", default_value = "yaml", parse(from_str))]
+    format: FormatVariant,
+}
+
+enum FormatVariant {
+    Yaml,
+    Json,
+    Custom(String),
+}
+
+impl<'a> From<&'a str> for FormatVariant {
+    fn from(format: &'a str) -> Self {
+        match format.trim().to_ascii_lowercase().as_str() {
+            "yaml" => FormatVariant::Yaml,
+            "json" => FormatVariant::Json,
+            _ => FormatVariant::Custom(format.to_string()),
+        }
+    }
+}
+
+custom_error! { pub Error
+    YamlFormattingFailed { source: serde_yaml::Error } = "failed to format output as YAML",
+    JsonFormattingFailed { source: serde_json::Error } = "failed to format output as JSON",
+    CustomFormattingFailed { source: GtmplError } = "failed to format output as custom format",
+}
+
+#[derive(Debug)]
+pub struct GtmplError(String);
+
+impl Display for GtmplError {
+    fn fmt(&self, f: &mut Formatter) -> Result<(), fmt::Error> {
+        self.0.fmt(f)
+    }
+}
+
+impl std::error::Error for GtmplError {}
+
+impl OutputFormat {
+    pub fn format_json(&self, data: JsonValue) -> Result<String, Error> {
+        Ok(match self.format {
+            FormatVariant::Yaml => serde_yaml::to_string(&data)?,
+            FormatVariant::Json => serde_json::to_string_pretty(&data)?,
+            FormatVariant::Custom(ref format) => {
+                let gtmpl_value = json_value_to_gtmpl(data);
+                gtmpl::template(format.as_str(), gtmpl_value).map_err(GtmplError)?
+            }
+        })
+    }
+}
+
+fn json_value_to_gtmpl(value: JsonValue) -> GtmplValue {
+    match value {
+        JsonValue::Null => GtmplValue::Nil,
+        JsonValue::Bool(boolean) => boolean.into(),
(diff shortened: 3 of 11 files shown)


src/bin/jcli_app/rest/v0/node/stats.rs
```diff
@@ -10,13 +10,19 @@ pub enum Stats {
...
     pub fn exec(self) {
-        let (addr, debug) = match self {
-            Stats::Get { addr, debug } => (addr, debug),
+        let (addr, debug, output_format) = match self {
+            Stats::Get {
+                addr,
+                debug,
+                output_format,
+            } => (addr, debug, output_format),
         };
```
mzabaluev: Could do it with less code:

```rest
let Stats::Get { addr, debug, output_format } = self;
```

src/bin/jcli_app/rest/v0/node/stats.rs
```diff
@@ -25,8 +31,8 @@ impl Stats {
         let builder = reqwest::Client::new().get(url);
         let response = RestApiSender::new(builder, &debug).send().unwrap();
         response.response().error_for_status_ref().unwrap();
-        let status: serde_json::Value = response.body().json().unwrap();
-        let status_yaml = serde_yaml::to_string(&status).unwrap();
-        println!("{}", status_yaml);
+        let status = response.body().json_value().unwrap();
+        let formatted = output_format.format(status).unwrap();
```
mzabaluev: Now that we have a custom template formatter accepting templates from the user, could this actually fail?

src/bin/jcli_app/utils/output_format.rs
```diff
@@ -0,0 +1,33 @@
...
+
+#[derive(StructOpt)]
+pub struct OutputFormat {
+    /// Format of output data. Possible values: json, yaml.
+    /// Any other value is treated as a custom format using values from output data structure.
+    /// Syntax example: "{node_id}: up for {stats.uptime}".
+    /// Full syntax reference: https://docs.rs/tinytemplate/1.0.2/tinytemplate/syntax/index.html
+    #[structopt(long = "output-format", default_value = "yaml")]
+    format: String,
+}
```
mzabaluev: For better type safety this could be:

```rest
pub enum OutputFormat {
    Json,
    Yaml,
    Template(String),
}
```

You'll need a custom `StructOpt` impl for it, though.

src/bin/jcli_app/utils/output_format.rs
```diff
@@ -0,0 +1,33 @@
...
+impl OutputFormat {
+    pub fn format(&self, data: impl Serialize) -> Result<String, Error> {
+        Ok(match self.format.trim().to_ascii_lowercase().as_str() {
+            "yaml" => serde_yaml::to_string(&data)?,
+            "json" => serde_json::to_string_pretty(&data)?,
+            _ => {
+                let mut template = tinytemplate::TinyTemplate::new();
+                let template_name = "default";
+                template.add_template(template_name, &self.format)?;
+                template.render(template_name, &data)?
```
mzabaluev: Could this template engine be reused and not instantiated for every line of output?

mzabaluev: Must `TinyTemplate` always haul around a map of named templates, or is there a way to instantiate and use just one self-contained template?

src/bin/jcli_app/rest/v0/node/stats.rs
CodeSandwich: That's brilliant!

src/bin/jcli_app/rest/v0/node/stats.rs
CodeSandwich: Of course, users can pass invalid template

src/bin/jcli_app/utils/output_format.rs
CodeSandwich: The `format` field is private and used in only one place, which wouldn't benefit from enum anyway. Isn't it an overkill?

src/bin/jcli_app/rest/v0/node/stats.rs
mzabaluev: That's destructuring `let` :smile:

src/bin/jcli_app/utils/output_format.rs
mzabaluev: Then you can match the output modes in a type-safe rather than stringized way.

src/bin/jcli_app/rest/v0/node/stats.rs
mzabaluev: Yeah, but the JSON and YAML formatters "can't fail" in normal circumstances.

src/bin/jcli_app/utils/output_format.rs
CodeSandwich: The engine is created and called at most once for the whole JSON fetched from REST. I think i see what you're thinking about: the template should be created during CLI args parsing, right?

Unfortunately yes, TinyTemplate always holds a whole collection of templates and can't be used without usage of keys.

src/bin/jcli_app/rest/v0/node/stats.rs
CodeSandwich: But it's done on a every possible (only) variant of an enum at once, I didn't know that compiler was able to do that

src/bin/jcli_app/utils/output_format.rs
mzabaluev: Ah, sorry, I thought it could be called repeatedly.

I wonder if there is some widespread template syntax that many people already know. Is this syntax mostly the same as Go templates?

mzabaluev:
I wonder if there is some widespread template syntax that many people already know. Is this syntax mostly the same as Go templates?

src/bin/jcli_app/rest/v0/node/stats.rs
CodeSandwich: Serde always leaves open door for failure, but we probably could rely on fact that `serde_json::Value` won't fail. But we always have TinyTemplate, which can and will fail often, whenever user makes a mistake.

mzabaluev: > But we also have TinyTemplate, which can and will fail often, whenever user makes a mistake.

Meaning that we might want handle the error in a graceful way here, rather than panicking.

CodeSandwich: There's no infrastructure for graceful error handling in `rest` command yet.  There are a lot of panicking places, which will be turned into `return`s. I don't think that this PR is the right scope to fix that.

src/bin/jcli_app/utils/output_format.rs
CodeSandwich: It's similar, but not more than probably 90% of templating tools, they're all basically the same

src/bin/jcli_app/rest/v0/node/stats.rs
mzabaluev: Yeah, it only works because there is a single variant, so the match is irrefutable.
You can also use the same syntax with structs, similar to how the cool kids do it in JavaScript these days.

src/bin/jcli_app/utils/output_format.rs
```diff
@@ -0,0 +1,33 @@
...
+    YamlFormattingFailed { source: serde_yaml::Error } = "failed to format output as YAML",
+    JsonFormattingFailed { source: serde_json::Error } = "failed to format output as JSON",
+    CustomFormattingFailed { source: tinytemplate::error::Error } = "failed to format output as custom format",
+}
+
+impl OutputFormat {
+    pub fn format(&self, data: impl Serialize) -> Result<String, Error> {
+        Ok(match self.format.trim().to_ascii_lowercase().as_str() {
+            "yaml" => serde_yaml::to_string(&data)?,
+            "json" => serde_json::to_string_pretty(&data)?,
```
mzabaluev: Then again, the literal match syntax is concise, so it's good as is.

CodeSandwich:
@mzabaluev Switched to `gtmpl`. I think that it was a good move, it's a more competent tool and does not require the change in JCLI output formats anymore. Also the syntax is more standard, which can be felt even in docs, the syntax has a name and serious docs page, no hand-waving.

src/bin/jcli_app/rest/v0/utxo/mod.rs
```diff
@@ -10,20 +10,24 @@ pub enum Utxo {
...
+        } = self;
         let url = addr.with_segments(&["v0", "utxo"]).unwrap().into_url();
         let builder = reqwest::Client::new().get(url);
         let response = RestApiSender::new(builder, &debug).send().unwrap();
         response.response().error_for_status_ref().unwrap();
-        let utxos: serde_json::Value = response.body().json().unwrap();
-        let utxos_yaml = serde_yaml::to_string(&utxos).unwrap();
-        println!("{}", utxos_yaml);
+        let status = response.body().json_value().unwrap();
+        let formatted = output_format.format(status).unwrap();
```
mzabaluev: This is a bit repetitive. Can `OutputFormat` accept the response body and then transform it as necessary? This could skip the JSON intermediate step.

src/bin/jcli_app/utils/output_format.rs
```diff
@@ -0,0 +1,83 @@
...
+            _ => {
+                let value = serde_json::to_value(data)
+                    .map(json_value_to_gtmpl)
+                    .map_err(|source| Error::JsonValueCreateFailed {
+                        source,
+                        filler: CustomErrorFiller,
```
mzabaluev: :confused:

src/bin/jcli_app/utils/output_format.rs
```diff
@@ -0,0 +1,83 @@
...
+    fn fmt(&self, f: &mut Formatter) -> Result<(), fmt::Error> {
+        self.0.fmt(f)
+    }
+}
+
+impl std::error::Error for GtmplError {}
+
+impl OutputFormat {
+    pub fn format(&self, data: impl Serialize) -> Result<String, Error> {
+        Ok(match self.format.trim().to_ascii_lowercase().as_str() {
```
mzabaluev: The `.trim().to_ascii_lowercase()` normalization should be done on construction of `OutputFormat`.

CodeSandwich: The normalization is not a valid step for custom format definiton, it can be used only for `json` and `yaml` flag detection

src/bin/jcli_app/utils/output_format.rs
CodeSandwich: This `filler` pattern is the best I could come up while fighting with `custom_error` macro. Without it there is a conflict of `From`s for variant with same `source`. And without `source` `Error::cause` doesn't work. Author of this macro assumed that `from` and `cause` functionalities should be synonymous.

src/bin/jcli_app/utils/output_format.rs
mzabaluev: The constructor could check the normalized value for these two matches and set the member value as is if it's something else.

src/bin/jcli_app/utils/output_format.rs
mzabaluev: You are talking about `cause` as if it's not deprecated. Is `custom_error` unaware of `Error::source`?

`custom_error` is only a convenience. If it does not cut it, there's always the good old way of manually defining `Error`, `Display`, etc.

I feel a growing urge to try `snafu` for error definition.

CodeSandwich: Right, I meant `Error::source`.

I think that this spot of ugliness is still a decent trade-off for not having to maintaining `source` and `Display` for every variant. Definitely migrating to a better tool is a great idea, `custom_error` isn't that good. Just not in this PR.

src/bin/jcli_app/utils/output_format.rs
CodeSandwich: This is the idea of making `OutputFormat` an enum coming back, isn't it?

src/bin/jcli_app/rest/v0/utxo/mod.rs
CodeSandwich: It could accept a `serde_json::Value`, so we don't serialize and deserialize pointlessly. I want to avoid accepting a REST body, this looks like a responsibility creep

src/bin/jcli_app/utils/output_format.rs
mzabaluev: No, just normalizing the input once in the proper place.

PR #420, opened 2019-06-06
Soak test for 100 transactions

dkijania:
first soak test. it puts 100 correct transaction into node and verifies if all of them are in a block

### DIFF
--- a/tests/common/process_utils/mod.rs
+++ b/tests/common/process_utils/mod.rs
@@ -168,6 +168,6 @@ pub fn run_process_until_response_matches<F: Fn(Output) -> bool>(
 }
 
 pub fn sleep(seconds: u64) {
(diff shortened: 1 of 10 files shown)


tests/common/process_utils/mod.rs
```diff
@@ -167,3 +166,8 @@ pub fn run_process_until_response_matches<F: Fn(Output) -> bool>(
     println!("Success: {}", &command_description);
     Ok(())
 }
+
+pub fn sleep(seconds: u64) {
+    let ten_millis = time::Duration::from_millis(1000 * &seconds);
```
NicolasDP: Duration has a [`from_secs`](https://doc.rust-lang.org/nightly/std/time/struct.Duration.html#method.from_secs) that does exactly that

PR #428, opened 2019-06-07
add current date so users are not suprised with very large epoch start

NicolasDP:


scripts/stakepool-single-node-test
```diff
@@ -59,7 +59,7 @@ STAKE_DELEGATION_CERT=$(cat stake_delegation.signcert)
 
 cat << EOF > genesis.yaml
 blockchain_configuration:
-  block0_date: 1550822014
+  block0_date: $(date +%s)
```
mzabaluev: Besides, could we repeat @CodeSandwich's good work and make this parameter format RFC 3339?

NicolasDP: I don't think we want that. I think here we really want to have _seconds since UNIX_EPOCH_. Especially, this genesis yaml file is not meant to be used/reused often by the users, I don't see a point to make it too fancy/shiny. I may be wrong.

mzabaluev: OK, there may more benefit in having a precise number here.

CodeSandwich: Since I'm already here: I would prefer RFC3339 too. It's the same precision, but  is human inspectable. If it's not changed often, it doesn't really matter, but we will keep consistency with other parts of the project.

vincenthz: I have a mild preference for seconds since epoch as well (just because it's not really suppose to be used by human), although having it in human format would be less a pain to deal with.

If this change to human time, we need to make sure, no one can specify anything more precise than the seconds, the rest of the code depends on this in some fragile way.

PR #444, opened 2019-06-10
Update network-grpc, ported to tower-hyper

mzabaluev:


mzabaluev:
Some changes still required with `AccountState` and `AccountIdentifier`, probably trivial.

mzabaluev:
Be warned: gRPC connections between nodes fail to get established after this change. I'm looking into the problem.

Issue #441, opened 2019-06-08
remove the shell ansi colours from scripts/stakepool-single-node-test


onicrom:
**Is your feature request related to a problem/context ? Please describe if applicable.**
If users are using non-standard shell terminals the colours show up unprocessed/raw shellcode format

**Describe the solution you'd like**
Remove shell colours from output.

**Additional context**
here is a sample of the output

faucet account: \033[0;32mSTUFF\033[0m
  * public: \033[0;33med25519_STUFFm\033[0m
  * secret: \033[0;31med25519e_STUFFv\033[0m
  * amount: \033[0;32m1000000000\033[0m
pool id: \033[0;32mSTUFF\033[0m
To start the node:
jormungandr --genesis-block block-0.bin --config config.yaml --secret poolsecret1.yaml

vincenthz:
it should be fixed already in the latest version of the script

Issue #388, opened 2019-05-30
Account is not created after successful utxo->account transaction


dkijania:
**Describe the bug**
after correct transaction (utxo->account) is send to node and node has **allow_account_creation** option set to **true**. then if we want to fetch account data using jcli command:
```jcl rest v0 account get ta1sv0rtc9rwrty05lt4mxur7egzvyac7qz2l5j65v8z37s5pqfntk969w7z99 -h http://127.0.0.1:8073/api```

we are recieving error:

```thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Error(Status(400), "http://127.0.0.1:8073/api/v0/account/ta1sv0rtc9rwrty05lt4mxur7egzvyac7qz2l5j65v8z37s5pqfntk969w7z99")', src\libcore\result.rs:997:5```

**To Reproduce**
1. Go to PR https://github.com/input-output-hk/jormungandr/pull/384
2. Run test **jcli::transaction::e2e::test_account_is_created_if_transaction_out_is_account** by command:
`cargo test jcli::transaction::e2e::test_account_is_created_if_transaction_out_is_account --features integration-test`

**Expected behavior**
test is passed

**Additional context**

vincenthz:
allow_account_creation is gone, and the jcli call should now be working

PR #445, opened 2019-06-10
Added scripts for bft node and send transaction

dkijania:


scripts/bft-single-node-test
```diff
@@ -0,0 +1,83 @@
...
+FAUCET_AMOUNT=1000000000
+ADDRTYPE="--testing"
+REST_PORT="8443"
+
+### MAKE EVERYTHING
+
+# faucet
+FAUCET_SK=$($CLI key generate --type=Ed25519Extended)
+FAUCET_PK=$(echo ${FAUCET_SK} | $CLI key to-public)
+FAUCET_ADDR=$($CLI address single ${ADDRTYPE} ${FAUCET_PK})
```
NicolasDP: make the faucet an account

scripts/bft-single-node-test
```diff
@@ -0,0 +1,83 @@
...
+echo  "--------"
+echo "To start the node as leader :"
+echo "$NODE --genesis-block block-0.bin --config config.yaml --secret secret.yaml"
+echo  "--------"
+echo "To start the node as standard node (without ability to create blocks) :"
+echo "$NODE --genesis-block block-0.bin --config config.yaml --secret secret.yaml"
+echo  "--------"
+echo "To interact with node :"
+echo "$CLI rest v0 node stats -h http://127.0.0.1:${REST_PORT}/api"
+echo  "--------"
```
NicolasDP: there are coloring options available from the other script, reuse

scripts/bft-single-node-test
dkijania: AFAIK accounts are not working currently (cannot get account state ) . See https://github.com/input-output-hk/jormungandr/issues/388

scripts/bft-single-node-test
dkijania: There is this issue: https://github.com/input-output-hk/jormungandr/issues/441. Observed on my machine also. that's why I've removed colouring

vincenthz: it has been fixed already

scripts/bft-single-node-test
vincenthz: they are working now

PR #453, opened 2019-06-11
Adding a utility that'll convert a between different addresses

sjmackenzie:


sjmackenzie:
depends on https://github.com/input-output-hk/rust-cardano/pull/724

PR #461, opened 2019-06-11
Stake key certificate does not exist anymore

enzoc4:
I think stake key registration certificates are no longer a thing? I'm not sure why or what changed, but at least they don't work for me.


vincenthz:
correct, the stake key registration is gone, replaced by creation-on-reference (sending coin to an account, or to a group address)

PR #474, opened 2019-06-12
Last minute updates

vincenthz:


docker/Dockerfile
```diff
@@ -0,0 +1,44 @@
...
+    mkdir -p ${ENV_PREFIX}/src && \
+    mkdir -p ${ENV_PREFIX}/bin && \
+    bash -c "curl https://sh.rustup.rs -sSf | bash -s -- -y" && \
+    export PATH=$HOME/.cargo/bin:$PATH && \
+    rustup install stable && \
+    rustup default stable && \
+    cd ${ENV_PREFIX} && \
+    git clone --recurse-submodules https://github.com/input-output-hk/jormungandr src && \
+    cd src && \
+    git checkout -b last-minutes-updates && \
```
mzabaluev: Why switch to the new branch?

onicrom: Why not make the branch a --build-arg when you build the docker container:

# default git branch 
ARG BRANCH=master
ENV ENV_BRANCH=${BRANCH}

git checkout -b ${ENV_BRANCH}

During the docker build  you can specify --build-arg BRANCH=last-minutes-updates

onicrom:
might be best to use ubuntu:bionic instead of the non LTS release of ubuntu.

mzabaluev:
> might be best to use ubuntu:bionic instead of the non LTS release of ubuntu.

We are Rust people, we need latest and greatest :)
Seriously though, as the toolchain is pulled by rustup, I don't know if we need anything really new.

docker/Dockerfile
vincenthz: damn, I forgot to remove it

vincenthz: I briefly look at your script btw @onicrom, I just pulled really quickly couple of improvement inspired by your docker, to the one I had locally but clearly this is being thrown together. down the line there's multiple improvements to make to the script, and make multiple variations

vincenthz:
no reason for choosing the ubuntu release, we don't need much from the system anyway. Also I don't expect this docker file to be really production ready; it's more for getting some docker stuff in the repo than a final product at this stage.

PR #477, opened 2019-06-12
jcli: 'remove allow_account_creation' from 'jcli genesis init'

### LINKED ISSUE #471: jcli: remove 'allow_account_creation' from the config generated with 'jcli genesis init > genesis.yaml'
**Describe the bug**
jcli genesis init > genesis.yaml is still including 'allow_account_creation' and it should not be used anymore. 

**To Reproduce**
1. run jcli genesis init > genesis.yaml  and check for 'allow_account_creation' inside the generated genesis file (it should not be present)

**Expected behavior**
run jcli genesis init > genesis.yaml  and check for 'allow_account_creation' inside the generated genesis file (it should not be present)

**Additional context**

rinor:
Remove **allow_account_creation** from `jcli genesis init`
Fix #471


vincenthz:
Thanks. I missed it in the grep I did for #471

Issue #470, opened 2019-06-12
Documentation: Small improvements


dorin100:
**Describe the bug**
1. Add example of submitting a Transaction for delegating stake (**both for Account and UTXO delegation**) - https://input-output-hk.github.io/jormungandr/stake_pool/delegating_stake.html

2. Add example for input when creating a Transaction and the source is an Account Address
- https://input-output-hk.github.io/jormungandr/jcli/transaction.html

3. Staking and Delegation - https://input-output-hk.github.io/jormungandr/stake_pool/delegating_stake.html
- it is not clear that: (i) rewards are not implemented yet, (ii) you cannot delegate part of a thing (account or utxo). in a case of an account you need to split the account into 2 accounts, in a case of utxo, you’d need to split your utxo into 2 utxos with 2 different grouping key.

4. Look and feel
  - Use same fonts/sizes/formats/capitalization in similar places
	- EX: https://input-output-hk.github.io/jormungandr/stake_pool/registering_stake.html
The paragraph titles have different sizes, one starts with a small letter, the other with capital one

5. Add links to the existing scripts - https://github.com/input-output-hk/jormungandr/tree/master/scripts (for Transaction, starting node)

**To Reproduce**
1. https://input-output-hk.github.io/jormungandr/stake_pool/delegating_stake.html.
- the section is ending in "The output can now be added in the transaction and submitted to a node." but it is not clear what values should you use for input,output, withnes when creating the delegation transaction

In order to Delegate the Stake from a Transaction to a Stake Pool, both for Account and UTXO addresses, you need to follow the below steps:
1. create a stake key pair - like here: https://input-output-hk.github.io/jormungandr/stake_pool/registering_stake.html
2. create and sign a Delegation Certificate like here: https://input-output-hk.github.io/jormungandr/stake_pool/delegating_stake.html
  - you need to use the public key generated at step 1 + desired stake pool ID (the one generated by the bootstrap script if you started the node using it)
3. create a delegation transaction using the below parameters
  - input
     - use the Personal Account (that you want to delegate)
     - as a value add the Transaction Fee (the remaining value of the Account will be Delegated) - somebody needs to pay the TxFee
     - EX
        - for Account: jCLI transaction add-account %sourceAccountAddress% 10 --staging transaction.tx
        - for UTXO: not tested yet
  - output
      - don't add any output
  - delegation certificate
      - add the signend delegation certificate before adding the witness and before finalizing the transaction
[...]

PR #487, opened 2019-06-13
Add example for account input and links to scripts

enzoc4:
related to #470 point 2 and point 5

### DIFF
--- a/doc/jcli/transaction.md
+++ b/doc/jcli/transaction.md
@@ -32,7 +32,11 @@ content information of a transaction:
 
 # Examples
 
-Let's say we have the following utxo
+The following example focuses on using an utxo as input, the few differences when transfering from an account will be pointed out when necessary.
+
+Let's use the following utxo as input and transfer 50 lovelaces to the destination address
+
+## Input utxo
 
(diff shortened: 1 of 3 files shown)


doc/jcli/transaction.md
```diff
@@ -32,7 +32,12 @@ content information of a transaction:
 
 # Examples
 
-Let's say we have the following utxo
+The following example focuses on using an utxo as input, the few differences when transfering from an account will be pointed out when necessary.
+There is also a script [here](https://github.com/input-output-hk/jormungandr/blob/master/scripts/send-transaction) to send a transaction from a faucet account to a specific address which could be used as a reference.
```
NicolasDP: I think this script is not really good. The `scripts` directory is more like a playground now:

* `bootstrap` generate everything for someone to start a new blockchain (genesis or bft);
* `bootstrap` produces scripts to test some actions (sending money, sending a certificate...);

the generated scripts can then be used to play with a blockchain and see some properties.

If @vincenthz is ok, I would rename the script directory as _playground_ and write a README in that directory explaining how to use the playground.

Issue #475, opened 2019-06-12
jcli transaction id not changing when adding certificate


enzoc4:
**Describe the bug**
When building a transaction with jcli, the id doesn't change when adding a certificate.

**To Reproduce**

### Create finalized transaction:

```sh
jcli transaction new > tx
jcli transaction add-input ee77f241c78d77a57056ae9cb13fb2b8d50c28b6609386180cc35db74fb7094a 0 1000 --staging tx
jcli transaction add-output ta1ssnr5pvt9e5p009strshxndrsx5etcentslp2rwj6csm8sfk24a2wu27hujgyzl4mlc736jaztykud5tqrxw5gd6esmw7rx5zujdzw3ahcaskk 100 --staging tx
jcli transaction finalize ta1sn9u0nxmnfg7af8ycuygx57p5xgzmnmgtaeer9xun7hly6mlgt3pjp0h0rrq782p7z4auve0s4l72egyc88f49mzm8lwuaw7xnw52udky79t3s --staging tx
```

### Get id
`jcli transaction id --staging tx`

Result: **7e650244c14155448449a14739873bbf396676c44e08a50a426554947cfe25be**

### Add certificate 

```sh
jcli transaction add-certificate cert1qyzlw7xxpuw5ru9tmcejlptlu4jsfswwn2tk9k07ae6audxag4cmv4qak5p5nc4urfd35uunnwwcdlz9qec30nynpsm2lwm0kz5n982pqyqyq6y5ltkusja2wt5nswms7mycvvph5w26g8fuycycxqynpug87cu70gx6cuq43xca5cc034wz5vh43dzwml9v0tfrtlzr4qdpwhm4zcrqfhytdc --staging tx
```
### Get id again

`jcli transaction id --staging tx`

Result: **7e650244c14155448449a14739873bbf396676c44e08a50a426554947cfe25be**

**Expected behavior**
The ids to be different

**Additional context**

When posting a transaction with a certificate to the node, I'm getting an error about invalid signature, and I think this may be the cause (because I need the transaction id for signing).

rinor:
It looks like `jcli transaction id ...` through **display_id(...)** function:
https://github.com/input-output-hk/jormungandr/blob/90ea1b9eac9a07e54fc9a061155affe3fcb30e2c/jcli/src/jcli_app/transaction/mod.rs#L136-L140

ends up using **transaction(...)** function which 'discards' the **extra** information related to the certificate, hence the transaction id (hash) is shown same.

[...]

PR #493, opened 2019-06-14
Automation test for issue #475

dkijania:


jormungandr-integration-tests/src/jcli/transaction/id.rs
```diff
@@ -0,0 +1,29 @@
...
+
+    let mut builder = JCLITransactionWrapper::new_transaction(FAKE_GENESIS_HASH);
+
+    builder
+        .assert_add_input(&FAKE_INPUT_TRANSACTION_ID, &0, &200)
+        .assert_add_output(&reciever.address, &150)
+        .assert_finalize_with_spending_address(&sender.address);
+
+    let transaction_id_before = builder.get_transaction_id();
+    builder.assert_add_certificate(&FAKE_CERTIFICATE);
```
vincenthz: certificate should be added before finalization, most definitely will not work; we had a bug that was requiring the certificate after finalization

jcli/.gitignore
```diff
@@ -0,0 +1 @@
+/target
```
vincenthz: already one at the root that ignore this already, what's the point of this ?

jormungandr-integration-tests/src/jcli/transaction/id.rs
dkijania: In steps to reproduce for issue #475 transaction is finalized prior to add-certificate, bug was accepted in that form and this is how it's working right now (on master from 14.06).  So `add-certificate` command should change transaction id only before finalize ? and if we use after finalize jcli should complain?

dkijania: ok. looks like after rebase it's working as you said

NicolasDP:
#475 is closed now. I am not sure if this is still relevant now.

dkijania:
I think it is. as a regression test

dkijania:
This PR is rejected, based on our new agreement i will implement it in lower layer. probably jcli

PR #497, opened 2019-06-14
Support GELF logging output.

### LINKED ISSUE #447: gelf logging support for slog
We'd like to have gelf logging support via slog so we can send logs to a 3rd party service that supports gelf. The main reason for this is slog-syslog does not support windows so the only way to setup logging on windows to send to a 3rd party server would be to instruct end users to install nixlog, configure it to read a log file and forward it to a syslog server. Having gelf support to be able to send directly to graylog would be much simpler for the user to enable.
Add gelf slog support (either with a custom implmenetation or 3rd party library)

jbgi:
Closes #447.

### DIFF
--- a/jormungandr/Cargo.toml
+++ b/jormungandr/Cargo.toml
@@ -64,6 +64,7 @@ jormungandr-lib = { path = "../jormungandr-lib" }
 strfmt = "0.1"
 gtmpl = "0.5.6"
 mktemp = "0.4.0"
+slog-gelf = "0.1.0"
 
 [dependencies.actix-web]
 version = "0.7.18"
--- a/jormungandr/src/main.rs
+++ b/jormungandr/src/main.rs
@@ -46,6 +46,7 @@ extern crate serde_yaml;
 #[macro_use(b, o, record_static, debug, info, warn, error, crit)]
 extern crate slog;
 extern crate slog_async;
+extern crate slog_gelf;
 #[cfg(feature = "systemd")]
 extern crate slog_journald;
 extern crate slog_json;
--- a/jormungandr/src/settings/logging.rs
+++ b/jormungandr/src/settings/logging.rs
@@ -1,6 +1,7 @@
 use crate::log::{AsyncableDrain, JsonDrain};
 use slog::{Drain, Logger};
 use slog_async::Async;
+use slog_gelf::Gelf;
 #[cfg(feature = "systemd")]
 use slog_journald::JournaldDrain;
 #[cfg(unix)]
@@ -13,6 +14,8 @@ pub struct LogSettings {
     pub verbosity: slog::Level,
     pub format: LogFormat,
     pub output: LogOutput,
+    pub backend: Option<String>,
+    pub logs_id: Option<String>,
 }
 
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
@@ -28,6 +31,7 @@ pub enum LogFormat {
 /// Output of the logger.
 pub enum LogOutput {
     Stderr,
+    Gelf,
     #[cfg(unix)]
     Syslog,
     #[cfg(feature = "systemd")]
@@ -52,6 +56,7 @@ impl FromStr for LogOutput {
     fn from_str(s: &str) -> Result<Self, Self::Err> {
         match &*s.trim().to_lowercase() {
             "stderr" => Ok(LogOutput::Stderr),
+            "gelf" => Ok(LogOutput::Gelf),
             #[cfg(unix)]
             "syslog" => Ok(LogOutput::Syslog),
             #[cfg(feature = "systemd")]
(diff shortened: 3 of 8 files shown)


jormungandr/Cargo.toml
```diff
@@ -63,6 +63,8 @@ jormungandr-utils = { path = "../jormungandr-utils" }
 strfmt = "0.1"
 gtmpl = "0.5.6"
 mktemp = "0.4.0"
+slog-gelf = { git = "https://github.com/jbgi/slog-gelf", rev = "2a43788a6449ed7e39b4cb2f083c5d1827d59e7b" }
```
NicolasDP: do we know why this was never released?

since you have forked it you could update the deps please? You are adding a lot of duplicated versions in the dependencies with this

jormungandr/src/settings/logging.rs
```diff
@@ -63,21 +68,31 @@ impl FromStr for LogOutput {
...
-        Ok(match self {
-            LogOutput::Stderr => format.decorate_stderr(),
+    fn to_logger(&self, format: &LogFormat, backend: &Option<String>) -> Result<Async, Error> {
+        match self {
+            LogOutput::Stderr => Ok(format.decorate_stderr()),
+            LogOutput::Gelf => 
+                match backend {
+                    Some(graylogHostPort) => {
+                        // We use hostname as source for now, but a better identifier should be used.
+                        let source = get_hostname().unwrap_or("unhostnamed".to_string());
```
NicolasDP: what about generating a random UUID instead ? So we can protect our users...

jormungandr/Cargo.toml
jbgi: I can update the deps in the repo, but for my understanding, in what sense is there duplicated versions? I was thinking cargo would use the versions from Cargo.lock and not from the upstream repo...

disassembler:
I've rebased in my fork (disassembler) but have no way to push to this PR.

jormungandr/src/settings/logging.rs
jbgi: Did that. But what about using the node id? Would this be too impractical to implement?

jormungandr/src/main.rs
```diff
@@ -44,7 +44,9 @@ extern crate serde_json;
 extern crate serde_yaml;
 #[macro_use(b, o, record_static, debug, info, warn, error, crit)]
 extern crate slog;
+extern crate hostname;
```
NicolasDP: now that you use uuid. do you still need this?

jbgi: fixed.

NicolasDP:
This UUID will be generated every time at start up. Is this what you really want?

How will the user be aware of the UUID to give to get support?

What about re-using the same UUID? Since this is a feature Devops wishes to include, why not have it set as parameter when starting GELF logging and have it generated by the devops script side of things?

mzabaluev:
I would provide the GELF logging id as an option in the config.yaml option map for GELF.

jbgi:
Yes, I was thinking of adding an `identifier` parameter (should the name be more gelf specific like `gelf_source`?) that would default to the generated uuid (and have it printed to stderr). But now I agree that it is simpler to generate it and just have it provided externally. Will update.

jormungandr/Cargo.toml
```diff
@@ -63,6 +63,7 @@ jormungandr-utils = { path = "../jormungandr-utils" }
 strfmt = "0.1"
 gtmpl = "0.5.6"
 mktemp = "0.4.0"
+slog-gelf = { git = "https://github.com/jbgi/slog-gelf", rev = "18197f7d5c9f439197111299509f02b24efe8985" }
```
vincenthz: unless you're planning to maintain it yourself going forward, it probably make sense  copied in this repository

mzabaluev: I think it would be better to make a fork in our GitHub org. But that's only second best to sending patches upstream, in case the original maintainer no longer works on the project.

jbgi: https://github.com/lancastr/slog-gelf/issues/1#issuecomment-503046771

mzabaluev: Ask and ye shall receive :slightly_smiling_face:

jbgi:
should be mergeable now.

vincenthz:
I think this is looking good, but it's missing a configurability option to make the gelf support optional

jormungandr/src/settings/start/config.rs
```diff
@@ -24,6 +24,8 @@ pub struct ConfigLogSettings {
     pub verbosity: Option<u8>,
     pub format: Option<LogFormat>,
     pub output: Option<LogOutput>,
+    pub backend: Option<String>,
```
vincenthz: that might be better as an Enum related to gelf instead of a random string.

jormungandr/src/settings/start/config.rs
```diff
@@ -24,6 +24,8 @@ pub struct ConfigLogSettings {
...
+    pub logs_id: Option<String>,
```
vincenthz: same thing

jormungandr/src/settings/logging.rs
```diff
@@ -28,6 +31,7 @@ pub enum LogFormat {
 /// Output of the logger.
 pub enum LogOutput {
     Stderr,
+    Gelf,
```
vincenthz: seems that if you make `Gelf(String, String)` you same yourself the trouble of extending the generic structure for something specific to the backend

mzabaluev: The structopt/YAML thing would need to be customized for this, I believe.

This whole enum needs love, but I can improve it later.

mzabaluev:
@vincenthz should we expect @jbgi to do the right `cfg`/optional dependency thing, or is it something we can do as a follow-up change?

jormungandr/src/settings/logging.rs
jbgi: yeah, I was not sure how the yaml parsing would workout if I did that. I can try.

jbgi:
> it's missing a configurability option to make the gelf support optional

@vincenthz so a feature like for journald support?

vincenthz:
I don't mind too much on the order, it should be easy to cfg() is out, so that it doesn't impose some transitient unseen burden

jormungandr/src/settings/logging.rs
vincenthz: ok nevermind, if it's too complicated we can push this through and revisit later

Issue #469, opened 2019-06-12
bootstrap script improvements


dorin100:
**Describe the bug**
1. automatically delete the storage file (when running bootstrap script - https://github.com/input-output-hk/jormungandr/blob/master/scripts/bootstrap)
2. add option/parameter to start the node with Faucet = UTXO Address

**To Reproduce**
1.  if you run the script 2 times and start the node, the second time you start the node, you will receive the below error:
```D:\iohk\otherProjects\jormungandr\scripts\boostrap>jormungandr --genesis-block block-0.bin --config config.yaml --secret pool-secret1.yaml
Jun 12 17:27:01.704 INFO storing blockchain in '"/d/iohk/otherProjects/jormungandr/scripts/boostrap/jormungandr-storage-test/blocks.sqlite"', task: init
Jun 12 17:27:01.729 INFO restoring state at tip 34c806dc532b84cb09a3d4ce9a4f3f2eb12773e0d9cd12391d23d251666bd434, task: bootstrap
Error while loading the blockchain state
Error in the blockchain storage: block not found
```

**Expected behavior**
1. you should not receive any error if you run the script multiple times, each time starting the node too. Basically, you do the same operations (run script, start node) so you should see the same result (node started successfully) 
2. besides the apiPort and consensus protocol, the Faucet Address Type would be a very useful parametrization for the end users.

**Additional context**

vincenthz:
happy to receive PR for improvement like this

onicrom:
I'll take a stab at this tonight after work

NicolasDP:
fixed by #498

PR #498, opened 2019-06-14
adding a few more flags/options to the bootstrap script

onicrom:
add -o flag to remove sqlite files if they exist (addressing one item from #469 )
add -d flag to make the sqlite directory configurable (default still ${PWD}
add printing block-0 hash to stdout

### DIFF
--- a/scripts/bootstrap
+++ b/scripts/bootstrap
@@ -17,13 +17,14 @@
 CLI=jcli
 NODE=jormungandr
 COLORS=1
+CLEAN=0
+DPATH="${PWD}"
 REST_PREFIX="api"
 REST_HOST="127.0.0.1"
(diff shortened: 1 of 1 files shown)


scripts/bootstrap
```diff
@@ -33,7 +35,7 @@ SECRET_PATH="."
 CONFIG_PATH="."
 ADD_STARTUP_SCRIPT=0
 
-while getopts 'bc:d:ghk:p:s:xC' c
+while getopts 'bc:d:ghk:p:s:xCod' c
```
vincenthz: there's already a `d:`, the ending `d` shouldn't be there

onicrom: will fix

Issue #485, opened 2019-06-13
Allow for better logging strategy


NicolasDP:
In the event of a large amount of Logs being emitted the user may see this kind of logs:

```
Jun 13 17:27:39.749 ERRO slog-async: logger dropped messages due to channel overflow, count: 97, sub_task: Leader Task, task: leadership
```

We can configure the [`Async`](https://crates.io/crates/slog_async) to have a better strategy to manage the logs. Especially we can allow the user to set a custom strategy.

mzabaluev:
Could it be that we lack buffering in the logging thread and the sink is sometimes too slow?

NicolasDP:
The `Async` object is doing the buffering for us. It has a channel to do that. Now the real problem, I believe. Is that we utilising the default settings which is to utilise 128 only:

https://github.com/slog-rs/async/blob/master/lib.rs#L233-L240

NicolasDP:
Reopen as wrongly closed

CodeSandwich:
@NicolasDP Do you have log output with new failure?

NicolasDP:
You fixed the symptoms, but not the problem. This ticket is to allow users to provide better logging strategy. 1024 may be way too large for certain users.

> We can configure the Async to have a better strategy to manage the logs. Especially we can allow the user to set a custom strategy.

CodeSandwich:
I was thinking about this issue and I think it's worth considering if this is worth configurability.

Is queue of 1024 entries too long? How many resources is extension of buffer going to use? My guess is that it's going to be an order of ~100KB. Are we going to run Jormungandr on platforms where memory is so constrained?

Is queue of 1024 entries too short? The worst case we've found required queue of less than 256 entries. It's 4 times as much, so it's probably reasonably safe. We're never going to be 100% safe as long as memory is finite.

I'm predicting that nobody's going to ever bother to configure this parameter. It's only going to litter documentation and complicate codebase.

NicolasDP:
hence `Priority - Low` label. This is a reminder we can do better. And also this is not about only providing (or not) the configuration to the user. It's about providing strategy on what log do dismiss (because this feature is available too).

CodeSandwich:
Slog-async currently provides only 3 strategies: drop and log dropping (we use it), drop silently and panic. Does it need configuration from user? Unfortunately slog-async doesn't allow picking which events to drop, it's always the newest. We can create an issue for that in slog-async, so it can drop lower priority items first.

[...]

PR #509, opened 2019-06-17
Increase logger async buffer from 128 to 1024 entries

CodeSandwich:
Fixes https://github.com/input-output-hk/jormungandr/issues/485


NicolasDP:
This PR didn't fix #485 at all

PR #519, opened 2019-06-17
bootstrap: fix printed example command

rinor:
bootstrap: fix rest printed example command (missing **get**)


vincenthz:
good catch thanks !

Issue #434, opened 2019-06-07
divide and reuse: jormungandr-lib as a library for shared data between jcli and jormugnadr


NicolasDP:
we have lot of code duplication between jormungandr, jcli and the tests, we need to harmonise them and consolidate them in a same library to facilitate the testing efforts and the daily development of the different projects.

PR #522, opened 2019-06-18
Unify jormungandr API Types to allow better reusability

NicolasDP:
This provides some interfaces and standard representation of some types. As an example on how to use this, this PR unifies the Fragment's Logs API across Jormungandr (the node) and the integration tests.

`jormungandr-lib` will replace `jormungandr-utils` on the long run.

On the long run, this library will provide interfaces for all the objects displayed to users or used in public APIs.

this PR bootstrap the work for #434


Cargo.toml
```diff
@@ -1,6 +1,7 @@
 [workspace]
 members = [
   "jormungandr-utils",
+  "jormungandr-lib",
```
mzabaluev: Is this different in purpose from jormungandr-utils so that this crate proliferation is justified?

mzabaluev: Alternatively, -utils could be morphed into this.

NicolasDP: it is going to replace it.

NicolasDP: it was stated in the commit message of the first commit, but not repeated in the PR description:

> add introduce jormungandr-lib, a set of components to replace jormungandr-utils

mzabaluev: I'd just do it right away then.

NicolasDP: too much code change/churn at once

PR #527, opened 2019-06-18
Move genesis.yaml initial state to single list

### LINKED ISSUE #403: Genesis file: preserve order of ledger operations
At the moment the genesis file specification only allow 2 fields:

* initial_funds
* initial_certs

which are process in a specific order chosen by the code. This create issue when funds and certs has complex relations (for example creating account first, then putting funds).

Genesis file specification need to move to sequence of operations (not using yaml dictionary, but sequence) (syntax for illustration purpose):

```
...
initial:
  - fund:
     utxo1
     utxo2
  - cert:
     cert1
     cert2
  - fund:
     utxo3
     account1
   - old_fund:
      utxo4
      utxo5  
```

CodeSandwich:
Fixes #403.

Also adds requirement that genesis.yaml can't have unknown fields. This will reject obsolete files instead of silently ignoring their content and generating invalid block.

Also updates docs for genesis.yaml, some fields were obsolete and some missing

### DIFF
--- a/jcli/src/jcli_app/block/DOCUMENTED_EXAMPLE.yaml
+++ b/jcli/src/jcli_app/block/DOCUMENTED_EXAMPLE.yaml
@@ -54,22 +54,19 @@ blockchain_configuration:
   # The speed to update the KES Key in seconds
   kes_update_speed: 43200 # 12hours
 
-# The initial state of the ledger. Each item is applied in order of this list
+# Initial state of the ledger. Each item is applied in order of this list
 initial:
-  # The initial deposits present in the blockchain.
-  #
-  # It can be UTxO addresses or account (if allow_account_creation is set).
+  # Initial deposits present in the blockchain
   - fund:
-      # this is the personal address of the developers, have a good heart and
-      # leave a good initial deposit for them
+      # UTxO addresses or account
       address: {initial_funds_address}
       value: 10000
 
-  # The initial certificate
+  # Initial certificate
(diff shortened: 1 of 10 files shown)


jcli/src/jcli_app/block/DOCUMENTED_EXAMPLE.yaml
```diff
@@ -54,21 +54,22 @@ blockchain_configuration:
...
-initial_funds:
-  # this is the personal address of the developers, have a good heart and
-  # leave a good initial deposit for them
-  - address: {initial_funds_address}
-    value: 10000
+# The initial state of the ledger. Each item is applied in order of this list
+initial:
+  # The initial deposits present in the blockchain.
+  #
+  # It can be UTxO addresses or account (if allow_account_creation is set).
```
rinor: Can we remove `allow_account_creation` mentioning, or it is still a valid directive?

CodeSandwich: No, it's not, they exist only in docs, which seem outdated

CodeSandwich:
Please don't merge it yet, the test failure is a bit too consistent. It usually fails (but sometimes passes), and before it was ok, so I should fix it.

NicolasDP:
update the `scripts/*` too

CodeSandwich:
Tests fixed and scripts updated

PR #541, opened 2019-06-19
Include UTxO and accounts

nahern:
Include sections on the concepts of UTxO and accounts


vincenthz:
this only add a placeholder

PR #544, opened 2019-06-19
Update blockchain.md

nahern:
Please add in these concepts up front and define:
- Delegation
- Stake pool


vincenthz:
this only add a placeholder

PR #545, opened 2019-06-19
Update node.md

nahern:
Mention files that the node needs to start 
Consider adding in logging considerations information here


vincenthz:
this only add placeholders

PR #551, opened 2019-06-20
Update README.md

nahern:
replace everything about running a node in the README by a link to the quickstart section
delete 'full'


vincenthz:
description doesn't seems to match what this is doing. I'm not sure I see a point of renaming "full node" to "node"

PR #520, opened 2019-06-17
Beginner friendly how-tos. related to #515

elviejo79:
Tries to solve @dorin100 request #515 

> I would add a new chapter named "How to start a Self-Node" that would contain details on how to execute the bootstrap script (https://github.com/input-output-hk/jormungandr/blob/master/scripts/bootstrap) in order to generate all the required files in order to start a node.

PR #553, opened 2019-06-20
adding How tos and other changes

nahern:


vincenthz:
@elviejo79 it doesn't change what I said in #520 , it doesn't seem appropriate in the user-guide and this more akin to a blog post. Potentially the howto part could go in the [wiki](https://github.com/input-output-hk/jormungandr/wiki).

Also as a side note the history is messy, there's quite a few typos (including in the instructions) and the content is reflecting old version (e.g. delegation field is now on 1 line).

PR #554, opened 2019-06-21
Pull missing blocks

mzabaluev:
Handle the case when the blockchain triage finds that a
block announced by a network peer is missing parents in local
storage.

WIP: processing of received chain headers is not implemented yet, will be added in a follow-up PR.

### DIFF
--- a/jormungandr/src/blockchain/chain.rs
+++ b/jormungandr/src/blockchain/chain.rs
@@ -267,6 +267,21 @@ impl Blockchain {
             .and_then(|mut iter| iter.next())
             .map(|leadership| leadership.1.into())
     }
+
+    /// Retrieves a list of checkpoint block hashes to pass to another
+    /// peer for recovery of missing chain blocks. The checkpoints
+    /// traverse back in the history with exponentially receding offsets.
+    pub fn get_checkpoints(&self) -> Result<Vec<HeaderHash>, StorageError> {
+        let tip = self.get_tip().unwrap();
+        let storage = self.storage.read().unwrap();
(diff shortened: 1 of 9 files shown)


jormungandr/src/blockchain/chain.rs
```diff
@@ -267,6 +267,21 @@ impl Blockchain {
...
+
+    /// Retrieves a list of checkpoint block hashes to pass to another
+    /// peer for recovery of missing chain blocks. The checkpoints
+    /// traverse back in the history with exponentially receding offsets.
+    pub fn get_checkpoints(&self) -> Result<Vec<HeaderHash>, StorageError> {
+        let tip = self.get_tip().unwrap();
+        let storage = self.storage.read().unwrap();
+        let tip_info = storage.get_block_info(&tip)?;
+        let mut checkpoints = Vec::new();
+        assert!(tip_info.depth > 0);
```
NicolasDP: This is WIP toward better error report? Or should it be an `assert_debug!` instead?

jormungandr/src/client.rs
```diff
@@ -65,33 +68,36 @@ fn handle_get_headers_range(
...
+    } else {
         /* Start at the newest checkpoint. */
         checkpoints.sort_unstable();
         let from = checkpoints.last().unwrap().1;
 
         // FIXME: handle checkpoint == genesis
 
         /* Send headers up to the maximum. */
-        let mut headers = vec![];
+        let mut header_count = 0usize;
```
NicolasDP: you should not need to specify the `usize` here.

PR #557, opened 2019-06-21
mark gelf as optional feature

vincenthz:


jbgi:
the documentation might need to be updated to reflect that gelf support is guarded by a feature flag now.

PR #558, opened 2019-06-21
re-generate lockfile and add CI to ensure Cargo.lock is up-to-date.

jbgi:
(0.2.2 fails to build with nix)


vincenthz:
probably best to not do a `cargo update` for a release, just going to update the minimum with `build`.

Does it make sense to combine rustfmt and lockfile in circle-ci ?

PR #559, opened 2019-06-21
Proposal to replace ENTRYPOINT with CMD in Dockerfile

wang2bo2:
This change will allow easier overriding, so we could do things like the following:

1. Start the node in the background
```bash
docker run -d --name j9r wang2bo2/j9r
```
2. Run scripts or jcli in separate containers:
```bash
docker run -t --rm --network=container:j9r wang2bo2/j9r create-account-and-delegate.sh
```
```bash
docker run -t --rm --network=container:j9r wang2bo2/j9r faucet-send-money.sh <account_addr> 10000000
```
```bash
docker run -t --rm --network=container:j9r wang2bo2/j9r jcli rest v0 account get <account_addr> -h "http://127.0.0.1:8448/api"
```

I have also made `${ENV_PREFIX}/bin` the `WORKDIR` because the scripts are referencing each other with `./`.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -6,6 +6,7 @@ LABEL description="Jormungandr latest"
 
 ARG PREFIX=/app
 ARG REST_PORT=8448
+ARG GRPC_PORT=8299
 
 ENV ENV_PREFIX=${PREFIX}
 
@@ -45,6 +46,6 @@ RUN mkdir -p ${ENV_PREFIX}/cfg && \
 EXPOSE ${REST_PORT}
 
 # expose gRPC port
-EXPOSE 8299 
+EXPOSE ${GRPC_PORT} 
 
 CMD [ "sh", "startup_script.sh" ]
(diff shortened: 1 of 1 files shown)


vincenthz:
Thanks, the use cases you listing above is definitely what we're aiming to do here, there's also some argument to bootstrap that could be docker ARG'ed

wang2bo2:
> Thanks, the use cases you listing above is definitely what we're aiming to do here, there's also some argument to bootstrap that could be docker ARG'ed

👌 I figured. I understand this part is still fledging; there're multiple other things could be improved:
1. Instead of cloning the repo's default branch in build, we could either use the local repo or add the version as an build arg and clone that version. These will help versioning/publishing these images.
2. For future deployments, it's probably easier to mount the config and secret instead of building them into the image. That means users are responsible for bootstrapping, and we don't have to call bootstrap in build.

wang2bo2:
> Thanks, the use cases you listing above is definitely what we're aiming to do here, there's also some argument to bootstrap that could be docker ARG'ed

BTW, congrats on getting it out!

PR #561, opened 2019-06-22
Update README.md

jschunick:
Additional install instructions required for a Windows system


vincenthz:
That doesn't appear windows specific nor marked as such (maybe WSL specific).
For example, when using git bash (msys2), I can follow the step in the README without adding anything.

PR #567, opened 2019-06-24
Docker: use alpine base image and versioned releases

wang2bo2:
If the docker image is meant for deployment rather than as a nightly testbed:
* This proposal reduced image size from 403 MB to 32 MB.
* Add build arg `VERSION`, so we can build image for a specific release
```bash
docker build --build-arg VERSION=0.2.2 -t j9r:0.2.2 docker/.
```


vincenthz:
ok, I think this has a very different use cases. I think we should have a way to start a released version, but also the latest version.

Maybe we need another dockerfile (and start making directories), one for releases and one for development

mzabaluev:
It's become common practice to have both Debian/Ubuntu based images and Alpine based images. Maybe also a Fedora container one for the 5 of us who care :grin:

onicrom:
> 
> 
> ok, I think this has a very different use cases. I think we should have a way to start a released version, but also the latest version.
> 
> Maybe we need another dockerfile (and start making directories), one for releases and one for development

We could make it another build-arg... build from source OR download a version, make the default download binary.  I can probably work on that tonight

wang2bo2:
Very good everyone; I agree.

On the deployment and development split: it makes sense. For the deployment image, it's anyway better to move config and secrets out while it's not necessarily so for the development image.

On different flavors of linux: sure. If people wants to build upon this image, they probably prefer a flavor they are comfortable with. Alpine's package manager is not for everyone. 🤷‍♂

vincenthz:
I'm happy to merge except it has a newline change in the original dockerfile, please fix it to make sure no one else is impacted by an unnecessary change

https://github.com/input-output-hk/jormungandr/pull/567/files#diff-ebacf6f6ae4ee68078bb16454b23247dR46

wang2bo2:
> I'm happy to merge except it has a newline change in the original dockerfile, please fix it to make sure no one else is impacted by an unnecessary change
> 
> https://github.com/input-output-hk/jormungandr/pull/567/files#diff-ebacf6f6ae4ee68078bb16454b23247dR46

The newline is back

PR #568, opened 2019-06-24
Update blockchain.md

nahern:
please add this paragraph at the end of consensus section.


dorin100:
This is meant to prevent issues like this: https://github.com/input-output-hk/shelley-testnet/issues/12

PR #569, opened 2019-06-24
bootstrap script ported to Powershell (Windows without WBS)

gufmar:
Does not (yet) include command line options, but automatically downloads the latest release on first run. 
Just start this script in an empty project folder.  (Windows users will like it ;)

On further runs the script checks for an updated release and ask the user if he wants to upgrade. Also, asks if the existing config/secret/storage folder should remain or recreated.


mzabaluev:
@dkijania Are you working together on this? How about splitting it into two more topical PRs?

gufmar:
Don't know what you mean with working together?
I created this out of interest how ps can do the same, and btw understanding better the detailed steps and many ways that can go wrong.

I'm open for all, as long as it fits in my free time. What PR splits do you have in mind? We may discuss first on gitter or telegram...

gufmar:
closed for a new clean attempt

PR #572, opened 2019-06-24
bootstrap script ported to Windows Powershell

gufmar:
Does not (yet) include command line options, but automatically downloads the latest release on first run.
Just start this script in an empty project folder. (Windows users will like it ;)

On further runs the script checks for an updated release and ask the user if he wants to upgrade. Also, asks if the existing config/secret/storage folder should remain or recreated.

### DIFF
--- a/scripts/bootstrap.ps1
+++ b/scripts/bootstrap.ps1
@@ -1,0 +1,288 @@
+#
+# The script creates a bft or genesis configuration with 1 faucet with a hardcoded amount
+# It is based on the bootstrap shell script at https://github.com/input-output-hk/jormungandr/tree/master/scripts
+#
+# Disclaimer:
+#
+#  The following use of Powershell script is for demonstration and understanding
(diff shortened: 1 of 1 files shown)


scripts/bootstrap.ps1
```diff
@@ -0,0 +1,288 @@
...
+$CONSENSUS="genesis_praos"
+$DATA_PATH="storage"
+$SECRET_PATH="secret"
+$CONFIG_PATH="config"
+$ADD_STARTUP_SCRIPT=0
+
+$STORAGE_PATH = $WORKDIR -replace "\\", "/"
+$STORAGE_PATH = $STORAGE_PATH + "/" + $DATA_PATH + "/"
+
+$BLOCK0_DATE=[int][double]::Parse((Get-Date -UFormat %s))
```
gufmar:
```suggestion
$BLOCK0_DATE=[int][double]::Parse((Get-Date(Get-Date).ToUniversalTime() -UFormat %s))
```

PR #574, opened 2019-06-24
Improve GELF logging backend configuration

mzabaluev:
Move config fields that make sense only for GELF under `gelf`, which must now be a property map used as the value of `logger.output`.
Removed `gelf` as an acceptable parameter for command line option `--log-output`, which did not work anyway due to lack of the required settings.
Improved documentation for config YAML overall.


vincenthz:
good, that looks like solving exactly what I asked in the initial PR :)

Issue #570, opened 2019-06-24
network-grpc 0.1.0-dev not found during build


mmahut:
**Describe the bug**

I'm trying to build the latest release 0.2.3 on NixOS. The build fails with:

```
   Compiling tower-grpc-build v0.1.0
     Running `CARGO=/nix/store/9lqkyz6ywr7i4xiws250799601mn9j46-cargo-1.34.2/bin/.cargo-wrapped CARGO_MANIFEST_DIR=/build/jormungandr-0.2.3-vendor/tower-grpc-build CARGO_PKG_AUTHORS='Tower Maintainers <team@tower-rs.com>' CARGO_PKG_DESCRIPTION='Code generation for tower-grpc
' CARGO_PKG_VERSION_MINOR=1 CARGO_PKG_VERSION_MAJOR=0 CARGO_PKG_VERSION_PATCH=0 CARGO_PKG_HOMEPAGE='https://github.com/tower-rs/tower-grpc' CARGO_PKG_VERSION=0.1.0 CARGO_PKG_VERSION_PRE= LD_LIBRARY_PATH='/build/jormungandr/target/release/deps:/nix/store/cn8p3kc4gv6l4407b5yjgb71w9g6hgpl-rustc-1.34.2/lib' CARGO_PKG_NAME=tower-grpc-build CARGO_PKG_REPOSITORY='https://github.com/tower-rs/tower-grpc' rustc --edition=2018 --crate-name tower_grpc_build /build/jormungandr-0.2.3-vendor/tower-grpc-build/src/lib.rs --color never --crate-type lib --emit=dep-info,link -C opt-level=3 --cfg 'feature="tower-hyper"' -C metadata=3faf55df822a9460 -C extra-filename=-3faf55df822a9460 --out-dir /build/jormungandr/target/release/deps -C linker=/nix/store/7nvy9l5jcp1c6z6n96l1a5x1bwbvamqn-gcc-wrapper-7.4.0/bin/cc -L dependency=/build/jormungandr/target/release/deps --extern codegen=/build/jormungandr/target/release/deps/libcodegen-06eee375f807e823.rlib --extern heck=/build/jormungandr/target/release/deps/libheck-d19926d2f846c990.rlib --extern prost_build=/build/jormungandr/target/release/deps/libprost_build-4234434a3f1ba85b.rlib --cap-lints warn`
[...]

Issue #56, opened 2018-12-10
Problem finding protoc on NixOS


qnikst:
When trying to install `jormungandr` on NixOS machine protobuf generation fails with the no such directory error:

```
     Running `/home/qnikst/workspace/tweag/jormungandr/target/debug/build/jormungandr-1c68e7af3d29fb06/build-script-build`
[jormungandr 0.0.1] No such file or directory (os error 2)
error: failed to run custom build command for `jormungandr v0.0.1 (/home/qnikst/workspace/tweag/jormungandr)`
process didn't exit successfully: `/home/qnikst/workspace/tweag/jormungandr/target/debug/build/jormungandr-1c68e7af3d29fb06/build-script-build` (exit code: 1)
--- stderr
No such file or directory (os error 2)
```
It happens that tower-grps-build is trying to access non-existing locale directories

```
24897 openat(AT_FDCWD, "/nix/store/fg4yq8i8wd08xg3fy58l6q73cjy8hjr2-glibc-2.27/share/locale/en_US.UTF-8/LC_MESSAGES/libc.mo", O_RDONLY) = -1 ENOENT (No such file or directory)
24897 openat(AT_FDCWD, "/nix/store/fg4yq8i8wd08xg3fy58l6q73cjy8hjr2-glibc-2.27/share/locale/en_US.utf8/LC_MESSAGES/libc.mo", O_RDONLY) = -1 ENOENT (No such file or directory)
24897 openat(AT_FDCWD, "/nix/store/fg4yq8i8wd08xg3fy58l6q73cjy8hjr2-glibc-2.27/share/locale/en_US/LC_MESSAGES/libc.mo", O_RDONLY) = -1 ENOENT (No such file or directory)
24897 openat(AT_FDCWD, "/nix/store/fg4yq8i8wd08xg3fy58l6q73cjy8hjr2-glibc-2.27/share/locale/en.UTF-8/LC_MESSAGES/libc.mo", O_RDONLY) = -1 ENOENT (No such file or directory)
24897 openat(AT_FDCWD, "/nix/store/fg4yq8i8wd08xg3fy58l6q73cjy8hjr2-glibc-2.27/share/locale/en.utf8/LC_MESSAGES/libc.mo", O_RDONLY) = -1 ENOENT (No such file or directory)
[...]

PR #582, opened 2019-06-26
Add notes on protobuf and C compilers to the install steps

### LINKED ISSUE #581: cargo install compile fail
after cfg-if v0.1.9
compiling error: linker 'cc' not found
no such file or dir (os err 2)
Screenshot (4)

mzabaluev:
Several of the dependencies fail to build without `cc`.

Resolves #581

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,12 +26,17 @@ Note:
   the compiler and linker executable `cc` is found in `PATH`.
 * On Linux with systemd: to enable logging to journald replace step 6
   with `cargo install --path . --features systemd`.
-* On Linux environments without glibc, such as NixOS or Alpine, the protobuf
-  compiler `protoc` needs to be installed and found in `PATH` or otherwise
-  specified in the environment variable `PROTOC`. For distribution or container
-  builds in general, it's a good practice to install protoc from the
-  official distribution package if available, otherwise the version bundled
-  with crate `prost-build` will be used.
+* The build requires the [Protocol Buffers][protobuf] compiler:
+  - On Linux environments without glibc such as Alpine, the protobuf compiler
+    `protoc` needs to be installed and found in `PATH` or otherwise
+    specified in the environment variable `PROTOC`.
+  - NixOS users should rely on [shell.nix](shell.nix) provided in this source
+    tree to pull the dependencies and set up the environment for the build.
+  - For distribution or container builds in general, it's a good practice to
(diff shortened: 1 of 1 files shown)


vincenthz:
I was going to merge this, but I wonder if this is actually accurate. should it'ld be a pre-requisite step for just linux/osx ?

gufmar:
On Windows with the VS C++ build tools 2019 it's also not that easy. The error message then complains about a missing linker.exe

mzabaluev:
@gufmar Yes, that's why I was vague about non-Linux in that notice :) IIRC the `cc` helper crate looks in the Windows registry to find an installed toolchain.

mzabaluev:
> I was going to merge this, but I wonder if this is actually accurate. should it'ld be a pre-requisite step for just linux/osx ?

I believe most of the crates that need a C linker also need it on the Windows targets. I'm not sure, but it might be a prerequisite for any crates that link -sys libraries.

mzabaluev:
The Visual C++ toolchain is also a prerequisite for installing rustup.

mzabaluev:
I have improved the install notes overall and added a notice about the protobuf compiler due to issues people are having with it (#56, #570, #590).

README.md
```diff
@@ -19,9 +19,19 @@ we recommend to use the most recent stable version of the rust compiler.
...
-* on linux and OSX: add `${HOME}/.cargo/bin` to your `${PATH}`
-* on linux with systemd: to enable logging to journald replace step 5. with `cargo install --path . --features systemd`
+* On Windows, you'll need to add the `%USERPROFILE%\.cargo\bin` into the
+  environment variable `PATH`.
+* On Linux and macOS: add `${HOME}/.cargo/bin` into your `PATH`.
+* Make sure the C compiler toolchain is installed and, on Unix (e.g. macOS),
+  the compiler and linker executable `cc` is found in `PATH`.
+* On Linux with systemd: to enable logging to journald replace step 6
+  with `cargo install --path . --features systemd`.
+* On Linux environments without glibc, such as NixOS or Alpine, the protobuf
```
edolstra: On NixOS you shouldn't install such tools manually but rely on `shell.nix` (which sets `PROTOC`).

mzabaluev: Thank you, I am chastised :bowing_man: Fixed in 00c0064.

PR #594, opened 2019-06-29
use binaries by default and support building source too

onicrom:
- Support binary or source installation of jormungandr and jcli
- Support changing the version, default v0.2.3
- Add README to provide info on how to use the --build-args

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -5,34 +5,54 @@ LABEL MAINTAINER IOHK
 LABEL description="Jormungandr latest"
 
 ARG PREFIX=/app
-ARG REST_PORT=8448
-
 ENV ENV_PREFIX=${PREFIX}
+ARG REST_PORT=8448
+ARG BUILD=false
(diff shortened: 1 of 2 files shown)


docker/Dockerfile
```diff
@@ -5,34 +5,55 @@ LABEL MAINTAINER IOHK
...
-    rm -rf /var/lib/apt/lists/*
-
-# install the node and jcli from source
-RUN mkdir -p ${ENV_PREFIX} && \
+    apt-get install -y git curl && \
+    mkdir -p ${ENV_PREFIX} && \
     mkdir -p ${ENV_PREFIX}/src && \
     mkdir -p ${ENV_PREFIX}/bin && \
+    cd ${ENV_PREFIX} && \
+    git clone --recurse-submodules https://github.com/input-output-hk/jormungandr src && \
```
vincenthz: this is just cloned to get the script files in the case of build=false .. that seems a bit overkill seems curl should grab them from the repository in raw mode instead

PR #595, opened 2019-07-01
updated delegation script 

dorin100:
These changes should help to avoid issues like input-output-hk/shelley-testnet#12

- updated delegation script in order to wait for transactions to be included into blocks
- added a comment for each operation in order to highlight the steps to have the delegation created and verified
- removed the debug command (set -ex) in order to keep only the steps into de log (otherwise the waitNewBlockCreated function is quite noisy)

### DIFF
--- a/scripts/create-account-and-delegate.shtempl
+++ b/scripts/create-account-and-delegate.shtempl
@@ -16,6 +16,7 @@ FEE_COEFFICIENT=####FEE_COEFFICIENT####
 ADDRTYPE="####ADDRTYPE####"
 STAKE_POOL_ID="####STAKE_POOL_ID####"
 SLOT_DURATION="####SLOT_DURATION####"
+TIMEOUT_NO_OF_BLOCKS=30
 
 ### COLORS
 if [ ${COLORS} -eq 1 ]; then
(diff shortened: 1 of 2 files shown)


dorin100:
These changes should also help to avoid issues like https://github.com/input-output-hk/shelley-testnet/issues/12

scripts/create-account-and-delegate.shtempl
```diff
@@ -34,11 +34,36 @@ if [ ${#} -ne 0 ]; then
...
+    echo $($CLI rest v0 tip get -h "${REST_URL}")
+}
+
+function waitNewBlockCreated {
+    echo "  ##Waiting for new block to be created (timeout = $(( 50*$SLOT_DURATION ))s)..."
+    COUNTER=0
+    initialTip=$(getTip)
+    actualTip=$(getTip)
+
+    while [[ ${actualTip} = ${initialTip} ]]; do
```
NicolasDP: this will always be true

NicolasDP: This will always be true:

```sh
    initialTip=$(getTip)
    actualTip=$(getTip)
```

dorin100:
ok. doing it now

PR #601, opened 2019-07-02
Pulling missing chain blocks from the network

mzabaluev:
Continuation of #554.

Adds a strategy of bumping the peer that has announced a block to the last in the linear order of peer entries. When blocks are fetched, the peers are to be picked from last to first.


vincenthz:
looks like a nice improvement. particularly nice to see abstraction like Peers to abstract the lowlevel choices of data structure

PR #605, opened 2019-07-04
Add proper error reporting to JCLI REST commands

CodeSandwich:
Fixes https://github.com/input-output-hk/jormungandr/issues/298

### DIFF
--- a/jcli/src/jcli_app/rest/mod.rs
+++ b/jcli/src/jcli_app/rest/mod.rs
@@ -1,5 +1,6 @@
 mod v0;
 
+use jcli_app::utils::{host_addr, output_format};
 use structopt::StructOpt;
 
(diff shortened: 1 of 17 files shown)


jcli/src/jcli_app/rest/mod.rs
```diff
@@ -10,8 +11,37 @@ pub enum Rest {
     V0(v0::V0),
 }
 
+const SERIALIZATION_ERROR_MSG: &'static str = "node returned malformed data";
```
NicolasDP: what is the rational for this const?

CodeSandwich: The same issue can be manifested in two kinds of errors and this ensures that it is presented in consistent way to the user

PR #615, opened 2019-07-08
Introduce `--quiet` CLI option and `quietness` configuration option.

jonathanknowles:
### CLI Option

The `--quiet` CLI option works in a similar way to the preexisting `--verbose` CLI option, but in reverse.

The more occurrences of `--quiet`, the quieter the output.

The `--quiet` and `--verbose` options are **mutually exclusive**.

### Configuration Option

The `quietness` configuration option works in a similar way to the preexisting `verbosity` configuration option, but in reverse.

The higher the level of `quietness`, the quieter the output.

The `quietness` and `verbosity` options are **mutually exclusive**.

### DIFF
--- a/jormungandr/src/settings/command_arguments.rs
+++ b/jormungandr/src/settings/command_arguments.rs
@@ -67,10 +67,17 @@ pub struct StartArguments {
     raw(setting = "structopt::clap::AppSettings::ColoredHelp")
 )]
 pub struct CommandLine {
-    /// activate the verbosity, the more occurrences the more verbose.
+    /// Activate quiet mode.
+    /// The more occurrences, the quieter the output.
+    /// (-q, -qq, -qqq)
+    #[structopt(short = "q", long = "quiet", parse(from_occurrences))]
+    pub quietness: u8,
+
+    /// Activate verbose mode.
+    /// The more occurrences, the more verbose the output.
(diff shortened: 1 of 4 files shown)


jonathanknowles:
Hi @NicolasDP 

Thanks for taking a look at this.

> Would not it be better to have a `--log-level=LOG_LEVEL` with default is _value to be determined_ ?

Actually, the Cardano Wallet CLI used to have a `--log-level` option, just as you've proposed. (We had a default of `INFO`). Though eventually the team decided to replace this with `--quiet` and `--verbose` options, as people felt these options were:

* user-friendlier (as users aren't necessarily familiar with the semantics of log levels) 
* standard across many CLI tools (http://tldp.org/LDP/abs/html/standard-options.html)

Of course, this is debatable.

In the end, I don't have a strong preference one way or the other. It's only important to us that we are able to make the output `jormungandr` quieter in the event that the user requests it. Having said that, I think it would be nice to have a similar "interface" across our tools.

NicolasDP:
Thanks for the feedback and explanation. I have created an issue to discuss/track changes that could be made to improve the log level settings in #622 . Which also cover the issue of inconsistency between the command line options (count the number of `--verbose` used) and the configuration file (use a number).

PR #616, opened 2019-07-08
make sure we use the latest stable available

NicolasDP:
it may be possible travis-ci image is not utilising the latest stable version (which is the one we officially support). In order to prevent cases where travis-ci image for rust does not contains the latest stable yet we can make sure to update it with `rustup`.

We don't need to handle cases for beta or nightly here as we only care about stable in travis ci

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -52,6 +52,7 @@ before_install:
   - rustup self update
 
 install:
+  - rustup udpate stable
   - sh ci/install.sh
   - source ~/.cargo/env || true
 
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -52,6 +52,7 @@ before_install:
   - rustup self update
 
 install:
+  - rustup udpate stable
```
mzabaluev: Typo: `update`
Would it be better to do this in `before_install`?

PR #617, opened 2019-07-08
Fix genesis.yaml consensus version naming

CodeSandwich:
Genesis.yaml parser was expecting consensus version "genesis_praos" instead of "genesis". This is in conflict with documentation and string parsing of chain-impl-mockchain enum.


NicolasDP:
`praos` would be better I think

NicolasDP:
we will need to talk about it. Would be better to fix the doc and remove the string parsing in the chain-libs/impl-mockchain

CodeSandwich:
It's no longer a trivial fix. I'll close this PR and open an issue to settle the details: https://github.com/input-output-hk/jormungandr/issues/618

Issue #586, opened 2019-06-27
Node crash if sending multiple transactions in the same slot


dorin100:
**Describe the bug**
Node is crashing if we send multiple transactions in the same slot.

**To Reproduce**
1. generate files using: bash bootstrap -p 8607 -b | tee initialValues.txt
2. start the node with the suggested command: jormungandr --genesis-block ./block-0.bin --config ./config.yaml --secret ./pool-secret1.yaml
3. run the attached script (multipleTransactions.sh)

**Expected behavior**
Node should not crash.

**Additional context**
The script (multipleTransactions.sh) is creating an Account and then it sends funds thru 10 transactions from Faucet to the Account (using faucet-send-money.sh script).
If we add timeouts between transactions (= slot_duration), everything is ok.

Node and script logs (export RUST_BACKTRACE=1) attached.
[scriptLogs.txt](https://github.com/input-output-hk/jormungandr/files/3333582/scriptLogs.txt)
[nodeLogs.txt](https://github.com/input-output-hk/jormungandr/files/3333583/nodeLogs.txt)
[multipleTransactions.sh.txt](https://github.com/input-output-hk/jormungandr/files/3333595/multipleTransactions.sh.txt)

This bug was created in order to track https://github.com/input-output-hk/shelley-testnet/issues/22

mzabaluev:
Thank you for the logs. @vincenthz have you narrowed it down to the specific unwrap?
It would help if you ran it on a debug build (`cargo run --bin jormungandr -- ...` in the source directory).

rinor:
**os**: Linux 4.19.0-5-amd64 #1 SMP Debian 4.19.37-5 (2019-06-19) x86_64 GNU/Linux
**rustc --version**: rustc 1.35.0 (3c235d560 2019-05-20)
**node build from**: 71d73d4e1b1d64b9f2563ead6eb84c93168b475a

**command**: `RUST_BACKTRACE=full cargo run --bin jormungandr -- ...`

[bt-full2_testnet.log](https://github.com/input-output-hk/jormungandr/files/3335232/bt-full2_testnet.log)

mzabaluev:
With #588 in, logging of the error should be more graceful now.

rinor:
**node bulid from:** d239f842b15a322e5186372ba847e9268f15a41a

[bt-full3_testnet.log](https://github.com/input-output-hk/jormungandr/files/3344996/bt-full3_testnet.log)

[...]

PR #619, opened 2019-07-08
Fix node crashing when multiple TXs for same account are in slot

CodeSandwich:
Fixes https://github.com/input-output-hk/jormungandr/issues/586

### DIFF
--- a/jormungandr/src/blockchain/process.rs
+++ b/jormungandr/src/blockchain/process.rs
@@ -37,16 +37,14 @@ pub fn handle_input(
         }
         BlockMsg::LeadershipBlock(block) => {
             let mut blockchain = blockchain.lock_write();
-            match chain::handle_block(&mut blockchain, block, true)
-                .map_err(|e| crit!(logger, "block processing failed: {:?}", e))?
-            {
-                HandledBlock::Rejected { reason } => {
+            match chain::handle_block(&mut blockchain, block, true) {
+                Ok(HandledBlock::Rejected { reason }) => {
                     warn!(logger,
                         "rejecting node's created block" ;
                         "reason" => reason.to_string(),
                     );
                 }
-                HandledBlock::MissingBranchToBlock { to } => {
+                Ok(HandledBlock::MissingBranchToBlock { to }) => {
                     // this is an error because we are in a situation
                     // where the leadership has created a block but we
                     // cannot add it in the blockchain because it is not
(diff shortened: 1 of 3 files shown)


jormungandr/src/blockchain/process.rs
```diff
@@ -33,20 +33,18 @@ pub fn handle_input(
         BlockMsg::LeadershipExpectEndOfEpoch(epoch) => {
             let blockchain = blockchain.lock_read();
             chain::handle_end_of_epoch_event(&blockchain, epoch)
-                .map_err(|e| crit!(logger, "end of epoch processing failed: {:?}", e))?;
+                .map_err(|e| crit!(logger, "end of epoch processing failed: {:?}", e));
```
NicolasDP: not keen on silently ignoring the error though. This is critical and if the end of epoch processing fails : it means we will not be able to start processing the epoch, the loop is broken and actions need to be performed to address the issue.

CodeSandwich: Won't the node just skip its slot leadership? It's not going to be a leader of the next one, so other node will produce it and failing node at some point will get ultimate ledger truth from the network. At least that's what I thought, could be wrong though.

NicolasDP: This part is not about a _slot_.

NicolasDP: have a look at the leadership doc https://github.com/input-output-hk/jormungandr/blob/e8f0cc2684a9bd5e7d8758fc39299280bd24e111/jormungandr/src/leadership/mod.rs#L3-L85

NicolasDP: you can also quickly generate the dev doc with `cargo doc --open --no-deps` and explore code documentation for a better experience.

mzabaluev: I believe an error here can be only a serious internal problem that should cause the whole task to call it quits.

NicolasDP: for now at least. we will need to see what else we can do. But not now.

CodeSandwich: Ah, right, this is exact line is not about being a leader. I've restored the task death in that case.

There's also similar case for `BlockMsg::LeadershipBlock`, where `chain::handle_block` failure stopped being a task killer. This one IS about being a leader, is that ok? https://github.com/input-output-hk/jormungandr/pull/619/files#diff-d93e6a17addd0650d4e7ded097ca6bb0L41

PR #620, opened 2019-07-08
Process events in the client connection

mzabaluev:
Poll every `network::Client` connection for events once it's established.


mzabaluev:
Previously added unsafe code is failing in tests, will investigate.

PR #623, opened 2019-07-09
CircleCI: Streamline the workflow

mzabaluev:
Remove the build jobs that don't currently do anything useful that needs to be done separately from test jobs.
Don't use the workspace, as build results no longer need to be handed off to other jobs (this may be resurrected when coverage is to be introduced).

Clean up `restore_cache` steps: only precise cache hits are used for the test jobs, as verified by the `--frozen` option given to cargo.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,19 +6,20 @@ jobs:
     working_directory: /mnt/crate
     steps:
       - checkout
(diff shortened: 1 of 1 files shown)


mzabaluev:
Wrong idea about the build directory, will resubmit a more careful change.

Issue #624, opened 2019-07-09
jormungandr-integration-tests break 'cargo test'


mzabaluev:
The current integration tests depend on `jormungandr` and `jcli` binaries to be pre-built in the target output directory. This does not happen if `cargo test` is run without a preceding `cargo build` that builds the binaries.

## To Reproduce

1. Change into the workspace source directory in the shell.
2. Run `cargo clean`.
3. Run `cargo test`.

## Expected behavior

Same result as `cargo clean; cargo build && cargo test` would produce.

mzabaluev:
Possible solutions:

- [assert_cmd](https://crates.io/crates/assert_cmd)
- [escargot](https://crates.io/crates/escargot)

lazaridiscom:
I hit on this, calling `cargo test` with the assumption that everything necessary will be done. It failed, and I thought "well, guess I have to look for the test-docu now". I did not, as I was to bored.

Later i tried again, wondering "why does it work now?" (It seems I used `cargo build` in between). Then I saw this issue.

PR #625, opened 2019-07-09
CircleCI: Streamline the workflow

mzabaluev:
Eliminate separate build jobs as these only add workspace persistence overhead with no artefacts usable outside of the tests.
In the test jobs, build the tests in a separate step to reduce non-test output in the test step.

Clean up `restore_cache` steps: only precise cache hits are used for build and test jobs, as verified by the `--frozen` option given to cargo.


mzabaluev:
Separate build steps are only needed due to #624.

PR #626, opened 2019-07-10
Rework handling of inbound blocks and headers from the network

mzabaluev:
Deboxed the service implementation after an API change in network-core.
Added back pressure and early error reporting to the peer.

Depends on input-output-hk/chain-libs#44


mzabaluev:
The two-stage chain pull process cannot work without support from blockchain storage: chain needs to be able to pre-validate the header chain without blocks, so that the header stream can be processed quickly and blocks requested out of order with batching and load balancing.

PR #631, opened 2019-07-11
Rename Message to Fragment

mzabaluev:
Adapt to the recent naming changes in chain-libs.


vincenthz:
merging, I don't think the failure has anything to do with this

Issue #410, opened 2019-06-04
REST: Add a way to dump the settings of the node (e.g. transaction limits, ...)


vincenthz:
related #86

NicolasDP:
things that could be added:

- block0 hash
- start date of the blockchain;
- current slot date;
- current consensus;
- current fee settings;

Issue #86, opened 2018-12-13
Blockchain Settings: maximum number of fragment per block


NicolasDP:
We have currently control over the block size limits except the limit by design (see [block format](https://github.com/input-output-hk/chain-libs/blob/master/chain-impl-mockchain/doc/format.md#block-header): size of content _unsigned integer of 32bits_ (2,147,483,647)). We need the protocol to enforce a reasonable number.

Previously we had in the settings an option to do something similar: number of transaction per block. We need now to have max size of a block settings.

- [x] add in the settings a maximum block size;
- [x] add a way to enforce the size; 
- [x] update the fragment selection algorithm to take this into account for bit-packing;
- [x] update the block0 configuration to provide this value (with minimum and maximum check);

---

### Old description before re-design of the blockchain ledger.

we are missing a mean to extract the blockchain's setting, here we are missing the maximum number of transaction per block.

In the current implementation of the `Update` trait, the ChainState is not giving the correct value. It is giving the total number of transaction ever seen in the blockchain.

see what is being introduced in [PR85](https://github.com/input-output-hk/jormungandr/pull/85/files#diff-18dec51afac534e0691e0e22e4455c8a)

CodeSandwich:
Is getting it via REST endpoint enough?

NicolasDP:
Not the same thing. Here it is related to the blockchain settings not being correct. I'll provide an update to this issue as we have changed some internal design since and it is no longer clear what this is about.

PR #634, opened 2019-07-12
Add REST endpoint for getting node settings

CodeSandwich:
Fixes https://github.com/input-output-hk/jormungandr/issues/410 and probably https://github.com/input-output-hk/jormungandr/issues/86

### DIFF
--- a/doc/jcli/rest.md
+++ b/doc/jcli/rest.md
@@ -206,3 +206,34 @@ value: 990
 * `delegation` is the Stake Pool Identifier the account is delegating to.
   it is possible this value is not set if there is no delegation certificate
   sent associated to this account.
+
+
+## Node settings
+
+Fetches node settings
+
+```
+jcli rest v0 node settings get <options>
+```
+
+The options are
+
+- -h <node_addr> - see [conventions](#conventions)
+- --debug - see [conventions](#conventions)
+- --output-format <format> - see [conventions](#conventions)
+
+
+YAML printed on success
+
+```yaml
+---
+block0Hash: 8d94ecfcc9a566f492e6335858db645691f628b012bed4ac2b1338b5690355a7  # block 0 hash of
+block0Time: "2019-07-09T12:32:51+00:00"         # block 0 creation time of
+consensusVersion: bft                           # currently used consensus
+currSlotStartTime: "2019-07-09T12:55:11+00:00"  # current slot start time
+fees:                                           # transaction fee configuration
+  certificate: 4                                # fee per certificate
+  coefficient: 1                                # fee per every input and output
+  constant: 2                                   # fee per transaction
+maxTxsPerBlock: 100                             # maximum number of transactions in block
+```
--- a/jormungandr/src/leadership/task.rs
+++ b/jormungandr/src/leadership/task.rs
@@ -7,6 +7,7 @@ use crate::{
     intercom::BlockMsg,
     leadership::{LeaderSchedule, Leadership},
     secure::enclave::{Enclave, LeaderId},
+    stats_counter::StatsCounter,
     utils::async_msg::MessageBox,
 };
 use chain_core::property::ChainLength as _;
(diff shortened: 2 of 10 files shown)


jormungandr/src/leadership/task.rs
```diff
@@ -67,17 +70,21 @@ impl Task {
             fragment_pool,
             epoch_receiver,
             block_message,
+            stats_counter,
         }
     }
 
     pub fn start(self) -> impl Future<Item = (), Error = ()> {
-        let handle_logger = self.logger.clone();
+        let params = HandleLeadershipParams {
```
mzabaluev: Can `Task` and `HandleLeadershipParams` simply reuse a common structure?

jormungandr/src/stats_counter.rs
```diff
@@ -39,11 +41,27 @@ impl StatsCounter {
...
-    pub fn get_uptime_sec(&self) -> u64 {
+    pub fn uptime_sec(&self) -> u64 {
         self.stats.start_time.elapsed().as_secs()
     }
+
+    pub fn set_slot_start_time(&self, time: SystemTime) {
+        self.stats
+            .slot_start_time
+            .write()
+            .unwrap_or_else(PoisonError::into_inner)
```
mzabaluev: Does this really need to survive a poisoned mutex?

CodeSandwich: The mutex should never be poisoned, but even if it is, the underlying data is not corruptable with panics. Could be this or `.unwrap`, doesn't matter. I personally like avoiding statements capable of panicking.

jormungandr/src/stats_counter.rs
```diff
@@ -12,6 +12,7 @@ struct StatsCounterImpl {
     tx_recv_cnt: AtomicUsize,
     block_recv_cnt: AtomicUsize,
     start_time: Instant,
+    slot_start_time: RwLock<Option<SystemTime>>,
```
vincenthz: RwLock for this seems totally overkill

CodeSandwich: I couldn't cram SystemTime into a single Atomic (it takes 128 bits) and the data in this structure must be both Sync and internally-mutable :(

vincenthz: I would put it in 2 atomic u64. it really doesn't matter is someone grab this value in the middle of an update, and not totally sure we need to keep the nanoseconds precision anyway

NicolasDP: In jormungandr there is `SecondsSinceUnixEpoch` which is a `u64` already. Does not contain the nanoseconds precision. And it provides compatibility back into/from `SystemTime`.

doc/jcli/rest.md
```diff
@@ -206,3 +206,34 @@ value: 990
...
   it is possible this value is not set if there is no delegation certificate
   sent associated to this account.
+
+
+## Node settings
+
+Fetches node settings
+
+```
+jcli rest v0 node settings get <options>
```
sjmackenzie: shouldn't have `node` in the command, also `jcli rest v0 settings get --output-format yaml -h <host>` doesn't work.

CodeSandwich: Good one about the docs.

But I can't reproduce JCLI not working:
```
./jcli rest v0 settings get --output-format yaml -h http://127.0.0.1:8443/api
---
block0Hash: 8d94ecfcc9a566f492e6335858db645691f628b012bed4ac2b1338b5690355a7
block0Time: "2019-07-09T12:32:51+00:00"
consensusVersion: bft
currSlotStartTime: ~
fees:
  certificate: 4
  coefficient: 1
  constant: 2
maxTxsPerBlock: 100
```
What's the failure on your computer?

sjmackenzie: Yeah I edited it out because I forgot to append the `/api` :-) sorry about the noise.

CodeSandwich: Whew, good to hear :)

PR #636, opened 2019-07-15
Simplify slot_start_time storage to seconds

CodeSandwich:
Addresses https://github.com/input-output-hk/jormungandr/pull/634#discussion_r303317996

### DIFF
--- a/jormungandr-lib/src/time.rs
+++ b/jormungandr-lib/src/time.rs
@@ -255,33 +255,31 @@ impl From<time::SystemTime> for SecondsSinceUnixEpoch {
         system_time
             .duration_since(time::UNIX_EPOCH)
             // duration since UNIX EPOCH will never go beyond boundaries
+            .map(|duration| duration.as_secs())
+            .map(SecondsSinceUnixEpoch::from_secs)
             .unwrap()
-            .as_secs()
-            .into()
     }
 }
 
-impl From<u64> for SystemTime {
-    fn from(seconds_since_epoch: u64) -> Self {
+impl SystemTime {
+    pub fn from_secs_since_epoch(secs: u64) -> Self {
         // here we can safely unwrap as we are adding from UNIX_EPOCH (0)
         // and SecondsSinceUnixEpoch is always a positive integer
         // and seconds will always be within bounds
         time::UNIX_EPOCH
-            .checked_add(time::Duration::from_secs(seconds_since_epoch))
+            .checked_add(time::Duration::from_secs(secs))
             .unwrap()
             .into()
     }
 }
 
-impl From<u64> for SecondsSinceUnixEpoch {
-    fn from(seconds_since_epoch: u64) -> Self {
-        SecondsSinceUnixEpoch(seconds_since_epoch)
+impl SecondsSinceUnixEpoch {
+    pub fn from_secs(secs: u64) -> Self {
+        SecondsSinceUnixEpoch(secs)
     }
-}
 
-impl From<SecondsSinceUnixEpoch> for u64 {
-    fn from(seconds_since_epoch: SecondsSinceUnixEpoch) -> Self {
-        seconds_since_epoch.0
+    pub fn to_secs(self) -> u64 {
+        self.0
     }
 }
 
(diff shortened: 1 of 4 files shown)


jormungandr-lib/src/time.rs
```diff
@@ -250,6 +250,17 @@ impl From<SystemTime> for time::SystemTime {
     }
 }
 
+impl From<time::SystemTime> for SecondsSinceUnixEpoch {
+    fn from(system_time: time::SystemTime) -> Self {
+        system_time
+            .duration_since(time::UNIX_EPOCH)
+            // duration since UNIX EPOCH will never go beyond boundaries
```
mzabaluev: I would not say never, but 2^64 - 1 seconds is a long time ;)

jormungandr-lib/src/time.rs
```diff
@@ -262,6 +273,18 @@ impl From<u64> for SystemTime {
     }
 }
 
+impl From<u64> for SecondsSinceUnixEpoch {
+    fn from(seconds_since_epoch: u64) -> Self {
```
mzabaluev: I'm a bit miffed about these super-easy `From` conversions, also for `SystemTime`. I think there should be a specially named method to make the code reader pause and understand the implications.

jormungandr-lib/src/time.rs
```diff
@@ -262,6 +273,18 @@ impl From<u64> for SystemTime {
...
+    }
+}
+
+impl From<SecondsSinceUnixEpoch> for u64 {
+    fn from(seconds_since_epoch: SecondsSinceUnixEpoch) -> Self {
+        seconds_since_epoch.0
+    }
+}
+
 impl From<time::Duration> for Duration {
```
mzabaluev: And why do we even have our own `Duration`...

jormungandr-lib/src/time.rs
CodeSandwich: I think that this assumption was made with pre-1970 dates in mind, but yeah, it's a rather safe one and truly unrecoverable

jormungandr-lib/src/time.rs
CodeSandwich: Probably for Serde support

jormungandr-lib/src/time.rs
```diff
@@ -273,15 +273,13 @@ impl From<u64> for SystemTime {
     }
 }
 
-impl From<u64> for SecondsSinceUnixEpoch {
-    fn from(seconds_since_epoch: u64) -> Self {
+impl SecondsSinceUnixEpoch {
+    pub fn from_secs_since_epoch(seconds_since_epoch: u64) -> Self {
```
mzabaluev: This looks a bit tautological, but OK.

jormungandr-lib/src/time.rs
```diff
@@ -273,15 +273,13 @@ impl From<u64> for SystemTime {
...
         SecondsSinceUnixEpoch(seconds_since_epoch)
     }
-}
 
-impl From<SecondsSinceUnixEpoch> for u64 {
-    fn from(seconds_since_epoch: SecondsSinceUnixEpoch) -> Self {
-        seconds_since_epoch.0
+    pub fn into_secs_since_epoch(self: SecondsSinceUnixEpoch) -> u64 {
```
mzabaluev: The `.into()` part was actually fine, as there is no precondition to observe during conversion to plain integer. I'd leave it as the standard conversion trait.

jormungandr-lib/src/time.rs
NicolasDP: no, `from_secs` would be better. It will already be `SecondsSinceUnixEpoc::from_secs`. No need to specify again `from_secs_since_epoch`.

jormungandr-lib/src/time.rs
CodeSandwich: Now it's obvious what's the meaning of the output number. The integer could be anything, e.g. miliseconds since epoch, or seconds since startup, but now it's clearly seconds since epoch.

CodeSandwich: Renamed to `to_secs`

PR #637, opened 2019-07-15
Upgrade Slog to 2.5.1

CodeSandwich:
The goal is to clean up some ugliness around handling `slog::FilterLevel`

### DIFF
--- a/jormungandr/Cargo.toml
+++ b/jormungandr/Cargo.toml
@@ -23,7 +23,7 @@ serde = "1.0"
 serde_derive = "1.0"
 serde_json = "1.0.38"
 serde_yaml = "0.8"
-slog = { version = "^2.4", features = [ "max_level_trace", "release_max_level_trace" ] }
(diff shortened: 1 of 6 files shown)


jormungandr/Cargo.toml
```diff
@@ -36,6 +36,7 @@ cryptoxide = "0.1"
 futures    = "0.1"
 http = "0.1.16"
 hyper = "0.12"
+lazy_static = "1.3"
```
NicolasDP: why did you move this line?

NicolasDP: my bad...

rinor:
slog [2.5.1](https://crates.io/crates/slog/2.5.1) has been **yanked** and is causing issues with jormunandr v0.3.1 source install. Can we use 2.5.0 or https://github.com/slog-rs/slog/commit/c8aa6770fd160330d032511ce1258f093d2d9df3.

CodeSandwich:
I've asked what's wrong with 2.5.1: https://github.com/slog-rs/slog/issues/219, when we know that, it will be easier to decide what to do next.

Issue #411, opened 2019-06-04
REST: add a shutdown control


vincenthz:
Add a way to shutdown the node from a REST call. This could also solve #15 

For now, we ignore any security aspect related to admin control, anyone that can connect to the REST can shutdown the node for now.

PR #643, opened 2019-07-18
Add node shutdown over REST

CodeSandwich:
Fixes https://github.com/input-output-hk/jormungandr/issues/411

### DIFF
--- a/jormungandr/src/rest/server/mod.rs
+++ b/jormungandr/src/rest/server/mod.rs
@@ -5,18 +5,17 @@ mod error;
 
 pub use self::error::Error;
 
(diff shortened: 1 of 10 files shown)


jormungandr/src/rest/server/mod.rs
```diff
@@ -47,20 +47,14 @@ impl Server {
             let _ = sender.send(server_handler);
             if run_system {
                 actix_system.run();
+                process::exit(0);
```
NicolasDP: This is the most hidden way to stop the node.

CodeSandwich: I can move it to main and trigger with MPSC channel

PR #644, opened 2019-07-19
Created 2 new scripts derived from bootstrap

dorin100:
Created 2 new scripts derived from bootstrap - `one-transaction-per-slot-1-account.shtempl` and `multiple-transactions-per-slot-2-accounts`

- `one-transaction-per-slot-1-account.shtempl` is sending a configured number of transactions, from faucet to a new account, 1 transaction per slot
- `multiple-transactions-per-slot-2-accounts` is sending a configured number of transactions (between 2 accounts), as many as possible per slot, and in the end, it checks the balances of both accounts.


vincenthz:
I'm not a fan to add those scripts in there, which are mostly stress testing script. Also shell is nice to do quick and dirty tools, it's far from ideal for making regression/stress test that will stand the test of time and are actually reliable. It would make sense to start combining rust library / bindings / jcli with some lightweight languages like python/ruby/javascript... for QA to do proper testing

PR #652, opened 2019-07-23
[Test] Recovery test for abrupt node restart

dkijania:
New test case with flow:

1) start node
2) do some transaction
3) restart node
4) compare if settings/utxo/account-states are the same like before restart


jormungandr-integration-tests/src/common/data/settings.rs
```diff
@@ -0,0 +1,27 @@
...
+#[derive(Debug, Serialize, Deserialize, Clone)]
+#[serde(rename_all = "camelCase")]
+pub struct Settings {
+    pub block0_hash: String,
+    pub block0_time: SystemTime,
+    pub curr_slot_start_time: SystemTime,
+    pub consensus_version: String,
+    pub fees: LinearFees,
+    pub max_txs_per_block: u8,
+}
```
NicolasDP: this is to decode the settings returned by the JCLI command?

I think it would be much better to have something in `jormungandr-lib` that can be used in both library.

(ping @CodeSandwich )

CodeSandwich: I've checked and format of this JSON is not engraved in any DTO, it's built with a `json!` macro ad-hoc inside REST handler. `jormungandr-ingegration-tests` depends on `jormungandr` anyway, so we probably could just create a public DTO in `jormungandr::rest::v0::???` for everyone to use.

CodeSandwich: @dkijania You can rebase onto `rest_settings_dto` to get that public DTO :)

dkijania: I've made some modifications because it looks like jormungandr-integration-tests cannot use structs from application. So i moved SettingsDto to jormungandr-lib

CodeSandwich: I think it's because `jormungandr` is defined as a `dev-dependency` of `integration-tests`, I'm not sure if that's what we want, could be. Or maybe we should remove this dependency completely, not sure really.

dkijania: i tried to move it as a `dependency` but with the same outcome

jormungandr-lib/src/interfaces/settings.rs
```diff
@@ -0,0 +1,31 @@
...
+            && self.max_txs_per_block == other.max_txs_per_block
+    }
+}
+
+#[derive(Clone, Debug, Deserialize, Serialize, PartialEq)]
+#[serde(rename_all = "camelCase")]
+pub struct SettingsFeesDto {
+    pub constant: u64,
+    pub coefficient: u64,
+    pub certificate: u64,
```
NicolasDP: There is already a `LinearFee` object to do the serialize/deserialize for serde in `jormungandr-lib`

It is used here: https://github.com/input-output-hk/jormungandr/blob/3c221c3c6865b19e5964d69e82e8c26bd8181c8b/jormungandr-lib/src/interfaces/block0_configuration/initial_config.rs#L59-L60

and defined there: https://github.com/input-output-hk/jormungandr/blob/3c221c3c6865b19e5964d69e82e8c26bd8181c8b/jormungandr-lib/src/interfaces/block0_configuration/initial_config.rs#L293-L299

It can be moved to its own module within the `interface`

jormungandr-lib/src/interfaces/settings.rs
```diff
@@ -1,14 +1,16 @@
...
+#[derive(Debug, Clone, Eq, Serialize, Deserialize)]
+#[serde(deny_unknown_fields, rename_all = "camelCase")]
 pub struct SettingsDto {
     pub block0_hash: String,
     pub block0_time: SystemTime,
     pub curr_slot_start_time: Option<SystemTime>,
     pub consensus_version: String,
-    pub fees: SettingsFeesDto,
+    #[serde(with = "LinearFeeDef")]
+    pub fees: LinearFee,
```
NicolasDP: excellent

Issue #412, opened 2019-06-04
REST: add leader control api


vincenthz:
We need endpoints for:

* [ ] Add way to add new leader in enclave
* [ ] Add ability to remove leader in enclave
* [ ] List leaders

Just like #411, ignore the security aspect

PR #654, opened 2019-07-24
Add leadership management REST API

CodeSandwich:
Fixes https://github.com/input-output-hk/jormungandr/issues/412


jormungandr/src/rest/mod.rs
```diff
@@ -23,6 +24,7 @@ pub struct Context {
     pub transaction_task: Arc<Mutex<MessageBox<TransactionMsg>>>,
     pub logs: Arc<Mutex<Logs>>,
     pub server: Arc<RwLock<Option<Server>>>,
+    pub enclave: Enclave,
```
vincenthz: not a big fan of this for fencing purpose, but for now it will do (and can be properly fenced later)

PR #657, opened 2019-07-25
network: Perform protocol handshake

mzabaluev:
Update chain-deps and make use of the new `Handshake` protocol method
to reject peers with incompatible protocol versions or wrong blockchains.

### DIFF
--- a/jormungandr/src/main.rs
+++ b/jormungandr/src/main.rs
@@ -60,7 +60,7 @@ extern crate structopt;
 extern crate test;
 
 use crate::{
-    blockcfg::Leader,
+    blockcfg::{HeaderHash, Leader},
     blockchain::BlockchainR,
     secure::enclave::Enclave,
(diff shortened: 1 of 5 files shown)


jormungandr/src/main.rs
```diff
@@ -101,6 +101,7 @@ fn start() -> Result<(), start_up::Error> {
 pub struct BootstrappedNode {
     settings: Settings,
     blockchain: BlockchainR,
+    block0_hash: HeaderHash,
```
CodeSandwich: This information should be already present in `BlockchainR`, in tip ledger static params, so it seems like redundancy

NicolasDP: not the same data. This is for the **Block 0**. The network needs to know what _node0_ we are talking about in order to be able to detect if a node is working on a different blockchain. (testnet vs mainnet for example)

Issue #546, opened 2019-06-19
REST: Add an endpoint to list stake pool


vincenthz:
useful to dump all the known stake pool ids in the system:

```
[
    stakepool1,
    stakepool2,
    stakepool3,
    ...
]
```

NicolasDP:
This will require some thoughts on the type of APIs to provide from the `Leadership` object (the one from chain-libs).

PR #660, opened 2019-07-26
Add stake pool getter to REST API

CodeSandwich:
Fixes https://github.com/input-output-hk/jormungandr/issues/546

BLOCKED BY https://github.com/input-output-hk/jormungandr/pull/657 AND https://github.com/input-output-hk/chain-libs/pull/71

### DIFF
--- a/jormungandr/src/rest/v0/handlers.rs
+++ b/jormungandr/src/rest/v0/handlers.rs
@@ -281,3 +281,16 @@ pub fn delete_leaders(
         false => Err(ErrorNotFound("Leader with given ID not found")),
     }
 }
+
+pub fn get_stake_pools(context: State<Context>) -> Result<impl Responder, Error> {
+    let blockchain = context.blockchain.lock_read();
+    let stake_pool_ids = blockchain
+        .tip
+        .ledger()
+        .map_err(|_| ErrorInternalServerError("Failed to get blockchain tip ledger"))?
+        .delegation()
+        .stake_pool_ids()
+        .map(|id| id.to_string())
+        .collect::<Vec<_>>();
+    Ok(Json(stake_pool_ids))
+}
--- a/doc/jcli/rest.md
+++ b/doc/jcli/rest.md
@@ -308,3 +308,26 @@ The options are
 
 - -h <node_addr> - see [conventions](#conventions)
 - --debug - see [conventions](#conventions)
+
+## Get stake pools
+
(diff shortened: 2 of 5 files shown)


jormungandr/src/main.rs
```diff
@@ -170,9 +170,17 @@ fn start_services(bootstrapped_node: BootstrappedNode) -> Result<(), start_up::E
             transaction_box: fragment_msgbox,
             block_box: block_msgbox,
         };
+        let block0 = bootstrapped_node
+            .blockchain
+            .lock_read()
+            .tip
+            .ledger()
```
vincenthz: I realize it was not made clear in the original issue, but I think this is was not the intended source of information. I think we want to list the stake pool at the moment it has been snapshotted for the current leadership, not the current stake pool.

that being said, we can fix it later and accept this as a putting most of the REST call in place. Or we can also offer both the current tip and the snapshot (possibly with a different rest endpoint)

jormungandr/src/rest/v0/handlers.rs
```diff
@@ -255,3 +255,16 @@ pub fn get_shutdown(context: State<Context>) -> Result<impl Responder, Error> {
         .stop();
     Ok(HttpResponse::Ok().finish())
 }
+
+pub fn get_stake_pools(context: State<Context>) -> Result<impl Responder, Error> {
+    let blockchain = context.blockchain.lock_read();
+    let stake_pool_ids = blockchain
+        .tip
+        .ledger()
```
vincenthz: I realize it was not made clear in the original issue, but I think this is was not the intended source of information. I think we want to list the stake pool at the moment it has been snapshotted for the current leadership, not the current stake pools defined in the tip ledger.

that being said, we can fix it later and accept this as a putting most of the REST call in place. Or we can also offer both the current tip and the snapshot (possibly with a different rest endpoint)

vincenthz:
is this still a WIP ?

CodeSandwich:
No, it's blocked, see description

vincenthz:
it has been merged

jormungandr/src/rest/v0/handlers.rs
CodeSandwich: Are both versions going to be usable?

CodeSandwich:
Only the first one ;)

CodeSandwich:
Thanks, let me rebase and clean up

PR #662, opened 2019-07-30
Boxing problem with custom_error

mzabaluev:
Fixing a regression in our usage of `custom_error` which hits on the next `cargo update` or, strangely, on `cargo install`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -13,7 +13,7 @@ dependencies = [
  "fnv 1.0.6 (registry+https://github.com/rust-lang/crates.io-index)",
  "futures 0.1.28 (registry+https://github.com/rust-lang/crates.io-index)",
  "libc 0.2.60 (registry+https://github.com/rust-lang/crates.io-index)",
- "log 0.4.7 (registry+https://github.com/rust-lang/crates.io-index)",
+ "log 0.4.8 (registry+https://github.com/rust-lang/crates.io-index)",
  "parking_lot 0.7.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "smallvec 0.6.10 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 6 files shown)


CodeSandwich:
Looks good to me.

To my understanding this is because custom_error introduced support for generic types in `source` field. Previously genericness of `Box` was disabling proper implementation of `.source` for this variant and was always returning `None`. Now it tries to return `&Box<&dyn Error>` and fails, because `Error` is not implemented for `&Error`, so `Box<&dyn Error>` doesn't implement `Error` and can't be uses as such unless manually dereferenced. I've created an issue on `custom_error`: https://github.com/lovasoa/custom_error/issues/20

CodeSandwich:
@mzabaluev The issue was fixed and published just now, the guy is incredible

rinor:
;) Double fixing causing double problem. ;)
```
error[E0277]: the trait bound `std::boxed::Box<bincode::ErrorKind>: std::borrow::Borrow<dyn std::error::Error>` is not satisfied
  --> jormungandr/src/network/p2p/topology.rs:17:1
   |
17 | / custom_error! {pub Error
18 | |     Encoding { source: bincode::Error } = "Serialization error",
19 | |     Io { source: io::Error } = "I/O Error",
20 | | }
   | |_^ the trait `std::borrow::Borrow<dyn std::error::Error>` is not implemented for `std::boxed::Box<bincode::ErrorKind>`
   |
   = help: the following implementations were found:
             <std::boxed::Box<T> as std::borrow::Borrow<T>>
   = note: required by `std::borrow::Borrow::borrow`
   = note: this error originates in a macro outside of the current crate (in Nightly builds, run with -Z external-macro-backtrace for more info)

error: aborting due to previous error
```

CodeSandwich:
Haha, just perfect, didn't think about it, I've encouraged him to break his crate for everyone :D Could you report this in the issue?

PR #668, opened 2019-08-01
[Tests] Set rust backtrace in e2e tests

dkijania:
Set `RUST_BACKTRACE=full`  system variable during e2e to help debugging issues during test run

### DIFF
--- a/jormungandr-integration-tests/build.rs
+++ b/jormungandr-integration-tests/build.rs
@@ -1,9 +1,9 @@
 use std::env;
-use std::fs::rename;
 
 fn main() {
     let jor_cli_name = option_env!("JOR_CLI_NAME").unwrap_or("jcli");
     let jormungandr_name = option_env!("JORMUNGANDR_NAME").unwrap_or("jormungandr");
     println!("cargo:rustc-env=JOR_CLI_NAME={}", jor_cli_name);
     println!("cargo:rustc-env=JORMUNGANDR_NAME={}", jormungandr_name);
+    println!("cargo:rustc-env=RUST_BACKTRACE={}", "full");
 }


jormungandr-integration-tests/src/common/jcli_wrapper/mod.rs
```diff
@@ -398,3 +398,10 @@ pub fn assert_get_rest_message_log(host: &str) -> Vec<FragmentLog> {
         serde_yaml::from_str(&content).expect("Failed to parse fragment log");
     fragments
 }
+
+pub fn assert_shutdown_node(host: &str) {
+    let output = process_utils::run_process_and_get_output(
+        jcli_commands::get_rest_shutdown_node_command(&host),
+    );
+    process_assert::assert_process_exited_successfully(output);
```
CodeSandwich: This will be failing randomly yet constantly. When REST shutdown endpoint responds, it means that the shutdown process has started, not ended!

dkijania: So basically to be sure node is not up. i need to send some message to it (like node get stats) and assure i got no connection error or this information could be retrieved from logs ?

dkijania:
I did a mistake and changes for shutdown should be in another pr : `passive_node_fix`. I'll remove this change from this pr and fix your comment in another pr

jormungandr-integration-tests/src/common/jcli_wrapper/mod.rs
CodeSandwich: Sorry, I've missed the point of this function. Assertion checks if JCLI command exited successfully, not node itself. Probably failures will be more subtle then, because if the node still runs for a moment after that, we won't get a direct panic from here.

I think that the perfect solution would be to check if the process crated on node start is down, that gives 100% reliability.

Looking for `info` `"Shutting down node"` in logs is almost as good, the node usually will be running after logging that just for a fraction of a second. The logs could be even flushed during process cleanup by OS, when node is already down. But it's still relying on something random.

Issue #672, opened 2019-08-01
Blockchain management and network support


NicolasDP:
We have made a huge progress regarding the self node: the ledger and the understanding of the different feature the node needs provide has been tested by the dev and the community (for ever grateful ❤️ ).

Now we arrive were things are starting to fall into place. The network protocol (the low level bits) are near finalisation. The ledger and the leadership scheduling is being stabilised in the `chain-libs`.

Now it is time to make `jormungandr` aware of other nodes: to start plugging the network and interconnect the nodes.

# improving the internal representation of the blockchain

There are undefined behaviour in the current code. Place holders that were left for the network sprint. Especially in the `blockchain` module: the management of receiving blocks from competitive nodes.

In order to do this we need to improve our data representation of the blockchain. Because of the way the leader election works with genesis praos we need to keep a couple of intermediate state representation.

The `blockchain` module is responsible to maintaining the blockchain (i.e.) the blocks,
the current working branches (we will come back to it in a bit), the different states
associated to every blocks, the epoch's data (the parameters, the stake active distribution
and the leadership schedule).

It is fairly easy to maintain the blocks of a blockchain. They all have the identifier of
the parent block. Storing them is another story though and is not covered here.

![Blockchain Data structure](https://raw.githubusercontent.com/input-output-hk/jormungandr/poc-new-blockchain-and-state-machine/doc/internal_design/blockchain_structure.png "Blockchain Data structure")

* `block0` or `blockx.y` are blocks of the blockchain. They `link` to the parent block except for the
  `block0` which may not have parents here (there is a special case where we could set a parent pointing
  to the block of a previously known state of the blockchain);
* `legder0` or `ledgerx.y` are states of the blockchain at a given `block`;
* `epoch x parameters` are the blockchain parameters that are valid for all the epoch `x`;
* `epoch N stake distribution` are the stake distribution as extracted from epoch `N`;
* `epoch x leadership` is the leadership schedule for the epoch `x`.

This may seem a bit overwhelming. Let's follow the flow of block creations and validation
on this blockchain:

## From the block 0

Let's start with first initialising the blockchain from the `block0`.

![Blockchain Data structure From block0](https://raw.githubusercontent.com/input-output-hk/jormungandr/poc-new-blockchain-and-state-machine/doc/internal_design/blockchain_structure_block0.png "Blockchain Data structure From block0")

The **first block**, the `block0`, is the block that contains the initial data of the blockchain.
[...]

PR #673, opened 2019-08-01
New blockchain data representation

NicolasDP:
This PR is part of the intermediate work for #672 

The new code is not yet enabled in the node as there are breaking changes in the way the data is represented.


jormungandr/src/blockchain/chain.rs
```diff
@@ -116,7 +117,13 @@ impl Blockchain {
 
                 let mut epoch = block_0.date().epoch;
                 let initial_leadership = Leadership::new(epoch, &state);
-                let mut leaderships = Leaderships::new(&block_0.header, initial_leadership);
+
+                let initial_gc_root = leaderships.add(
```
mzabaluev: This seems unused, why is this change done to the legacy blockchain code?

NicolasDP: because it was wrong the way it works. We never kept the initial gc_root. GC Run of the multiverse would have created a cascade of error. This is placeholder for when we do the switch to the new data structure.

PR #675, opened 2019-08-01
Implement bootstrap for new blockchain API

mzabaluev:
Part of #672

### DIFF
--- a/jormungandr/src/network/protocols/bootstrap.rs
+++ b/jormungandr/src/network/protocols/bootstrap.rs
@@ -1,6 +1,8 @@
 use super::super::{grpc, BlockConfig};
 use crate::blockcfg::{Block, HeaderHash};
-use crate::blockchain::protocols::{Blockchain, Error as BlockchainError, PreCheckedHeader, Ref};
+use crate::blockchain::protocols::{
+    Blockchain, Branch, Error as BlockchainError, PreCheckedHeader, Ref,
+};
 use crate::settings::start::network::Peer;
 use chain_core::property::HasHeader;
 use network_core::client::{block::BlockService, Client as _};
@@ -66,25 +68,24 @@ impl error::Error for Error {
 pub fn bootstrap_from_peer(
     peer: Peer,
     blockchain: Blockchain,
-    tip: Ref,
+    mut branch: Branch,
     logger: &Logger,
 ) -> Result<Ref, Error> {
     info!(logger, "connecting to bootstrap peer {}", peer.connection);
     let bootstrap = grpc::connect(peer.address(), None)
         .map_err(Error::Connect)
         .and_then(|client: Connection<BlockConfig>| client.ready().map_err(Error::ClientNotReady))
-        .and_then(|mut client| {
+        .join(branch.get_ref().map_err(|_| unreachable!()))
+        .and_then(|(mut client, tip)| {
             let tip_hash = *tip.hash();
             client
                 .pull_blocks_to_tip(&[tip_hash])
                 .map_err(Error::PullRequestFailed)
                 .and_then(|stream| bootstrap_from_stream(blockchain, tip, stream, logger.clone()))
-        });
+        })
+        .and_then(move |tip| branch.update_ref(tip).map_err(|_| unreachable!()));
 
-    current_thread::block_on_all(bootstrap).map(|tip| {
-        debug!(logger, "bootstrap complete");
-        tip
-    })
+    current_thread::block_on_all(bootstrap)
 }
 
 fn bootstrap_from_stream<S>(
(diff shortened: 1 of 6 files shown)


mzabaluev:
Draft: it introduces a new entry point in `network::protocols::bootstrap` alongside the legacy implementation, not yet replacing it.

The entry point takes a `Ref` to the current tip; I'm not sure it's going to be the right way to invoke it.

jormungandr/src/network/protocols/bootstrap.rs
```diff
@@ -0,0 +1,107 @@
...
+use crate::settings::start::network::Peer;
+use chain_core::property::HasHeader;
+use network_core::client::{block::BlockService, Client as _};
+use network_grpc::client::Connection;
+use slog::Logger;
+use std::fmt::Debug;
+use tokio::prelude::*;
+use tokio::runtime::current_thread;
+
+pub fn bootstrap_from_peer(peer: Peer, blockchain: Blockchain, tip: Ref, logger: &Logger) {
```
NicolasDP: so as you said in your comment in the PR, you are not sure about utilising a `Ref` here.

So indeed we will need the `Ref` so we know what is our current tip hash and we can actually query the peer. On another hand you also need to update the `Branch` you are building, continue to update it so when we are down the other task will know where to start from. Otherwise the `Tip` branch is still at your original `Tip` but not the updated (with this function) `Tip`.

So at the end this function needs to returns the last created `Ref` and then update the `Branch`.

so `bootstrap_from_stream` would return `Item = Ref` and then you would update the branch with this new `Ref`

jormungandr/src/network/protocols/bootstrap.rs
```diff
@@ -0,0 +1,107 @@
...
+        .and_then(move |pre_checked| match pre_checked {
+            PreCheckedHeader::AlreadyPresent { header, .. } => {
+                warn!(logger, "block {} is already present", header.hash());
+                Err(())
+            }
+            PreCheckedHeader::MissingParent { header, .. } => {
+                warn!(logger, "received a disconnected block {}", header.hash());
+                Err(())
+            }
+            PreCheckedHeader::HeaderWithCache { header, parent_ref } => Ok((header, parent_ref)),
```
NicolasDP: of course it implements `IntoFuture`. neat. I'll simplify some code here and there (and remove some `future::ok` and `future:err` or `future::result`).

jormungandr/src/network/protocols/bootstrap.rs
```diff
@@ -0,0 +1,141 @@
...
+            BlockMissingParent(_) => None,
+            ApplyBlockFailed(e) => Some(e),
+        }
+    }
+}
+
+pub fn bootstrap_from_peer(
+    peer: Peer,
+    blockchain: Blockchain,
+    tip: Ref,
```
NicolasDP: I think this should be the `Branch` here.

jormungandr/src/network/protocols/bootstrap.rs
```diff
@@ -0,0 +1,141 @@
...
+    let bootstrap = grpc::connect(peer.address(), None)
+        .map_err(Error::Connect)
+        .and_then(|client: Connection<BlockConfig>| client.ready().map_err(Error::ClientNotReady))
+        .and_then(|mut client| {
+            let tip_hash = *tip.hash();
+            client
+                .pull_blocks_to_tip(&[tip_hash])
+                .map_err(Error::PullRequestFailed)
+                .and_then(|stream| bootstrap_from_stream(blockchain, tip, stream, logger.clone()))
+        });
```
NicolasDP: once completion, update the branch with the returned `Ref` from `bootstrap_from_stream`

jormungandr/src/network/protocols/bootstrap.rs
```diff
@@ -0,0 +1,141 @@
...
+pub fn bootstrap_from_peer(
+    peer: Peer,
+    blockchain: Blockchain,
+    tip: Ref,
+    logger: &Logger,
+) -> Result<Ref, Error> {
+    info!(logger, "connecting to bootstrap peer {}", peer.connection);
+    let bootstrap = grpc::connect(peer.address(), None)
+        .map_err(Error::Connect)
+        .and_then(|client: Connection<BlockConfig>| client.ready().map_err(Error::ClientNotReady))
```
NicolasDP: you might need to call `Branch.get_ref` (which is a future) to get the current `Ref` from the `Branch`

mzabaluev:
Ready for review now. I have left the new entry point commented out alongside `start_up::load_blockchain`, which will also need some fixing (tried locally, but ended up discarding code unrelated to this PR).

Issue #548, opened 2019-06-20
REST: make or add an swagger/openapi description


vincenthz:
to facilitate documentation, usability from other languages, etc.. we should have an openapi description of our APIs. the best approach would be to have it automatically generated from our server endpoints

PR #692, opened 2019-08-14
Openapi validator 1st level

CodeSandwich:
1st level of OpenAPI validator, very basic, only checks if route is present in OpenAPI spec. Part of https://github.com/input-output-hk/jormungandr/issues/548

### DIFF
--- a/jcli/src/jcli_app/utils/open_api_verifier.rs
+++ b/jcli/src/jcli_app/utils/open_api_verifier.rs
@@ -1,0 +1,146 @@
+use openapiv3::{OpenAPI, PathItem, ReferenceOr};
+use reqwest::Request;
+use std::env;
+use std::fs::File;
+use std::io;
+
+pub struct OpenApiVerifier(VerifierMode);
+
+enum VerifierMode {
+    AcceptAll,
+    Verify(Verifier),
+}
+
+struct Verifier {
+    openapi: OpenAPI,
+    path: String,
+}
+
+custom_error! { pub Error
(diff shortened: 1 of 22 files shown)


jcli/src/jcli_app/utils/open_api_verifier.rs
```diff
@@ -0,0 +1,146 @@
...
+use std::io;
+
+pub struct OpenApiVerifier(OpenApiVerifierMode);
+
+enum OpenApiVerifierMode {
+    AcceptAll,
+    Verify(OpenApiVerifierImpl),
+}
+
+struct OpenApiVerifierImpl {
```
mzabaluev: Let's not use "impl" in type names. `VerifierState`?

CodeSandwich: It's not quite state, more like implementation of the mode. I think I quite often see `...Inner` in Rust ecosystem, would `OpenApiVerifierInner` be ok? Why not `Impl`? Too much like if it was implementor of an `OpenApiVerifier` interface?

NicolasDP: I agree with @mzabaluev

CodeSandwich: I've renamed it to `pub OpenApiVerifier` -> `VerifierMode` -> `Verifier`, it's less verbose and names without `OpenApi` prefix are private anyway, so their OpenApi context is always clear

Issue #694, opened 2019-08-14
Explorer mode


enzoc4:
# Add an explorer node for jormungandr

## Objective

Allow Jormungandr to work in an explorer node, configurable by command line settings. This requires the creation of a separate module that only is used when the feature is on, as implementing this would require to store things that are not necessary (and are not stored right now) and would waste space for other use cases, like stake pools.

The new component should receive messages from the Block task to acknowledge any new blocks, read all the transactions and update the state accordingly. It should also communicate with the Blockchains component, with read-only access.
The state could be kept in memory using the Multiverse to later add the necessary branching logic. The important data to keep is the transactions, to be able to query for specific ones by its id, and the block information given it's block number.

A graphql interface would be provided to access the information.

# Planned query support:

Edit: See schema below

### Possible:
Delegated pool from Address

### Missing details (to consider):
- Pagination
- Filters for the lists

## Tasks

| PR | Description |
| ------------- | ------------- |
| #702  | Write a new Explorer task, conditionally enabled, that receives messages from the Block task when new blocks are added, and keeps an in-memory data structure with the required state. |
| #727  | And endpoint with graphql schema |
| #771 | Use imhamt and multiverse in explorer for keeping multiple branches |

## Current status (including #771, not on master. Right now the explorer is not doing anything in master)

The explorer can be enabled by passing `--enable-explorer` to jormungandr when starting. Or by setting:
```yaml
explorer:
    enabled: true
```
in the node configuration.

Process all the blocks produced by the leadership task. Also restore the longest branch from storage.
The explorer listens for querys on `/explorer/graphql` 
The graphiql (interactive playground) is served at `/explorer/graphiql`

### Query support:

At this point is possible to query:
- Blocks by id and chain length
Example 
```graphql
{
  blockByChainLength(length: "79") {
    date {
      slot
      epoch {
        id
      }
    }
  }
}
```
- Epochs by number

```graphql
{
  epoch(id: "0") {
    lastBlock {
      id
      chainLength
    }
  }
}
```
- Get latest block.

Example: 

```graphql
{
  status {
    latestBlock {
      id
      chainLength
    }
  }
}
```

- Transactions by id and by address

For example: 
```graphql
{
  transaction(id: "3b30d56c916e2974b3d8ba0d088dce54ea1c1bf5d907f8df2f4ee0a6b9484b0e") {
    inputs {
      amount
      address {
        id
      }
    }
    outputs {
      amount
      address {
        id
      }
    }
  }
[...]

PR #702, opened 2019-08-16
Add explorer mode startup config

enzoc4:
# About #694. (Draft)

## Summary

- Add flag `--enable-explorer` to the command line start arguments.
Note: Maybe a config from the configuration file is better.
- Add explorer Task
- Send messages from the Block task when the feature was enabled (only when a Leadership block is processed, for now)

### DIFF
--- a/jormungandr/src/explorer.rs
+++ b/jormungandr/src/explorer.rs
@@ -1,0 +1,86 @@
+use super::blockchain::Blockchain;
+use crate::blockcfg::Block;
+use crate::blockcfg::ChainLength;
+use crate::intercom::ExplorerMsg;
+use crate::utils::task::{Input, ThreadServiceInfo};
+use chain_impl_mockchain::multiverse::{GCRoot, Multiverse};
+use futures::Future;
+use futures::IntoFuture;
+
+use chain_core::property::Block as _;
+
+pub struct Process {
+    multiverse: Multiverse<State>,
+}
+
+struct State {
+    block: Block,
+}
+
+impl State {
+    pub fn new(block: Block) -> Self {
+        Self { block }
+    }
+}
+
+impl Process {
+    pub fn new() -> Self {
+        Self {
+            multiverse: Multiverse::<State>::new(),
+        }
+    }
+
+    pub fn handle_input(
+        &mut self,
+        info: &ThreadServiceInfo,
+        blockchain: &Blockchain,
+        input: Input<ExplorerMsg>,
+    ) {
+        let logger = info.logger();
+        let bquery = match input {
+            Input::Shutdown => {
+                return;
+            }
+            Input::Input(msg) => msg,
+        };
+
+        match bquery {
+            ExplorerMsg::NewBlock(header) => {
+                //I don't really understand what's the purpose of the gcroot yet
+                let _gcroot: Result<GCRoot, ()> = blockchain
+                    .storage()
+                    .get(header.hash())
+                    .or_else(|storage_error| {
+                        error!(logger, "{}", storage_error);
+                        Err(())
+                    })
+                    .and_then(|block| {
+                        debug!(
+                            logger,
+                            "storing blockhash: {} with chain_length: {}",
+                            header.hash(),
+                            header.chain_length()
+                        );
+
+                        block
+                            .map(|some_block| self.store_block(some_block))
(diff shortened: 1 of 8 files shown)


jormungandr/src/explorer.rs
```diff
@@ -0,0 +1,86 @@
...
+        let bquery = match input {
+            Input::Shutdown => {
+                return;
+            }
+            Input::Input(msg) => msg,
+        };
+
+        match bquery {
+            ExplorerMsg::NewBlock(header) => {
+                //I don't really understand what's the purpose of the gcroot yet
```
NicolasDP: so the multiverse is a way to handle multiple/possible forks. It allows you to follow different branches where these branches might still share some common data (here the same parent block for example). This cannot grow indefinitely, so there is some sort of garbage collection available on demand (this is not used at the moment, we are kind of trying to see how far we can go in term of memory waster :D in the node).

Now if you drop the `GCRoot`, if someone calls Multiverse's function `gc` https://github.com/input-output-hk/chain-libs/blob/09b67a1d7117e103edc7f8b62b5a659af87692dc/chain-impl-mockchain/src/multiverse.rs#L159 then the code `Block` will be removed from the `Multiverse`.

jormungandr/src/settings/command_arguments.rs
```diff
@@ -61,6 +61,11 @@ pub struct StartArguments {
     /// or from the network.
     #[structopt(long = "genesis-block-hash", parse(try_from_str))]
     pub block_0_hash: Option<HeaderHash>,
+
+    // TODO: Explorer mode description
+    // Maybe a config file variable is better
```
NicolasDP: A config file boolean is good too. But the command line option still have the priority over the config file (i.e. if someone set `--enable-explorer=false` then we don't allow the explorer module and handlers in the REST interface).

jormungandr/src/intercom.rs
```diff
@@ -350,5 +350,12 @@ pub enum NetworkMsg {
     },
 }
 
+/// Messages to the explorer task
+
```
NicolasDP: no trailing space here

jormungandr/src/intercom.rs
```diff
@@ -350,5 +350,12 @@ pub enum NetworkMsg {
...
+#[derive(Debug)]
+pub enum ExplorerMsg {
+    NewBlock(Header),
```
NicolasDP: Just send the block hash here. You don't need the `header`

jormungandr/src/explorer.rs
```diff
@@ -0,0 +1,86 @@
...
+    pub fn new() -> Self {
+        Self {
+            multiverse: Multiverse::<State>::new(),
+        }
+    }
+
+    pub fn handle_input(
+        &mut self,
+        info: &ThreadServiceInfo,
+        blockchain: &Blockchain,
```
NicolasDP: Since you have access to the `Blockchain` you also have access to the `RefCache` and the `Ref`. So something interesting is that you don't need to store the block in the explorer `State`. You can simply hold a `Ref` and query the `Storage` for the content of the block.

I.e., something you can do instead is to have the following:

```rust
struct Process {
    multiverse: Multiverse<Ref>, // storage by block hash

    chain_length_to_hash: HashMap<ChainLength, Vec<Ref>>,
}
```

and to have the `Ref` passed on by the `ExplorerMsg`'s `NewBlock`.

jormungandr/src/explorer.rs
```diff
@@ -0,0 +1,86 @@
...
+                            "storing blockhash: {} with chain_length: {}",
+                            header.hash(),
+                            header.chain_length()
+                        );
+
+                        block
+                            .map(|some_block| self.store_block(some_block))
+                            .unwrap_or_else(|| Err(error!(logger, "Block is not in storage")))
+                    })
+                    .wait();
```
NicolasDP: Good stuff on utilising the `Future` APIs. Now try to remove this `wait` here and return a `Future` instead. You can do that by changing the function you used to spawn the explorer task.

jormungandr/src/explorer.rs
enzoc4: Alright, I understand now. Although from what I see, `gc` is only implemented for the Ledger specialization.

PR #703, opened 2019-08-19
Mempool and Leadership logs GC setting and fixes

### LINKED ISSUE #591: The Node stops creating blocks (BFT and Genesis)
**Describe the bug**
After 15 mins (tens of transactions), the Node stops creating blocks  (BFT and Genesis) 

**To Reproduce**
**Even simpler**: 
1. start node
2. create 1 transaction and wait 15 mins while blocks are created
3. after 15 mins, blocks are no longer created

1. generate files using bootstrap script: `bash bootstrap -p 8607 -b | tee initialValues.txt`
2. start node: `jormungandr --genesis-block ./block-0.bin --config ./config.yaml --secret ./pool-secret1.yaml`
3. run attached script (`one-transaction-per-slot.sh`) - after some time, the node stops creating new blocks; In this state, no new transactions are accepted and the script will fail; 

**Expected behavior**
Especially because we are in BFT mode, as long as the node is ON, the leader should create blocks.

**Additional context**
Machine used: Win10
I ran the script in Genesis mode too and the behavior was the same (there still was stake in the system - on Faucet account)

As @rinor observed, if you don't make any transaction, the blocks will continue to be created even after 15 mins. 

Attached:
- print screen - 
nodeLogsTime

- node logs (logger:  verbosity: 4) - 
[nodeLogs.txt](https://github.com/input-output-hk/jormungandr/files/3337767/nodeLogs.txt)

- storage file - 
[blocks.sqlite.txt](https://github.com/input-output-hk/jormungandr/files/3337768/blocks.sqlite.txt)

- one-transaction-per-slot.sh script - 
[one-transaction-per-slot.sh.txt](https://github.com/input-output-hk/jormungandr/files/3337878/one-transaction-per-slot.sh.txt)

NicolasDP:
- allow configuration of the different durations for the mempool and leadership logs
- add some documentations for the new settings
- fixes #591 : don't report the log purge as _not ready_ if there are not expired items yet or we will block all threads.

### DIFF
--- a/doc/SUMMARY.md
+++ b/doc/SUMMARY.md
@@ -16,6 +16,8 @@
 - [Configuration](./configuration/introduction.md)
     - [Logging](./configuration/logging.md)
     - [Node network](./configuration/network.md)
+    - [Fragment Pool](./configuration/mempool.md)
+    - [Leader Events](./configuration/leadership.md)
 
 - [jcli](./jcli/introduction.md)
     - [Cryptographic keys](./jcli/key.md)
--- a/doc/configuration/leadership.md
+++ b/doc/configuration/leadership.md
@@ -1,0 +1,13 @@
+The `leadership` field in your node config file is not mandatory, by default it is set
+as follow:
+
+```yaml
+leadership:
+    log_ttl: 1h
+    garbage_collection_interval: 15m
+```
+
+* `log_ttl` describes for how long the node will keep logs of leader events.
+  This is link to the data you receives from the REST leadership logs end point;
+* `garbage_collection_interval` describes the interval between 2 garbage collection
+  runs: i.e. when the node removes item logs that have timed out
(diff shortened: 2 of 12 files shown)


mzabaluev:
> fixes #591 : don't report the log purge as not ready if there are not expired items yet or we will block all threads.

I think it is wrong here as well: the meaning of `NotReady` is temporal and it applies to this poll's iteration. The delay queue has scheduled a wakeup at the next expiration deadline when it returns `NotReady`, so the task will re-poll then. The blocking issue may be elsewhere, maybe in how multiple `Ready` and `NotReady` poll results are aggregated. If, **but only if**, any of the other pollables is `Ready`, the outer poll can continue iteration. Conversely, the outer poll should return `NotReady` **only if** some sub-poll returned `NotReady` or the next wakeup was primed in the context directly.

NicolasDP:
Well yes it is elsewhere. It is in the sense that DelayQueue is behind a `Lock` and in order to purge it:

1. you need to **acquire** the lock;
2. and then `poll` the delay queue;

But if the delay queue is blocking for an event to finish, the lock remains acquired. Preventing us to add new items in the delay queue (no new item in the Logs), blocking us to finish the leadership task or the fragment or whatever else we needed to complete other task.

mzabaluev:
Ah, but maybe it should be the other way around: you first wait for an item to expire and then go locking?

Issue #698, opened 2019-08-16
serve the leader logs through the Rest API


NicolasDP:
Since the recent changes in the `Leadership` (#685 ) now log leader events happening in the node.

The logs are stored in https://github.com/input-output-hk/jormungandr/blob/650a005c6b3577921c1f754c2d91c0e038e31d11/jormungandr/src/leadership/logs.rs#L12

And the function https://github.com/input-output-hk/jormungandr/blob/650a005c6b3577921c1f754c2d91c0e038e31d11/jormungandr/src/leadership/logs.rs#L101-L104

returns an array of https://github.com/input-output-hk/jormungandr/blob/650a005c6b3577921c1f754c2d91c0e038e31d11/jormungandr-lib/src/interfaces/leadership_log.rs#L17-L24

Which is already JSON/Human readable friendly.

We need to add a REST end point `/leaderships/logs` to serve the logs (just like we serve the `fragment/logs`).

The leadership logs are available in the `main` here: https://github.com/input-output-hk/jormungandr/blob/650a005c6b3577921c1f754c2d91c0e038e31d11/jormungandr/src/main.rs#L113

It can be passed to the Rest server for the handle to access the logs.

NicolasDP:
This tool will help to look at #591

PR #707, opened 2019-08-19
Add leadership logs fetching over REST API and JCLI

CodeSandwich:
Fixes https://github.com/input-output-hk/jormungandr/issues/698

### DIFF
--- a/jormungandr/src/main.rs
+++ b/jormungandr/src/main.rs
@@ -52,6 +52,7 @@ use crate::{
     settings::start::Settings,
     utils::{async_msg, task::Services},
 };
+use futures::Future;
 use settings::{start::RawSettings, CommandLine};
 use slog::Logger;
 use std::sync::{Arc, Mutex};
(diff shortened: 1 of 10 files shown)


jormungandr/src/main.rs
```diff
@@ -201,14 +202,11 @@ fn start_services(bootstrapped_node: BootstrappedNode) -> Result<(), start_up::E
...
-        use tokio::prelude::*;
-
+        let leadership_logs = leadership_logs.clone();
+        let enclave = leadership::Enclave::new(enclave.clone());
         let fragment_pool = fragment_pool.clone();
-        let block_task = block_task.clone();
         let blockchain_tip = blockchain_tip.clone();
-
-        let enclave = leadership::Enclave::new(enclave.clone());
-        let stats_counter = stats_counter.clone();
```
NicolasDP: except for `leadership_logs` that needed to be cloned, there is no useful changes here. Why did you re-organise the order of these lines?

CodeSandwich: I've changed order so it matches order of usage few lines below. It's easier to follow what comes from where. I've touched this block anyway.

NicolasDP: but now you are breaking git history with non necessary code changes.

CodeSandwich: I can revert it, but I don't agree that fighting entropy is unnecessary. It's not worth a separate PR and I've already touched this block, it'll be dirty anyway.

PR #715, opened 2019-08-22
Testing scenario managing test flow

NicolasDP:
When running the scenario we can now print the flow of execution of the tests.

Example of usage demonstrating 2 nodes. 1 running as BFT leader and creating a block every seconds (with epoch length is 10 blocks per epoch). We test that the tip of node1 has been fetched already by node2 (the passive node) demonstrating node2 has synced. Node 1 is shutdown before we get the block from node2

scenario testing 3


jormungandr-scenario-tests/src/node.rs
```diff
@@ -15,43 +26,162 @@ error_chain! {
...
+    progress_bar: ProgressBarController,
+    node_settings: NodeSetting,
+    status: Arc<Mutex<Status>>,
 }
 
 const NODE_CONFIG: &str = "node_config.yaml";
 const NODE_SECRET: &str = "node_secret.yaml";
 
-impl Node {
+impl NodeController {
```
mzabaluev: Can you please document the purpose of `NodeController` as separate from `Node`?

Issue #719, opened 2019-08-23
Extended info on --version


rinor:
**Is your feature request related to a problem/context ? Please describe if applicable.**
This is a **JIP** and expands a bit over https://github.com/input-output-hk/jormungandr/issues/645 and https://github.com/input-output-hk/jormungandr/issues/606.
Right now the reported version from `jormungandr --version` compiled from master is the same as the one from the latest release so there is no way to tell if the binary is up to date or not. Furthermore since master is a fast moving branch, it would be nice to have more info reported form --version.

**Describe the solution you'd like**
The following is just an example and may be tweeked as needed.
```
jormungandr 0.3.3 (master-3326b9b+, debug build, linux [x86_64]) - [rustc 1.37.0 (eae3437df 2019-08-13)]
\_________/ \___/  \____/ \_____/|  \_________/  \___/  \____/      \_________________________________/
 |           |      |      |     |   |            |      |           |
 |           |      |      |     |   |            |      |           +- rustc --version
 |           |      |      |     |   |            |      +- std::env::consts::ARCH
 |           |      |      |     |   |            +- std::env::consts::OS
 |           |      |      |     |   +- Checks debug_assertions
 |           |      |      |     +- Adds a "+" if the working tree is not clean
 |           |      |      +- Commit hash
 |           |      +- Current branch name
 |           +- Package version from Cargo.toml
 +- Package name from Cargo.toml
```

**Additional context**
To produce the proposed output, tried to create a cargo build script `build.rs` that (ab)uses `CARGO_PKG_VERSION`, but due to zero rust experience wasn't able to properly complete it.
` ¯\_(ツ)_/¯`
In the example the only dependency is the `git` binary and there is no need to change any other part of the code.
The following works, but needs proper error handling. The missing part is that if `git` binary is missing or there is an error, the build should continue and the reported version should the same as today `jormungandr 0.3.3` 

```rust
use std::env::consts::{ARCH, OS};
use std::process::Command;

#[cfg(debug_assertions)]
const BUILD_TYPE: &'static str = "debug";
#[cfg(not(debug_assertions))]
const BUILD_TYPE: &'static str = "release";

fn main() {
    let version_string = format!(
        "{} ({}-{}{}, {}, {} [{}]) - [{}]",
        env!("CARGO_PKG_VERSION"),
        git_branch_name(),
        git_commit_hash(),
        if git_working_tree_isclean() { "" } else { "+" },
        BUILD_TYPE,
        OS,
[...]

PR #720, opened 2019-08-23
jormungandr/jcli --version extended

rinor:
Initial implementation of JIP https://github.com/input-output-hk/jormungandr/issues/719

### DIFF
--- a/jcli/build.rs
+++ b/jcli/build.rs
@@ -1,0 +1,82 @@
+use std::env::consts::{ARCH, OS};
+use std::process::Command;
+
+#[cfg(debug_assertions)]
+const BUILD_TYPE: &'static str = "debug";
+#[cfg(not(debug_assertions))]
+const BUILD_TYPE: &'static str = "release";
+
+fn main() {
+    let version_string = format!(
+        "{} ({}-{}{}, {}, {} [{}]) - [{}]",
+        env!("CARGO_PKG_VERSION"),
+        git_branch_name(),
+        git_commit_hash(),
+        if git_working_tree_isclean() { "" } else { "+" },
+        BUILD_TYPE,
+        OS,
+        ARCH,
+        compiler_version()
+    );
+
+    println!("cargo:rustc-env=CARGO_PKG_VERSION={}", version_string);
+}
+
+fn compiler_version() -> String {
+    let output = Command::new("rustc")
+        .arg("--version")
+        .current_dir(env!("CARGO_MANIFEST_DIR"))
+        .output()
+        .unwrap();
+
+    assert!(output.status.success());
+
+    String::from_utf8_lossy(&output.stdout)
+        .trim_end()
+        .to_string()
+}
+
+fn git_commit_hash() -> String {
+    let output = Command::new("git")
+        .arg("rev-parse")
+        .arg("--short")
+        .arg("HEAD")
+        .current_dir(env!("CARGO_MANIFEST_DIR"))
+        .output()
+        .unwrap();
+
+    assert!(output.status.success());
+
+    String::from_utf8_lossy(&output.stdout)
+        .trim_end()
+        .to_string()
+}
+
+fn git_branch_name() -> String {
+    let output = Command::new("git")
+        .arg("rev-parse")
+        .arg("--abbrev-ref")
+        .arg("HEAD")
+        .current_dir(env!("CARGO_MANIFEST_DIR"))
+        .output()
+        .unwrap();
+
+    assert!(output.status.success());
+
+    String::from_utf8_lossy(&output.stdout)
+        .trim_end()
+        .to_string()
+}
+
+fn git_working_tree_isclean() -> bool {
+    let status = Command::new("git")
+        .arg("diff")
+        .arg("--quiet")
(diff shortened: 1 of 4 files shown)


jcli/build.rs
```diff
@@ -0,0 +1,82 @@
...
+use std::process::Command;
+
+#[cfg(debug_assertions)]
+const BUILD_TYPE: &'static str = "debug";
+#[cfg(not(debug_assertions))]
+const BUILD_TYPE: &'static str = "release";
+
+fn main() {
+    let version_string = format!(
+        "{} ({}-{}{}, {}, {} [{}]) - [{}]",
```
mzabaluev: This would be easier to understand if it used named format patterns.

jcli/build.rs
```diff
@@ -0,0 +1,82 @@
...
+        .unwrap();
+
+    assert!(output.status.success());
+
+    String::from_utf8_lossy(&output.stdout)
+        .trim_end()
+        .to_string()
+}
+
+fn git_working_tree_isclean() -> bool {
```
mzabaluev: Nitpick: `is_clean`?

jcli/build.rs
rinor: is this ok?
```rust
    let version_string = format!(
        "{pkg_version} ({branch_name}-{commit_hash}{dirty}, {build_type}, {os_type} [{cpu_arch}]) - [{rustc_version}]",
        pkg_version = env!("CARGO_PKG_VERSION"),
        branch_name = git_branch_name(),
        commit_hash = git_commit_hash(),
        dirty = if git_working_tree_is_clean() { "" } else { "+" },
        build_type = BUILD_TYPE,
        os_type = OS,
        cpu_arch = ARCH,
        rustc_version = compiler_version()
    );
```

jcli/build.rs
```diff
@@ -0,0 +1,82 @@
+use std::env::consts::{ARCH, OS};
+use std::process::Command;
+
+#[cfg(debug_assertions)]
+const BUILD_TYPE: &'static str = "debug";
+#[cfg(not(debug_assertions))]
+const BUILD_TYPE: &'static str = "release";
+
+fn main() {
+    let version_string = format!(
```
CodeSandwich: There's a problem of code duplication, the 2 `build.rs`s are literally a copy-paste.

I think that this could be solved by moving this string generation to a new macro in `jormungandr-lib`.

The only things that really must be retrieved in build.rs are git info, rustc version and `PROFILE`. All the rest can be fetched with `env!` macro in production code. All of git info, rustc version and `PROFILE` are the same across all crates, so it can be retrieved only once: in `jormungandr-lib`'s `build.rs`. It can be then stored in a constant. The macro can then use the global constant from `jormungandr-lib` and local crate properties fetched with `env!` making its output relevant wherever it's used.

What do you think about it, will it fly?

rinor: > ... git info ... use the global constant from `jormungandr-lib` ...  will it fly?

Yes and No, since it depends on the build case scenarios.
Let's focus only on `git info`, since the other ones work fine in any case.

Works fine (global const contains correct `git info`) on:
 - Official/CI releases (since they build everything everytime)
 - Local first build
 - Local full rebuilds
 - Local incremental builds caused by changes that require `jormungandr-lib` to be rebuild
 - ...

My understanding is that in _local user (re)builds_:

1. if there are changes (pulled/created) that need `jormungandr-lib` to be rebuilt, `jormungandr`/`jcli` binaries will be always rebuilded - global const contains **correct/actual** `git info`
2. if there are changes (pulled/created) that need only `jormungandr`/`jcli` binary to be rebuild, `jormungandr-lib` will not be rebuilded by default - global const contains **incorrect/old** `git info`!!!

If my understanding is correct, (**2**) may be problematic/confusing, if not I apologize for my misunderstanding.

jcli/build.rs
```diff
@@ -0,0 +1,82 @@
...
+        BUILD_TYPE,
+        OS,
+        ARCH,
+        compiler_version()
+    );
+
+    println!("cargo:rustc-env=CARGO_PKG_VERSION={}", version_string);
+}
+
+fn compiler_version() -> String {
```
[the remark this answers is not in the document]
rinor: Yes, `rustc_version` crate is much better. compiler version was put as an example and probably is not needed and can be removed since the documentation already states that:

> Currently the minimum supported version of the rust compiler is 1.35, however we recommend to use the most recent stable version of the rust compiler.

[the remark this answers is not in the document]
CodeSandwich: If we assume that compiler is never an issue then we can remove it, but I think that it sometimes might be useful, especially if somebody uses nightly compiler from specific date. If it's not a burden, I'd keep it.

jcli/build.rs
CodeSandwich: Your understanding is 100% correct, that will be a problem. To remove this duplication we probably would have to create a new crate intended only for use in `build.rs`s exposing this functionality. To be honest this could be quite useful when published on `crates.io`, I couldn't find any tool like that. Let's keep it like that for now.

NicolasDP:
superseeded by #730

Issue #618, opened 2019-07-08
Genesis block0_consensus accepted values


CodeSandwich:
# Problem
Genesis.yaml parser expects consensus version "genesis_praos". This is in conflict with documentation and string parsing of chain-impl-mockchain enum, which require "genesis" instead.

# Solutions
1. Fix YAML parser to match docs
2. Drop both "genesis" and "genesis_praos" and change them to "praos" in all places (@NicolasDP)

There is also an idea to drop string parsing in chain-impl-mockchain enum altogether (again @NicolasDP), but I haven't checked yet if this is possible and at what cost.

PR #721, opened 2019-08-24
doc: renaming genesis to genesis_praos

mmahut:
This change has been omitted from documentation.


rinor:
Probably because of https://github.com/input-output-hk/jormungandr/issues/618 and https://github.com/input-output-hk/jormungandr/pull/617

mmahut:
True, right now it is still ``genesis_praos``, so until it is fixed we should correct it in the documentation, so it reflects the reality.

PR #725, opened 2019-08-26
Openapi verifier lvl2

CodeSandwich:
Adds verification of request method. Also fixes config CORS lifetime from usize to u64 and reverts JCLI error directly depending on Reqwest.


jormungandr/src/rest/mod.rs
```diff
@@ -84,7 +84,7 @@ where
 {
     let mut cors = Cors::for_app(app);
     if let Some(max_age_secs) = cors_cfg.max_age_secs {
-        cors.max_age(max_age_secs);
+        cors.max_age(max_age_secs as usize);
```
mzabaluev: Ah, that's where the problem is. Bad dependency API.

PR #728, opened 2019-08-28
Openapi validator lvl3

CodeSandwich:
Adds validation of request body

### DIFF
--- a/jcli/src/jcli_app/utils/debug_flag.rs
+++ b/jcli/src/jcli_app/utils/debug_flag.rs
@@ -1,3 +1,5 @@
+use jcli_app::utils::rest_api::{RestApiRequestBody, RestApiResponse};
+use reqwest::Request;
 use structopt::StructOpt;
 
 #[derive(StructOpt)]
@@ -9,10 +11,24 @@ pub struct DebugFlag {
 }
 
 impl DebugFlag {
-    pub fn debug_writer(&self) -> Option<impl std::io::Write> {
-        match self.debug {
-            true => Some(std::io::stderr()),
-            false => None,
+    pub fn write_request(&self, request: &Request, body: &RestApiRequestBody) {
+        if !self.debug {
+            return;
+        }
+        eprintln!("{:#?}", request);
+        if body.has_body() {
+            eprintln!("Request body:\n{}", body)
+        }
+    }
+
+    pub fn write_response(&self, response: &RestApiResponse) {
+        if !self.debug {
+            return;
+        }
+        eprintln!("{:#?}", response.response());
+        let body = response.body();
+        if !body.is_empty() {
+            eprintln!("Response body:\n{}", body)
         }
     }
 }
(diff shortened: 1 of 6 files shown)


jcli/src/jcli_app/utils/debug_flag.rs
```diff
@@ -9,10 +11,24 @@ pub struct DebugFlag {
 }
 
 impl DebugFlag {
```
mzabaluev: The type name does not match the purpose, rename to `DebugWriter` or something like that?

jcli/src/jcli_app/utils/debug_flag.rs
```diff
@@ -9,10 +11,24 @@ pub struct DebugFlag {
...
+        if !self.debug {
+            return;
+        }
+        eprintln!("{:#?}", request);
+        if body.has_body() {
+            eprintln!("Request body:\n{}", body)
+        }
+    }
+
+    pub fn write_response(&self, response: &RestApiResponse) {
```
mzabaluev: Odd that `RestApiResponse` contains the http response member, but `write_request` has to be passed disjoint request and body.

jcli/src/jcli_app/utils/debug_flag.rs
```diff
@@ -9,10 +11,24 @@ pub struct DebugFlag {
...
+        if !self.debug {
+            return;
+        }
+        eprintln!("{:#?}", response.response());
```
mzabaluev: `RestApiResponse::response` is a bit tautological, maybe the method should say that this is the lower level message, like `response.http()`?

jcli/src/jcli_app/utils/debug_flag.rs
CodeSandwich: It's still primarily a flag passed from CLI, which happens to encapsulate the logic of what this flag is supposed to do. If this argument doesn't convince you, I'll rename it.

mzabaluev: Ah, I didn't see how it was the option value.

jcli/src/jcli_app/utils/debug_flag.rs
CodeSandwich: Yes, it's an oddity coming from the fact that when body is passed to a `Request`, it's lost forever. All you can do is get the `Body`, which has no access to underlying data and that's by design. In order to do anything with the body data it must be passed separately before it's added to a `Request`.

jcli/src/jcli_app/utils/debug_flag.rs
CodeSandwich: It's a simple `reqwest::Response` getter, which is why it's `.response()`. I can rename it, but I'm not sure if it's worth the effort.

jcli/src/jcli_app/utils/debug_flag.rs
CodeSandwich: I can wrap these 2 values with a `RestApiRequest`, but I think that we won't gain much. A few inner functions will be getting 1 argument less, but the new structure won't have any logic to provide, it will be nothing more than a tuple.

PR #729, opened 2019-08-29
Update stats and scenario testing

NicolasDP:
- [ ] missing: update the open API doc

### DIFF
--- a/jormungandr-scenario-tests/src/main.rs
+++ b/jormungandr-scenario-tests/src/main.rs
@@ -2,7 +2,7 @@
 extern crate jormungandr_scenario_tests;
 
 use jormungandr_scenario_tests::{prepare_command, style, Context, Seed};
-use std::path::PathBuf;
+use std::{path::PathBuf, thread, time::Duration};
 use structopt::StructOpt;
 
 #[derive(StructOpt, Debug)]
@@ -55,7 +55,7 @@ fn main() {
 
     introduction(&context);
 
-    scenario_1(context.derive());
+    scenario_2(context.derive());
 }
 
 fn introduction<R: rand_core::RngCore>(context: &Context<R>) {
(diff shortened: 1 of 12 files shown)


jormungandr-scenario-tests/src/main.rs
```diff
@@ -133,3 +133,67 @@ pub fn scenario_1(mut context: Context<ChaChaRng>) {
 
     controller.finalize();
 }
+
+pub fn scenario_2(mut context: Context<ChaChaRng>) {
```
mzabaluev: Could the name be more informative as to what the scenario tests?

jormungandr-scenario-tests/src/main.rs
```diff
@@ -55,7 +55,7 @@ fn main() {
 
     introduction(&context);
 
-    scenario_1(context.derive());
+    scenario_2(context.derive());
```
mzabaluev: Has `scenario_1` fallen into dead code then?

jormungandr-scenario-tests/src/scenario/controller.rs
```diff
@@ -21,7 +22,7 @@ pub struct ControllerBuilder {
 }
 
 pub struct Controller {
-    settings: Arc<Settings>,
+    settings: Settings,
```
mzabaluev: Good.

jormungandr-scenario-tests/src/wallet/account.rs
```diff
@@ -46,10 +46,19 @@ impl Wallet {
...
+        writeln!(w, "{}", self.signing_key().to_bech32_str())
+    }
+
     pub fn address(&self, discrimination: Discrimination) -> Address {
         self.identifier().to_address(discrimination).into()
     }
 
+    pub fn increment_counter(&mut self) {
+        let v : u32 = self.internal_counter.into();
+        self.internal_counter = account::SpendingCounter::from(v + 1);
```
mzabaluev: `SpendingCounter` might benefit from an impl of `AddAssign` to simplify this.

jormungandr-scenario-tests/src/main.rs
NicolasDP: nop. they will be cleaned up when I have a better control of the different scenario and their execution

jormungandr-scenario-tests/src/main.rs
NicolasDP: soon

jormungandr-scenario-tests/src/wallet/account.rs
NicolasDP: agree

PR #731, opened 2019-09-01
Openapi lvl4

CodeSandwich:
Adds validation of request URL

### DIFF
--- a/jcli/src/jcli_app/utils/open_api_verifier.rs
+++ b/jcli/src/jcli_app/utils/open_api_verifier.rs
@@ -26,6 +26,14 @@ struct Verifier {
     path: String,
 }
 
+type PathWildcardValues<'a> = HashMap<&'a str, &'a str>;
+
+enum SegmentMatch<'a> {
+    None,
+    Exact,
+    Wildcard(&'a str),
+}
+
 custom_error! { pub Error
     VerificationFailed { source: VerifierError, path: String }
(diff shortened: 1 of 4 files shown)


jcli/src/jcli_app/utils/open_api_verifier.rs
```diff
@@ -130,46 +171,70 @@ fn verify_path(
...
+fn url_matches_path<'a>(url: &'a str, path: &'a str) -> Option<PathWildcardValues<'a>> {
+    let mut url_segments = url.trim_matches('/').rsplit('/');
+    let path_segments = path.trim_matches('/').rsplit('/');
+    let mut wildcards = HashMap::new();
+    for path_segment in path_segments {
+        let url_segment = url_segments.next()?;
+        match url_segment_matches_path_segment(url_segment, path_segment) {
+            SegmentMatch::None => return None,
+            SegmentMatch::Exact => continue,
+            SegmentMatch::Wildcard(wildcard) => wildcards.insert(wildcard, url_segment),
```
mzabaluev: Why is a hash table used here? What happens when there is a wildcard segment identical to another segment used elsewhere in the path?

CodeSandwich: The key is wildcard name and wildcard names must be unique. The HashMap gives convenient access to wildcard values taken from URL using wildcard names taken from OpenAPI definition.

PR #733, opened 2019-09-02
fix issue with --full-version expecting the --config

### LINKED ISSUE #732: `--full-version` and `--source-version` fail since `--config <node_config>` is mandatory
As it is right now `--full-version` and `--source-version` fail since `--config <node_config>` is mandatory:
```
$ jormungandr --full-version                                         
error: The following required arguments were not provided:
    --config <node_config>

USAGE:
    jormungandr --full-version --config <node_config>

For more information try --help
```
Need to pass `--config` to get output:
```
$ jormungandr --full-version --config ./config_trace.yaml
jormungandr 0.3.3 (master-80631699, debug, linux [x86_64]) - [rustc 1.37.0 (eae3437df 2019-08-13)]
```

_Originally posted by @rinor in https://github.com/input-output-hk/jormungandr/pull/730#issuecomment-526952701_

NicolasDP:
fix #732 

also improve the error reporting in case of a time dilation in the current running epoch. This will help users mitigate the issue if they find it while operating a stake pool.

### DIFF
--- a/jormungandr/src/main.rs
+++ b/jormungandr/src/main.rs
@@ -340,15 +340,16 @@ pub struct InitializedNode {
 
 fn initialize_node() -> Result<InitializedNode, start_up::Error> {
     let command_line = CommandLine::load();
-    let raw_settings = RawSettings::load(command_line)?;
 
-    if raw_settings.full_version() {
+    if command_line.full_version {
         println!("{}", env!("FULL_VERSION"));
         std::process::exit(0);
-    } else if raw_settings.source_version() {
+    } else if command_line.source_version {
         println!("{}", env!("SOURCE_VERSION"));
         std::process::exit(0);
     }
+
+    let raw_settings = RawSettings::load(command_line)?;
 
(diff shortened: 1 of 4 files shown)


jormungandr/src/main.rs
```diff
@@ -340,16 +340,17 @@ pub struct InitializedNode {
...
 
-    if raw_settings.full_version() {
+    if command_line.full_version {
         println!("{}", env!("FULL_VERSION"));
         std::process::exit(0);
-    } else if raw_settings.source_version() {
+    } else if command_line.source_version {
         println!("{}", env!("SOURCE_VERSION"));
         std::process::exit(0);
     }
```
rinor: These can be removed!
https://github.com/input-output-hk/jormungandr/blob/44c8266329c75cb399c10e65d4344f309e43e030/jormungandr/src/settings/start/mod.rs#L54-L61

NicolasDP: thanks for spotting that

PR #736, opened 2019-09-03
make the node use multiaddr for the listen_addr

NicolasDP:


jormungandr/src/settings/start/mod.rs
```diff
@@ -145,8 +145,11 @@ fn generate_network(
     let p2p = &config.p2p;
     network::Configuration {
         public_id: p2p.public_id.clone(),
-        public_address: p2p.public_address.clone(),
-        listen: p2p.listen.clone(),
+        public_address: Some(p2p.public_address.clone()),
+        listen_address: p2p
+            .listen_address
+            .clone()
+            .and_then(|addr| addr.to_socketaddr()),
```
mzabaluev: This would make any multiaddr value that does not resolve to `SocketAddr` silently ignored. Raise an error if `None` is returned?

NicolasDP: Good point. Thanks for spotting this.

PR #737, opened 2019-09-03
Fragment process clean up

CodeSandwich:
Clean up fragment process, most notably:
- Move source of truth about which fragments are already pooled from logs to pool itself
- Fix logger and pool overwriting duplicate entries while inserting
- Remove need for obtaining and dropping pool and logger locks for each fragment, now it's once for all fragments
- Fix FragmentSelectionAlgorithm dropping last block if block is full

I hope these are good changes and I didn't misunderstand anything. Network API trimming comping up next.

### DIFF
--- a/jormungandr/src/fragment/pool.rs
+++ b/jormungandr/src/fragment/pool.rs
@@ -44,6 +44,24 @@ impl Pool {
         })
     }
 
+    /// Returns number of registered fragments
+    pub fn insert_all(
+        &mut self,
+        origin: FragmentOrigin,
+        fragments: impl IntoIterator<Item = Fragment>,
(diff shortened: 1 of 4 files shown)


jormungandr/src/fragment/pool.rs
```diff
@@ -24,33 +25,41 @@ impl Pool {
...
-                            guard.insert(fragment);
-
-                            let log = FragmentLog::new(id.into(), origin);
-                            logs.insert(log).map(|()| true)
-                        },
-                    ))
-                }
-            })
+    /// Returns number of registered fragments
+    pub fn insert_all(
```
mzabaluev: Good idea.

PR #739, opened 2019-09-03
Outline expected schema

enzoc4:
# Add possible schema design matching the documented scope in #694 

This are mostly schema design changes, there are also a few things like moving the struct field syntax to the method resolvers, as that is more flexible and it's easier to not compute unnecessary things.

For reference, the resulting schema from this changes looks like this:

graph

Note: As mentioned in #694, the pagination requirements are still yet to be addressed for the plural fields (blocks in Epoch, for example)


vincenthz:
looking good on the diagram, but couple of comments:
* stake distribution in epoch is only for a special case of consensus. in BFT, there's no stake distribution (or alternatively the empty distribution)
* replace "lovelaces" by "coins" or "Value"
* coefficient in fees are not expressed in coins, they are most of the ratio integer type (aka a scaler for scalar multiplication).

PR #740, opened 2019-09-04
Update chain-deps, use ContentService to subscribe the client to fragments

mzabaluev:
The fragment subscription was not done previously.

Rename remaining use of "message" in network subsystem to "fragment".

### DIFF
--- a/jormungandr/src/network/client.rs
+++ b/jormungandr/src/network/client.rs
@@ -6,11 +6,12 @@ use super::{
     subscription, Channels, ConnectionState, GlobalStateR,
 };
 use crate::{
-    blockcfg::{Block, Header, HeaderHash},
+    blockcfg::{Block, Fragment, Header, HeaderHash},
     intercom::{self, BlockMsg, ClientMsg},
 };
 use futures::prelude::*;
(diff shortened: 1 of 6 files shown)


jormungandr/src/network/client.rs
```diff
@@ -70,13 +72,21 @@ where
...
+                let content_req = service.content_subscription(peer_comms.subscribe_to_fragments());
+                service
+                    .ready()
+                    .map(move |service| (service, peer_comms, block_req, content_req))
             })
+            .and_then(
+                move |(mut service, mut peer_comms, block_req, content_req)| {
+                    let gossip_req = service.gossip_subscription(peer_comms.subscribe_to_gossip());
+                    block_req
+                        .join(gossip_req)
```
CodeSandwich: `join` is going to yield result only if both `gossip_req` and `block_req` are received, right? Is this correct, should we process them grouped and only if both requests were received?

mzabaluev: This resolves successfully when all requests succeed.

I also need to plug in the content request. I'll push a new commit shortly.

PR #741, opened 2019-09-04
Implementation for test : Passvie to Leader (L2001)

dkijania:
PR with failing test.

TestCaseID: L2001
Desc:

Alice sends transaction to Node A (Passive node) | Node A should propagate transaction to Node B which then should create block with given transaction (when it is elected as slot leader) and should propagate block to Node A

In order to execute it:

1) build jormungandr-scenario-tests
2) run command:  jormungandr-scenario-tests --root-dir L2001 --document

### DIFF
--- a/jormungandr-scenario-tests/src/main.rs
+++ b/jormungandr-scenario-tests/src/main.rs
@@ -55,7 +55,7 @@ fn main() {
 
     introduction(&context);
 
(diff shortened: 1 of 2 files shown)


dkijania:
This PR was meant only to fasten reproduction for mentioned issue. there will be another PR which will try to better handle code. Thus closing this one

PR #744, opened 2019-09-04
Implementation for test case : T3005

dkijania:
PR with failing test.

TestCaseID: T3005
Desc:
Topology: Point to Point
Action:  Keep sending transaction to Node 1
Result:  Node 4 is in sync with Node 1

In order to execute it:

1) build jormungandr-scenario-tests
2) run command:  jormungandr-scenario-tests --root-dir T3005--document

### DIFF
--- a/jormungandr-scenario-tests/src/main.rs
+++ b/jormungandr-scenario-tests/src/main.rs
@@ -55,7 +55,7 @@ fn main() {
 
     introduction(&context);
 
-    scenario_2(context.derive());
+    scenario_point_to_point(context.derive());
 }
 
 fn introduction<R: rand_core::RngCore>(context: &Context<R>) {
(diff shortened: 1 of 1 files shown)


jormungandr-scenario-tests/src/main.rs
```diff
@@ -134,6 +134,71 @@ pub fn scenario_1(mut context: Context<ChaChaRng>) {
...
+                account "unassigned1" with   500_000_000,
+                account "delegated1" with  2_000_000_000 delegates to "Leader1",
+            ],
+        }
+    };
+
+    let mut controller = scenario_settings.build(context).unwrap();
+
+    controller.monitor_nodes();
+    let leader4 = controller.spawn_node("Leader4", true).unwrap();
```
mzabaluev: It would be nice for manual repro if the enumeration followed the order in which the nodes should be launched. I had to find out from code which one was the "seed" node.

jormungandr-scenario-tests/src/main.rs
```diff
@@ -134,6 +134,71 @@ pub fn scenario_1(mut context: Context<ChaChaRng>) {
...
+        if let Ok(status) = status {
+            if status.is_in_a_block() {
+                wallet1.confirm_transaction();
+            } else {
+                break;
+            }
+        } else {
+            break;
+        }
+    }
```
mzabaluev: It would be good to have a timeout; if none of the break conditions are reached, this scenario continues forever. But this probably should be a global framework feature.

jormungandr-scenario-tests/src/main.rs
NicolasDP: One thing you can do is add the `--document` to the command line of the scenario. It will create a `.dot` file that you can convert into png with graphviz. The file is in the test directory when you ran with `--root-dir`.

there is still a million things to write on this tool. We have each command available and in order available in the controller. we just need to dump it in a script file to reproduce the tests.

dkijania:
This PR was meant only to fasten reproduction for mentioned issue. there will be another PR which will try to better handle code. Thus closing this one

Issue #742, opened 2019-09-04
Fragment propagation on the network fails


dkijania:
**Describe the bug**

Given network scenario:

![image](https://user-images.githubusercontent.com/20424186/64251405-c4d57900-cf18-11e9-8467-abc27a08b44b.png)

Fails.

1) One observation is that when sending transaction from `unassigned1` to `delegated1` via Node1 creates transaction which is pending forever 

```
INFO ][Passive@127.0.0.1:8001]: GET 'fragment/logs'
[INFO ][Passive@127.0.0.1:8001]: fragment logs (1)
[INFO ][Passive@127.0.0.1:8001]: Fragment '512fddc692ada5e8ca28027a3a45407b9dd18b33b21ecc8c4cf4f6ed1a76d7ba' is still pending
[INFO ][Passive@127.0.0.1:8001]: GET 'fragment/logs'
[INFO ][Passive@127.0.0.1:8001]: fragment logs (1)
[INFO ][Passive@127.0.0.1:8001]: Fragment '512fddc692ada5e8ca28027a3a45407b9dd18b33b21ecc8c4cf4f6ed1a76d7ba' is still pending
[INFO ][Passive@127.0.0.1:8001]: GET 'fragment/logs'
[INFO ][Passive@127.0.0.1:8001]: fragment logs (1)
[INFO ][Passive@127.0.0.1:8001]: Fragment '512fddc692ada5e8ca28027a3a45407b9dd18b33b21ecc8c4cf4f6ed1a76d7ba' is still pending
[INFO ][Passive@127.0.0.1:8001]: GET 'fragment/logs'
[INFO ][Passive@127.0.0.1:8001]: fragment logs (1)
```

2) Also we can observe network errors in logs:
```
[INFO ][Passive@127.0.0.1:8001]: Sep 04 11:25:29.007 WARN GetBlocks response stream failed: Error { code: Unknown, source: Status { code: Unknown, message: "grpc-status header missing, mapped from HTTP status code 200" } }, node_id: 22741290214948620289923328334713000481, peer_addr: 127.0.0.1:12000, task: network
```

As a result funds are not transferred between accounts.

[full logs.txt](https://github.com/input-output-hk/jormungandr/files/3574288/full.logs.txt)

**Mandatory Information**
1. `jcli --full-version` output;
`jormungandr 0.3.3 (master-f39cb72, release, windows [x86_64]) - [rustc 1.36.0 (a53f9df32 2019-07-03)]`
[...]

PR #748, opened 2019-09-04
Fragment validator

CodeSandwich:
- Removes unused fragment process message types
- Adds fragment transaction balance validation\
Part of https://github.com/input-output-hk/jormungandr/issues/742

### DIFF
--- a/jormungandr/src/fragment/pool.rs
+++ b/jormungandr/src/fragment/pool.rs
@@ -3,7 +3,12 @@ use crate::{
     fragment::{selection::FragmentSelectionAlgorithm, Fragment, Logs},
 };
 use chain_core::property::Fragment as _;
-use futures::future::Either::{A, B};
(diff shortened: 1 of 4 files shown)


jormungandr/src/fragment/pool.rs
```diff
@@ -93,6 +110,20 @@ impl Pool {
     }
 }
 
+fn is_fragment_valid(fragment: &Fragment) -> bool {
```
NicolasDP: Don't we have something already in chain-libs?

CodeSandwich: I was looking for something like that, but couldn't find anything. All there was was witness count validation during deserialization, so I didn't have to check that and the balance check, which I used. Is there something better? Maybe we should and this check should end up in `chain-libs`.

CodeSandwich:
Finally done :)

PR #749, opened 2019-09-04
go through the whole list of trusted peers on the network

NicolasDP:
Here we try to connect to any of our trusted peers for the initial bootstrap. This is not a mandatory action to do as it only helps us get closer to the general state of the whole network.

### DIFF
--- a/jormungandr/src/network/mod.rs
+++ b/jormungandr/src/network/mod.rs
@@ -371,23 +371,40 @@ pub fn bootstrap(
     blockchain: NewBlockchain,
     branch: Branch,
     logger: &Logger,
-) -> Result<(), bootstrap::Error> {
+) -> Result<bool, bootstrap::Error> {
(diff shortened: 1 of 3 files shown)


jormungandr/src/network/mod.rs
```diff
@@ -371,23 +371,40 @@ pub fn bootstrap(
...
+                    warn!(logger, "Unable to reach peer for initial bootstrap" ; "reason" => err.to_string());
+                }
+                Err(err) => {
+                    error!(logger, "with initial bootstrap" ; "reason" => err.to_string());
+                }
+                Ok(_) => {
+                    info!(logger, "initial bootstrap completed");
+                    bootstrapped = true;
+                }
+            }
```
mzabaluev: This loop will try bootstrapping from each peer in the list, also after one bootstrap succeeds. Is this intended?

PR #764, opened 2019-09-09
Small update to improve perf and memory usage of blockchain cache

NicolasDP:
This is a rather simple change but might break a couple of pending work in different PR.

also contains a fix that should have been in #703

### DIFF
--- a/jormungandr/src/blockchain/process.rs
+++ b/jormungandr/src/blockchain/process.rs
@@ -16,7 +16,7 @@ use futures::future::Either;
 use slog::Logger;
 use tokio::{prelude::*, sync::mpsc::Sender};
 
-use std::convert::identity;
+use std::{convert::identity, sync::Arc};
 
 pub fn handle_input(
     info: &TokioServiceInfo,
@@ -70,11 +70,13 @@ pub fn handle_input(
                 });
 
             if let Some(msg_box) = explorer_msg_box {
+                /*
                 msg_box
                     .try_send(ExplorerMsg::NewBlock(new_block_ref))
                     .unwrap_or_else(|err| {
                         error!(info.logger(), "cannot add block to explorer: {}", err)
                     });
+                */
             };
             stats_counter.add_block_recv_cnt(1);
         }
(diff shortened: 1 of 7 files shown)


jormungandr/src/blockchain/process.rs
```diff
@@ -70,11 +70,13 @@ pub fn handle_input(
                 });
 
             if let Some(msg_box) = explorer_msg_box {
+                /*
                 msg_box
                     .try_send(ExplorerMsg::NewBlock(new_block_ref))
                     .unwrap_or_else(|err| {
                         error!(info.logger(), "cannot add block to explorer: {}", err)
                     });
+                */
```
mzabaluev: Will this be re-enabled in a follow-up change?

jormungandr/src/blockchain/reference_cache.rs
```diff
@@ -121,11 +121,15 @@ impl RefCacheData {
         }
     }
 
-    fn poll_purge(&mut self) -> Poll<(), timer::Error> {
-        while let Some(entry) = try_ready!(self.expirations.poll()) {
-            self.entries.remove(entry.get_ref());
+    pub fn poll_purge(&mut self) -> Poll<(), timer::Error> {
+        loop {
+            match self.expirations.poll()? {
+                Async::NotReady => return Ok(Async::Ready(())),
```
mzabaluev: It looks like it does not makes sense for this to be a poll, but with std futures, there will be an explicit context to pass down, so OK.

jormungandr/src/rest/v0/handlers.rs
```diff
@@ -19,14 +19,15 @@ use bytes::{Bytes, IntoBuf};
...
+use std::sync::Arc;
 
 pub use crate::rest::Context;
 
 macro_rules! ActixFuture {
     () => { impl Future<Item = impl Responder + 'static, Error = impl Into<Error> + 'static> + 'static }
 }
 
-fn chain_tip_fut<'a>(context: &State<Context>) -> impl Future<Item = Ref, Error = Error> {
+fn chain_tip_fut<'a>(context: &State<Context>) -> impl Future<Item = Arc<Ref>, Error = Error> {
```
mzabaluev: Maybe `Ref` could encapsulate the `Arc` under the hood, but maybe explicit is better.

NicolasDP: sure that would make sense. there are pro and cons to encapsulate and hide the `Arc`.

jormungandr/src/blockchain/process.rs
NicolasDP: yes, thanks, I need to talk with @enzoc4 about it.

PR #767, opened 2019-09-09
provide better checkpoints than before

NicolasDP:
previously the checkpoints was only providing something not really useful for the other nodes to work with and we often fall back into re-downloading the whole blockchain.

Now we can leverage our great multiverse/hamt data structure and utilise the backward reference to previous block of interest. Skipping an increasingly larger of amount of epochs so after a year we still don't need to provide hundreds of header hash but a handful.

### DIFF
--- a/jormungandr/src/blockchain/checkpoints.rs
+++ b/jormungandr/src/blockchain/checkpoints.rs
@@ -1,0 +1,63 @@
+use crate::{blockcfg::HeaderHash, blockchain::Ref};
+use std::sync::Arc;
+
+/// list of pre-computed checkpoints from a given [`Ref`].
+///
+/// [`Ref`]: ./struct.Ref.html
+pub struct Checkpoints(Vec<HeaderHash>);
+
+impl Checkpoints {
+    /// create a new list of checkpoints from the given starting point (tip).
+    ///
+    /// The function make use of the `Ref` object pointing to previous epoch in order
+    /// to populate _interestingly_ spaced checkpoints.
+    ///
+    /// For now the algorithm provide the current block, the parent, the last block of the
+    /// previous epoch, the last block of the epoch before that... until the block0.
+    pub fn new_from(from: Arc<Ref>) -> Self {
+        let mut checkpoints = vec![from.hash(), from.block_parent_hash().clone()];
+
+        let mut ignore_prev = 0;
+        let mut current_ref = from;
+        while let Some(prev_epoch) = current_ref.last_ref_previous_epoch() {
+            current_ref = Arc::clone(prev_epoch);
+
+            for _ in 0..ignore_prev {
+                if let Some(prev_epoch) = current_ref.last_ref_previous_epoch() {
+                    current_ref = Arc::clone(&prev_epoch);
+                } else {
+                    break;
+                }
+            }
+
+            ignore_prev += 1;
+
+            checkpoints.push(current_ref.hash());
+        }
+
+        Checkpoints(checkpoints)
+    }
+
+    pub fn iter(&self) -> impl Iterator<Item = &HeaderHash> {
+        self.0.iter()
+    }
+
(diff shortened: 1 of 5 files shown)


jormungandr/src/blockchain/checkpoints.rs
```diff
@@ -0,0 +1,70 @@
...
+    /// to populate _interestingly_ spaced checkpoints.
+    ///
+    /// For now the algorithm provide the current block, the parent, the last block of the
+    /// previous epoch, the last block of the epoch before that... until the block0.
+    pub fn new_from(from: Arc<Ref>) -> Self {
+        let mut checkpoints = vec![from.hash(), from.block_parent_hash().clone()];
+
+        let mut ignore_prev = 0;
+        let mut current_ref = from;
+        while let Some(prev_epoch) = current_ref.last_ref_previous_epoch() {
```
mzabaluev: If the starting block's parent is the last in the previous epoch, will it be added twice?

jormungandr/src/blockchain/checkpoints.rs
```diff
@@ -0,0 +1,70 @@
...
+
+impl Checkpoints {
+    /// create a new list of checkpoints from the given starting point (tip).
+    ///
+    /// The function make use of the `Ref` object pointing to previous epoch in order
+    /// to populate _interestingly_ spaced checkpoints.
+    ///
+    /// For now the algorithm provide the current block, the parent, the last block of the
+    /// previous epoch, the last block of the epoch before that... until the block0.
+    pub fn new_from(from: Arc<Ref>) -> Self {
```
mzabaluev: This should be just `new` IMHO, no need to have strange constructor names.

jormungandr/src/blockchain/checkpoints.rs
NicolasDP: good catch!

jormungandr/src/blockchain/checkpoints.rs
```diff
@@ -0,0 +1,70 @@
...
+        let mut checkpoints = vec![from.hash(), from.block_parent_hash().clone()];
+
+        let mut ignore_prev = 0;
+        let mut current_ref = from;
+        while let Some(prev_epoch) = current_ref.last_ref_previous_epoch() {
+            current_ref = Arc::clone(prev_epoch);
+
+            for _ in 0..ignore_prev {
```
mzabaluev: This walks the epochs linearly, but skips incremental numbers of them. As the blockchain grows, the time to compute this every time will grow as well.

jormungandr/src/blockchain/checkpoints.rs
NicolasDP: `new` would mean `empty`. here we are creating checkpoints *from* the given object.

jormungandr/src/blockchain/checkpoints.rs
NicolasDP: but it does not matter because the epoch grow slowly. 1 every 3 or 4 days. This will work for now and we can find more precise construction later on

mzabaluev: Should we have some sort of O(1)-ish random access by epoch number which can be used here?

NicolasDP: we have time for that really. Right now after 2 years of blockchain, there is 143 epochs : https://cardanoexplorer.com

NicolasDP: It's cheaper that way. Maintaining an indexed accessor to epoch or checkpoints would make things more difficult for very little gain compared to the current approach.

With the current approach from the multiverse and the HAMT we have both rollbacks and branch for free. And again, unless in a testing mode, you are not going to create that many epochs per year. As counter intuitive as it feels the linked list is way faster here (and more certain).

PR #771, opened 2019-09-09
Use imhamt and multiverse in explorer

enzoc4:
# Changes in this PR

## Change the explorer data structures to make handling rollbacks easier. 
Basically something like this: 

```rust
pub struct ExplorerDB {
    multiverse: Multiverse<State>,
    longest_chain_tip: Lock<Option<Block>>,
}

type ExplorerBlock = Block;
type Hamt<K, V> = imhamt::Hamt<DefaultHasher, K, V>;

struct State {
    transactions: Hamt<FragmentId, HeaderHash>;,
    blocks: Hamt<HeaderHash, ExplorerBlock>,
    addresses: Hamt<ChainLength, HeaderHash>,
    epochs: Hamt<Address, Hamt<FragmentId, ()>>,
    chain_lengths: Hamt<Epoch, EpochData>,
}
```
## Create a general function/method `apply_block` to make restoring easier.

This function takes the parent of this block and updates the stake accordingly. It currently only logs an error if this previous state is not found. 

## Re-enable the block processing

And also start processing blocks from the network
 
## Add function to load blocks from storage when restarting the node.

By using the longest chain known by the node at the point of shutdown.

## Expand query support

At this point is possible to query:
- Blocks by id and chain length
Example 
```graphql
{
  blockByChainLength(length: "79") {
    date {
      slot
      epoch {
        id
      }
    }
  }
}
```
- Epochs by number

```graphql
{
  epoch(id: "0") {
    lastBlock {
      id
      chainLength
    }
  }
}
```
- Get latest block.

Example: 

```graphql
{
  status {
    latestBlock {
      id
      chainLength
    }
  }
}
```

- Transactions by id and by address

For example: 
```graphql
{
  transaction(id: "3b30d56c916e2974b3d8ba0d088dce54ea1c1bf5d907f8df2f4ee0a6b9484b0e") {
    inputs {
      amount
      address {
        id
      }
    }
    outputs {
      amount
      address {
        id
      }
    }
  }
}
```

By address: 

```graphql
{
  address(bech32: "ca1skmmqxvaesaew6aygcpkshrs0et80rh6hutkzlt6k6dpgewyarwv7zl5efq") {
    transactions {
      id
    }
  }
}
```


jormungandr/src/explorer/mod.rs
```diff
@@ -66,112 +89,303 @@ impl Explorer {
...
+                .insert_or_update(
+                    output.address.clone(),
+                    [txid].iter().cloned().collect(),
+                    |set| {
+                        let mut new_set: HashSet<FragmentId> = set.iter().cloned().collect();
+                        new_set.insert(txid);
+                        std::result::Result::<_, Infallible>::Ok(Some(new_set))
+                    },
+                )
+                // FIXME: I still don't know when a insert or update can fail
```
NicolasDP: `insert_or_update` will fail if you return an error in the callback. But you said `Infaillible` so it won't. easy.

jormungandr/src/explorer/mod.rs
```diff
@@ -66,112 +89,303 @@ impl Explorer {
...
+                .unwrap();
+        }
+        for input in inputs {
+            match input.to_enum() {
+                InputEnum::AccountInput(id, _value) => {
+                    // TODO: How do I get an Address from an AccountIdentifier?
+                    // Can I do it without knowing the Discrimination?
```
NicolasDP: easy: from the block0/genesis block.

But there is another problem/question for you is that you also don't know what `prefix` to use for the address in its bech32 representation.

a question for @vincenthz to help with maybe.

From my point of view we can either:

* add this in the configuration file of the node (under explorer: address prefix);
* let the user decide from the GraphQL query;
* just ignore it and return the public keys only?

jormungandr/src/explorer/mod.rs
```diff
@@ -66,112 +89,303 @@ impl Explorer {
...
+fn apply_block_to_blocks(blocks: Blocks, block: &Block) -> Result<Blocks> {
+    let block_id = block.id();
+    blocks
+        .insert(block_id, (*block).clone())
+        .map_err(|_| Error::from(ErrorKind::BlockAlreadyExists(format!("{}", block_id))))
+}
+
+fn apply_block_to_addresses(addresses: Addresses, block: &Block) -> Result<Addresses> {
+    // FIXME: Probably there is a cleaner way to do this
+    let txs = block.contents.iter().filter_map(|fragment| match fragment {
```
NicolasDP: you are ignoring the certificate contents here. I'd have this pattern match in the loop below and do the right thing depending of the kind of fragment it is

NicolasDP: You will also need to split this function into small ones to help have something more _precise_.

jormungandr/src/explorer/mod.rs
```diff
@@ -66,112 +89,303 @@ impl Explorer {
...
-        let chain_length = new_block_ref.chain_length();
-        let header_hash = new_block_ref.hash();
+        let blocks = apply_block_to_blocks(Blocks::new(), &block0)?;
+        let epochs = apply_block_to_epochs(Epochs::new(), &block0)?;
+        let chain_lengths = apply_block_to_chain_lengths(ChainLengths::new(), &block0)?;
 
-        // Clone things to move into closures, this is just cloning locks
-        let mut map = self.chain_length_to_hash.clone();
-        let multiverse = self.multiverse.clone();
+        // XXX: I think the block0 doesn't have transactions
```
NicolasDP: it does, they are the initial funds (from ICO or whatever...)

jormungandr/src/explorer/mod.rs
enzoc4: for sure. I need to improve that piece of code.

jormungandr/src/explorer/mod.rs
enzoc4: my mistake then. I'll change that

PR #772, opened 2019-09-10
Fall back to block 0 when no starting checkpoints match

mzabaluev:
Also, fail early on `Storage::get_with_info` errors instead of treating it as if the block was not found.

### DIFF
--- a/jormungandr/src/client.rs
+++ b/jormungandr/src/client.rs
@@ -6,35 +6,50 @@ use chain_core::property::HasHeader;
 use chain_storage::store;
 use tokio::prelude::*;
 
-pub fn handle_input(
-    _info: &ThreadServiceInfo,
-    storage: &Storage,
-    blockchain_tip: &Branch,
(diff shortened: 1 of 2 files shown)


jormungandr/src/client.rs
```diff
@@ -47,37 +62,44 @@ fn handle_get_block_tip(blockchain_tip: &Branch) -> Result<Header, Error> {
...
-        })
-        .max_by_key(|&(depth, _)| depth)
-        .map(|(_, hash)| *hash)
+fn find_latest_checkpoint(
+    checkpoints: &[HeaderHash],
+    storage: &Storage,
+    block0_hash: &HeaderHash,
+) -> Result<HeaderHash, Error> {
+    // Filter out the checkpoints that don't exist in the storage;
+    // among the checkpoints present, find the latest by chain length.
```
NicolasDP: I think we should assume they are already sorted from the client point of view as the best to worse choice. Right?

PR #784, opened 2019-09-16
Link network fragment subscription with fragment pool

CodeSandwich:
- Push fragments from network fragment subscription to fragment pool
- Broadcast new fragments added to pool over network fragment subscription

### DIFF
--- a/jormungandr/src/fragment/pool.rs
+++ b/jormungandr/src/fragment/pool.rs
@@ -1,29 +1,38 @@
 use crate::{
     blockcfg::{HeaderContentEvalContext, Ledger, LedgerParameters},
     fragment::{selection::FragmentSelectionAlgorithm, Fragment, Logs},
+    intercom::{NetworkMsg, PropagateMsg},
+    utils::async_msg::MessageBox,
 };
 use chain_core::property::Fragment as _;
-use chain_impl_mockchain::transaction::{AuthenticatedTransaction, Balance};
-use chain_impl_mockchain::value::Value;
-use futures::future::{
-    self,
-    Either::{A, B},
-};
+use chain_impl_mockchain::transaction::AuthenticatedTransaction;
 use jormungandr_lib::interfaces::{FragmentLog, FragmentOrigin};
 use std::time::Duration;
-use tokio::{prelude::*, sync::lock::Lock, timer};
+use tokio::{
+    prelude::{
+        future::{
+            self,
+            Either::{A, B},
+        },
+        stream, Async, Future, Poll, Sink, Stream,
+    },
+    sync::lock::Lock,
+    timer,
+};
 
 #[derive(Clone)]
 pub struct Pool {
     logs: Logs,
     pool: Lock<internal::Pool>,
+    network_msg_box: MessageBox<NetworkMsg>,
 }
 
 impl Pool {
-    pub fn new(ttl: Duration, logs: Logs) -> Self {
+    pub fn new(ttl: Duration, logs: Logs, network_msg_box: MessageBox<NetworkMsg>) -> Self {
         Pool {
             logs,
             pool: Lock::new(internal::Pool::new(ttl)),
+            network_msg_box,
         }
     }
 
@@ -32,7 +41,7 @@ impl Pool {
     }
 
     /// Returns true if fragment was registered
-    pub fn insert(
+    pub fn insert_and_propagate(
         &mut self,
         origin: FragmentOrigin,
         fragment: Fragment,
@@ -42,6 +51,7 @@ impl Pool {
         }
         let mut pool_lock = self.pool.clone();
         let mut logs = self.logs.clone();
+        let mut network_msg_box = self.network_msg_box.clone();
         B(self
             .logs
             .exists(fragment.id())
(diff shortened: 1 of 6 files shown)


jormungandr/src/fragment/pool.rs
```diff
@@ -1,80 +1,54 @@
...
-                        let fragment_log = FragmentLog::new(id.into(), origin);
-                        let insert_future = logs.insert(fragment_log).map(|_| true);
-                        B(insert_future)
-                    }),
-                )
-            }))
-    }
-
     /// Returns number of registered fragments
     pub fn insert_all(
```
mzabaluev: The method name could be changed to indicate that messaging also takes place. `insert_and_propagate_all`?

jormungandr/src/fragment/pool.rs
```diff
@@ -85,11 +59,18 @@ impl Pool {
...
+                    let count = new_fragments.len();
+                    let fragment_logs = new_fragments
+                        .iter()
+                        .map(move |fragment| FragmentLog::new(fragment.id().into(), origin))
+                        .collect::<Vec<_>>();
+                    for fragment in new_fragments {
+                        let network_msg = NetworkMsg::Propagate(PropagateMsg::Fragment(fragment));
+                        network_msg_box.try_send(network_msg).unwrap_or_else(|err| {
+                            error!(logger, "cannot propagate fragment to network: {}", err)
+                        });
```
mzabaluev: `MessageBox` wraps `tokio::sync::mpsc::Sender`, so using `try_send` this way can get messages dropped on the floor if the capacity of the buffer overflows.

Try to return `new_fragments` as the result of the poll future and add an `add_then` to process them in a sequence of futures returned by `network_msg_box.send(...)`.

jormungandr/src/fragment/pool.rs
```diff
@@ -85,12 +65,20 @@ impl Pool {
...
-                    logs.insert_all(fragment_logs).map(move |_| count)
+                    let new_fragments = pool.insert_all(new_fragments);
+                    let count = new_fragments.len();
+                    let fragment_logs = new_fragments
+                        .iter()
+                        .map(move |fragment| FragmentLog::new(fragment.id().into(), origin))
+                        .collect::<Vec<_>>();
+                    stream::iter_ok(new_fragments)
+                        .map(|fragment| NetworkMsg::Propagate(PropagateMsg::Fragment(fragment)))
+                        .for_each(move |network_msg| network_msg_box.try_send(network_msg))
```
mzabaluev: More async, but still `try_send`? It's not recommended by the `Sender` docs. You can make it overflow-proof by doing `send` (of trait `Sink`) inside the `for_each` closure, if I'm not mistaken. Or, better and shorter, feed the whole stream of `PropagateMsg` into the message box with `send_all`.

CodeSandwich: Sorry, a typo, I wanted to use `send`. `send_all` looks better

CodeSandwich: `send_all` and `forward` are the only way to send whole stream at once into a sink, but both of them close the sink after flushing. There's no workaround for that other than writing custom stream consumer that pushes items one by one: https://github.com/rust-lang-nursery/futures-rs/issues/462

We can just ignore this fact, because luckily Tokio MPSC does nothing on close. This is ugly, illegal and relies on undocumented behavior. I guess that we can use that temporarily. If MPSC behavior changes before we migrate to std futures infrastructure (which is not even there), the tests will probably notice that.

mzabaluev: No, better just do `send` inside `for_each`.

mzabaluev: An incidental observation is that `SendAll` does not have the poll-after-done bug that `Forward` has in futures 0.1.29, being otherwise equivalent. I can remove the fuse workaround this way.

jormungandr/src/fragment/pool.rs
```diff
@@ -85,12 +65,22 @@ impl Pool {
...
-                    logs.insert_all(fragment_logs).map(move |_| count)
+                    let new_fragments = pool.insert_all(new_fragments);
+                    let count = new_fragments.len();
+                    let fragment_logs = new_fragments
+                        .iter()
+                        .map(move |fragment| FragmentLog::new(fragment.id().into(), origin))
+                        .collect::<Vec<_>>();
+                    stream::iter_ok(new_fragments)
+                        .map(|fragment| NetworkMsg::Propagate(PropagateMsg::Fragment(fragment)))
+                        .fold(network_msg_box, |network_msg_box, fragment_msg| {
```
mzabaluev: I wanted to tell you that cloning the box can be avoided this way, well done :)

jormungandr/src/fragment/pool.rs
```diff
@@ -85,12 +65,22 @@ impl Pool {
...
+                            network_msg_box.send(fragment_msg)
+                        })
+                        .map_err(move |err: <MessageBox<_> as Sink>::SinkError| {
```
mzabaluev: Hmm, interesting that it should need explicit type annotation...

jormungandr/src/network/subscription.rs
```diff
@@ -57,7 +59,8 @@ where
     tokio::spawn(
         inbound
             .for_each(move |fragment| {
-                // TODO: implement
+                let msg = TransactionMsg::SendTransaction(FragmentOrigin::Network, vec![fragment]);
+                transaction_box.try_send(msg).unwrap();
```
mzabaluev: I did not get to this place last time. Would it benefit from the same treatment?

jormungandr/src/fragment/pool.rs
CodeSandwich: I was surprised too

jormungandr/src/network/subscription.rs
CodeSandwich: This is coming from the outside, so if we run out of MPSC buffer, we probably should just drop it instead of queuing on Tokio.

I'll change it so it ignores full queue error, but returns error in other cases.

mzabaluev: No, the correct thing is to implement back pressure that will pause the inbound stream. I've looked at the similar piece of code that processes incoming blocks and it's wrong in the same way, except that blocks are large and will probably be processed faster than they are received.

mzabaluev: I can change both places in a follow-up PR.

CodeSandwich: Thanks! So this comment is resolved?

mzabaluev: Yes.

PR #787, opened 2019-09-17
network: Replace forward combinator with send_all

mzabaluev:
An equivalent way to forward a stream into sink that does not need the fuse workarounds of #780.

### DIFF
--- a/jormungandr/src/network/client.rs
+++ b/jormungandr/src/network/client.rs
@@ -272,9 +272,7 @@ where
                             error!(err2_logger, "sending to block task failed: {:?}", e);
                         })
                         .and_then(move |_| {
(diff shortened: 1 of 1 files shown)


jormungandr/src/network/client.rs
```diff
@@ -357,29 +350,24 @@ where
...
+                    });
+                    InboundProcessing::with_unary(
+                        block_box.clone(),
+                        logger.clone(),
+                        |block, reply| BlockMsg::NetworkBlock(block, reply),
+                    )
+                    .sink_map_err(move |e| {
+                        warn!(sink_err_logger, "network block validation failed: {:?}", e)
+                    })
+                    .send_all(stream)
```
NicolasDP: much simpler to read as well

PR #794, opened 2019-09-19
fix silly bug in the checkpoint

NicolasDP:
This bug was introduced with the latest release...


NicolasDP:
adding network label because this affects the nodes on the network

Issue #795, opened 2019-09-20
MemPool: the obsolete fragments should definitely be purged


NicolasDP:
The root cause is that when block is received, its fragments aren't removed from the fragment pool. This should be harmless in greater scheme of things, but the obsolete fragments should definitely be purged to prevent clutter and memory leak. I'll fix that tomorrow.

_Originally posted by @CodeSandwich in https://github.com/input-output-hk/jormungandr/issues/742#issuecomment-533327820_

CodeSandwich:
Possible real solutions, supporting rollback:
1. Update the mempool to match the same model as blockchain, storing the items in the IMHAMT and utilising the multiverse to track the different potentialities
2. Push fragments back from block into pool on rollback

NicolasDP:
2. On rollback: just select from the multiverse the right imhamt. O(1). No need to push fragments.

PR #798, opened 2019-09-20
Remove fragments added to block from fragment pool V1

CodeSandwich:
The naive fix for https://github.com/input-output-hk/jormungandr/issues/795. Whenever block is received from the network and added to the blockchain remove all its fragments from fragment pool. No rollback support. Should be enough to fix https://github.com/input-output-hk/jormungandr/issues/742.


CodeSandwich:
I'm starting vacation. If there are some issues, please take over the branch and fix them, thank you :hugs:

PR #801, opened 2019-09-21
only update the block counter once the block has been validated

NicolasDP:
It is possible some other nodes _propagate_ common blocks to us. Reduce the blockcounter stats to only what we have actually validated.

### DIFF
--- a/jormungandr/src/blockchain/process.rs
+++ b/jormungandr/src/blockchain/process.rs
@@ -99,7 +99,6 @@ pub fn handle_input(
             future.wait().unwrap();
         }
         BlockMsg::NetworkBlock(block, reply) => {
-            stats_counter.add_block_recv_cnt(1);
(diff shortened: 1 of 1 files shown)


jormungandr/src/blockchain/process.rs
```diff
@@ -99,14 +99,14 @@ pub fn handle_input(
...
         BlockMsg::NetworkBlock(block, reply) => {
-            stats_counter.add_block_recv_cnt(1);
             let fragment_ids = block.fragments().map(|f| f.id()).collect::<Vec<_>>();
             let future = process_network_block(blockchain.clone(), block, info.logger().clone());
             match future.wait() {
                 Err(e) => {
                     reply.reply_error(network_block_error_into_reply(e));
                 }
                 Ok(maybe_updated) => {
+                    stats_counter.add_block_recv_cnt(1);
```
mzabaluev: Should this be incremented only when the tip is actually updated, i.e. `maybe_updated` is not `None`?

PR #805, opened 2019-09-23
Get the trusted peer list from the CLI and make `--config` optional

NicolasDP:
This makes the config optional and we can now get the list of trusted peers from the command line.

also a quick hack for node without a `listen_address`.


jormungandr/src/settings/start/mod.rs
```diff
@@ -124,28 +125,54 @@ impl RawSettings {
...
             explorer,
         })
     }
 }
 
 fn generate_network(
-    _command_arguments: &StartArguments,
-    config: &Config,
+    command_arguments: &StartArguments,
+    config: &Option<Config>,
```
mzabaluev: Consider initializing `Config` with default values in case the file is missing, rather than dragging the `Option` through the APIs.

PR #807, opened 2019-09-23
reduce the error level when peers have invalid addresses

NicolasDP:
this is creating a lot of noise for no necessary reasons and scaring users.


mzabaluev:
Looks OK. #810 should remove outright unusable addresses from gossip.

PR #810, opened 2019-09-23
Filter private addresses from gossip

### LINKED ISSUE #799: Filter addresses that are not reachable
something to add ASAP. It's important to filter addresses that are not reachable `10.0.0.0/8` or `0.0.0.0`.

mzabaluev:
Resolves #799

### DIFF
--- a/jormungandr/src/network/subscription.rs
+++ b/jormungandr/src/network/subscription.rs
@@ -10,7 +10,8 @@ use crate::{
 };
 use futures::prelude::*;
 use jormungandr_lib::interfaces::FragmentOrigin;
-use network_core::{error as core_error, gossip::Gossip};
+use network_core::error as core_error;
+use network_core::gossip::{Gossip, Node as _};
 use slog::Logger;
 
 pub fn process_block_announcements<S>(
@@ -109,7 +110,7 @@ where
                 trace!(logger, "received gossip: {:?}", gossip);
                 let nodes = gossip
                     .into_nodes()
-                    .filter(|node| filter_gossip_node(node, &state.config));
+                    .filter(|node| filter_gossip_node(node, &state.config, &logger));
                 state.topology.update(nodes);
                 Ok(())
             })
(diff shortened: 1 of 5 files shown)


mzabaluev:
This results in log spam if the connecting node's reported address in gossip is not acceptable.
Should this be fixed by:
1. disconnecting such nodes completely rather than just filtering out bad gossip items;
2. removing the log message, to detriment of the network admin getting a clue about the configuration problem?

jormungandr/src/settings/start/config.rs
```diff
@@ -79,6 +79,10 @@ pub struct P2pConfig {
     /// help the different modules of the P2P topology engine to determine the
     /// best possible neighborhood.
     pub topics_of_interest: Option<BTreeMap<Topic, InterestLevel>>,
+
+    /// Whether to allow non-public IP addresses on the network
+    #[serde(default)]
+    pub allow_private_addresses: bool,
```
NicolasDP: what is the default value? It's not obvious to me

could you add it in the documentation too? I'd like to see the default set to _false_ by default. Is it the case?

jormungandr/src/network/subscription.rs
```diff
@@ -105,12 +107,40 @@ where
...
+        node.is_global()
+    };
+    if !accepted {
+        if let Some(addr) = node.address() {
+            warn!(
+                logger,
+                "node {} has unacceptable IP address {}, pruned from gossip",
+                node.id(),
+                addr.ip()
+            );
```
NicolasDP: We will get a lot of logs, could we filter them all and then says something like:

filtering out these `[addresses]` from the gossips from `address` of the one who sent us these gossips?

jormungandr/src/settings/start/config.rs
mzabaluev: Yes, it should be false by default, but I don't know how to specify this explicitly for serde-yaml. I'll add the information to the docs.

NicolasDP: no worries. Just needs it to be documented in the comment message

jormungandr/src/network/subscription.rs
mzabaluev: Done in a follow-up comment. The peer address should be already added to the `Logger` context of the connection. If not, it should be fixed elsewhere.

PR #811, opened 2019-09-23
Make bump_peer_for_block_fetch miss non-fatal

mzabaluev:
In `network::p2p::comm::PeerMap::bump_peer_for_block_fetch`,
don't panic if the node to bump is no longer in the map.
This can happen in the wild if an announcement is processed after the peer comms entry is removed from the map due to (probably) a subscription stream having closed.


NicolasDP:
I believe I saw an issue refering the problem, @dkijania @mzabaluev do you remember which one it was?

dkijania:
closes #818

PR #817, opened 2019-09-24
Mitigation for the cases where we don't have local public address and listen address

NicolasDP:
this will create something we don't want to do for the long run but mitigate the issues at hands.

### DIFF
--- a/jormungandr/src/network/grpc/server.rs
+++ b/jormungandr/src/network/grpc/server.rs
@@ -4,15 +4,11 @@ use network_grpc::server::{self, Server};
 use tokio::prelude::*;
 
 pub fn run_listen_socket(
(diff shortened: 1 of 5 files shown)


jormungandr/src/network/grpc/server.rs
```diff
@@ -4,11 +4,15 @@ use network_grpc::server::{self, Server};
 use tokio::prelude::*;
 
 pub fn run_listen_socket(
-    listen: Listen,
+    listen: Option<Listen>,
```
mzabaluev: It's better to supply the fake `Listen` address outside of the function.

NicolasDP:
Closing this PR as it is not the solution anyway.

Issue #837, opened 2019-09-26
improve jcli certificate new-stake-pool-certificate error handling


NicolasDP:
In `jcli` code, it is possible to check that the list of `owner` is at list the number of `threshold`

https://github.com/input-output-hk/jormungandr/blob/c5a92b7a6d0a6c23d208b60553c2ba110368cd70/jcli/src/jcli_app/certificate/new_stake_pool_registration.rs#L51-L67

i.e. `self.owners.len() >= self. management_threshold` and if not, returns an appropriate error in https://github.com/input-output-hk/jormungandr/blob/c5a92b7a6d0a6c23d208b60553c2ba110368cd70/jcli/src/jcli_app/certificate/mod.rs#L13-L22

This would improve the user experience a lot.

_Originally posted by @NicolasDP in https://github.com/input-output-hk/jormungandr/issues/836#issuecomment-535396362_

Issue #832, opened 2019-09-25
(documentation) Update the documentation for `jcli certificate new stake-pool-registration` 


dorin100:
**Describe the bug**
There were added new mandatory parameters to `jcli certificate new stake-pool-registration` that are not documented in the official documentation (https://input-output-hk.github.io/jormungandr/stake_pool/registering_stake_pool.html#creating-your-stake-pool-certificate)

**Mandatory Information**
```
jormungandr 0.5.0 (HEAD-98922c77, release, windows [x86_64]) - [rustc 1.36.0 (a53f9df32 2019-07-03)]
jcli 0.5.0 (HEAD-98922c77, release, windows [x86_64]) - [rustc 1.36.0 (a53f9df32 2019-07-03)]
```

**To Reproduce**
1. https://input-output-hk.github.io/jormungandr/stake_pool/registering_stake_pool.html#creating-your-stake-pool-certificate

```
$ jcli certificate new stake-pool-registration --kes-key ${POOL_KES_PK} --vrf-key ${POOL_VRF_PK} --serial 1010101010 > stake_pool.cert
error: The following required arguments were not provided:
    --start-validity <SECONDS-SINCE-START>
    --management-threshold <THRESHOLD>

USAGE:
    jcli.exe certificate new stake-pool-registration [OPTIONS] --kes-key <KES_KEY> --start-validity <SECONDS-SINCE-START> --serial <SERIAL> --management-threshold <THRESHOLD> --vrf-key <VRF_KEY> [--] [output]

For more information try --help
```

**Expected behavior**
All command parameters are clearly described in the official documentation. A correct and functional example should also be provided. 

**Additional context**
It seems that the certificate is not created correctly also if we don't use `--owner` parameter so I suppose this should be mandatory too. (and documented)

dorin100:
When I am trying to create a stake pool using some random values for `--start-validity` and `--management-threshold`, the transaction for submitting the stake pool registration certificate is rejected as below. I am not sure that the values of the 2 parameters is causing it but it can be.

```jcli certificate new stake-pool-registration --kes-key ${POOL_KES_PK} --vrf-key ${POOL_VRF_PK} --serial 1010101010 --start-validity 1000000 --management-threshold 10 > stake_pool.cert```

```
fragment_id: 2508faa998eb8e6486e616cc07055b66f1f7c3afaf2a993d16d4ca50f5e3475c
  last_updated_at: "2019-09-25T13:38:31.002804300+00:00"
[...]

PR #838, opened 2019-09-26
JCLI: management_threshold validity

rinor:
Fix: https://github.com/input-output-hk/jormungandr/issues/837
Helps: https://github.com/input-output-hk/jormungandr/issues/832

### DIFF
--- a/jcli/src/jcli_app/certificate/mod.rs
+++ b/jcli/src/jcli_app/certificate/mod.rs
@@ -19,6 +19,8 @@ custom_error! {pub Error
     OutputInvalid { source: std::io::Error, path: PathBuf }
         = @{{ let _ = source; format_args!("invalid output file path '{}'", path.display()) }},
     InvalidCertificate { source: CertificateFromStrError } = "Invalid certificate",
+    ManagementThresholdInvalid { got: usize, max_expected: usize }
+        = "invalid management_threshold value, expected between at least 1 and {max_expected} but got {got}",
 }
 
 #[derive(StructOpt)]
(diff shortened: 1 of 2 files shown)


jcli/src/jcli_app/certificate/mod.rs
```diff
@@ -19,6 +19,8 @@ custom_error! {pub Error
     OutputInvalid { source: std::io::Error, path: PathBuf }
         = @{{ let _ = source; format_args!("invalid output file path '{}'", path.display()) }},
     InvalidCertificate { source: CertificateFromStrError } = "Invalid certificate",
+    ManagementThresholdInvalid { got: String, expected: String }
```
NicolasDP:
```suggestion
    ManagementThresholdInvalid { got: usize, expected: usize }
```

jcli/src/jcli_app/certificate/mod.rs
```diff
@@ -19,6 +19,8 @@ custom_error! {pub Error
...
+        = @{{ format_args!("invalid management_threshold value, expected between 1 - {}, got {}", expected, got) }},
```
NicolasDP:
```suggestion
        = "invalid management_threshold value, expected between at least {expected} but got {got}"
```

NicolasDP:
thank you @rinor. Just a couple of things that needs to be updated.

jcli/src/jcli_app/certificate/mod.rs
rinor: AFAIK management_threshold can be in the range [1 - number of owners].
`expected` here contains the number of owners, probably should rename it to `max_expected` and maybe make 1 as `min_expected`.

Shouldn't the message be **at most** and not **at least**?
If yes it will look akward when `management_threshold = 0`

PR #873, opened 2019-10-01
stop the node if we detect a panic in a runtime

NicolasDP:
Currently if the node experience a panic, we may not detect it due to log overloading. This will make sure the node stops as soon as one of the runtime environment detects a panic or any exceptions.

### DIFF
--- a/jormungandr/src/utils/task.rs
+++ b/jormungandr/src/utils/task.rs
@@ -174,7 +174,19 @@ impl Services {
             executor: executor,
         };
 
-        runtime.spawn(f(future_service_info));
+        use std::panic::AssertUnwindSafe;
+
+        let future = AssertUnwindSafe(f(future_service_info))
+            .catch_unwind()
+            .map(|_| ())
+            .map_err(|err| {
+                if let Some(string) = err.downcast_ref::<String>() {
+                    eprintln!("{}", string);
+                }
+                std::process::exit(66);
+            });
+
+        runtime.spawn(future);
 
         let task = Service::new_runtime(name, runtime, now);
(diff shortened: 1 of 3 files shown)


jormungandr/src/utils/task.rs
```diff
@@ -183,7 +183,19 @@ impl Services {
             executor: executor,
         };
 
-        runtime.spawn(f(future_service_info));
+        use std::panic::AssertUnwindSafe;
+
+        let future = AssertUnwindSafe(f(future_service_info))
```
mzabaluev: I would not assert it for a generic closure, but add `UnwindSafe` bound for the `F` closure type, and address unwind safety in tasks on a case-by-case basis, if required by the compiler.

NicolasDP: tried that and that is 2 years of software development to resolve it that way

mzabaluev: Not sure what is required besides, in the worst case, shifting the `AssertUnwindSafe` wrapper to every use site for the few tasks we have, to leave audit markers for re-evaluation later if this causes problems.
But let's move it forward.

NicolasDP:
Yes. I think now that the panic actually excite the program we can see there was a test that should have failed a while ago. We didn't see it before but the panic was always there, but the node continued on the other tasks.

I had updated the PR with the necessary changes in chain-storage-sqlite to not unwrap on every error 🙈

jormungandr/src/utils/task.rs
NicolasDP: I have created an issue to track that down at some point

PR #878, opened 2019-10-01
Add certificate query for transaction

enzoc4:
Summary

- Store the certificates in the transactions
- Move custom scalars to a submodule
- Add the certificate as a field in the Transaction

Example query:

```graphql
{
  transaction(id: "4eb6662f057d204c2f31386d765d4a38a1e6c7fb00778a9d5d9e7233a261f4ac") {
    id
    certificate {
      __typename
      ... on StakeDelegation {
        account {
          id
        }
        pool {
          id
        }
      }
      ... on PoolRegistration {
        pool {
          id
        }
        serial
        startValidity
        managementThreshold
        owners
      }
    }
  }
}
```

Result currently implemented as: 

```graphql
union Certificate = StakeDelegation | OwnerStakeDelegation | PoolRegistration
```

The problem with this interface is that it is that graphql unions can't have unions or interfaces as types, so it is not possible to map  the PoolManagement to a union too:

```rust
pub enum PoolManagement {
    Update(PoolOwnersSigned<PoolUpdate>),
    Retirement(PoolOwnersSigned<PoolRetirement>),
}
```

One alternative is to make two types:

```graphql
type PoolManagementUpdate {}
type PoolManagementRetirement {}
```

The other is to implement `Certificate` as a interface with a tag (interfaces must have at least one field) instead as a union:

```graphql
enum CertificateType {
  StakeDelegation
  OwnerStakeDelegation
  PoolRegistration
  PoolManagement
}

interface Certificate {
  type: CertificateType
}
```
One cons of this alternative is that the `type` field is redundant with the auto-generated  `__typename`

### DIFF
--- a/jormungandr/src/explorer/graphql/mod.rs
+++ b/jormungandr/src/explorer/graphql/mod.rs
@@ -1,15 +1,19 @@
 mod error;
+mod scalars;
 use self::error::ErrorKind;
 use super::indexing::{EpochData, ExplorerBlock, ExplorerTransaction};
 use crate::blockcfg::{self, FragmentId, HeaderHash};
-use chain_impl_mockchain::value;
+use chain_impl_mockchain::certificate;
 pub use juniper::http::GraphQLRequest;
-use juniper::EmptyMutation;
-use juniper::FieldResult;
-use juniper::RootNode;
-use std::convert::{TryFrom, TryInto};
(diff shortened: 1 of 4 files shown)


vincenthz:
@enzoc4 if this is better, we could flatten the PoolManagement into 2 fragment types.

Issue #888, opened 2019-10-03
500 Internal Server error querying next-id


disassembler:
**Describe the bug**

I got this error running cardano-wallet against testnet (I know it's not supported yet). I was able to reproduce on a separate node with just `jcli` command below, so figure the bug is still probably worth looking into.

```
jcli rest v0 block c79855ee1585d688e17278601d5a95a2f8608ad33aeddf6b07c73610f3a665df next-id get
failed to make a REST request
  |-> node internal error
  |-> http://10.40.33.1:3101/api/v0/block/c79855ee1585d688e17278601d5a95a2f8608ad33aeddf6b07c73610f3a665df/next_id: Server Error: 500 Internal Server Error
```

Log shows:
```
Error occurred during request handling, status: 500 Internal Server Error Tip block is parent of a block
```
**Mandatory Information**
1. `jcli --full-version` output;
2. `jormungandr --full-version` output;
[nix-shell:~/work/iohk/jormungandr-nix]$ jcli --full-version
jcli 0.5.5 (, release, linux [x86_64]) - [rustc 1.37.0 (eae3437df 2019-08-13)]

[nix-shell:~/work/iohk/jormungandr-nix]$ jormungandr --full-version
jormungandr 0.5.5 (, release, linux [x86_64]) - [rustc 1.37.0 (eae3437df 2019-08-13)]
**To Reproduce**
Sync testnet
run query

**Expected behavior**

**Additional context**

CodeSandwich:
I was unable to reproduce this bug. Based on the code I suspect that the problem manifests whenever queried block is known, but it's not in tip's chain. Returning error is correct behavior in such situation (to avoid deciding on branches) except that it should be just 400, no need for 500.

PR #894, opened 2019-10-04
Return HTTP 400 from next_id when block is not in tip chain

CodeSandwich:
Fixes https://github.com/input-output-hk/jormungandr/issues/888

### DIFF
--- a/doc/openapi.yaml
+++ b/doc/openapi.yaml
@@ -94,7 +94,9 @@ paths:
                 type: string
                 format: binary
         400:
-          description: Block with given ID was not found
+          description: Block ID malformed
+        404:
+          description: Block with given ID was not found in chain of the tip
   /api/v0/fragment/logs:
     get:
       description: Gets logs from node message pool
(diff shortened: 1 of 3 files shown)


jormungandr/src/rest/v0/handlers.rs
```diff
@@ -177,7 +177,7 @@ pub fn get_block_next_id(
                     .stream_from_to(block_id, tip.hash())
                     .then(|res| match res {
                         Ok(Some(stream)) => Ok(stream.map_err(ErrorInternalServerError)),
-                        Ok(None) => Err(ErrorInternalServerError("Tip block is parent of a block")),
+                        Ok(None) => Err(ErrorBadRequest("Block is not in chain of tip")),
```
mzabaluev: I'd use "the tip" (I know, articles, difficult for us Slavic speakers).

doc/openapi.yaml
```diff
@@ -94,7 +94,7 @@ paths:
                 type: string
                 format: binary
         400:
```
mzabaluev: This needs to be changed to 404 too, I think.

Issue #785, opened 2019-09-17
Trusted peers should accept DNS name


disassembler:
**Is your feature request related to a problem/context ? Please describe if applicable.**
We define the rendezvous nodes in a DNS round robin this way we can change/scale the IP's in the DNS round robin 
**Describe the solution you'd like**
Allow DNS names in trusted peers.

**Additional context**

mzabaluev:
It's possible to encode DNS in multiaddr, so at the very least we need to extract the DNS components and make the HTTP service URL out of them, relying on the resolver in H2 (which can further be customized if necessary).

NicolasDP:
this could be done easily in the `CLI` by adding support to enter multiaddr but with a [FQDN](https://en.wikipedia.org/wiki/Fully_qualified_domain_name) instead of the associated `ip`.

the multiaddr addr format first entry needs to be enforced:

i.e.:

- `/ip4/example.com/tcp/8080` still need to only resolve into `ip4` (`A`) : `/ip4/192.168.0.1/tcp/8080`
- same thing for `ip6` (`AAAA` only)

Also if a query returns multiple entries, we need to resolve into that multiple entries:

```
sh$ dig iohk.io
iohk.io.		300	IN	A	104.24.29.11
iohk.io.		300	IN	A	104.24.28.11
```

so if one enters in the trusted peer list `/ip4/iohk.io/tcp/9999` we expect this to resolve into a list of the following: `/ip4/104.24.29.11/tcp/9999` and `/ip4/104.24.28.11/tcp/9999`

NicolasDP:
one could use the standard library value:

https://doc.rust-lang.org/stable/std/net/trait.ToSocketAddrs.html

onicrom:
I'm not a huge fan of this.  It requires more trust in at least 1 or 2 third parties (owner of the TLD and owner of the nameservers hosting the domain) -- it also assumes those 2 entities are not attacked.  Not to mention increasing complexity of the node.

mark-stopka:
@onicrom, there is trust established in DNS thru DNSSEC, in a near future we can also expect new DNS services built on top of a distributed ledgers... 

Also, support for DNS FQDN resolution does not force users to use it...

onicrom:
DNSSEC adoption has been SUPER slow: https://stats.labs.apnic.net/dnssec
Currently this list of TLDs do NOT support DNSSSEC:
[...]

PR #897, opened 2019-10-06
Update network.md and Jormungandr User Guide

PacificGreen:
Propose changing https://input-output-hk.github.io/jormungandr/introduction.html and also testing the validity of the Ports ASAP.  I still believe they should be different with public on 3000 with the rest of the gossip peers and anything internal should be on a different port like 3100.  This would in effect make the public gossip server like a dns server with the actual jormangandr node like apache behind the firewall but all my testing has lead me to understand that I can only sync and make blocks if they are the same.  There is a huge amount of confusion surrounding this on the telegram group with no concrete answer.  There are people putting 8080 and even snmp ports in.

### DIFF
--- a/doc/configuration/network.md
+++ b/doc/configuration/network.md
@@ -20,13 +20,10 @@ p2p:
 
 - `trusted_peers`: (optional) the list of nodes' [multiaddr][multiaddr] to connect to in order to
     bootstrap the p2p topology (and bootstrap our local blockchain);
-- `public_address`: [multiaddr][multiaddr] the address to listen from and accept connection
-    from. This is the public address that will be distributed to other peers
-    of the network that may find interest into participating to the blockchain
-    dissemination with the node;
-- `listen_address`: (optional) [multiaddr][multiaddr] specifies the address the node
-    will listen to to receive p2p connection. Can be left empty and the node will listen
-    to whatever value was given to `public_address`.
+-  public_address: [multiaddr] The address to listen from and accept connections. This is the external public address that will be distributed to other gossip peers on the network.  Public address port must match listen address port.
+-  listen_address:  [multiaddr] Internal address that the node will listen to for 
+    receiving p2p connections.  This connection will typically be port forwarded from 
+    your router or cable modem. Listen address port must match public address port.
 - `topics_of_interest`: the different topics we are interested to hear about:
     - `messages`: notify other peers this node is interested about Transactions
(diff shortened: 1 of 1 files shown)


doc/configuration/network.md
```diff
@@ -20,13 +20,10 @@ p2p:
...
 - `trusted_peers`: (optional) the list of nodes' [multiaddr][multiaddr] to connect to in order to
     bootstrap the p2p topology (and bootstrap our local blockchain);
-- `public_address`: [multiaddr][multiaddr] the address to listen from and accept connection
-    from. This is the public address that will be distributed to other peers
-    of the network that may find interest into participating to the blockchain
-    dissemination with the node;
-- `listen_address`: (optional) [multiaddr][multiaddr] specifies the address the node
-    will listen to to receive p2p connection. Can be left empty and the node will listen
-    to whatever value was given to `public_address`.
+-  public_address: [multiaddr] The address to listen from and accept connections. This is the external public address that will be distributed to other gossip peers on the network.  Public address port must match listen address port.
```
mmahut: Why would public port need to match the listening port?

doc/configuration/network.md
```diff
@@ -20,13 +20,10 @@ p2p:
...
+-  listen_address:  [multiaddr] Internal address that the node will listen to for 
+    receiving p2p connections.  This connection will typically be port forwarded from
```
mmahut: I would say you are "typically" running an server with a public IP address, not port forwarding.

doc/configuration/network.md
NicolasDP: I don't see why would that be needed.

mzabaluev:
Without looking into every discussion around this, I think there is much confusion among users trying to run nodes under NAT in a variety of home network setups. This generally requires a stable port forwarding with the public IP address assignment that can be put into `public_address`. The local IP address (or simply `0.0.0.0`/`::0`) and the local mapped port is then put into `listen_address`. If you don't have a port mapping and the public address is bindable on the local host, you don't need `listen_address`.

The port number is not important to jormungandr or the P2P protocol, but various network gateways may have policies restricting access to certain well known ports used by other protocols, or conversely, blocking access to any ports by default.

Support for running a node in a NAT configuration should definitely be improved by, among other things, supporting DNS public addresses (#785) which can be backed by DynDNS, also possibly UPnP IGD and similar hole-punching technologies that are sadly still prevalent.

PR #901, opened 2019-10-07
Handle listening error in run_listen_socket

mzabaluev:
Remove the unimplemented panic and return an error for the `Future`, resulting in the network thread to gracefully terminate.


NicolasDP:
the network thread terminates, but what of the node?

mzabaluev:
This only eliminates the panic.
We should have some common infrastructure that catches when a task terminates and exits the process.

PR #910, opened 2019-10-08
JCLI: transaction info - remove args positionality

rinor:
Convert *positional* arguments to normal ones.
Why? Readability and practicity.
Ex: If we want to get only the **balance** value without input/output lines:
```
jcli transaction info 
  --staging tx.staging
  --fee-certificate 10000
  --fee-coefficient 50
  --fee-constant 1000
  --format {balance}
  "output_file.txt" // forced usage if we need to use the following params
  false
  false
  false
  ""
  ""
  ""

// need to read output_file.txt for result
```

VS:

```
jcli transaction info
  --staging tx.staging
  --fee-certificate 10000
  --fee-coefficient 50
  --fee-constant 1000
  --format {balance}
  --format-account-input ""
  --format-utxo-input ""
  --format-output ""
```

**labels**: jcli, enhancement, breaking-change

### DIFF
--- a/jcli/src/jcli_app/transaction/info.rs
+++ b/jcli/src/jcli_app/transaction/info.rs
@@ -18,6 +18,7 @@ pub struct Info {
     pub fee: common::CommonFees,
 
     /// write the info in the given file or print it to the standard output
+    #[structopt(long = "output")]
     pub output: Option<PathBuf>,
 
     /// formatting for the output to displays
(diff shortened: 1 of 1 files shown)


NicolasDP:
THanks, I am a bit concerned about the breaking change. But it is one of the small breaking change that can be done without breaking everything up.

PR #914, opened 2019-10-09
[Tests] Extract common folder from test configuration

dkijania:
For future development there is a must to extract common folder from test cfg and use it as dev module

### DIFF
--- a/jormungandr-integration-tests/src/common/jcli_wrapper/certificate/wrapper.rs
+++ b/jormungandr-integration-tests/src/common/jcli_wrapper/certificate/wrapper.rs
@@ -7,6 +7,9 @@ use crate::common::{
 use chain_impl_mockchain::{
     certificate::PoolId, testing::builders::cert_builder::build_stake_pool_retirement_cert,
 };
+
+extern crate jcli;
+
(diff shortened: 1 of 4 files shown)


CodeSandwich:
For me this PR in unreviewable. Almost 5000 added lines over 18 files is way more than I can digest in a single hit. I can be the one to click "merge", but I'm not able to actually review the changes. This PR is clearly much more than 1 added feature, so it can and should be split.

dkijania:
Ok, I'll  try to split PR into smaller pieces

dkijania:
I think you can take a look right know. I've added a little too many dependencies that are not needed at this moment, but once I prepare another PR, just after this they will be used

jormungandr-integration-tests/src/common/jcli_wrapper/certificate/wrapper.rs
```diff
@@ -10,6 +10,9 @@ use chain_impl_mockchain::{
     certificate::PoolId, key::EitherEd25519SecretKey,
     testing::builders::cert_builder::build_stake_pool_retirement_cert,
 };
+
+extern crate jcli;
```
CodeSandwich: This is useless I think

PR #916, opened 2019-10-09
Explorer scenario test

enzoc4:
# Summary

Experimental test case:

- Add a test case that creates a transaction and asks the explorer for it
- Add a loop for running all the tests

Things to improve
-  Probably a way to enable and disable the explorer in the `prepare_scenario!` macro, as it's always enabled this way.
- I'm not sure catching the panics is really the best way of handling failures.
- Reporting errors better?
- Timeouts, maybe
- Ergonomics for registering tests, instead of just adding them all in the `main` function?
- Introduce some kind of way for making graphql queries easily.
- Query more fields of the transaction


enzoc4:
@dkijania I think we should close this PR, right?

dkijania:
oh, yes. forgot do it

Issue #825, opened 2019-09-24
The user experience is low after using the single command to start the node


dorin100:
**Describe the bug**
When using the provided single command to start the node, the node seems to start but we don't have any way of interacting with it.

**Mandatory Information**
``
jormungandr 0.5.0 (HEAD-98922c77, release, windows [x86_64]) - [rustc 1.36.0 (a53f9df32 2019-07-03)]
jcli 0.5.0 (HEAD-98922c77, release, windows [x86_64]) - [rustc 1.36.0 (a53f9df32 2019-07-03)]
``

**To Reproduce**
1. start the node using `jormungandr --trusted-peer "valid_trusted_peer" --genesis-block-hash 11e340f9c20a4bcdc19103d9794413be81c9a713374997b574e9f9d66419a2b2`

**Expected behavior**
All the values from the node config should have default values or add command parameters for most of them. At this moment I don't have (or don't have) a rest port to communicate with the node so I cannot have any clue about how/if the node is working. 

**Additional context**
- node logs:
```
D:\iohk\otherProjects\jormungandr\scripts\connect-to-cluster>jormungandr --trusted-peer "/ip4/3.123.xx.xx/tcp/3000" --genesis-block-hash 11e340f9c20a4bcdc19103d9794413be81c9a713374997b574e9f9d66419a2b2
Sep 24 19:21:10.454 INFO Starting jormungandr 0.5.0 (HEAD-98922c77, release, windows [x86_64]) - [rustc 1.36.0 (a53f9df32 2019-07-03)], task: init
Sep 24 19:21:10.468 WARN Node started without path to the stored secret keys (not a stake pool or a BFT leader), task: init                                                                                              
Sep 24 19:21:10.471 INFO storing blockchain in memory, task: init                                                                                                                                                        Sep 24 19:21:10.474 INFO fetching block 11e340f9c20a4bcdc19103d9794413be81c9a713374997b574e9f9d66419a2b2 from 3.123.xx.xx:3000, task: init                                                                             
Sep 24 19:21:10.540 INFO connecting to bootstrap peer 3.123.xx.xx:3000, peer: /ip4/3.123.xx.xx/tcp/3000, task: bootstrap
Sep 24 19:21:10.803 INFO initial bootstrap completed, peer: /ip4/3.123.xx.xx/tcp/3000, task: bootstrap
[...]

PR #925, opened 2019-10-11
Add --rest-listen-address Jormungandr CLI arg

CodeSandwich:
Fixes https://github.com/input-output-hk/jormungandr/issues/825

### DIFF
--- a/jormungandr/src/settings/command_arguments.rs
+++ b/jormungandr/src/settings/command_arguments.rs
@@ -1,5 +1,6 @@
 use crate::settings::LOG_FILTER_LEVEL_POSSIBLE_VALUES;
 use slog::FilterLevel;
+use std::net::SocketAddr;
 use std::path::PathBuf;
 use structopt::StructOpt;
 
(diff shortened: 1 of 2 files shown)


jormungandr/src/settings/command_arguments.rs
```diff
@@ -45,6 +46,14 @@ pub struct StartArguments {
     pub explorer_enabled: bool,
 }
 
+#[derive(StructOpt, Debug)]
+pub struct RestArguments {
+    /// REST API listening address.
+    /// If not configured anywhere, defaults to REST API being disabled
+    #[structopt(long = "rest-listen-address")]
+    pub listen: Option<SocketAddr>,
```
mzabaluev: Nitpick: could be `--rest-listen` for brevity and commonality with config; also we have `*_address` settings that tend to be multiaddr and this is not (neither it should be).

CodeSandwich: This was name proposed by @NicolasDP in the issue, but I agree with you, I'll fix it.

Issue #860, opened 2019-09-29
Node not receiving updates until...


onicrom:
**Describe the bug**
Creating a new node, setting the listen/public tcp port to 3100 will initially download the chain history but then fail to receive updates.  If you change the listen/public tcp port to any other valid tcp port (3111) and restart the node starts to receive updates.

**Mandatory Information**
jcli 0.5.2 (master-23a9fe1c, release, linux [x86_64]) - [rustc 1.36.0 (a53f9df32 2019-07-03)]\
jormungandr 0.5.2 (master-23a9fe1c, release, linux [x86_64]) - [rustc 1.36.0 (a53f9df32 2019-07-03)]

**To Reproduce**
- run node with config like this:
```
log:
  format: "plain"
  level: "info"
  output: "stderr"
p2p:
  listen_address: "/ip4/0.0.0.0/tcp/3100"
  public_address: "/ip4/x.x.x.x/tcp/3100"
  topics_of_interest:
    blocks: "high"
    messages: "high"
  trusted_peers:
    - "/ip4/136.244.106.95/tcp/3000"
    - "/ip4/91.121.85.221/tcp/3100"
    - "/ip4/68.183.155.37/tcp/3100"
    - "/ip4/3.123.177.192/tcp/3000"
    - "/ip4/3.123.155.47/tcp/3000"
    - "/ip4/3.115.57.216/tcp/3000"
    - "/ip4/3.112.185.217/tcp/3000"
    - "/ip4/18.139.40.4/tcp/3000"
    - "/ip4/18.140.134.230/tcp/3000"
rest:
  listen: "127.0.0.1:3101"
storage: "/path/to/storage"
```

- stop node, change config to this:

```
log:
  format: "plain"
  level: "info"
  output: "stderr"
p2p:
  listen_address: "/ip4/0.0.0.0/tcp/3333"
  public_address: "/ip4/x.x.x.x/tcp/3333"
  topics_of_interest:
    blocks: "high"
    messages: "high"
  trusted_peers:
    - "/ip4/136.244.106.95/tcp/3000"
    - "/ip4/91.121.85.221/tcp/3100"
    - "/ip4/68.183.155.37/tcp/3100"
    - "/ip4/3.123.177.192/tcp/3000"
    - "/ip4/3.123.155.47/tcp/3000"
    - "/ip4/3.115.57.216/tcp/3000"
    - "/ip4/3.112.185.217/tcp/3000"
[...]

PR #926, opened 2019-10-11
Fix node that do not sync when no Public Address is set

NicolasDP:
will help with #860 

This address multiple items:

* put in place the initial tooling for node id authentication;
* put the required system for the non publicly reachable node to continue to receive propagation

This comes with a breaking change though...

Now the trusted peers will need to create cryptographic material to identify them selves for the other nodes (see `private_id` in the `config.yaml` for the node):

When a node wants to be publicly reachable and wants to advertise themselves as trusted peers they will have to give a public key along their public address:

to generate that key:

```
jcli key generate --type Ed25519
```

and add the output of that command in the node's config file:

```yaml
p2p:
    private_id: ed25519_sk1z30an6rtv26z8fyjdgqgcjtxt9fph7n7pw93g86ayrntx62vgz7sy4hfjs
    public_address: /ip4/x.y.z.t/tcp/8010
```

That node will them publicly announce the `public_address` and the associated public key of the `private_id`, to get the public key:

```
echo ed25519_sk1z30an6rtv26z8fyjdgqgcjtxt9fph7n7pw93g86ayrntx62vgz7sy4hfjs | jcli key to-public
```

And now to connect to that nodes user will need to either add the associated public key of the trusted peer:

```yaml
p2p:
    trusted_peers:
        - address /ip4/x.y.z.t/tcp/8010
          ed25519_pk1hs2208lr8w5lq9wmfm674d5zafaplr0xl7pml40nacg6jp2z0umsqzs3fr
```

or add it in the command line parameter:

```
jormungandr --trusted-peer /ip4/x.y.z.t/tcp/8010@ed25519_pk1hs2208lr8w5lq9wmfm674d5zafaplr0xl7pml40nacg6jp2z0umsqzs3fr --genesis-block-hash a0f92...
```

### DIFF
--- a/doc/configuration/introduction.md
+++ b/doc/configuration/introduction.md
@@ -12,9 +12,12 @@ log:
   format: json
 p2p:
   trusted_peers:
-    - "/ip4/104.24.28.11/tcp/8299"
-    - "/ip4/104.24.29.11/tcp/8299"
+    - address: "/ip4/104.24.28.11/tcp/8299"
+      id: ed25519_pk1tyz2drx5kvsp49fk0gunkl8ktlhjc976zj5erqwavnpfmrexxxnscxln75
+    - address: "/ip4/104.24.29.11/tcp/8299"
+      id: ed25519_pk13j4eata8e567xwdqp6wjeu8wa7dsut3kj0u3tgulrsmyvveq9qxqeqr3kc
   public_address: "/ip4/127.0.0.1/tcp/8080"
+  private_id: ed25519_sk1n649x7zrmt38q6dtqkvj769wru4vpkfnrppw6d83qd7jh3uhux7qwhg8q3
   topics_of_interest:
     messages: low
     blocks: normal
--- a/jormungandr/src/network/p2p/topology.rs
+++ b/jormungandr/src/network/p2p/topology.rs
@@ -84,6 +84,10 @@ impl Node {
 }
 
 impl NodeData {
+    pub fn poldercast_address(&self) -> &Option<poldercast::Address> {
+        self.0.address()
+    }
+
     pub fn has_valid_address(&self) -> bool {
         let addr = match self.address() {
             None => return false,
--- a/jormungandr/src/network/p2p/topology.rs
+++ b/jormungandr/src/network/p2p/topology.rs
@@ -29,13 +29,15 @@ impl From<bincode::Error> for Error {
     }
 }
 
-#[derive(Clone, Debug)]
 pub struct Node(poldercast::Node);
+
+#[derive(Clone, Debug)]
+pub struct NodeData(poldercast::NodeData);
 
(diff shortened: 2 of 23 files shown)


doc/configuration/introduction.md
```diff
@@ -12,9 +12,12 @@ log:
...
 p2p:
   trusted_peers:
-    - "/ip4/104.24.28.11/tcp/8299"
-    - "/ip4/104.24.29.11/tcp/8299"
+    - address: "/ip4/104.24.28.11/tcp/8299"
+      id: ed25519_pk1tyz2drx5kvsp49fk0gunkl8ktlhjc976zj5erqwavnpfmrexxxnscxln75
+    - address: "/ip4/104.24.29.11/tcp/8299"
+      id: ed25519_pk13j4eata8e567xwdqp6wjeu8wa7dsut3kj0u3tgulrsmyvveq9qxqeqr3kc
   public_address: "/ip4/127.0.0.1/tcp/8080"
+  private_id: ed25519_sk1n649x7zrmt38q6dtqkvj769wru4vpkfnrppw6d83qd7jh3uhux7qwhg8q3
```
mzabaluev: Is this the private key? Considered a low-grade secret so it's OK to put it into the config file?

jormungandr/src/network/p2p/topology.rs
```diff
@@ -29,13 +29,15 @@ impl From<bincode::Error> for Error {
     }
 }
 
-#[derive(Clone, Debug)]
 pub struct Node(poldercast::Node);
 
+#[derive(Clone, Debug)]
+pub struct NodeData(poldercast::NodeData);
```
mzabaluev: Losing track a bit as to why there is now `Node` and `NodeData` when the useful things are now done using the latter. Some doc comments would be nice.

NicolasDP: I understand, I am differentiating the `Node`, ourselves, from the `Data` that we have on other `Node` (or ourselves).

So now, locally everyone has a `Node` and a set of `NodeData`. The `Node` contains more info than before so adding an object to superseed the `NodeData` was necessary (i.e. we don't want to gossip private keys for example).

rinor:
:fire: simply beautiful :fire:

jormungandr/src/network/p2p/topology.rs
mzabaluev: I see. Maybe a name like `SelfNode` would help understand the difference.

hemants1:
Quick question: how does the new node want to join the network know the public ID of trusted peers?  Is the trusted peers details such as [address] and [id] published anywhere?

mzabaluev:
@hemants1 Yes, the expectation is that the list of trusted peers for a testnet is made available _somewhere_.

Issue #932, opened 2019-10-13
Trusted peers ids


aidik:
**Describe the bug**
After this commit (https://github.com/input-output-hk/jormungandr/commit/60526824d9b8b4555fc8e17764d7f2d441b544fd#diff-45a5dd2c4b220bfd700f85800472ccc9R16) address and id are needed for trusted_peers configuration. The two peers in the "example" are either only for demonstration or currently down. I'm unavare of a way how to obtain a node_id from other modes on the network. For example for this list https://testnet.iohkdev.io/cardano/shelley/get-started/configuring-your-network/

That leads to the following loop of errors:
```
Oct 13 22:20:39.073 INFO error connecting to peer, reason: Connection timed out (os error 110), peer_addr: 104.24.29.11:8299, task: network
Oct 13 22:20:39.073 INFO error connecting to peer, reason: Connection timed out (os error 110), peer_addr: 104.24.28.11:8299, task: network
Oct 13 22:20:39.073 INFO client P2P connection closed, node_id: 8cab9eafa7cd35e339a00e9d2cf0eeef9b0e2e3693f915a39f1c36463320280c, peer_addr: 104.24.29.11:8299, task: network
Oct 13 22:20:39.074 INFO client P2P connection closed, node_id: 5904a68cd4b3201a95367a393b7cf65fef2c17da14a99181dd64c29d8f2631a7, peer_addr: 104.24.28.11:8299, task: network
```

**Mandatory Information**
1. `jcli 0.5.6+lock (master-d2f1170b+, release, linux [x86_64]) - [rustc 1.38.0 (625451e37 2019-09-23)]`
2. `jormungandr 0.5.6+lock (master-d2f1170b+, release, linux [x86_64]) - [rustc 1.38.0 (625451e37 2019-09-23)]`

**To Reproduce**
1. add 
```
  trusted_peers:
    - address: "/ip4/104.24.28.11/tcp/8299"
      id: ed25519_pk1tyz2drx5kvsp49fk0gunkl8ktlhjc976zj5erqwavnpfmrexxxnscxln75
    - address: "/ip4/104.24.29.11/tcp/8299"
      id: ed25519_pk13j4eata8e567xwdqp6wjeu8wa7dsut3kj0u3tgulrsmyvveq9qxqeqr3kc
```
to your config file.
2. run jormungandr
3. observe above mentioned error loop
[...]

PR #934, opened 2019-10-13
Don't advise build from master but from latest release tag.

aidik:
In the context of the closing response in issue #932 it is no longer advised to build from the master branch, but to use the latest released tag for builds.


onicrom:
*sad cowboy face*

NicolasDP:
Don’t be sad. Seasoned cowboys can still ignore this and use master. Just more risky. But there is land to discover and gold to find in the unexplored far west.

PR #938, opened 2019-10-14
add PeerStats's creation time

NicolasDP:


doc/openapi.yaml
```diff
@@ -383,6 +383,10 @@ paths:
                       description: Hex-encoded node ID
                       type: string
                       pattern: '[0-9a-fA-F]+'
+                    establishedAt:
```
CodeSandwich: This field required, it should be on `requiredFields` list

PR #958, opened 2019-10-15
Peers with IDs

aidik:
Are those peers trusted? I have no idea, I compile a list of configs shared by various people on Telegram Cardano Shelley Testnet & StakePool Workgroup


psychomb:
Hi @aidik, 
My feelings is that it is probably better to let IOHK IDs create the node/IDs pairs for official docs. For example, I find my node in your list but its associated ID is not correct anymore since yesterday. 
Likewise, people might change their pk IDs whenever they feel like it and this could be quite chaotic to follow.

NicolasDP:
not a good _ID_ 🤣

NicolasDP:
Thanks guys. I appreciate the intent, but this is not a good idea to share things like that. Especially when the value are not necessarily consistent or true.

the `ID`s are meant to be floating, except for a few trusted peers. So better not add random `ID`s for now.

however when we reached stability and then mainnet we will be able to consider such change.

PR #960, opened 2019-10-15
Add block search by stake pool

enzoc4:
# Summary

- Add field to query the Stake Pool or BFT leader that produced a block. For example:

```graphql
{
  blockByChainLength(length: "1") {
    id
    leader {
      __typename
      ... on Pool {
        id
      }
    }
  }
}
```

The caveat here is that the bft indexing is not implemented, because I couldn't find a way to get the leader id from the block, so at this moment it won't work in bft mode.

Also, not sure if `leader` is the best name for the field, as leader is a concept more related to slots than blocks, other options could be `creator`, `producer`, `created_by`.

- Add query for all the blocks (paginated) produced by a certain stake pool

```graphql
{
  stakePool(id:"c52e336e3a23548c2c0d3b5d97849303f52931f3b6a8b85d64fa180307e6995f") {
    blocks(first: 10) {
      edges {
        node {
          chainLength
          id
        }
      }
    }
  }
}
```

### DIFF
--- a/jormungandr/src/explorer/graphql/connections.rs
+++ b/jormungandr/src/explorer/graphql/connections.rs
@@ -1,8 +1,7 @@
 use super::error::ErrorKind;
 use super::scalars::BlockCount;
-use super::{Block, Context, ExplorerDB};
-use blockcfg;
-use futures::Future;
+use super::{Block, Context};
+use blockcfg::{self, HeaderHash};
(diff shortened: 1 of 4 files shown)


enzoc4:
@NicolasDP did you find time to look at this? Particularly on the naming, but also about the bft part.

Issue #923, opened 2019-10-11
Suspected file descriptor leak (ERRO Error while accepting connection on 0.0.0.0:3000: Os)


bjarnekvae:
**Describe the bug**
First error in logs are "ERRO Error while accepting connection on 0.0.0.0:3000: Os { code: 24, kind: Other, message: "Too many open files" }, task: network", after a while its starts spamming "ERRO cannot send PullHeaders request to network: send failed because receiver is gone, task: block" and node stops receving blocks

**Mandatory Information**
1. jcli 0.5.6 (HEAD-7ab929e+, release, linux [x86_64]) - [rustc 1.38.0 (625451e37 2019-09-23)]
2. jormungandr 0.5.6 (HEAD-7ab929e+, release, linux [x86_64]) - [rustc 1.38.0 (625451e37 2019-09-23)]

**To Reproduce**
Dont know how to reproduce, happens after the node has run a few hours

**Expected behavior**
No error

**Additional context**
See appendix for output logs and strace logs. I have four examples as I happens every time I try to run my node.

[errPullHeader.zip](https://github.com/input-output-hk/jormungandr/files/3717092/errPullHeader.zip)

mzabaluev:
Thank you for the captures, will investigate.

mzabaluev:
Massively duplicate connections are still present; working on a better fix.

bjarnekvae:
I'm running 0.6.0 with strace, will serve you fresh logs when(/if?) it crashes :)

bjarnekvae:
Error just happened on my other server, running a other setup with a old-style mechanical HDD, dont know if it is relevant. Running v0.6.1 

Here are the logs:
[err_24.zip](https://github.com/input-output-hk/jormungandr/files/3732813/err_24.zip)

bjarnekvae:
Just happened on my other server, so its not hardware related as this one uses a SSD. Still on v0.6.1

Logs:
[err_24_2.zip](https://github.com/input-output-hk/jormungandr/files/3734347/err_24_2.zip)

vincenthz:
@bjarnekvae it's not storage related, it is a network subsystem issue as indicated in the log : `task: network" `

mzabaluev:
One probable cause has been fixed in #981 (and fixes of the fix that have followed).
Please retry with version 0.6.5 or later.

bjarnekvae:
running 0.6.5 with strace now

bjarnekvae:
[...]

PR #981, opened 2019-10-18
Rework pending client connections

mzabaluev:
Add state to peer map entry already when the client begins connecting. These bits of state keep track of the progress and suspend propagation until the connection future is resolved, or alternatively, when an incoming subscription on the server results in canceling the client-side connection.

This removes another way of client connection attempts to pile up, and should cope better with "flare" conditions when peers attempt connecting each other at the same time.

Fixes (one cause of) #923


jormungandr/src/network/client/connect.rs
```diff
@@ -0,0 +1,414 @@
...
+    T: BlockService + FragmentService + GossipService,
+{
+    fn new() -> Self {
+        SubscriptionRequests {
+            blocks: None,
+            fragments: None,
+            gossip: None,
+        }
+    }
+}
```
NicolasDP: seems like this should be an impl of `Default` instead

jormungandr/src/network/client/connect.rs
```diff
@@ -0,0 +1,414 @@
...
+{
+    fn new() -> Self {
+        InboundSubscriptionStaging {
+            node_id: None,
+            block_events: None,
+            fragments: None,
+            gossip: None,
+        }
+    }
+
```
NicolasDP: same here... this should be an impl of `Default`

jormungandr/src/network/client/connect.rs
```diff
@@ -0,0 +1,414 @@
...
+    fn try_complete(&mut self) -> Option<InboundSubscriptions<T>> {
+        match (&self.block_events, &self.fragments, &self.gossip) {
+            (&Some(_), &Some(_), &Some(_)) => Some(InboundSubscriptions {
```
NicolasDP: what about using a more classic obvious:

```
if self.blocks_events.is_some() && self.fragments.is_some() && self.gossip.is_some() {
    Some(/* all that it `take` */)
} else {
    None
}```

PR #983, opened 2019-10-18
[Tests] Stability fix for test net e2e test

dkijania:


jormungandr-integration-tests/src/common/jormungandr/starter.rs
```diff
@@ -4,14 +4,21 @@ use self::custom_error::custom_error;
...
 use crate::common::process_assert;
 use crate::common::process_utils::{self, output_extensions::ProcessOutput, ProcessError};
-use std::process::{Child, Command, Output};
-
+use std::{
+    path::PathBuf,
+    process::{Child, Command, Output},
+    time::{Duration, Instant},
+};
 custom_error! {pub StartupError
```
mzabaluev: I advise converting uses of `custom_error` to [thiserror](https://crates.io/crates/thiserror) when you have an opportunity. Just as easy to use, but without the ugly custom syntax and you can explicitly control which `From` impls get generated.

jormungandr-integration-tests/src/common/jormungandr/starter.rs
```diff
@@ -39,7 +46,11 @@ fn try_to_start_jormungandr_node(
 
     match proces_start_result {
         Ok(_) => return Ok(process),
```
mzabaluev: Drive-by style remark: if `proces_start_result` (typo: should be `process_*`) returns `Ok(())`, it's better to match it explicitly, so that when you decide it should return something useful, it won't be silently ignored here.

jormungandr-integration-tests/src/common/jormungandr/starter.rs
```diff
@@ -185,3 +196,63 @@ pub fn assert_start_jormungandr_node_as_passive_fail(
...
+        &config,
+        timeout_value,
+        |logger: &JormungandrLogger| {
+            logger.message_logged_multiple_times("initial bootstrap completed", 2)
+        },
+        |logger: &JormungandrLogger| logger.contains_error(),
+    )
+}
+
+fn start_jormungandr_node_as_passive_with_timeout_and_log_checks<F: 'static, G: 'static>(
```
mzabaluev: Looking at the proliferation of `start_jormungandr_node_*` functions, I wonder if this API could be made more composable with a builder-style API dialing in the features you want with methods changing the configuration dynamically:

```rust
impl Starter {
    pub fn new() -> Self {
        // init with default values ...
    }

    pub fn timeout(&mut self, timeout: Duration) -> &mut Self {
        self.timeout = timeout;
        self
    }

    pub fn passive(&mut self) -> &mut Self {
        self.passive = true;
        self
    }

    pub fn start(self) -> Result<JormungandrProcess, StartupError> {
        // ...
    }
}
```

mzabaluev: It could also be `fn start(&self)` so that you could reuse all or most of the configuration to launch multiple instances.

dkijania: Geat idea. i'll implement it in separate pr.

jormungandr-integration-tests/src/common/jormungandr/starter.rs
dkijania: I'll implement it in separate PR

PR #994, opened 2019-10-19
Fix a hangup in network client polling

mzabaluev:
Also added a logging adapter for streams.


rinor:
I couln't resist, it's just beautiful. 3 active stake pools + 1 passive/explorer in perfect sync (private testnet)
perfect_sync

Issue #995, opened 2019-10-19
Node not getting blocks until after a few hours


NotToFly:
After bootstarp stuck on connecting to peer. no updates no errors. Rust and Nix node the same.

**Mandatory Information**
1. jcli 0.6.3 (master-bb925b7a+, release, linux [x86_64]) - [rustc 1.38.0 (625451e37 2019-09-23)
2. jormungandr 0.6.3 (master-bb925b7a+, release, linux [x86_64]) - [rustc 1.38.0 (625451e37 2019-09-23)

**To Reproduce**
1. Start Jormungandr with
jormungandr --genesis-block-hash adbdd5ede31637f6c9bad5c271eec0bc3d0cb9efb86a5b913bb55cba549d0770 --config node-config.yaml --secret node-secret.yaml --log-level=info

Not the last release related. was like that for a while.

![19-Oct-19 12-05-07 PM](https://user-images.githubusercontent.com/53143161/67144066-7848c180-f26a-11e9-9c22-c2d84148a3ad.png)

mzabaluev:
Please update to 0.6.5 and see if the problem persists.

jbax:
I'm having the same issue here with 0.6.5, right now `uptime: 3183` and no blocks yet. Usually it starts syncing after 4 to 6 hours - it has been like that since version 0.5.4 which is the version I began testing with.

mzabaluev:
Setting log level to `debug` would enable dumping of protocol messages received by the node (among gobs of other information). Please check if the node receives any kind of block events, provided that it's even connected to any other nodes.

jbax:
Are there any messages in specific from the log that I should look for?

NotToFly:
Updated to 0.6.5. 
Uptime 7200 
Debug screen below

![19-Oct-19 6-33-37 PM](https://user-images.githubusercontent.com/53143161/67149122-49007780-f29f-11e9-922b-35120bb0fe0a.png)

jbax:
My uptime is at 31645 right now. The log shows what you are looking for:

```
Oct 20 08:43:53.931 DEBG scheduling Read for: 0, scope: global
Oct 20 08:43:53.931 DEBG send; frame=Data { stream_id: StreamId(7) }, scope: global
Oct 20 08:43:53.931 DEBG send; frame=Data { stream_id: StreamId(7) }, scope: global
Oct 20 08:43:53.931 DEBG scheduling Read for: 0, scope: global
[...]

PR #1010, opened 2019-10-22
network: Rewrite the client state machine

### LINKED ISSUE #923: Suspected file descriptor leak (ERRO Error while accepting connection on 0.0.0.0:3000: Os)

mzabaluev:
More disciplined polling loop organization in the form of `ProcessingOutcome` and `Progress`.
Don't spin off separate tasks for processing inbound fragment and gossip streams, as this may cause the background HTTP client to linger after the `Client`'s future has resolved.
Expose the sinks and the gossip processing state in `mod subscription` to enable code reuse in the client.

Might fix #923 and #995.

### DIFF
--- a/jormungandr/src/network/subscription.rs
+++ b/jormungandr/src/network/subscription.rs
@@ -1,6 +1,6 @@
 use super::{
     p2p::topology::{NodeData, NodeId},
-    GlobalState, GlobalStateR,
+    GlobalStateR,
 };
 use crate::{
     blockcfg::{Fragment, Header},
@@ -9,7 +9,6 @@ use crate::{
     utils::async_msg::MessageBox,
 };
 use futures::prelude::*;
-use futures::sink;
 use jormungandr_lib::interfaces::FragmentOrigin;
 use network_core::error as core_error;
 use network_core::gossip::{Gossip, Node as _};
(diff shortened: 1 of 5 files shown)


jormungandr/src/network/subscription.rs
```diff
@@ -20,136 +19,74 @@ pub fn process_block_announcements<S>(
...
-                core_error::Code::Aborted,
-                "the node stopped processing blocks",
-            )
-        })
-    }
+    let stream = inbound.map_err(move |e| {
+        debug!(
+            stream_err_logger,
+            "block subscription stream failure";
+            "error" => ?e,
```
NicolasDP:
```suggestion
            "reason" => ?e,
```

just so we remain consistent?

mzabaluev: That's inconsistent by me :grin: I've been using `"error" => ?e` in debug logs and `"reason" => %e` in info logs which should produce more user-friendly messages.

I think it's time to deliberate and lay down the law on this.

Issue #997, opened 2019-10-19
Switch from slog to tracing


mzabaluev:
[`tracing`](https://github.com/tokio-rs/tracing) is a logging framework designed for async applications.
In the long term, we might want to switch to it to avoid issues with mixing asynchronous code with synchronous logging calls.

We use `slog-async`, but it is still problematic: the slog `Drain` API can't provide back pressure and polling like a truly asynchronous sink and therefore is forced to drop messages on the floor to avoid blocking.

The developers of the tower stack also plan to switch to `tracing` for scoped logging in their libraries (currently they use `log` which is even more appalling), so the integration of logging will be natural when they do.

eugene-babichenko:
@mzabaluev shall we close it?

PR #1014, opened 2019-10-22
Logging improvements in the network task

mzabaluev:
Various changes to improve what gets logged, at what level, and how slog keywords are used.

### DIFF
--- a/jormungandr/src/network/p2p/comm.rs
+++ b/jormungandr/src/network/p2p/comm.rs
@@ -441,11 +441,10 @@ impl Peers {
                         Err(e) => {
                             debug!(
                                 self.logger,
-                                "propagation to peer {} failed: {}",
-                                id,
-                                e.kind()
+                                "propagation to peer failed, unsubscribing peer";
+                                "node_id" => %id,
+                                "reason" => %e.kind()
                             );
-                            debug!(self.logger, "unsubscribing peer {}", id);
                             entry.remove();
                             true
                         }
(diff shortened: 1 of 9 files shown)


jormungandr/src/network/p2p/comm.rs
```diff
@@ -441,11 +441,10 @@ impl Peers {
                         Err(e) => {
                             debug!(
                                 self.logger,
-                                "propagation to peer {} failed: {}",
-                                id,
-                                e.kind()
+                                "propagation to peer failed, unsubscribing peer";
+                                "node_id" => %id,
+                                "reason" => %e.kind()
```
NicolasDP: on the same spirit it would be good to leverage the `std::error::Error` trait and have a way to display the `source` of the error too if any (recursively). This would be useful for the whole node actually.

mzabaluev: slog does not have error chain utilities out of the box, and it would probably be messy to marshal the chain into a single message (splitting across messages is even worse). Dealing with dynamically created metadata in slog is cumbersome, it's really optimized for instantiating everything in the log site with macros.

CodeSandwich: This makes duplicate keys in single message almost inevitable. Slog is really dumb when it comes to that, e.g. JSON output doesn't deduplicate and generates malformed output, human readable, but impossible to machine-process.

mzabaluev: Our long term solution should be `tracing` (#997).

PR #1015, opened 2019-10-22
[Tests] Grpc mock 

dkijania:
Added implementation for server and client mock. 
Added cli for server and client.
Proto file generation is now on demand (comment build.rs action). T


jormungandr-integration-tests/mock/proto/node.proto
```diff
@@ -0,0 +1,144 @@
+syntax = "proto3";
```
mzabaluev: Would it be possible to use the source of truth .proto file from its chain-deps location, by relative path? The copy creates a maintenance burden.

jormungandr-integration-tests/src/mock/client.rs
```diff
@@ -0,0 +1,233 @@
...
+            client: client,
+            host: host.to_owned(),
+            port: port,
+        }
+    }
+
+    pub fn get_blocks(&self, blocks_id: &Vec<Hash>) -> Vec<ChainBlock> {
+        self.get_blocks_stream(blocks_id)
+            .into_future()
+            .wait_drop_metadata()
```
mzabaluev: Now I remember again why I rejected that gRPC stack, it's so fugly.
There is also another one that binds to the C library, check if that might be better.

On the other hand, it could be useful for dissimilar interop testing.

jormungandr-integration-tests/src/mock/client.rs
```diff
@@ -0,0 +1,233 @@
...
+        self.pull_headers_stream(from, to)
+            .into_future()
+            .wait_drop_metadata()
+            .unwrap()
+            .into_iter()
+            .map(|x| read_into(x.get_content()))
+            .collect()
+    }
+
+    pub fn pull_headers_get_err(&self, from: Option<Hash>, to: Option<Hash>) -> grpc::Error {
```
mzabaluev: `pull_headers` could just return `Result`, so the tests can call `expect` and `expect_err` on it and you don't need two methods.

jormungandr-integration-tests/src/mock/client.rs
```diff
@@ -0,0 +1,233 @@
...
+
+    pub fn get_tip(&self) -> ChainHeader {
+        let resp = self
+            .client
+            .tip(grpc::RequestOptions::new(), TipRequest::new());
+        let (_, tip, _) = resp.wait().unwrap();
+        read_into(&tip.get_block_header())
+    }
+
+    pub fn upload_blocks_internal(
```
mzabaluev: A public internal method, tut-tut.

jormungandr-integration-tests/src/mock/client.rs
```diff
@@ -0,0 +1,233 @@
...
+extern crate chain_impl_mockchain;
+extern crate futures;
+extern crate futures_cpupool;
+extern crate grpc;
+extern crate hex;
+extern crate protobuf;
+
+use crate::mock::{
+    grpc::*,
+    proto::{node::*, node_grpc::*},
```
mzabaluev: So many globs, can't see what comes from where.

jormungandr-integration-tests/src/mock/client.rs
```diff
@@ -0,0 +1,233 @@
...
+    }
+
+    pub fn get_headers(&self, blocks_id: &Vec<Hash>) -> Vec<ChainHeader> {
+        self.get_headers_stream(blocks_id)
+            .into_future()
+            .wait_drop_metadata()
+            .unwrap()
+            .into_iter()
+            .map(|x| read_into(x.get_content()))
+            .collect()
```
mzabaluev: You could try to begin writing macros to save this much typing.

Also maybe, a generic wrapper that encapsulates the response:

```rust
pub struct StreamingResponse<T> {
    inner: grpc::StreamingResponse<T>,
}
```

The wrapper would then have accessor methods for getting the streamed data in a preferred way.

```rust
impl<T> StreamResponse<T> {
    pub fn to_vec(self) -> Vec<T> {
        self.into_iter()
            .map(|x| read_into(x.get_content()))
            .collect()
    }
}
```

jormungandr-integration-tests/src/bin/mock_server_app.rs
```diff
@@ -0,0 +1,46 @@
...
+        "Mock server started with genesis hash: {}, tip hash: {}, and protocol: {}, log_file:{:?} ",
+        genesis_hash,
+        tip,
+        protocol,
+        path_log_file.as_os_str()
+    );
+
+    loop {
+        thread::park();
+    }
```
mzabaluev: Is this the approved way to run that server? Can't it run on a Tokio runtime?

jormungandr-integration-tests/src/bin/mock_server_app.rs
```diff
@@ -0,0 +1,46 @@
+extern crate base64;
+extern crate futures;
+extern crate futures_cpupool;
```
mzabaluev: Why the imports? I don't see it used anywhere.

jormungandr-integration-tests/src/mock/client.rs
dkijania: I used macros

jormungandr-integration-tests/src/bin/mock_server_app.rs
dkijania: The purpose was just to fire it up in simplest way. There is no tokio runtime crate in jormungandr-integration-tests,  At the time of writing it I couldn't find any reason why it should be more than that

jormungandr-integration-tests/build.rs
```diff
@@ -5,17 +5,16 @@ use std::env;
 
 fn main() {
     // generate grpc mock
-    /*
-    protoc_rust::run(protoc_rust::Args {
+
+    /*protoc_rust::run(protoc_rust::Args {
```
mzabaluev: Is there a problem with running protoc automatically so that this needs to be commented out?

If you do, also add the below line to this function to avoid excessive rebuilds:

```rust
    println!("cargo:rerun-if-changed=../../chain-libs/network-grpc/proto/node.proto");
```

jormungandr-integration-tests/src/mock/client.rs
```diff
@@ -65,17 +76,7 @@ impl JormungandrClient {
...
-            .into_future()
-            .wait_drop_metadata()
-            .unwrap()
-            .into_iter()
-            .map(|x| read_into(x.get_content()))
-            .collect()
-    }
-
-    fn get_headers_stream(&self, blocks_id: &Vec<Hash>) -> grpc::StreamingResponse<Header> {
+    pub fn get_headers(&self, blocks_id: &Vec<Hash>) -> grpc::StreamingResponse<Header> {
```
mzabaluev: Much more concise now, thanks.

jormungandr-integration-tests/src/mock/mod.rs
```diff
@@ -5,17 +5,17 @@ extern crate grpc;
 extern crate hex;
 extern crate protobuf;
 
+#[macro_use]
 pub mod client;
 pub mod convert;
 pub mod proto;
 pub mod server;
 
+use chain_core::mempack::{ReadBuf, Readable};
```
mzabaluev: Commingling `use` of external things and `pub use` of reexports is confusing. This line is better where it was before.

jormungandr-integration-tests/src/mock/proto/node.rs
```diff
@@ -2678,7 +2678,7 @@ static file_descriptor_proto_data: &'static [u8] = b"\
     \x08announce\x18\x01\x20\x01(\x0b2\x17.iohk.chain.node.HeaderH\0R\x08ann\
     ounce\x125\n\x07solicit\x18\x02\x20\x01(\x0b2\x19.iohk.chain.node.BlockI\
     dsH\0R\x07solicit\x12?\n\x07missing\x18\x03\x20\x01(\x0b2#.iohk.chain.no\
-    de.PullHeadersRequestH\0R\x07missingB\x06\n\x04item2\xb3\x07\n\x04Node\
+    de.PullHeadersRequestH\0R\x07missingB\x06\n\x04item2\xb4\x07\n\x04Node\
```
mzabaluev: This generated crap in git is precisely why it's better to invoke protoc in the build script.

jormungandr-integration-tests/src/bin/mock_server_app.rs
mzabaluev: OK, if this works...

jormungandr-integration-tests/build.rs
dkijania: I think the problem is with CI, which doesn't have protoc installed:
```
Caused by:
  process didn't exit successfully: `/mnt/crate/target/debug/build/jormungandr-integration-tests-c4c28dc0b4a16e82/build-script-build` (exit code: 101)
--- stderr
thread 'main' panicked at 'protoc: Custom { kind: NotFound, error: "failed to spawn `\"protoc\" \"--version\"`: No such file or directory (os error 2)" }', src/libcore/result.rs:1165:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
```

https://circleci.com/gh/input-output-hk/jormungandr/14884

mzabaluev: Filed #1020 about this problem.

PR #1019, opened 2019-10-23
Safeguards against connecting to self

### LINKED ISSUE #1016: Old gossip may cause the node to connect to self
When a node changes its ID and its old ID is still in gossip, it may attempt to connect to itself. We should detect this and quash any such old gossip, too.

Apart from being obviously wrong, it may be a cause of weird `accept` errors on Mac OS (#953).

mzabaluev:
Pending changes in poldercast to not cause the assertion to fail.

Fixes #1016

### DIFF
--- a/jormungandr/src/network/mod.rs
+++ b/jormungandr/src/network/mod.rs
@@ -352,11 +352,15 @@ fn connect_and_propagate_with<F>(
         }
     };
     let node_id = node.id();
-    assert_ne!(
-        node_id,
-        state.topology.node().id(),
-        "the topology should not tell the node to connect to itself",
-    );
+    // TODO: turn this into an assertion once poldercast is fixed
+    // to never do this.
+    if node_id == state.topology.node().id() {
+        warn!(
+            state.logger(),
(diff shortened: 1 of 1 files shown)


jormungandr/src/network/mod.rs
```diff
@@ -352,6 +352,11 @@ fn connect_and_propagate_with<F>(
         }
     };
     let node_id = node.id();
+    assert_ne!(
```
NicolasDP:
```suggestion
    debug_assert_ne!
```

what about a debug assert instead?

and returning early with an error in the log if that is the case? Just would like to avoid crash now on the node for something that we can _work around_ if it happens

mzabaluev: Could do, but this is meant to just double-check after poldercast which should not give us the self node in views. I think I'd try to go in, make a fix in poldercast and update Cargo.toml with the new version for this branch.

mzabaluev: @NicolasDP is already reworking poldercast, replaced this with a warning and an early return until that is done.

Issue #893, opened 2019-10-04
Signature validation tool


deepfire:
For the stake pool registry we need to be able to validate external signatures for files.

**Describe the solution you'd like**
At a minimum, something like the following command should terminate with exit status `0` or `1` ,
denoting success or failure of signature verification.

`jcli check-signature --pubkey BASE16 --content FILENAME --signature-type SIGFORMAT --signature FILENAME`

Additionally, in case of failure, the tool could print the reason on the standard error:

- whether it was an error in some of inputs (pubkey or signature), or
- whether it was a signature verification error.

NicolasDP:
I believe by SIGFORMAT, you mean the signature scheme. We don't need it as we can get that from a bech32 encoded pubkey with with consistent prefix (i.e. same as any generated keys in `jcli`).

so it will be:

add `jcli key sign`:

* `--secret-key <PATH_TO_SECRET_FILE>: bech32 encoded secret key as outputed by `jcli key generate`.
* `--data <PATH_TO_DATA_FILE>`: unless present, will be read from the standard input. Bytes... no specific format expected.
* `--output <PATH_TO_WRITE_OUTPUT_TO>: unless present will be printed to the standard output. In bech32.

and add `jcli key verify`:

* `--public-key <PATH_TO_PUBLIC_KEY>`: bech32 encoded public key as outputed by `jcli key to-public`.
* `--data <PATH_TO_DATA_FILE>`: unless present, will be read from the standard input. Bytes... no specific format expected.
* `--signature <SIGNATURE_FILE>`: as printed by `jcli key sign`.

This function will return `0` on success, something else otherwise and optionally an error message on the standard error.

deepfire:
@NicolasDP, I'd prefer us to have an additional option of supplying the pubkey as a base16 string -- not even a file.

deepfire:
Integrated in https://github.com/input-output-hk/testnet-stake-pool-registry-validator .

PR #1022, opened 2019-10-24
Add JCLI sign and verify tools

CodeSandwich:
Fixes https://github.com/input-output-hk/jormungandr/issues/893

### DIFF
--- a/jcli/src/jcli_app/key.rs
+++ b/jcli/src/jcli_app/key.rs
@@ -1,14 +1,15 @@
 use bech32::{u5, Bech32, FromBase32, ToBase32};
 use chain_crypto::{
-    AsymmetricKey, AsymmetricPublicKey, Curve25519_2HashDH, Ed25519, Ed25519Bip32, Ed25519Extended,
-    SumEd25519_12,
+    bech32::Bech32 as _, AsymmetricKey,
+    AsymmetricPublicKey, Curve25519_2HashDH, Ed25519, Ed25519Bip32, Ed25519Extended, SecretKey,
+    SigningAlgorithm, SumEd25519_12, VerificationAlgorithm,
 };
 use hex::FromHexError;
 use jcli_app::utils::io;
 use rand::{rngs::EntropyRng, SeedableRng};
 use rand_chacha::ChaChaRng;
 use std::{
-    io::{BufRead, BufReader, Write},
+    io::{BufRead, BufReader, Read, Write},
     path::{Path, PathBuf},
 };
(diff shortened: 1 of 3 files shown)


jcli/src/jcli_app/key.rs
```diff
@@ -214,24 +258,75 @@ impl FromBytes {
...
+    fn sign<A>(self, secret_bytes: &[u8]) -> Result<(), Error>
+    where
+        A: SigningAlgorithm,
+        <A as AsymmetricKey>::PubAlg: VerificationAlgorithm,
+    {
+        let secret = SecretKey::<A>::from_binary(secret_bytes)?;
+        let mut data = Vec::new();
+        io::open_file_read(&self.data)?.read_to_end(&mut data)?;
+        let signature = secret.sign(&data);
+        io::open_file_write(&self.output)?.write_all(signature.as_ref())?;
```
NicolasDP: can we have the signature in `bech32` ?

jcli/src/jcli_app/key.rs
```diff
@@ -99,6 +110,40 @@ pub struct ToPublic {
...
+#[derive(StructOpt, Debug)]
+pub struct Sign {
+    /// path to file with bech32-encoded secret key
+    ///
+    /// supported key formats are: ed25519, ed25519bip32, ed25519extended and sumed25519_12
+    #[structopt(long = "secret-key", short = "s")]
+    secret_key: PathBuf,
+
+    /// path to file with data to sign, if no value is passed, standard input will be used
+    #[structopt(long = "data", short = "d")]
```
NicolasDP: actually, now thinking about it. ideally the command would be as follow:

```
jcli key sign --secret-key my_secret_key.key contract.pdf
```

and 

```
jcli key verify --public-key pub.key --signature contact.signature contract.pdf
```

i.e. having `data` as a  positioned argument.

jcli/src/jcli_app/key.rs
```diff
@@ -99,6 +110,40 @@ pub struct ToPublic {
...
+    /// supported key formats are: ed25519, ed25519bip32 and sumed25519_12
+    #[structopt(long = "public-key", short = "p")]
+    public_key: PathBuf,
+
+    /// path to file with signed data, if no value is passed, standard input will be used
+    #[structopt(long = "data", short = "d")]
+    data: Option<PathBuf>,
+
+    /// path to file with signature
+    #[structopt(long = "signature", short = "s")]
```
NicolasDP: the short here is confusing with the short for the `sign` command. better no shorts than wearing the same one as everyone else

PR #1025, opened 2019-10-24
Sanity check the addresses we are connecting to

### LINKED ISSUE #953: Crash on Jormungandr 0.6.0 on Mac OSX Catalina
Oct 15 03:19:04.372 ERRO Error while accepting connection on 0.0.0.0:3100: Os { code: 22, kind: InvalidInput, message: "Invalid argument" }, task: network
thread 'network9' panicked at 'Box<Any>', jormungandr/src/main.rs:225:30
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace.

Oct 15 04:31:30.659 ERRO Error while accepting connection on 0.0.0.0:3100: Os { code: 22, kind: InvalidInput, message: "Invalid argument" }, task: network
thread 'network2' panicked at 'Box<Any>', jormungandr/src/main.rs:225:30
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace.

This is my crash output while running Mac OSX Catalina 10.15 and using zsh for a shell vice bash since Catalina uses zsh by default. 12 Core Xeon processor with 64 GB RAM. The node crashed after about 5 hours of run time, then again after 1 hour and 12 minutes. Was running solid for days on 0.5.6.

mzabaluev:
Refuse to connect to our known listening address and the public address.

Fixes #953

### DIFF
--- a/jormungandr/src/network/mod.rs
+++ b/jormungandr/src/network/mod.rs
@@ -333,14 +333,7 @@ fn send_gossip(state: GlobalStateR, channels: Channels) {
     }
 }
 
-fn connect_and_propagate_with<F>(
-    node: topology::NodeData,
-    state: GlobalStateR,
-    channels: Channels,
-    modify_comms: F,
-) where
-    F: FnOnce(&mut PeerComms),
-{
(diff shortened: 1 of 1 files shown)


NicolasDP:
This is no longer needed as we are already handling this from the poldercast PoV now

PR #1035, opened 2019-10-25
Implement pagination for transactions

enzoc4:
# Summary

- Add pagination for transactions by address
    - For this was necessary to remember the order, so the previous `Set` is replaced with a sequence.
- Add pagination for transactions in block
   - These are sorted by the order they are in the block
- Use offsets for pagination instead of opaque cursors in all connections (but still maintaining the Connection model)

EDIT:
This is a breaking change for the resolvers that return transactions,  the one in `Address` and the one in `Block`.
Something like this would need to be made to get the same result as before (for transactions in the genesis block):
```graphql
{
  blockByChainLength(length: "0") {
    transactions {
      edges {
        node {
          id
        }
      }
    }
  }
}
```

### DIFF
--- a/jormungandr/src/explorer/graphql/mod.rs
+++ b/jormungandr/src/explorer/graphql/mod.rs
@@ -117,13 +117,7 @@ impl Block {
                     let to = usize::try_from(range.upper_bound).unwrap();
 
                     (from..=to)
-                        .map(|i| {
-                            (
-                                transactions[i].id().clone(),
-                                i.try_into()
-                                    .expect("tried to paginate more than 2^32 elements"),
-                            )
-                        })
+                        .map(|i| (transactions[i].id().clone(), i.try_into().unwrap()))
                         .collect::<Vec<(FragmentId, u32)>>()
                 }
             },
--- a/jormungandr/src/explorer/graphql/mod.rs
+++ b/jormungandr/src/explorer/graphql/mod.rs
@@ -83,7 +83,7 @@ impl Block {
             .sort_unstable_by_key(|tx| tx.offset_in_block);
 
         let lower_bound = 0u32;
-        let upper_bound = transactions.len().checked_sub(1).unwrap_or(0);
+        let upper_bound = transactions.len();
 
         TransactionConnection::new(
             lower_bound,
@@ -100,12 +100,10 @@ impl Block {
                 let from = usize::try_from(from).unwrap();
                 let to = usize::try_from(to).unwrap();
 
-                transactions[from..min(transactions.len(), to.checked_add(1).unwrap())]
-                    .iter()
-                    .enumerate()
-                    .map(|(i, tx)| {
+                (from..to)
+                    .map(|i| {
                         (
-                            tx.id().clone(),
+                            transactions[i].id().clone(),
                             i.try_into()
                                 .expect("tried to paginate more than 2^32 elements"),
                         )
@@ -205,36 +203,53 @@ impl From<blockcfg::BlockDate> for BlockDate {
 #[derive(Clone)]
 struct Transaction {
     id: FragmentId,
-    in_block: Option<HeaderHash>,
+    block_hash: Option<HeaderHash>,
 }
 
 impl Transaction {
     fn from_id(id: FragmentId, context: &Context) -> FieldResult<Transaction> {
-        let in_block = Self::get_in_block(&id, context).ok_or(ErrorKind::NotFound(format!(
-            "transaction not found: {}",
-            &id,
-        )))?;
+        let block_hash = context
+            .db
+            .find_block_hash_by_transaction(&id)
+            .wait()
+            .unwrap()
+            .ok_or(ErrorKind::NotFound(format!(
+                "transaction not found: {}",
+                &id,
+            )))?;
 
         Ok(Transaction {
             id,
-            in_block: Some(in_block),
+            block_hash: Some(block_hash),
         })
     }
 
     fn from_valid_id(id: FragmentId) -> Transaction {
-        Transaction { id, in_block: None }
+        Transaction {
+            id,
+            block_hash: None,
+        }
     }
 
-    fn get_in_block(id: &HeaderHash, context: &Context) -> Option<HeaderHash> {
-        context.db.find_block_by_transaction(&id).wait().unwrap()
+    fn get_block_hash(id: &HeaderHash, context: &Context) -> Option<HeaderHash> {
+        context
+            .db
+            .find_block_hash_by_transaction(&id)
+            .wait()
+            .unwrap()
     }
 
     fn get_block(&self, context: &Context) -> FieldResult<ExplorerBlock> {
-        let block_id = match self.in_block {
+        let block_id = match self.block_hash {
             Some(block_id) => block_id,
-            None => Self::get_in_block(&self.id, context).ok_or(ErrorKind::InternalError(
-                "Transaction's block was not found".to_owned(),
-            ))?,
+            None => context
+                .db
+                .find_block_hash_by_transaction(&self.id)
+                .wait()
+                .unwrap()
+                .ok_or(ErrorKind::InternalError(
+                    "Transaction's block was not found".to_owned(),
+                ))?,
         };
 
         context.db.get_block(&block_id).wait()?.ok_or(
@@ -831,7 +846,7 @@ impl Query {
 
         BlockConnection::new(
             block0,
-            longest_chain.into(),
+            u32::from(longest_chain) + 1,
             first,
             last,
             before,
--- a/jormungandr/src/explorer/graphql/mod.rs
+++ b/jormungandr/src/explorer/graphql/mod.rs
@@ -96,17 +96,18 @@ impl Block {
             after,
             |from: u32, to: u32| {
                 use std::cmp::min;
-                transactions[usize::try_from(from).unwrap()
-                    ..min(
-                        transactions.len(),
-                        usize::try_from(to.checked_add(1).unwrap()).unwrap(),
-                    )]
+
+                let from = usize::try_from(from).unwrap();
+                let to = usize::try_from(to).unwrap();
+
+                transactions[from..min(transactions.len(), to.checked_add(1).unwrap())]
                     .iter()
                     .enumerate()
                     .map(|(i, tx)| {
                         (
                             tx.id().clone(),
-                            i.try_into().expect("tried to paginate more than 2^32 elements"),
+                            i.try_into()
+                                .expect("tried to paginate more than 2^32 elements"),
                         )
                     })
                     .collect::<Vec<(FragmentId, u32)>>()
--- a/jormungandr/src/explorer/graphql/mod.rs
+++ b/jormungandr/src/explorer/graphql/mod.rs
@@ -1,7 +1,10 @@
 mod connections;
 mod error;
 mod scalars;
-use self::connections::{BlockConnection, PaginationArguments, TransactionConnection};
+use self::connections::{
+    BlockConnection, InclusivePaginationInterval, PaginationArguments, PaginationInterval,
+    TransactionConnection,
+};
 use self::error::ErrorKind;
 use super::indexing::{BlockProducer, EpochData, ExplorerBlock, ExplorerTransaction};
 use super::persistent_sequence::PersistentSequence;
@@ -82,9 +85,6 @@ impl Block {
             .as_mut_slice()
             .sort_unstable_by_key(|tx| tx.offset_in_block);
 
-        let lower_bound = 0u32;
-        let upper_bound = transactions.len();
-
         let pagination_arguments = PaginationArguments {
             first,
             last,
@@ -93,25 +93,39 @@ impl Block {
         }
         .validate()?;
 
-        TransactionConnection::new(
-            lower_bound,
-            upper_bound
-                .try_into()
-                .expect("tried to paginate more than 2^32 elements"),
-            pagination_arguments,
-            |from: u32, to: u32| {
-                let from = usize::try_from(from).unwrap();
-                let to = usize::try_from(to).unwrap();
+        let boundaries = if transactions.len() > 0 {
+            PaginationInterval::Inclusive(InclusivePaginationInterval {
+                lower_bound: 0u32,
+                upper_bound: transactions
+                    .len()
+                    .checked_sub(1)
+                    .unwrap()
+                    .try_into()
+                    .expect("tried to paginate more than 2^32 elements"),
+            })
+        } else {
+            PaginationInterval::Empty
+        };
 
-                (from..to)
-                    .map(|i| {
-                        (
-                            transactions[i].id().clone(),
-                            i.try_into()
-                                .expect("tried to paginate more than 2^32 elements"),
-                        )
-                    })
-                    .collect::<Vec<(FragmentId, u32)>>()
+        TransactionConnection::new(
+            boundaries,
+            pagination_arguments,
+            |range: PaginationInterval<u32>| match range {
+                PaginationInterval::Empty => vec![],
+                PaginationInterval::Inclusive(range) => {
+                    let from = usize::try_from(range.lower_bound).unwrap();
+                    let to = usize::try_from(range.upper_bound).unwrap();
+
+                    (from..=to)
+                        .map(|i| {
+                            (
+                                transactions[i].id().clone(),
+                                i.try_into()
+                                    .expect("tried to paginate more than 2^32 elements"),
+                            )
+                        })
+                        .collect::<Vec<(FragmentId, u32)>>()
+                }
             },
         )
     }
@@ -399,8 +413,14 @@ impl Address {
                 "Expected address to be indexed".to_owned(),
             ))?;
 
-        let lower_bound = 0u64;
-        let upper_bound: u64 = transactions.len();
+        let boundaries = if transactions.len() > 0 {
+            PaginationInterval::Inclusive(InclusivePaginationInterval {
+                lower_bound: 0u64,
+                upper_bound: transactions.len(),
+            })
+        } else {
+            PaginationInterval::Empty
+        };
 
         let pagination_arguments = PaginationArguments {
             first,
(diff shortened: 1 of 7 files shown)


jormungandr/src/explorer/graphql/mod.rs
```diff
@@ -17,11 +16,13 @@ use std::str::FromStr;
...
+    BlockCount, ChainLength, EpochNumber, IndexCursor, PoolId, PublicKey, Serial, Slot,
+    TimeOffsetSeconds, Value,
 };
 
 use crate::explorer::{ExplorerDB, Settings};
 
+#[derive(Clone)]
 pub struct Block {
     hash: HeaderHash,
 }
```
mzabaluev: This is meant to have other fields as needed to send information about blocks, it's not just here to have a hash, right?

jormungandr/src/explorer/graphql/mod.rs
```diff
@@ -63,16 +64,54 @@ impl Block {
...
-                id: id.clone(),
-                in_block: self.hash.clone(),
-            })
-            .collect())
+    pub fn transactions(
+        &self,
+        first: Option<i32>,
+        last: Option<i32>,
+        before: Option<IndexCursor>,
+        after: Option<IndexCursor>,
```
mzabaluev: These four would be good to group as fields in a pagination parameter struct.

jormungandr/src/explorer/graphql/mod.rs
```diff
@@ -63,16 +64,54 @@ impl Block {
...
+        context: &Context,
+    ) -> FieldResult<TransactionConnection> {
+        let explorer_block = self.get_explorer_block(&context.db)?;
+        let mut transactions: Vec<&ExplorerTransaction> =
+            explorer_block.transactions.values().collect();
+
+        // TODO: This may be expensive at some point, but I can't rely in
+        // the HashMap's order (also, I'm assuming the order in the block matters)
```
mzabaluev: There's `LinkedHashMap` if that's of any help, it is already in use in `network`.

jormungandr/src/explorer/graphql/mod.rs
```diff
@@ -63,16 +64,54 @@ impl Block {
...
+        transactions
+            .as_mut_slice()
+            .sort_unstable_by_key(|tx| tx.offset_in_block);
+
+        let lower_bound = 0u32;
+        let upper_bound = transactions.len().checked_sub(1).or(Some(0)).unwrap();
```
mzabaluev: `.checked_sub(1).unwrap_or(0)` would be more concise.

jormungandr/src/explorer/graphql/mod.rs
```diff
@@ -63,16 +64,54 @@ impl Block {
...
+
+        TransactionConnection::new(
+            lower_bound,
+            upper_bound
+                .try_into()
+                .expect("less than 2^32 transactions per block"),
```
mzabaluev: The message is for a panic, so it should say what the problem is when the operation fails. This is one unfortunately named method.

jormungandr/src/explorer/graphql/mod.rs
```diff
@@ -63,16 +64,54 @@ impl Block {
...
-            .iter()
-            .map(|(id, _tx)| Transaction {
-                id: id.clone(),
-                in_block: self.hash.clone(),
-            })
-            .collect())
+    pub fn transactions(
+        &self,
+        first: Option<i32>,
+        last: Option<i32>,
```
mzabaluev: Is there meaning behind these parameters being potentially negative?

jormungandr/src/explorer/graphql/mod.rs
```diff
@@ -63,16 +64,54 @@ impl Block {
...
+            first,
+            last,
+            before,
+            after,
+            |from: u32, to: u32| {
+                use std::cmp::min;
+                transactions[usize::try_from(from).unwrap()
+                    ..min(
+                        transactions.len(),
+                        usize::try_from(to.checked_add(1).unwrap()).unwrap(),
```
mzabaluev: It would be better for readability if these long sanitization expressions were on separate lines as `let from = ...`.

jormungandr/src/explorer/graphql/mod.rs
```diff
@@ -63,16 +64,54 @@ impl Block {
...
+                .try_into()
+                .expect("less than 2^32 transactions per block"),
+            first,
+            last,
+            before,
+            after,
+            |from: u32, to: u32| {
+                use std::cmp::min;
+                transactions[usize::try_from(from).unwrap()
+                    ..min(
```
mzabaluev: You can use an inclusive range operator `..=`, so that you don't have to do arithmetics on the last index.

jormungandr/src/explorer/graphql/mod.rs
```diff
@@ -63,16 +64,54 @@ impl Block {
...
+                    )]
+                    .iter()
+                    .enumerate()
+                    .map(|(i, tx)| {
+                        (
+                            tx.id().clone(),
+                            i.try_into().expect("less than 2^32 transactions per block"),
+                        )
+                    })
+                    .collect::<Vec<(FragmentId, u32)>>()
```
mzabaluev: `.enumerate()` counts from 0. If you meant it to be so, what's the purpose of collecting 0-based enumeration indices into a vector that is already indexed this way?

jormungandr/src/explorer/graphql/mod.rs
```diff
@@ -162,28 +201,42 @@ impl From<blockcfg::BlockDate> for BlockDate {
...
+            id,
+            in_block: Some(in_block),
+        })
+    }
+
+    fn from_valid_id(id: FragmentId) -> Transaction {
+        Transaction { id, in_block: None }
+    }
+
+    fn get_in_block(id: &HeaderHash, context: &Context) -> Option<HeaderHash> {
```
mzabaluev: This has no relation to the `Transaction` type, so it's better be a method on `Context` or, if you don't define that (can't quickly find where it's coming from), as a free function.

jormungandr/src/explorer/graphql/mod.rs
```diff
@@ -162,28 +201,42 @@ impl From<blockcfg::BlockDate> for BlockDate {
...
     fn get_block(&self, context: &Context) -> FieldResult<ExplorerBlock> {
-        context.db.get_block(&self.in_block).wait()?.ok_or(
+        let block_id = match self.in_block {
+            Some(block_id) => block_id,
+            None => Self::get_in_block(&self.id, context).ok_or(ErrorKind::InternalError(
+                "Transaction's block was not found".to_owned(),
+            ))?,
+        };
+
+        context.db.get_block(&block_id).wait()?.ok_or(
```
mzabaluev: Why do you need a separate lookup just to check if the block is there, before obtaining it? Is the database error not descriptive enough to detect this condition?

jormungandr/src/explorer/graphql/mod.rs
```diff
@@ -162,28 +201,42 @@ impl From<blockcfg::BlockDate> for BlockDate {
...
+
+    fn from_valid_id(id: FragmentId) -> Transaction {
+        Transaction { id, in_block: None }
+    }
+
+    fn get_in_block(id: &HeaderHash, context: &Context) -> Option<HeaderHash> {
+        context.db.find_block_by_transaction(&id).wait().unwrap()
     }
 
     fn get_block(&self, context: &Context) -> FieldResult<ExplorerBlock> {
```
mzabaluev: Same as above.

jormungandr/src/explorer/graphql/mod.rs
```diff
@@ -320,18 +373,38 @@ impl Address {
...
     }
 
-    fn transactions(&self, context: &Context) -> FieldResult<Vec<Transaction>> {
-        let ids = context
+    fn transactions(
+        &self,
+        first: Option<i32>,
+        last: Option<i32>,
+        before: Option<IndexCursor>,
+        after: Option<IndexCursor>,
```
mzabaluev: Here, again, using a struct would save some typing and guessing which parameter in the function call refers to what.

jormungandr/src/explorer/graphql/mod.rs
```diff
@@ -621,8 +694,8 @@ impl Epoch {
         &self,
         first: Option<i32>,
         last: Option<i32>,
-        before: Option<BlockCursor>,
-        after: Option<BlockCursor>,
+        before: Option<IndexCursor>,
+        after: Option<IndexCursor>,
```
mzabaluev: And here, and everywhere else.

jormungandr/src/explorer/graphql/mod.rs
```diff
@@ -633,23 +706,36 @@ impl Epoch {
         let lower_bound = context
             .db
             .get_block(&epoch_data.first_block)
-            .map(|block| block.expect("The block to be indexed").chain_length)
+            .map(|block| u32::from(block.expect("The block to be indexed").chain_length))
```
mzabaluev: Why are block information structs convertible to integer indices? A `From` impl does not make sense here and might cause confusion.

jormungandr/src/explorer/graphql/scalars.rs
```diff
@@ -92,3 +117,33 @@ impl From<chain_time::TimeOffsetSeconds> for TimeOffsetSeconds {
...
     }
 }
+
+impl From<u32> for TransactionCount {
+    fn from(n: u32) -> TransactionCount {
+        TransactionCount(format!("{}", u32::from(n)))
+    }
+}
+
+impl From<u32> for IndexCursor {
```
mzabaluev: If this is not checked in any way on conversion, why the hassle with newtypes?

jormungandr/src/explorer/graphql/mod.rs
mzabaluev: Not same as above, sorry.

jormungandr/src/explorer/graphql/mod.rs
enzoc4: Well, probably. Right now I'm just asking the "db" for the block's body directly in every resolver, as the "db" is not actually a db but an object in memory. My original idea was to have something like:
```rust
body: Option<ExplorerBlock>
```
or maybe something more granular.

Your concern was about the `Clone` or the fact this could only be a newtype?

mzabaluev: I was wondering if it's only a newtype with a name that means something else, but yeah, if it's supposed to have other fields later on, it's OK.

jormungandr/src/explorer/graphql/scalars.rs
enzoc4: I can't expose a `u32` directly to graphql. It only has i32 and floats, so I need to pass it as a string. I'm just using this to have a name (different from String) for it in the graphql schema.  Do you think the `From` implementation is unnecessary?

jormungandr/src/explorer/graphql/mod.rs
enzoc4: what structs? `ChainLength` is a newtype wrapping a `u32`

jormungandr/src/explorer/graphql/mod.rs
mzabaluev: Yeah, the resolver API being generic (the pagination param names are just a React convention, right?) does not mean it has to percolate into our internal Rust code organization.

jormungandr/src/explorer/graphql/mod.rs
mzabaluev: Ah, right.

jormungandr/src/explorer/graphql/mod.rs
mzabaluev: Agree. The `Transaction::get_in_block` non-methodic method could be named `check_hash_in_block` or something like that.

jormungandr/src/explorer/graphql/mod.rs
mzabaluev: I see. I guess we'll have to live with it.

jormungandr/src/explorer/graphql/mod.rs
mzabaluev: OTOH, if you say these functions are consumed by juniper and there is no way to change that, let it be.

jormungandr/src/explorer/graphql/mod.rs
mzabaluev: Right, I did not read this carefully enough, there's a field access at the end of it.
The result of `get_block` should be meant to be lightweight enough for this to be good going forward, otherwise there should be data store methods to query the header information specifically.

jormungandr/src/explorer/graphql/scalars.rs
```diff
@@ -33,6 +36,28 @@ pub struct PublicKey(pub String);
 #[derive(juniper::GraphQLScalarValue)]
 pub struct TimeOffsetSeconds(pub String);
 
+// u32 should be enough to count blocks and transactions (the only two cases for now)
+#[derive(Clone)]
+pub struct IndexCursor(pub u32);
+
+juniper::graphql_scalar!(IndexCursor where Scalar = <S> {
+    description: ""
```
mzabaluev: Hmm, we ought to have some nice GraphQL metadata to our API if we can get it for free here.

jormungandr/src/explorer/graphql/scalars.rs
mzabaluev: I get the purpose now, thanks. It's a good practice actually.

jormungandr/src/explorer/persistent_sequence.rs
```diff
@@ -0,0 +1,36 @@
+use imhamt::Hamt;
+use std::collections::hash_map::DefaultHasher;
+
+// Use a Hamt to store a sequence, the indexes can be used for pagination
+// XXX: Maybe there is a better data structure for this?
+#[derive(Clone)]
+pub struct PersistentSequence<T> {
+    // Could be usize, but it's only used to store blocks and transactions for now
```
mzabaluev: Could there be more than 2^32 blocks at some future date, or is it within an epoch only?

In general, it only makes sense to be stingy on this if we seriously target wasm32 and small devices, otherwise just use `u64` wherever 1) it's not about a size of a data structure in memory; 2) you are not sure that the number is always going to be so ridiculously small as to not overflow `u32` even in intermediate calcualtions.

jormungandr/src/explorer/graphql/mod.rs
mzabaluev: Right; I thought these could be Pythonic wraparound indices, but this just looks like juniper being too Javascriptey.

jormungandr/src/explorer/graphql/scalars.rs
enzoc4: You are right

jormungandr/src/explorer/graphql/mod.rs
enzoc4: I'm not really storing the header (`get_block` gives you an `ExplorerBlock`, which is a version with only the things the API is exposing), if blocks get too big it may be better to do what you say, but I'm not sure if I should address that now, as most blocks won't have that many transactions. (and they are stored in memory)

jormungandr/src/explorer/graphql/mod.rs
```diff
@@ -63,16 +64,56 @@ impl Block {
...
+            |from: u32, to: u32| {
+                let from = usize::try_from(from).unwrap();
+                let to = usize::try_from(to).unwrap();
+
+                (from..to)
+                    .map(|i| {
+                        (
+                            transactions[i].id().clone(),
+                            i.try_into()
+                                .expect("tried to paginate more than 2^32 elements"),
```
mzabaluev: This is now infallible as `both` from and `to` are `u32`?

jormungandr/src/explorer/graphql/mod.rs
```diff
@@ -162,28 +202,51 @@ impl From<blockcfg::BlockDate> for BlockDate {
...
+                .db
+                .find_block_hash_by_transaction(&self.id)
+                .wait()
+                .unwrap()
+                .ok_or(ErrorKind::InternalError(
+                    "Transaction's block was not found".to_owned(),
+                ))?,
+        };
+
+        context.db.get_block(&block_id).wait()?.ok_or(
```
mzabaluev: OK, though I suspect this dynamism is still not warranted and you can instead have two types for a transaction before its block is found and a transaction with a confirmed block, and transition from the first to the last as needed with a `self`-consuming method.

jormungandr/src/explorer/graphql/scalars.rs
```diff
@@ -69,6 +96,12 @@ impl TryFrom<EpochNumber> for blockcfg::Epoch {
     }
 }
 
+impl From<u64> for BlockCount {
+    fn from(number: u64) -> BlockCount {
+        BlockCount(format!("{}", number))
```
mzabaluev: I wonder if juniper requires these newtypes to wrap strings, or these can be the wrapped integers represented with `.to_string()` in the GraphQL-facing APIs as necessary.

enzoc4: I can do that, it just requires a bit more of boilerplate, as I can't use the derive macro. I can wrap whatever type I want, I just need to provide the serialization and deserialization functions to the `juniper::graphql_scalar!` macro. (Or maybe there is a custom attribute for the derive, but I didn't find it in the documentation)

jormungandr/src/explorer/graphql/mod.rs
enzoc4: yes, good eye

jormungandr/src/explorer/graphql/mod.rs
enzoc4: I'll think about this, it's definitely interesting. The resolver must be implemented for just one type, but I'm thinking I can wrap an enum or something like that, like (just an on the fly example):

```rust
enum Transaction {
    UnconfirmedBlock { id: FragmentId },
    ConfirmedBlock { id: FragmentId, block: Block }
} 

impl Transaction {
    fn transition_method(self) -> Transaction { ... } 
}

#[juniper::object(
    Context = Context
)]
impl Transaction { resolvers here... }
```
Or maybe something using generics...

mzabaluev: The enum is not much an improvement over having an `Option<BlockHash>` as a member: same dynamism mostly, except now you also need to use an accessor method to get the reference to the invariable part for methods that don't have anything to do with the block hash.

Generics would get you into boxed objects and dynamic dispatch, as you still need to be able to hold values of both types in one API slot I guess? If so, don't bother.

NicolasDP:
LGTM, @enzoc4 could you rebase your PR to resolve the conflicts?

enzoc4:
@NicolasDP done

PR #1058, opened 2019-11-01
Lay ground for policy management and update

NicolasDP:
Utilise the latest version of the `poldercast` library which includes policy management and some quarantine policy to prevent use or gossiping of invalid nodes.

### DIFF
--- a/jormungandr/src/network/client/connect.rs
+++ b/jormungandr/src/network/client/connect.rs
@@ -1,6 +1,6 @@
 use super::super::{
     grpc,
-    p2p::{comm::PeerComms, topology},
+    p2p::{comm::PeerComms, Gossip as NodeData, Id},
     Channels, ConnectionState,
 };
 use super::{Client, ClientBuilder, GlobalStateR, InboundSubscriptions};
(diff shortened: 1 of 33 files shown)


jormungandr/src/network/p2p/topology.rs
```diff
@@ -1,341 +1,91 @@
...
-    }
-
-    /// this is the function to utilise in order to select gossips to share
-    /// with a given node
-    pub fn select_gossips(&self, gossip_recipient: &NodeData) -> impl Iterator<Item = NodeData> {
+    pub fn view(&self) -> Vec<Node> {
         let mut topology = self.lock.write().unwrap();
         topology
-            .select_gossips(&gossip_recipient.0)
+            .view(None, poldercast::Selection::Any)
```
mzabaluev: Looks mighty so far. I can only hope it's not a poorly scalable bag of locks internally, because traversal of all these locks has to happen in the network task.

NicolasDP: there is no locks inside poldercast

jormungandr/src/network/client/connect.rs
```diff
@@ -1,14 +1,13 @@
 use super::super::{
     grpc,
-    p2p::{comm::PeerComms, topology},
+    p2p::{comm::PeerComms, Gossip as NodeData, Id},
```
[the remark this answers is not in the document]
NicolasDP: you used `NodeId` for the trait's associated type.

Issue #1057, opened 2019-11-01
Passive node does not sync (receive blocks) - `v0.7.0-rc3` (local cluster, 2 nodes)


dorin100:
**Describe the bug**
Passive does not do not receive any blocks from Leader/Stake Pool node.

**Mandatory Information**
```
jcli 0.7.0-rc3 (HEAD-466c0fb0, release, windows [x86_64]) - [rustc 1.38.0 (625451e37 2019-09-23)]
jormungandr 0.7.0-rc3 (HEAD-466c0fb0, release, windows [x86_64]) - [rustc 1.38.0 (625451e37 2019-09-23)]
```

**To Reproduce**
1. Start node 1 (leader node - configs attached) - `jormungandr ---genesis-block block-0.bin --config node_config.yaml --secret node_secret.yaml`
2. Start node 2 (passive node - config attached) - `jormungandr ---config node_config.yaml --genesis-block-hash a087451eceeaef00ddf7d69c2084b961bd2bdcb4c891a11db16b779199a6bdaf`

**Expected behavior**
The passive node should be in sync with the leader node.

**Additional context**
Configuration details:
- VM with windows
- 2 nodes connected together (direct connectivity) - 1 passive + 1 leader (with 1 stake pool configured inside genesis.yaml)

Attachments:
- leader node
[node1_faucet.zip](https://github.com/input-output-hk/jormungandr/files/3797870/node1_faucet.zip)
[leader_node_logs.txt](https://github.com/input-output-hk/jormungandr/files/3797871/leader_node_logs.txt)

- passive node
[node2_stake_pool.zip](https://github.com/input-output-hk/jormungandr/files/3797872/node2_stake_pool.zip)
[passive_node_logs.txt](https://github.com/input-output-hk/jormungandr/files/3797873/passive_node_logs.txt)

Other results:
```
$ netstat -an | grep -E ":(3000|3001|3002)"
  TCP    127.0.0.1:3001         0.0.0.0:0              LISTENING
  TCP    127.0.0.1:3001         127.0.0.1:63822        ESTABLISHED
  TCP    127.0.0.1:3002         0.0.0.0:0              LISTENING
  TCP    127.0.0.1:63822        127.0.0.1:3001         ESTABLISHED
```
```
Node 1 network stats:
---
- establishedAt: "2019-11-01T11:38:03.092771900+00:00"
  lastBlockReceived: ~
  lastFragmentReceived: ~
[...]

PR #1059, opened 2019-11-01
Docs: allow_private_addresses - add to sample cfg

rinor:
Related to https://github.com/input-output-hk/jormungandr/issues/1057

### DIFF
--- a/doc/configuration/introduction.md
+++ b/doc/configuration/introduction.md
@@ -21,4 +21,5 @@ p2p:
   topics_of_interest:
     messages: low
     blocks: normal
+  allow_private_addresses: false
 ```


doc/configuration/introduction.md
```diff
@@ -21,4 +21,5 @@ p2p:
   topics_of_interest:
     messages: low
     blocks: normal
+  allow_private_addresses: true
```
NicolasDP:
```suggestion
  allow_private_addresses: false
```

By default I would put false (especially because this is the default setting anyway). This is because under normal settings the users don't want to accept nodes that are advertising unreachable addresses.

This isn not true obviously in the context of testing locally or in a private cluster.

This will change btw, we are looking at some strategy to allow any addresses to be set anyway.

rinor: Ok. true was selected just because the public_address has a loopback address

PR #1063, opened 2019-11-02
Transaction update (major breaking changes)

vincenthz:
### Chain-libs change

* Keep the transaction in binary format and slice it efficiently on demand (**API BREAKING CHANGE**)
* Get rid of AuthenticatedTransaction (now just Transaction) (**API BREAKING CHANGE**)
* Add InputOutputs builder which is not parametrised to the type anymore, but expect it explicitly when calculating fees.
* Redo the transaction builder to be a state machine: 1) set payload 2) set input-outputs 3) set witnesses 4) set payload-auth (**API BREAKING CHANGE**)
* Bind certificate with transaction (**BREAKING CHANGE**)
* simplify tests and add more efficient kitchen sink construct (e.g. TestLedger) to allow less repetition and more state tracking
* Flatten PoolManagement into PoolUpdate/PoolRetirement (**BREAKING CHANGE**)
* Simplify the fees calculation (**API BREAKING CHANGE**)

### jcli usage change

`jcli certificate sign`:

* now take explicit secret keys using the -k/--key flag
* verify that the secret key is matching the associated public key (if possible)
* is now only used for the block0 creation, as for certificate using the transaction system, we know need `transaction auth`.

`jcli transaction auth`:

* new subcommand that need to happens after sealing
* take explicit secret keys using the -k/--key flags
* similar to `jcli certificate sign` but apply to an almost built transaction
* only needed for StakeDelegation, PoolRegistration, PoolRetirement, PoolUpdate. Not for OwnerStakeDelegation

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1686,6 +1686,7 @@ dependencies = [
  "rand_core 0.3.1",
  "serde",
  "serde_yaml",
+ "typed-bytes",
 ]
 
 [[package]]
--- a/jcli/src/jcli_app/block/mod.rs
+++ b/jcli/src/jcli_app/block/mod.rs
@@ -1,4 +1,4 @@
-use chain_core::property::{Block as _, Deserialize, HasFragments, Serialize};
+use chain_core::property::{Block as _, Deserialize, Serialize};
 use chain_impl_mockchain::{
     block::Block,
     ledger::{self, Ledger},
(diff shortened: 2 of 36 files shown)


rinor:
@vincenthz just fell in 'love' with `transaction auth` :boom:. Much...much cleaner flow separation. Awesome and thank you.

Issue #674, opened 2019-08-01
JCLI's transaction `info` and `id` not really doing what they say


NicolasDP:
since change in #633 `jcli transaction id` does not return the transaction `ID` in the sense to retrieve it in the UTxO or to use it in the input of a transaction.

In order to make things right we ought to update this:

- [x] rename current `jcli transaction id` into `jcli transaction data-for-witness`
- [ ] write function `jcli transaction id` to actually create the fragment id;
- [ ] update `jcli transaction info` to return the fragment id.

rinor:
Based on the latest changes https://github.com/input-output-hk/chain-libs/pull/64 and https://github.com/input-output-hk/jormungandr/pull/726, thank you @vincenthz,
with `self` being an instance of:
https://github.com/input-output-hk/jormungandr/blob/ec5eba4611754f0eec40441cbbd8f71115f891db/jcli/src/jcli_app/transaction/staging.rs#L34-L40

would the following be the correct way to create/get the fragment id?
```rust
pub fn id(&self) -> FragmentId {
    let finalizer = chain::txbuilder::TransactionFinalizer::new(self.transaction());
    finalizer.to_fragment().unwrap().hash() //.to_raw().id()
}
```
If yes, is `panic` (through unwrap) the desired action in case `to_fragment()` errors?

rinor:
Answering to myself:
> would the following be the correct way to create/get the fragment id?

No, it is wrong. `to_fragment()` will try to finalize the transaction, which is only possible if all witnesses have been filled. If not it will fail with an `Err` value: MissingWitnessAt { index: 0 }

rinor:
I think this would be the right moment to implement also this (due to recent breaking-changes).

Question: 
Is it necessary for the (new fragment related) `transaction id` to be available at all transaction states (StagingKind) or it suffices just on `sealed`

rinor:
Done in https://github.com/input-output-hk/jormungandr/pull/1073

NicolasDP:
Thanks again @rinor

PR #1069, opened 2019-11-03
jcli: transaction (id -> data-for-witness) and new id

rinor:
1. temporary disable some unavaliable variables from `info` format
2. replace variable `id` -> `sign-data-hash` on `info` format variables list.
3. `transaction id` -> `transaction data-for-witness` (TODO: documentation)
4. new _limited_ `transaction id` (fragment id), 
   works only on `sealed` tx, hence not included on `info`.

partial attempt on https://github.com/input-output-hk/jormungandr/issues/674

### DIFF
--- a/jcli/src/jcli_app/transaction/info.rs
+++ b/jcli/src/jcli_app/transaction/info.rs
@@ -26,12 +26,12 @@ pub struct Info {
     /// formatting for the output to displays
     /// user "{name}" to display the variable with the named `name'.
     ///
-    /// available variables: id, num_inputs, num_outputs, num_witnesses, fee
-    /// balance, input, output and status
+    /// available variables: sign-data-hash, num_inputs, num_outputs, num_witnesses
+    // id, fee, balance, input, output and status
     ///
     #[structopt(
         long = "format",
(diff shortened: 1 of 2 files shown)


vincenthz:
for upgrade pain limitation purpose, we really need to keep the `id` subcommand so that we don't break all the script altogether and instead just add a deprecated notice on `stderr`

rinor:
> for upgrade pain limitation purpose, we really need to keep the `id` subcommand so that we don't break all the script altogether and instead just add a deprecated notice on `stderr`

Agree, so `jcli transaction id` and `jcli transaction data-for-witness` will have the same functionality, for now.

I would still like to have a way to get the fragmentId before sending the transaction. 
Is it acceptable to add `jcli transaction fragment-id ...`?

vincenthz:
yes, a great addition indeed

rinor:
Will close this one, since the new proposal isn't a breaking-change, and wait for https://github.com/input-output-hk/jormungandr/tree/tx_info to be merged before a new pr.

rinor:
new PR https://github.com/input-output-hk/jormungandr/pull/1073

PR #1070, opened 2019-11-04
Restore REST TX info functionality

CodeSandwich:
Restores full TX info functionality from before fallout except for displaying ID of an incomplete TX, this had to be removed. There's no cleanup or UI changes, this will be in separate PR.

### DIFF
--- a/jcli/src/jcli_app/transaction/info.rs
+++ b/jcli/src/jcli_app/transaction/info.rs
@@ -31,7 +31,7 @@ pub struct Info {
     ///
     #[structopt(
         long = "format",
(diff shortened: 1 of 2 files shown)


jcli/src/jcli_app/transaction/info.rs
```diff
@@ -31,7 +31,7 @@ pub struct Info {
     ///
     #[structopt(
         long = "format",
-        default_value = "Transaction `{id}' ({status})\n  Input:   {input}\n  Output:  {output}\n  Fees:    {fee}\n  Balance: {balance}\n"
+        default_value = "Transaction ({status})\n  Input:   {input}\n  Output:  {output}\n  Fees:    {fee}\n  Balance: {balance}\n"
```
rinor: You could replace `{id}` with `{sign-data-hash}` instead of deleting it.

PR #1073, opened 2019-11-04
jcli transaction - add data-for-witness (alias id) and fragment-id

rinor:
- [x]  add `data-for-witness` providing same functionality as `id`
- [x] `id` is now an **alias** of `data-for-witness` with a deprecation message on:
 - _help_
 - _usage_ (_stderr_)
- [x] add `fragment-id` from a given **sealed** transaction.
- [x] change `id` to `data-for-witness` on tests cmd args.
- [x] update docs mentioning also `id` deprecation
- [ ] update tests to reflect `data-for-witness` name change from `id` (is it worth it?)

### DIFF
--- a/doc/jcli/transaction.md
+++ b/doc/jcli/transaction.md
@@ -26,9 +26,12 @@ There are also functions to help decode and display the
 content information of a transaction:
 
 * `info`
-* `id` to get the **Transaction ID** of the transaction
+* `data-for-witness` get the data to sign from a given transaction
+* `fragment-id` get the **Fragment ID** from a transaction in *sealed* state
 * `to-message` to get the hexadecimal encoded message, ready to send with `cli rest message`
 
+**DEPRECATED**:
+* `id` get the data to sign from a given transaction (use `data-for-witness` instead)
 
 # Examples
 
@@ -41,9 +44,9 @@ Let's use the following utxo as input and transfer 50 lovelaces to the destinati
 
 | Field                     | Value        |
 | ------------------------- |:------------:|
-| UTXO's transaction ID     | 55762218e5737603e6d27d36c8aacf8fcd16406e820361a8ac65c7dc663f6d1c| 
-| UTXO's output index       | 0     | 
-| associated address        |  ca1q09u0nxmnfg7af8ycuygx57p5xgzmnmgtaeer9xun7hly6mlgt3pjyknplu    | 
+| UTXO's transaction ID     | 55762218e5737603e6d27d36c8aacf8fcd16406e820361a8ac65c7dc663f6d1c|
+| UTXO's output index       | 0     |
+| associated address        |  ca1q09u0nxmnfg7af8ycuygx57p5xgzmnmgtaeer9xun7hly6mlgt3pjyknplu    |
 | associated value          | 100             |
 
 ## Destination address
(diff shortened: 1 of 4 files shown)


NicolasDP:
if you could update the doc too that would be great.

jcli/src/jcli_app/transaction/mod.rs
```diff
@@ -158,11 +164,22 @@ impl Transaction {
 }
 
 fn display_id(common: common::CommonTransaction) -> Result<(), Error> {
+    eprintln!("DEPRECATED: use 'data-for-witness' instead\n");
```
NicolasDP: good idea printing on stderr, the message won't be caught by scripts.

however, no need for the extra `\n`.

```suggestion
    eprintln!("DEPRECATED: use 'data-for-witness' instead");
```

jcli/src/jcli_app/transaction/mod.rs
```diff
@@ -44,9 +44,13 @@ pub enum Transaction {
     Finalize(finalize::Finalize),
     /// Finalize the transaction
     Seal(seal::Seal),
-    /// get the Transaction ID from the given transaction
-    /// (if the transaction is edited, the returned value will change)
+    /// get the Fragment ID from the given 'sealed' transaction
+    FragmentId(common::CommonTransaction),
```
NicolasDP: This is where things are interesting, this will be also the `Id` that can be use to lookup the status of a transaction within the blockchain. May it be a certificate or a basic transaction.

jcli/src/jcli_app/transaction/mod.rs
rinor: `stderr` was suggested by @vincenthz https://github.com/input-output-hk/jormungandr/pull/1069#issuecomment-549266328 :)

newline removed.

doc/jcli/transaction.md
```diff
@@ -26,9 +26,12 @@ There are also functions to help decode and display the
 content information of a transaction:
 
 * `info`
-* `id` to get the **Transaction ID** of the transaction
+* `data-for-witness` get the data to sign from a given transaction
+* `fragment-id` get the **Fragment ID** from a transaction in *sealed* state
 * `to-message` to get the hexadecimal encoded message, ready to send with `cli rest message`
 
+**DEPRECATED**:
```
NicolasDP: Well no. It gives you the data to for witness

rinor: My bad, I ment **TO BE DEPRECATED**, or should I remove that notice completely?

NicolasDP: no **DEPRECATED** is fine. I wrote that in the middle of the night and miss the line I wanted to comment to. It was the line below.

doc/jcli/transaction.md
```diff
@@ -26,9 +26,12 @@ There are also functions to help decode and display the
...
+* `id` to get the Transaction ID of the transaction (use `data-for-witness` instead)
```
NicolasDP: Here you say that `id` gives the `Transaction ID`. While actually it gives the `data to sign`, the `data-for-witness`.

```suggestion
* `id` to get the Data to sign the transaction (use `data-for-witness` instead)
```

rinor: updated with the same info as `data-for-witness`

PR #1078, opened 2019-11-05
Reimplemented chain pull

mzabaluev:
If a tree falls in the asynchronous forest and there is no one polling to observe it, did it really fell?

### DIFF
--- a/jormungandr/src/blockchain/chain.rs
+++ b/jormungandr/src/blockchain/chain.rs
@@ -49,11 +49,13 @@ See Internal documentation for more details: doc/internal_design.md
 [`Branch`]: ./struct.Branch.html
 */
 
+use super::{branch::Branches, reference_cache::RefCache};
 use crate::{
     blockcfg::{
-        Block, Block0Error, Epoch, Header, HeaderHash, Leadership, Ledger, LedgerParameters,
+        Block, Block0Error, BlockDate, ChainLength, Epoch, Header, HeaderHash, Leadership, Ledger,
+        LedgerParameters,
     },
-    blockchain::{Branch, Branches, Checkpoints, Multiverse, Ref, RefCache, Storage},
+    blockchain::{Branch, Checkpoints, Multiverse, Ref, Storage},
     start_up::NodeStorage,
 };
 use chain_impl_mockchain::{leadership::Verification, ledger};
@@ -61,6 +63,9 @@ use chain_storage::error::Error as StorageError;
 use chain_time::TimeFrame;
 use std::{convert::Infallible, sync::Arc, time::Duration};
 use tokio::prelude::*;
+
+// derive
+use thiserror::Error;
 
 error_chain! {
     foreign_links {
@@ -82,9 +87,14 @@ error_chain! {
             description("Block0 is not yet in the storage")
         }
 
-        MissingParentBlockFromStorage(header: Header) {
+        MissingParentBlock(hash: HeaderHash) {
             description("missing a parent block from the storage"),
-            display("Missing a block from the storage. The process was recovering the blockchain and the block parent block '{}' was not already in the cache", header.block_parent_hash()),
+            display(
+                "Missing a block from the storage. The node was recovering \
+                 the blockchain and the parent block '{}' was not \
+                 already stored",
+                hash,
+            ),
         }
 
         NoTag (tag: String) {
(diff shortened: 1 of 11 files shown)


jormungandr/src/blockchain/chain.rs
```diff
@@ -49,11 +49,13 @@ See Internal documentation for more details: doc/internal_design.md
 [`Branch`]: ./struct.Branch.html
 */
 
+use super::{branch::Branches, reference_cache::RefCache};
```
NicolasDP: why did these include moves?

jormungandr/src/blockchain/reference_cache.rs
```diff
@@ -124,7 +124,7 @@ impl RefCacheData {
     pub fn poll_purge(&mut self) -> Poll<(), timer::Error> {
         loop {
             match self.expirations.poll()? {
-                Async::NotReady => return Ok(Async::Ready(())),
+                Async::NotReady => return Ok(Async::NotReady),
```
NicolasDP: no it is a `Ready(())` that you want to return, the reason being that otherwise the cache will never free from the purging and will wait to have more expired values.

In the case of **purging** the expiration queue you want to unblock the purging task as soon as you have had all the elements who needs to be removed. `NotReady` is returned as soon as there are still elements in the expiration poll that haven't expired. So no, the previous value was correct.

jormungandr/src/network/mod.rs
```diff
@@ -16,12 +16,7 @@ mod subscription;
...
 
 mod buffer_sizes {
-    // Size of chunks to split processing of chain pull streams.
-    // Apart from sizing data chunks for intercom messages, it also
-    // determines how many blocks will be requested per each GetBlocks request
-    // distributed between different peers.
-    //
-    // This may need to be made into a configuration parameter.
+    // Size of buffer for processing of header push/pull streams.
     pub const CHAIN_PULL: usize = 32;
```
NicolasDP: don't you want to have this the same value as the one in the `src/blockchain/mod.rs` ?

jormungandr/src/blockchain/reference_cache.rs
mzabaluev: It's a big no-no to return `Ready` when all sources (in this context) polled `NotReady`.
This way, the re-poll of this task has been scheduled in the reactor to when the next expiration comes up, but returning `Ready` would leed to it being polled again to busy loop. I had the same problem with client code.

jormungandr/src/network/mod.rs
mzabaluev: Not necessarily; this is a (minimal) buffer size for the MPSC channel, and the new one determines how many headers the blockchain task processes before it drops out. This should be implemented better and take into account the epoch and leadership schedule, so that we first request all blocks till the end of the next epoch, pause the stream, then when the blocks are received, somehow get a wakeup in the task that was waiting for them and process more headers with full validation.

jormungandr/src/blockchain/reference_cache.rs
NicolasDP: Theoretically I would say yes. But the way it is used in this case is not the same. We are polling all elements that are ready to be removed. Not polling the element that we want to keep. And this is what this is doing here. If you return `NotReady` the task will wait for **all** the element in the expirations to be polled until there is none. The issue goes back to #703 where I have made the same assumption you are making about how it works here.

jormungandr/src/blockchain/chain.rs
mzabaluev: I removed re-exports from the top of `mod blockchain` that were only used internally in the module, and minimized exports in general. This allows to catch dead code lints in more places.

jormungandr/src/network/mod.rs
NicolasDP: ok

jormungandr/src/blockchain/reference_cache.rs
mzabaluev: I thought the `DelayQueue` works just like all the other async objects: if the poll returns `NotReady`, it has nothing to yield now, but has scheduled a wakeup for when it might have more. This is what the docs on the `DelayQueue` say. So re-polling before the scheduled time makes no sense. There is one edge case when a new item is inserted and its expiration time is shorter than a previously scheduled wakeup, but I'd suppose [it is handled](https://docs.rs/tokio-timer/0.2.11/src/tokio_timer/delay_queue.rs.html#411) in the implementation?

NicolasDP: yes but that's not how it is used, we poll every `N` seconds to `purge` the expired objects. This is needed because we cannot keep hold of the object all the time. Otherwise we cannot add new entries in the delay queue. Going back to the deadlock fixed by #703

mzabaluev: One thing is that we need to sit down and convert all these tasks to run properly async, without `wait` calls in the message handlers. Then these polls will work as they are intended to and returning `Ready()` on non-ready queue will busyloop. Even now, I believe polling every N seconds will yield the expired entry; where did the deadlock come from?

mzabaluev: Removed this change for now.

Issue #1090, opened 2019-11-08
Mined transactions are not propagated to other nodes - RC5, local cluster with 2 nodes


dorin100:
**Describe the bug**
Mined transactions are not propagated to other nodes.

**Mandatory Information**
```
jcli 0.7.0-rc5 (HEAD-d4ac194b, release, windows [x86_64]) - [rustc 1.39.0 (4560ea788 2019-11-04)]
jormungandr 0.7.0-rc5 (HEAD-d4ac194b, release, windows [x86_64]) - [rustc 1.39.0 (4560ea788 2019-11-04)]
```

**To Reproduce**
1. Start node1 --> `jormungandr ---genesis-block block-0.bin --config node_config.yaml --secret node_secret.yaml`
2. wait for node1 to create some blocks
3. Start node2 --> `jormungandr ---config node_config.yaml --genesis-block-hash 55dedb3c348adb8b7ff3ef1a9ae3a114c8609f338c42d5c44ee806cd290a85c8`
4. Send some funds from faucet to a new account --> `bash send-money.sh ca1sku4yc68q9cuwlsg296l6yrk6ccth2pzqu26r4s5r8enymyfeyptgl5vxk4 28000000000000 9001 ed25519e_sk18r7nd20gaxjfgmahyqu2vngv98leqefcdcft2nevcakpf999spx55t4ph8ryqslp6ac7uryekjcqsqzl63rjpmh0k92dvquesweq38cc8a0wc`
5. Wait for 1 new block to be created

- on node1  (9001 - stake pool):
```
$ jcli rest v0 account get ca1sku4yc68q9cuwlsg296l6yrk6ccth2pzqu26r4s5r8enymyfeyptgl5vxk4 --host "http://127.0.0.1:9001/api"
---
counter: 0
delegation:
  pools: []
value: 28000000000000
```
```
$ jcli rest v0 message logs --host "http://127.0.0.1:9001/api"
---
- fragment_id: 4dafd8a161052003376babc518971443f124de240a7ade5a92817d1e74103ab2
  last_updated_at: "2019-11-08T08:54:10.005584+00:00"
  received_at: "2019-11-08T08:54:08.286583300+00:00"
  received_from: Rest
  status:
    InABlock:
      block: 1f498921bfb2656773d708897763797dcd46dd06e457fe14a2d0df6b996a0a5a
      date: "5504.25"
```
```
$ jcli rest v0 network stats  get --host "http://127.0.0.1:9001/api"
---
[...]

PR #1095, opened 2019-11-08
Revert "Flip the switch on PushHeaders"

NicolasDP:
This reverts commit 36be4687781c7f8bcceb6b9bd65c322845b9a4e8.

temporary revert a feature that is creating #1090


mzabaluev:
Not really a breaking change, both protocol requests are still supported.

NicolasDP:
looks like I need to revert more in order to have something that actually sync

PR #1097, opened 2019-11-08
breaking change: add pool permission + operators

vincenthz:
Add pool operators where the owners can delegate administration operations over pool parameters (e.g. updating the pool cryptographic keys).

no permission are defined, so for now the operator cannot do anything.


jcli/src/jcli_app/certificate/new_stake_pool_registration.rs
```diff
@@ -30,6 +30,14 @@ pub struct StakePoolRegistration {
         required = true
     )]
     pub owners: Vec<PublicKey<Ed25519>>,
+    /// public key of the operators(s)
+    #[structopt(
+        long = "operators",
```
rinor: Not that important, but if possible can we keep also this in *singular* like we have it for *owner*.

vincenthz: good catch

PR #1114, opened 2019-11-12
Leeway for schedule to happen

NicolasDP:
allow up to 50milli seconds early wake at the leader schedule


rinor:
What about **late** case scenario?

NicolasDP:
There is no late case scenario here. If the `Delay` is accurate up to milliseconds. The smallest duration of a time slot is one second.

PR #1126, opened 2019-11-13
make the blockchain::process fully async

NicolasDP:
this will prevent the node to crash if there is an error with normal chain processing operations (rejecting a block is normal operation).

### DIFF
--- a/jormungandr/src/blockchain/process.rs
+++ b/jormungandr/src/blockchain/process.rs
@@ -34,49 +34,50 @@ pub fn handle_input(
     explorer_msg_box: Option<&mut MessageBox<ExplorerMsg>>,
     input: Input<BlockMsg>,
 ) -> impl Future<Item = (), Error = ()> {
-    future::result(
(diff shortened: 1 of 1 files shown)


jormungandr/src/blockchain/process.rs
```diff
@@ -83,40 +83,61 @@ pub fn run_handle_input(
...
 
-            if let Some(msg_box) = explorer_msg_box {
-                msg_box
-                    .try_send(ExplorerMsg::NewBlock(block))
-                    .unwrap_or_else(|err| error!(logger, "cannot add block to explorer: {}", err));
-            };
+            let notify_explorer = propagate_new_block.and_then(move |new_block_ref| {
+                if let Some(msg_box) = explorer_msg_box {
+                    msg_box
+                        .try_send(ExplorerMsg::NewBlock(block))
```
mzabaluev: This could be a `.send` as long as we're creating a future here anyway.

PR #1136, opened 2019-11-15
Implement BIP32 key derivation

### LINKED ISSUE #450: Add key derivation capability in the `jcli key` for bip32 key types
if this is a private key, we need to be able to do soft and hard derivation. If this is a public key only soft derivation

## soft derivation

```
cat secret-bip32.key | jcli key derive 0
cat secret-bip32.key | jcli key to-public | jcli key derive 0
```

an interesting property is that soft deriving from the private key or the public key should always give the same derived public key

```
cat secret-bip32.key | jcli key derive 0 > pubkey1.
cat secret-bip32.key | jcli key to-public | jcli key derive 0 > pubkey2
```

Here `pubkey1 == pubkey2`

## hard derivation:

```
cat secret-bip32.key | jcli key derive 2147483648
cat secret-bip32.key | jcli key to-public | jcli key derive 2147483648 # should fail
```

eugene-babichenko:
Resolves #450

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1431,6 +1431,7 @@ dependencies = [
  "chain-time 0.1.0",
  "clap 2.33.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "custom_error 1.7.1 (registry+https://github.com/rust-lang/crates.io-index)",
+ "ed25519-bip32 0.1.5 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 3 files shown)


eugene-babichenko:
@NicolasDP Now `ed25519-bip32` is imported into `jcli`. Should I leave it as is, or is it better to make a wrapper for derivation in `chain-crypto`?

NicolasDP:
No, no need. It's good like this.

ilap:
Good news. Are you going to implement some master key generation from  entropy/seed?

Issue #913, opened 2019-10-09
Convert REST server to a Tokio service


mzabaluev:
Currently last thing the `start` function does in `main.rs` is blocking in wait for the REST server to terminate. This prevents systematic handling of asynchronous task errors and termination, as attested in #911 (have to trigger a panic and its associated unsafe unwrap handler to exit the process).

The REST service should be converted to plug into the task infrastructure just like the rest of them.

PR #1141, opened 2019-11-17
Shutdown node when any service terminates

CodeSandwich:
Previously it was just REST service, now any error or shutdown triggers node shutdown. Adds logging of Tokio task panics. Part of https://github.com/input-output-hk/jormungandr/issues/913


jormungandr/src/utils/task.rs
```diff
@@ -345,3 +340,51 @@ impl<Msg> TaskMessageBox<Msg> {
...
+        let _ = self.sender.send(());
+    }
+}
+
+fn log_service_panic(logger: &Logger, error: &dyn Any) {
+    let reason_logger = error
+        .downcast_ref::<&str>()
+        .map(|reason| logger.new(o!("reason" => *reason)));
+    let panic_logger = reason_logger.as_ref().unwrap_or(logger);
+    error!(panic_logger, "Task panicked");
```
NicolasDP:
```suggestion
    crit!(panic_logger, "Task panicked");
```

jormungandr/src/utils/task.rs
```diff
@@ -100,25 +100,29 @@ impl Services {
...
-        let handler = thread::Builder::new()
+        let finish_notifier = self.finish_listener.notifier();
+        thread::Builder::new()
             .name(name.to_owned())
             // .stack_size(2 * 1024 * 1024)
             .spawn(move || {
-                info!(thread_service_info.logger, "starting task");
-                f(thread_service_info)
+                info!(logger, "starting task");
+                if let Err(error) = catch_unwind(AssertUnwindSafe(|| f(thread_service_info))) {
```
mzabaluev: That's another `AssertUnwindSafe` I'm not sure at all is safe. Have you tried to set the bound on `F` so that `catch_unwind` does not require this wrapper? I tried to do it previously, but there was a lot to push through that closure boundary that did not confrom, so I gave up.

CodeSandwich: @mzabaluev So I did a little research.

I've tried forcing tasks to be `UnwindSafe` and it's not feasible. Huge part of `futures` and `tokio` infrastructure is `!UnwindSafe`. E.g. `tokio::Lock` and `futures::mpsc`, which aren't realistically replaceable. `async_std` provides `UnwindSafe` `Mutex`, which may be an alternative to `tokio::Lock`, but its implementation of `channel` is still `!UnwindSafe`.

But this doesn't seem to be a problem at all in context of thread executor. The unwind boundary makes sense only in context of current thread. If thread is created and its content panics, its closure doesn't have to be unwind safe, thread boundary is much stronger and can replace unwind boundary. E.g. `std::thread::spawn` doesn't need closure to be `UnwindSafe`, the fact that it's sent to another thread is enough even though thread technically catches panics. Here we can guarantee that nothing not unwind safe is used after panic is caught, so it's perfectly safe.

It's the same in case of tokio tasks. Whenever there's a panic, we catch it and finish execution, but because it's the only future on executor, it stops right away guaranteeing to never touch `!UnwindSafe` data anymore.

NicolasDP: We should not need to run `thread`s in the module. The last one still using a non `tokio` task is `client.rs` I believe. Better do the little leap and finish the client.rs work to make it asynchronous. That way we can remove this function and all the unsafe. Also this would simplify this module and the future work on porting to standard future APIs.

mzabaluev: What would be the plan to catch panics once everything is running in an executor's thread pool?
We want to exit as soon as it happens, but the behavior without `.catch_unwind` is that the pool recovers from loss of a thread dropping the parked task, or is it not?

mzabaluev: Or did you mean we do run tasks as futures wrapped in the `.catch_unwind()` combinator, not in separate threads?

NicolasDP: Yes, last one. It's time to clean up the last module and to make it async too, in a future task.

CodeSandwich: Tokio runtime always catches panics and there's nothing we can do about it. What we can do (and I propose it here) is shut down everything whenever any future stops for whatever reason. `.catch_unwind` is used only to log the panic for the user.

Right now we have 2 tasks running on thread executor: client and REST.

NicolasDP:
Merging this for the short term win. still need a proper handling of #913 and to clean up the client task to fully async #1160

rinor:
This if failing for me to compile:
```
error: reached the type-length limit while instantiating `<futures::Map<futures::MapErr<fu...ask::ServiceFinishNotifier], ()>`
   --> /home/rinor/.cargo/registry/src/github.com-1ecc6299db9ec823/futures-0.1.29/src/future/mod.rs:411:5
    |
411 | /     fn map_err<F, E>(self, f: F) -> MapErr<Self, F>
412 | |         where F: FnOnce(Self::Error) -> E,
413 | |               Self: Sized,
414 | |     {
415 | |         assert_future::<Self::Item, E, _>(map_err::new(self, f))
416 | |     }
    | |_____^
    |
    = note: consider adding a `#![type_length_limit="1140126"]` attribute to your crate

error: aborting due to previous error

error: could not compile `jormungandr`.
```

Issue #1020, opened 2019-10-24
Need a CI build container with protoc installed


mzabaluev:
Our builds depend on a `protoc` binary to generate code from the protobuf/gRPC .proto file. Pre-built protoc binaries are bundled with `prost` and these get used in builds that depend on that crate. This is far from ideal, and furthermore, #1015 is about to introduce a dependency on another gRPC stack for the tests (arguably a good thing for dissimilar interop testing), which does not bundle `protoc`, meaning that it's not possible to generate code in the build script.

The solution is to create a custom Docker image for CircleCI which will have protoc installed on top of the rust base image.

PR #1145, opened 2019-11-18
Custom Dockerfile for CircleCI

eugene-babichenko:
**TBD**

- [x] Upload docker image to a registry (**who to ask?**: @NicolasDP or @mzabaluev )
- [ ] Add links to the uploaded image in CircleCI config
- [ ] Remove all of the tooling installation from the CircleCI config (already done in Dockerfile)
- [ ] Uncomment protoc related lines in https://github.com/input-output-hk/jormungandr/blob/master/jormungandr-integration-tests/build.rs

relates to #1020

### DIFF
--- a/docker/ci/Dockerfile.nightly
+++ b/docker/ci/Dockerfile.nightly
@@ -1,0 +1,12 @@
+FROM rustlang/rust:nightly
+
+LABEL org.label-schema.schema-version = "1.0"
+LABEL org.label-schema.vendor = "Input Output HK Limited"
+LABEL org.label-schema.description = "CI image for Jormungandr (nightly Rust)"
+
+RUN apt-get update && \
+    apt-get install -y protobuf-compiler && \
+    apt-get clean && \
+    cargo install cargo-audit && \
+    rm -rf $HOME/.cargo/git && \
+    rm -rf $HOME/.cargo/registry
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -33,6 +33,10 @@ install:
   - curl -sSf -o rustup-init.exe https://win.rustup.rs/
   - rustup-init.exe -y --default-host %TARGET% --default-toolchain %RUST_VERSION%
   - set PATH=%PATH%;C:\Users\appveyor\.cargo\bin
+  - curl --silent --output protobuf.zip https://github.com/protocolbuffers/protobuf/releases/download/v3.10.1/protoc-3.10.1-win64.zip
+  - ps: Expand-Archive -Path protobuf.zip -DestinationPath protobuf -Force
+  - del protobuf.zip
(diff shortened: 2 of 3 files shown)


eugene-babichenko:
@NicolasDP where is our docker registry?

docker/ci/Dockerfile.nightly
```diff
@@ -0,0 +1,5 @@
+FROM rust:nightly
+RUN apt-get update && \
+    apt-get install -y protobuf-compiler && \
+    rustup component add rustfmt && \
+    cargo install cargo-audit
```
mzabaluev: I'd add an apt-get cache cleanup to be really neat about the image size.

mzabaluev: Also `cargo install` pulls loads of crates to compile, these caches should better be cleaned from the image as well.

We might want to build these images in CircleCI, where apt-get and cargo caches can be reused between runs like we do to speed up `cargo test` steps.

mzabaluev:
Consider using [circleci/rust](https://hub.docker.com/r/circleci/rust) as the base for at least the stable image, as it has `rustfmt` installed and some other niceties that the CircleCI team thinks images should have in their pipelines.

docker/ci/Dockerfile.nightly
```diff
@@ -0,0 +1,12 @@
+FROM rustlang/rust:nightly
+
+LABEL org.label-schema.schema-version = "1.0"
+LABEL org.label-schema.vendor = "Input Output HK Limited"
+LABEL org.label-schema.description = "CI image for Jormungandr (nightly Rust)"
```
mzabaluev: :+1:

mzabaluev:
I guess we need to merge the branch with these Dockerfiles first, so we can set up Docker Hub builds for them and those should be set up against the master branch.

NicolasDP:
What about : https://github.com/input-output-hk/jormungandr/tree/master/docker

NicolasDP:
we have an existing docker directory and we could definitely reuse or at least concentrate docker images in the same directory, at least.

eugene-babichenko:
I'm not sure about that. At least, I don't want to mix CI and production images in the same directory.

Or, if we do, it's better to make like that (now there is just `Dockerfile` in the root directory and `alpine` with `Dockerfile` in it):

```
circleci-nightly.dockerfile
circleci-stable.dockerfile
production.dockerfile
production-alpine.dockerfile
```

eugene-babichenko:
@mzabaluev Do you mean automated builds? Then it makes sense to pin the exact versions of Rust and protoc in docker images and update them when we need to.

NicolasDP:
what's the status on this?

PR #1153, opened 2019-11-18
allow setting the number of unreachable nodes to contact for propagation

### LINKED ISSUE #1150: add config for the number of allowed non reachable nodes at a time
Add setting option in the `p2p` config to set the number of unreachable nodes to be contacted at a time.

Here we need to call the [`RandomDirectConnections::with_max_view_length`](https://docs.rs/poldercast/0.9.7/poldercast/custom_layers/struct.RandomDirectConnections.html#method.with_max_view_length)

https://github.com/input-output-hk/jormungandr/blob/74310fd946ce12710ca1f1d31f7271f3629c4727/jormungandr/src/network/p2p/topology.rs#L55-L57

NicolasDP:
fix #1150

### DIFF
--- a/jormungandr/src/network/p2p/topology.rs
+++ b/jormungandr/src/network/p2p/topology.rs
@@ -52,9 +52,13 @@ impl P2pTopology {
         topology.add_layer(Cyclon::default());
     }
 
-    pub fn set_custom_modules(&mut self) {
+    pub fn set_custom_modules(&mut self, unreachable_size: Option<usize>) {
         let mut topology = self.lock.write().unwrap();
-        topology.add_layer(custom_layers::RandomDirectConnections::default());
+        if let Some(size) = unreachable_size {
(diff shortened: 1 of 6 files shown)


jormungandr/src/network/p2p/topology.rs
```diff
@@ -52,9 +52,13 @@ impl P2pTopology {
         topology.add_layer(Cyclon::default());
     }
 
-    pub fn set_custom_modules(&mut self) {
+    pub fn set_custom_modules(&mut self, unreachable_size: Option<usize>) {
```
mzabaluev: There could be a struct so that the custom module parameter is not anonymous. Maybe just pass it a reference to `P2pConfig`?

mzabaluev:
The setting name could be more informative: the nodes are not really "unreachable", just not currently subscribed, and the meaning of "size" is unclear. Something like `max_connect_to_propagate`?

NicolasDP:
It's magic ❗️

PR #1157, opened 2019-11-19
Add warning if blockchain is not moving up

### LINKED ISSUE #863: add warning if blockchain is not moving up
add a period check process that check that the slot id of the last block is close to the latest slot id expected related to a given time clock.

In pseudo code:
```
loop {
  delay(waited-time)
  if tip.slot + waited-time-in-slot < tip.time.expected_slot(current-time) {
      warning-log
      set-status-of-node = disconnected 
  }
}
```

eugene-babichenko:
Resolves #863

### DIFF
--- a/jormungandr/src/blockchain_stuck_notifier.rs
+++ b/jormungandr/src/blockchain_stuck_notifier.rs
@@ -1,0 +1,51 @@
+use crate::{blockchain, utils::task::TokioServiceInfo};
+use chain_time::{
+    era::{EpochPosition, EpochSlotOffset},
+    Epoch,
+};
+use futures::prelude::*;
+use std::time::{Duration, SystemTime};
+use tokio::timer::Interval;
+
+pub fn check_last_block_time(
+    service_info: TokioServiceInfo,
+    blockchain_tip: blockchain::Tip,
+) -> impl Future<Item = (), Error = ()> {
+    let logger = service_info.logger().clone();
+    let err_logger = logger.clone();
+
+    // TODO set from configuration
+    let check_period = Duration::from_secs(180);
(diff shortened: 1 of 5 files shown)


eugene-babichenko:
I am not sure about how the interval should be configured here? https://github.com/input-output-hk/jormungandr/pull/1157/files#diff-9e8437da934ee12bac80dd5af0b1c857R17

jormungandr/src/blockchain_stuck_notifier.rs
```diff
@@ -0,0 +1,50 @@
...
+use tokio::timer::Interval;
+
+pub fn check_last_block_time(
+    service_info: TokioServiceInfo,
+    blockchain_tip: blockchain::Tip,
+) -> impl Future<Item = (), Error = ()> {
+    let logger = service_info.logger().clone();
+
+    // TODO set from configuration
+    let check_period = Duration::from_secs(180);
```
eugene-babichenko: I am not sure about how the interval should be configured here?

NicolasDP: Make it a parameter on the node config file.

jormungandr/src/settings/start/config.rs
```diff
@@ -30,6 +30,9 @@ pub struct Config {
     pub p2p: P2pConfig,
 
     pub explorer: Option<Explorer>,
+
+    /// the time interval with no blockchain updates after which alerts are thrown
+    pub no_blockchain_updates_warning_interval: Option<u64>,
```
NicolasDP: why not using the `Duration` type?

PR #1159, opened 2019-11-19
Ground work for header chain validation

mzabaluev:
Use the full `Blockchain` in `CandidateForest` to validate headers.
Currently this is only useful for skipping through existing blocks, as its header pre-validation API expects parent blocks of a header to be present in storage.


mzabaluev:
This also fixes a potential bug where a header pull that is completely redundant because blocks are already present would result in a wrong candidate branch landing.

Issue #1158, opened 2019-11-19
Expose the parameters of the stake pool in the graphQL data


NicolasDP:
expose the different parameter of the stake pool data if it is not already there.

* the owners (the threshold too);
* the stake pool's reward account if any;
* the stake pool's operator key if any;
* the tax and the % redistributed;
* the start validity time;

NicolasDP:
done in #1161

PR #1161, opened 2019-11-19
Expose stake pool parameters

enzoc4:
Implement #1158 

Currently all the fields mentioned in that issue are present in the registration certificate, so this adds a way to get the registration certificate from a certain Pool. If those parameters can actually change after the registration, then it may be better to add this fields directly in the pool (but keep them in the certificate)

Like this
```graphql
{
  stakePool(id: "935f26b40001c2c3b716d865df4d92487262414e26c5cabe3bb6b01bdbba088a") {
    registration {
      serial
      startValidity
      managementThreshold
      owners
      operators
      rewards {
        fixed
        ratio {
          numerator
          denominator
        }
        maxLimit
      }
      rewardAccount
    }
  }
}
```

### DIFF
--- a/jormungandr/src/explorer/graphql/mod.rs
+++ b/jormungandr/src/explorer/graphql/mod.rs
@@ -582,12 +582,32 @@ impl PoolRegistration {
     }
 
     /// Reward account
-    pub fn reward_account(&self, context: &Context) -> Option<String> {
-        // TODO: What's the best way to show this? As an Address?
+    pub fn reward_account(&self, context: &Context) -> Option<Address> {
+        use chain_impl_mockchain::transaction::AccountIdentifier;
+        let discrimination = context.db.blockchain_config.discrimination;
+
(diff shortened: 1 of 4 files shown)


jormungandr/src/explorer/graphql/mod.rs
```diff
@@ -571,8 +571,64 @@ impl PoolRegistration {
...
+            .collect()
+    }
+
+    pub fn rewards(&self) -> TaxType {
+        TaxType(self.registration.rewards)
+    }
+
+    /// Reward account
+    pub fn reward_account(&self, context: &Context) -> Option<String> {
+        // TODO: What's the best way to show this? As an Address?
```
enzoc4: @NicolasDP what do you think of this? reward_account is an `AccountIdentifier`, but that's not in the graphql API at this moment. Is `Address` a good type for something like this or should it be something more specific?

PR #1162, opened 2019-11-20
Rise type_length_limit

CodeSandwich:
Complex futures seem to exhaust default rustc limit. Fixes https://github.com/input-output-hk/jormungandr/pull/1141#issuecomment-555946899 and an existing workaround.

### DIFF
--- a/jormungandr/src/main.rs
+++ b/jormungandr/src/main.rs
@@ -1,3 +1,7 @@
+// Rustc default type_length_limit is too low for complex futures, which generate deeply nested
+// monomorphized structured with long signatures. This value is enough for current project.
+#![type_length_limit = "10000000"]
+
 extern crate actix_net;
 extern crate actix_threadpool;
 extern crate actix_web;
--- a/jormungandr/src/utils/task.rs
+++ b/jormungandr/src/utils/task.rs
@@ -187,16 +187,13 @@ impl Services {
             executor,
         };
 
-        let finish_notifier_ok = self.finish_listener.notifier();
-        let finish_notifier_err = self.finish_listener.notifier();
+        let finish_notifier = self.finish_listener.notifier();
         // This AssertUnwindSafe is safe, because after `catch_unwind`
         // its content is never used in executor thread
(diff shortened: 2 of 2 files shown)


jormungandr/src/utils/task.rs
```diff
@@ -194,9 +194,7 @@ impl Services {
         let future = AssertUnwindSafe(f(future_service_info))
             .catch_unwind()
             .map_err(move |error| log_service_panic(&logger, &*error))
-            // use of `then` triggers type-length limit compilation error
-            .map(|_| finish_notifier_ok.notify())
-            .map_err(|_| finish_notifier_err.notify());
```
NicolasDP: you can remove the unused variable `finish_notifier_err`.

jormungandr/src/main.rs
```diff
@@ -1,3 +1,5 @@
+#![type_length_limit = "11401260"]
```
NicolasDP: can you document why you picked this value, why it is there?

mzabaluev:
There may easily be some buffer realized as an array of generic values, and with a large item value the size multiplies. This would need to be reworked.

jormungandr/src/main.rs
NicolasDP: `// Rustc default type_length_limit is too low for complex futures`

still not enough. Can't we narrow it down to something sensible instead? Why do we need such a high value in first place?

CodeSandwich:
@mzabaluev This is not about size of structures in memory, but about length of names of monomorphized structures. With every `and_then`, `map` etc. wrapping the memory size is usually unchanged, because closures can be slimmed down to 0-sized, but the type name grows and is more and more nested.

jormungandr/src/main.rs
CodeSandwich: This one was suggested by compiler minus a zero in the end. The exact value doesn't matter, but it should give some headroom, so as futures get more complex, compiler keeps accepting them. I've rounded it to a less random looking one.

NicolasDP:
merging in for the sake of unblocking development.

PR #1164, opened 2019-11-20
allow adding `public_peers` with DNS name

### LINKED ISSUE #785: Trusted peers should accept DNS name

eugene-babichenko:
Resolves #785

### DIFF
--- a/jormungandr/src/settings/start/mod.rs
+++ b/jormungandr/src/settings/start/mod.rs
@@ -1,5 +1,6 @@
 pub mod config;
 pub mod network;
+pub mod trusted_peer;
 
 use self::config::{Config, Leadership};
 pub use self::config::{Cors, Rest};
--- a/jormungandr/src/settings/start/mod.rs
+++ b/jormungandr/src/settings/start/mod.rs
@@ -6,6 +6,7 @@ pub use self::config::{Cors, Rest};
 use self::network::Protocol;
 use crate::rest::Error as RestError;
 use crate::settings::logging::{LogFormat, LogOutput, LogSettings, LogSettingsEntry};
+use crate::settings::start::config::AddressError;
 use crate::settings::{command_arguments::*, Block0Info};
 use jormungandr_lib::interfaces::Mempool;
 use slog::{FilterLevel, Logger};
@@ -233,7 +234,17 @@ fn generate_network(
             .clone()
             .unwrap_or(vec![])
             .into_iter()
-            .map(Into::into)
+            .filter_map(|tp| {
+                tp.address.to_addresses().ok().map(|addrs| {
+                    addrs.into_iter().map(move |address| {
+                        crate::settings::start::network::TrustedPeer {
+                            id: tp.id.clone(),
+                            address: address.0,
+                        }
+                    })
+                })
+            })
+            .flatten()
             .collect(),
         protocol: Protocol::Grpc,
         policy: p2p.policy.clone(),
--- a/jormungandr/src/settings/start/mod.rs
+++ b/jormungandr/src/settings/start/mod.rs
@@ -6,7 +6,6 @@ pub use self::config::{Cors, Rest};
 use self::network::Protocol;
 use crate::rest::Error as RestError;
 use crate::settings::logging::{LogFormat, LogOutput, LogSettings, LogSettingsEntry};
-use crate::settings::start::config::AddressError;
 use crate::settings::{command_arguments::*, Block0Info};
 use jormungandr_lib::interfaces::Mempool;
 use slog::{FilterLevel, Logger};
@@ -237,16 +236,12 @@ fn generate_network(
             .unwrap_or(vec![])
             .into_iter()
             .filter_map(|tp| {
-                tp.address
-                    .to_addresses()
-                    .ok()
-                    .and_then(std::convert::identity)
-                    .map(|addrs| {
-                        addrs.into_iter().map(move |addr| TrustedPeer {
-                            id: tp.id.clone(),
-                            address: addr.0,
-                        })
+                tp.address.to_addresses().ok().map(|addrs| {
+                    addrs.into_iter().map(move |addr| TrustedPeer {
+                        id: tp.id.clone(),
+                        address: addr.0,
                     })
+                })
             })
             .flatten()
             .collect(),
--- a/jormungandr/src/settings/start/mod.rs
+++ b/jormungandr/src/settings/start/mod.rs
@@ -182,6 +182,8 @@ fn generate_network(
     command_arguments: &StartArguments,
     config: &Option<Config>,
 ) -> Result<network::Configuration, Error> {
+    use crate::settings::start::network::TrustedPeer;
+
     let mut p2p = if let Some(cfg) = config {
         cfg.p2p.clone()
     } else {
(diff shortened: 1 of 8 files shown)


NicolasDP:
We cannot use that changes yet as we need to remove the need to use the trusted peers node id first.

jormungandr/src/settings/start/trusted_peer.rs
```diff
@@ -0,0 +1,146 @@
...
+            .collect();
+
+        Ok(addresses)
+    }
+}
+
+impl std::str::FromStr for TrustedPeer {
+    type Err = String;
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let mut split = s.split('@');
```
mzabaluev: I'm not sure we need to make an official microformat for this.

jormungandr/src/settings/start/mod.rs
```diff
@@ -233,7 +236,15 @@ fn generate_network(
             .clone()
             .unwrap_or(vec![])
             .into_iter()
-            .map(Into::into)
+            .filter_map(|tp| {
+                tp.address.to_addresses().ok().map(|addrs| {
+                    addrs.into_iter().map(move |addr| TrustedPeer {
+                        id: tp.id.clone(),
```
mzabaluev: Umm, do we have provisions elsewhere for the duplicated peer entries?
These are going to overwrite each other in `PeerMap`. To do it properly, there should be a "happy eyeballs" style async connection routine for a single peer, racing up to two connection trains – one going through a shuffled list of IPv4 addresses, the other through IPv6 – but only to the point where one connects on TCP, then the other needs to be discarded and the protocol handshake with subscriptions should proceed on the winning connection.

Provisions may need to be made in network-grpc to enable two-stage connections.

mzabaluev: OK, maybe happy eyeballs is a little overkill as there is not much interactive difference in how many milliseconds it takes to establish a connection. So shuffle the list of all addresses for a single peer and go through them with TCP connection attempts until one succeeds, that would do it.

jormungandr/src/settings/start/trusted_peer.rs
```diff
@@ -0,0 +1,146 @@
...
+            Some(AddrComponent::DNS4(fqdn)) => format!("{}:{}", fqdn, port)
+                .to_socket_addrs()
+                .map_err(|e| AddressError::DnsLookupError { source: e })?
+                .into_iter()
+                .filter_map(|r| match r {
+                    SocketAddr::V4(addr) => Some(AddrComponent::IP4(*addr.ip())),
+                    _ => None,
+                })
+                .collect(),
+            Some(AddrComponent::DNS6(fqdn)) => format!("{}:{}", fqdn, port)
```
mzabaluev: It's... weird that multiaddr does not support just `dns/` which would resolve both A and AAAA records. Which seems to be what's going to happen under the hood anyway, except that you discard the addresses of the wrong type.

jormungandr/src/settings/start/trusted_peer.rs
```diff
@@ -0,0 +1,146 @@
...
+
+custom_error! {pub AddressError
+    DnsLookupError { source: std::io::Error } = "failed to resolve DNS name {source}",
+    NoPortSpecified = "no TCP port specified",
+    NoAppropriateDNSFound = "the address was resolved, but it doesn't provide IPv4 or IPv6 addresses",
+    UnsupportedProtocol = "the provided protocol is unsupported, please use one of ip4/ip6/dns4/dns6",
+}
+
+impl TrustedAddress {
+    pub fn to_addresses(&self) -> Result<Vec<Address>, AddressError> {
```
mzabaluev: Please add a prominent (doc) comment notice that a blocking network resolver roundtrip happens here, so that we remember to avoid calling this in async code.

jormungandr/src/settings/start/mod.rs
mzabaluev: Even better perhaps, we could let `hyper` deal with DNS addresses because we give it an URL to connect to.

jormungandr/src/settings/start/trusted_peer.rs
mzabaluev: Converting from multiaddr to an FQDN URL for `hyper`, as per https://github.com/input-output-hk/jormungandr/pull/1164#discussion_r349581675, would erase the IP address preference, so I'm not sure what would be more proper here.

mzabaluev: Theoretically, `HttpConnector::set_local_address` [in hyper](https://docs.rs/hyper/0.12.35/hyper/client/connect/struct.HttpConnector.html#method.set_local_address) can be used to restrict connections to IPv4 or IPv6, but I'm not sure it works with the internal resolver.

mzabaluev:
> We cannot use that changes yet as we need to remove the need to use the trusted peers node id first.

We still need _some_ way to identify a node with multiple IP addresses, including the corner cases when we receive its information from gossip with a different, maybe partially overlapping, set of addresses. Any of the old IP addresses may be taken over by other, logically different nodes, so I'm not sure intersecting on any IP address would give you a moral right to put it under the same node entry and reuse the connection.

jormungandr/src/settings/start/mod.rs
mark-stopka: How much headache would it be to implement DNSSEC signature validation within Jormungandr and WARN if provided FQDN is not signed? Seems (I checked just three) Rust libraries do not support DNSSEC validation, does `hyper` rely on OS DNS resolver or a Rust library outside of the OS?

mark-stopka:
> > We cannot use that changes yet as we need to remove the need to use the trusted peers node id first.
> 
> We still need _some_ way to identify a node with multiple IP addresses, including the corner cases when we receive its information from gossip with a different, maybe partially overlapping, set of addresses. Any of the old IP addresses may be taken over by other, logically different nodes, so I'm not sure intersecting on any IP address would give you a moral right to put it under the same node entry and reuse the connection.

That's what was good on the initial private key, public key design, now when node_id is just a random string, anybody can replicate someone else's node_id.

jormungandr/src/settings/start/mod.rs
mzabaluev: `hyper` supports custom resolvers, and previous commits on this branch used `trust-dns-resolver` which supports DNSSec. @eugene-babichenko @NicolasDP what do you think of re-integrating it back, either in this form or hooking into `hyper`?

eugene-babichenko: > Umm, do we have provisions elsewhere for the duplicated peer entries?

No, and actually this is my main concern about this pull request which I discussed with @NicolasDP. The **correct** approach is testing IPs until we find a reachable one and then just go for that address. But we must do it right after we got the whole list of addresses, because we also need to feed the node data to `poldercast`. We can rework poldercast crate a bit though to make it able to accept DNS instead of bare IPs.

> Even better perhaps, we could let `hyper` deal with DNS addresses because we give it an URL to connect to.

If the whole network stuff including gossiping is done via `hyper` (that's what I can see from the code), I think it's totally fine.

> what do you think of re-integrating it back, either in this form or hooking into `hyper`?

The whole point of the last commit was to remove an additional dependency, but I feel like we will return if want to support DNSSEC. I'm more into hooking that stuff into `hyper`, just because it sounds way more elegant.

_To sum everything up_, my approach would be to allow using DNS just everywhere in the code. Let's say, make it a first-class citizen along with the IPs. But this may require some modification of external dependencies, specifically `poldercast`.

jormungandr/src/settings/start/trusted_peer.rs
eugene-babichenko: Either approach requires us to make the whole system treat FQDNs the same way as IPs and probably bring everything to `network-grpc` (see https://github.com/input-output-hk/jormungandr/pull/1164#discussion_r349745898).

eugene-babichenko:
closing this in favor of #2193

PR #1169, opened 2019-11-21
Add pool retirement and pool update certificates graphql types

enzoc4:
- Add missing `PoolRetirement` and `PoolUpdate`
- Rename (breaking change) `pool` for `pools` in `StakeDelegation` and `OwnerStakeDelegation` certificates, as it returns a list (plural) now.

### DIFF
--- a/jormungandr/src/explorer/graphql/mod.rs
+++ b/jormungandr/src/explorer/graphql/mod.rs
@@ -483,7 +483,7 @@ impl StakeDelegation {
             .map(|addr| Address::from(&ExplorerAddress::New(addr)))
     }
 
-    pub fn pool(&self, context: &Context) -> Vec<Pool> {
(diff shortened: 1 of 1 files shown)


jormungandr/src/explorer/graphql/mod.rs
```diff
@@ -585,12 +585,63 @@ impl OwnerStakeDelegation {
...
+
+    pub fn start_validity(&self) -> TimeOffsetSeconds {
+        self.pool_update.start_validity.into()
+    }
+
+    // TODO: Previous keys?
+    // TODO: Updated keys?
+}
+
 // TODO can we use jormungandr-lib Certificate ?
```
enzoc4: @vincenthz I can't implement juniper's `GraphQLType` for a type in a different crate. Maybe I could try with a newtype

PR #1173, opened 2019-11-21
Transform REST server into Tokio service

CodeSandwich:
Closes https://github.com/input-output-hk/jormungandr/issues/913. Actix-web is monolithic, the only way to run it is in custom thread. This PR wraps it in a future making it controllable as a regular Tokio task:

- finish future when server stops (causing stopping the whole node)
- stop server when future is dropped (because other task has finished or crashed and whole node is being stopped)

### DIFF
--- a/jormungandr/src/rest/server/mod.rs
+++ b/jormungandr/src/rest/server/mod.rs
@@ -10,11 +10,15 @@ use actix_web::{
     actix::{Addr, System},
     server::{self, IntoHttpHandler, StopServer},
 };
+use futures::sync::oneshot::{self, Receiver};
+use futures::{Async, Future, Poll};
 use native_tls::{Identity, TlsAcceptor};
 use std::{
     fs,
     net::{SocketAddr, ToSocketAddrs},
     path::PathBuf,
+    sync::mpsc,
+    thread,
 };
 
 pub type ServerResult<T> = Result<T, Error>;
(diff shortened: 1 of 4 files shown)


jormungandr/src/rest/server/mod.rs
```diff
@@ -10,40 +10,88 @@ use actix_web::{
...
+            Err(_) => Ok(Async::Ready(())),
+            Ok(ok) => Ok(ok),
+        }
     }
+}
+
+impl Drop for Server {
+    fn drop(&mut self) {
+        self.stopper.stop();
+        let _ = self.wait();
```
NicolasDP: not sure about that though...

NicolasDP: If there is a deadlock in there this is risky.

CodeSandwich: Me neither. I've decided that it's better than having REST server randomly killed in the middle of network operation. Timeout would be the perfect solution, but at that point Tokio runtime is already down and it's not easy without its reactor.

mzabaluev: Wait, what? Why do you need to do this? It should be OK to drop "disgracefully", that's how cancellation of everything works in futures.

mzabaluev: If there is a graceful shutdown, it should be done with sending a `Shutdown` message to the task.

CodeSandwich: :man_facepalming: that's so obvious!

Issue #967, opened 2019-10-16
Node panicked - Os { code: 1455, kind: Other, message: "The paging file is too small for this operation to complete." }


dorin100:
**Describe the bug**
Node panicked after ~7 hours

**Mandatory Information**
1. used version `0.6.1` from release files (OS: windows VM)
```
jormungandr 0.6.1 (HEAD-a0de8d47, release, windows [x86_64]) - [rustc 1.38.0 (625451e37 2019-09-23)]
jcli 0.6.1 (HEAD-a0de8d47, release, windows [x86_64]) - [rustc 1.38.0 (625451e37 2019-09-23)]
```

**To Reproduce**
1. Start the node as a leader and let it run

**Expected behavior**
The node should not crash

**Additional context**
- stack trace
```
Oct 16 13:34:47.593 INFO incoming P2P connection closed, peer_addr: 96.248.209.37:63183, task: network
thread 'network0' panicked at 'called `Result::unwrap()` on an `Err` value: Os { code: 1455, kind: Other, message: "The paging file is too small for this operation to complete." }', src\libcore\result.rs:1084:5
stack backtrace:
   0:           0x7d2d93 - <unknown>
   1:           0x7d1f9a - <unknown>
   2:           0x7d8442 - <unknown>
   3:           0x7e4949 - <unknown>
   4:           0x7ef88c - <unknown>
   5:           0x7f5e1f - <unknown>
   6:           0x4713ac - <unknown>
   7:           0x97db67 - <unknown>
   8:           0x92d6a2 - <unknown>
   9:           0x8422a3 - <unknown>
  10:           0x7b6c5c - <unknown>
  11:           0x7ba4c3 - <unknown>
  12:           0x7b8628 - <unknown>
  13:           0x79c236 - <unknown>
  14:           0x79fa31 - <unknown>
  15:           0x7b61c4 - <unknown>
  16:           0x7b7072 - <unknown>
  17:           0x7e3e86 - <unknown>
  18:           0x7e548c - <unknown>
  19:     0x7ff965a537e4 - <unknown>
  20:     0x7ff96850cb81 - <unknown>
called `Result::unwrap()` on an `Err` value: Os { code: 1455, kind: Other, message: "The paging file is too small for this operation to complete." }
Oct 16 13:34:51.781 ERRO Error accepting connection: Either the application has not called WSAStartup, or WSAStartup failed. (os error 10093), scope: global
```
- node logs:
[...]

PR #1178, opened 2019-11-22
Convert the client task to async, use bounded channels in intercom

### LINKED ISSUE #1160: Convert the client task to full async and remove the header pull limit
There is a limit of maximum 2000 items to stream when serving peer's requests for missing headers or `PullBlocksToTip`. This is done to prevent indefinite blocking in the client task which still processes one request at a time even though most of its internals has been converted to futures already. We need to convert the rest of the client task to run on the thread pool executor, so that we can serve chain streams as long as they need to be without blocking other requests.

mzabaluev:
Remove the future wait calls in the client task, closing the gap between async `chain` API and async streams to and from the network task.
As a collateral fix, `intercom::ReplyStreamHandle` is converted to use a bounded sender under the hood, removing one cause of unbounded RAM usage.

Resolves #1160, may fix some remaining memory hogging issues such as #967.

### DIFF
--- a/jormungandr/src/blockchain/branch.rs
+++ b/jormungandr/src/blockchain/branch.rs
@@ -97,7 +97,7 @@ impl Branch {
         }
     }
 
-    pub fn get_ref(&self) -> impl Future<Item = Arc<Ref>, Error = Infallible> {
+    pub fn get_ref<E>(&self) -> impl Future<Item = Arc<Ref>, Error = E> {
         let mut branch = self.inner.clone();
         future::poll_fn(move || Ok(branch.poll_lock())).map(|guard| guard.reference().clone())
     }
--- a/jormungandr/src/blockchain/branch.rs
+++ b/jormungandr/src/blockchain/branch.rs
@@ -97,7 +97,7 @@ impl Branch {
         }
     }
 
-    pub fn get_ref<E>(&self) -> impl Future<Item = Arc<Ref>, Error = E> {
+    pub fn get_ref(&self) -> impl Future<Item = Arc<Ref>, Error = Infallible> {
         let mut branch = self.inner.clone();
         future::poll_fn(move || Ok(branch.poll_lock())).map(|guard| guard.reference().clone())
     }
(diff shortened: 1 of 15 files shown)


jormungandr/src/utils/async_msg.rs
```diff
@@ -61,6 +73,51 @@ impl<Msg> Sink for MessageBox<Msg> {
...
+                }
+            } else {
+                try_ready!(self
+                    .mbox
+                    .poll_complete()
+                    .map_err(|e| self.handle_mbox_error(e)));
+                return Ok(().into());
+            }
+        }
+    }
```
CodeSandwich: I wonder if this could be reduced to `mbox.clone().send().map_err(<log it>)`?

mzabaluev: It has to be a named type, but a closure in `.map_err(...)` does not currently allow that as far as I know.

CodeSandwich: I mean the whole `SendTask` structure with custom `poll` implementation. How is it better than gluing together a few built-in `Stream` wrappers?

CodeSandwich: Oh, the problem is that it must be stored in a struct field, so it needs a concrete, named type.

jormungandr/src/blockchain/branch.rs
```diff
@@ -97,7 +97,7 @@ impl Branch {
         }
     }
 
-    pub fn get_ref<E>(&self) -> impl Future<Item = Arc<Ref>, Error = E> {
+    pub fn get_ref(&self) -> impl Future<Item = Arc<Ref>, Error = Infallible> {
```
NicolasDP: Please revert. The infaillible type of for convertion. Having a type parameter to say any error type make it easier to use and avoid the pattern matching to map an unreachable.

This is an issue of the legacy future.

mzabaluev: I see, but then it makes type inference fail in some other use sites, so you have to uglify the code with type annotations. Pick your poison.

If pattern matching is too ugly, we can use this little [crate](https://crates.io/crates/unwrap-infallible).

NicolasDP: I still think this is the wrong approach. Adding a type to fix the problem of the `Future` API is the wrong approach. Especially that we will still need to remove it down the line with the port to new Future API.

CodeSandwich: `Result<_, Infallible>` is exactly the new `Future` API

PR #1196, opened 2019-11-25
remote syslog via UDP

### LINKED ISSUE #525: Logging to syslog via UDP
Would it be possible to have an option who let jormungandr log to udp localhost instead of unix sockets? For example with https://docs.rs/syslog/4.0.1/syslog/
There is an existing free open-source syslog server also available for windows hosts. It can be configured to accept/filter/reformat and also redirect messages in many different ways. (For example redirect to a global testnet syslog server)
He can also parse new messages out from a logfile written by Jormungandr but the direct, fast and cool way would be to send the messages directly to the syslog service. 
Will help then also in a multiple proxy node setup.

_Originally posted by @gufmar in https://github.com/input-output-hk/jormungandr/issues/511#issuecomment-502949717_

eugene-babichenko:
Resolves #525 

An example configuration:

```yaml
log:
  - output:
      syslogudp:
        host: 127.0.0.1:514
        hostname: jormungandr
    level: info
    format: plain
  - output: stderr
    level: debug
    format: plain
```

### DIFF
--- a/jormungandr/src/settings/logging.rs
+++ b/jormungandr/src/settings/logging.rs
@@ -49,6 +49,11 @@ pub enum LogOutput {
     Stderr,
     #[cfg(unix)]
     Syslog,
+    #[cfg(unix)]
+    SyslogUdp {
+        host: String,
+        hostname: String,
+    },
     #[cfg(feature = "systemd")]
     Journald,
     #[cfg(feature = "gelf")]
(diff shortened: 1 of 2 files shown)


eugene-babichenko:
Should this be unix-only, or Windows host should also work with this?

jormungandr/src/settings/logging.rs
```diff
@@ -146,6 +151,36 @@ impl LogOutput {
                     Err(e) => Err(Error::SyslogAccessFailed(e)),
                 }
             }
+            #[cfg(unix)]
+            LogOutput::SyslogUdp { host, hostname } => {
+                use std::net::{IpAddr, Ipv4Addr, SocketAddr};
+
+                format.require_plain()?;
+
+                let mut local_port = 1024;
```
NicolasDP: start from `30_000`

Issue #1125, opened 2019-11-13
Network timeouts on blocks (i.e. genesis block for incentivised testnet)


lewnelson:
**Describe the bug**
When fetching genesis block data on the GraphQL API if the genesis block contains hundreds of thousands of addresses on transactions then there are network performance issues.

See explorer https://shelley-testnet-explorer-staging.netlify.com/en/blocks/ last page of table for example.

*Example request which fails*

POST https://explorer.incentivized-testnet.iohkdev.io/explorer/graphql
```
{
    block(id:"7be0014ffe4e5fd74aeb2916bd85cbfdaf52c2d738e4883f4cf52f6f695a6e2d") {
        transactions {
            edges {
                node {
                    inputs {
                        amount
                        address {
                            id
                        }
                    }
                    outputs {
                        amount
                        address {
                            id
                        }
                    }
                }
            }
        }
    }
}
```

Removing the inputs and outputs from the transaction solves the issue. However for the explorer we use the transaction inputs and outputs amounts to calculate fees and output. This may be mitigated from the genesis block, but won't be as easy to do on random blocks given we don't know the size of the data.

**Mandatory Information**
Incentivised node https://explorer.incentivized-testnet.iohkdev.io/
1. `jcli --full-version` output;
2. `jormungandr --full-version` output;

**To Reproduce**
1. Go to the explorer https://shelley-testnet-explorer-staging.netlify.com/en/blocks/
2. Click on the last page for the table
3. Check the network tab to see the timeout

**Expected behavior**
Expected the network not to timeout handling the large amount of data. It may be beneficial to query for totals instead of doing the calculations from the data received.

**Additional context**

enzoc4:
Although I know of a few things that can be optimized for that query, I think in this case it's just not feasible to send that amount of data in one chunk. I think adding the aggregations on the backend is the only way this is going to work, maybe for the entire block, or for each transaction. Any thoughts @NicolasDP ?

PR #1191, opened 2019-11-25
Certificate fees

eugene-babichenko:
Integrate changes from input-output-hk/chain-libs#168 to completely resolve input-output-hk/chain-libs#9

In addition to that this PR integrates a couple of unrelated changes from `chain-libs` just to make everything compile and run.

PR #1198, opened 2019-11-25
Add input output to block and initial fees

enzoc4:
- Add total input and output aggregations to blocks. To avoid the need of iterating though all the transactions as mentioned in #1125 , although I'm not sure how much worth it is
- Implement fee getter. This are only the initial fees and doesn't track updates.


NicolasDP:
FYI @eugene-babichenko , while working on the #1191 this will need updated in the explorer too

PR #1199, opened 2019-11-26
[Tests] another attempt to stabilize tests

dkijania:


jormungandr-integration-tests/src/common/jormungandr/starter.rs
```diff
@@ -255,18 +255,20 @@ impl Starter {
...
+            self.custom_errors_found()?;
+            /* if self.stop() {
+                println!("attempt stopped due to error signal recieved");
                 logger.print_raw_log();
                 return Err(StartupError::ErrorInLogsFound {
                     log_content: file_utils::read_file(&self.config.log_file_path),
                 });
-            }
-            self.custom_errors_found()?;
+            }*/
```
mzabaluev: If this is not meant to be here eventually, just remove the whole thing. We have version control to recover old bits of code.

mzabaluev: Besides, the method names in `StartupVerification` look like actions, while they are meant to check the outcome. Rename them to something like `check_if_stopped()` for clarity?

jormungandr-integration-tests/src/common/jormungandr/starter.rs
```diff
@@ -45,8 +45,8 @@ pub enum Role {
 }
 
 pub trait StartupVerification {
-    fn stop(&self) -> bool;
-    fn success(&self) -> bool;
+    fn check_if_stopped(&self) -> bool;
+    fn check_if_succeed(&self) -> bool;
```
mzabaluev: nitpick: `if_succeeded`?

dkijania: done :)

PR #1222, opened 2019-11-27
Expose resource usage limits on UNIX systems

### LINKED ISSUE #1215: detect node's environment/system settings at boot or on demand
at startup check the node's environment/system settings.

only for linux and mac os: [getrlimit](https://linux.die.net/man/2/getrlimit)

eugene-babichenko:
Resolves #1215

### DIFF
--- a/jormungandr/src/diagnostic.rs
+++ b/jormungandr/src/diagnostic.rs
@@ -1,0 +1,50 @@
+use crate::start_up::Error;
+use std::fmt::{self, Display, Formatter};
+
+#[derive(Debug, Clone)]
+pub struct Diagnostic {
+    pub open_files_limit: u64,
+    pub cpu_usage_limit: u64,
+}
+
+impl Diagnostic {
+    #[cfg(unix)]
+    pub fn new() -> Result<Self, Error> {
+        Ok(Self {
+            open_files_limit: getrlimit(libc::RLIMIT_NOFILE)
+                .map_err(|e| Error::DiagnosticError { source: e })?,
+            cpu_usage_limit: getrlimit(libc::RLIMIT_CPU)
+                .map_err(|e| Error::DiagnosticError { source: e })?,
+        })
+    }
+}
+
+impl Display for Diagnostic {
+    fn fmt(&self, formatter: &mut Formatter<'_>) -> Result<(), fmt::Error> {
+        write!(
+            formatter,
+            "limit for open files (RLIMIT_NOFILE): {}",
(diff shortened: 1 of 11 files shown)


jormungandr/src/main.rs
```diff
@@ -470,9 +470,38 @@ fn initialize_node() -> Result<InitializedNode, start_up::Error> {
...
+        rlim_max: 0,
+    };
+
+    unsafe {
+        Errno::result(getrlimit(RLIMIT_NOFILE, &mut limits as *mut rlimit))?;
+        if limits.rlim_cur >= DEFAULT_RLIMIT_NOFILE {
+            return Ok(());
+        }
+        limits.rlim_cur = DEFAULT_RLIMIT_NOFILE;
+        Errno::result(setrlimit(RLIMIT_NOFILE, &mut limits))?;
```
mzabaluev: I think the failure should be ignored or logged as `warn!`.

eugene-babichenko:
@NicolasDP I doubt about adding this to `/node/stats` because with the current implementation these values will not be available on non-unix systems (e.g. Windows). If we are okay with optional fields in json, the I will move those values to `/node/stats`.

NicolasDP:
Some other kind of data can be exposed on windows platforms. It does not make it less useful/interesting. The value that cannot be _collected_ can be left to `None` and displayed as `N/A`.

NicolasDP:
Looking at this, missing documentation in the openapi file.

jormungandr/src/diagnostic.rs
```diff
@@ -0,0 +1,73 @@
+use crate::start_up::Error;
+use std::fmt::{self, Display, Formatter};
+
+#[cfg(unix)]
+pub type DiagnosticError = nix::Error;
+
+#[cfg(not(unix))]
+custom_error! {pub DiagnosticError {} = "diagnostic error"}
```
mzabaluev: Please convert this to a derive with `thiserror`. Let's bring this error stuff to a single good standard.

mzabaluev: It's OK to do it this way if `DiagnosticError` is meant for internal consumption, otherwise I think there might be an API mismatch problems on different platforms and so the error would better be a newtype.

eugene-babichenko: Oops, just missed that we already have it in the project. Btw, why not the `failure` crate?

NicolasDP:
@eugene-babichenko , just need a rebase 😅

PR #1228, opened 2019-11-28
Fix nightly for introduction of built-in never type

CodeSandwich:
Previously nightly was failing due to ICE, now it's because the newly introduced never type does not allow conversions that we were relying on.

### DIFF
--- a/jormungandr/src/explorer/graphql/mod.rs
+++ b/jormungandr/src/explorer/graphql/mod.rs
@@ -46,10 +46,13 @@ impl Block {
     }
 
     fn get_explorer_block(&self, db: &ExplorerDB) -> FieldResult<ExplorerBlock> {
-        db.get_block(&self.hash).wait()?.ok_or(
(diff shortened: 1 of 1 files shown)


jormungandr/src/explorer/graphql/mod.rs
```diff
@@ -46,10 +46,13 @@ impl Block {
     }
 
     fn get_explorer_block(&self, db: &ExplorerDB) -> FieldResult<ExplorerBlock> {
-        db.get_block(&self.hash).wait()?.ok_or(
-            ErrorKind::InternalError("Couldn't find block's contents in explorer".to_owned())
-                .into(),
-        )
+        db.get_block(&self.hash)
+            .wait()
+            .unwrap_or_else(|e| match e {})
```
mzabaluev: To be replaced by `.into_ok()` when that is stabilized: https://github.com/rust-lang/rust/pull/66045

PR #1229, opened 2019-11-28
Switch Circle-CI from Rust nightly to beta

CodeSandwich:
This prevents CI from collapsing whenever there's a ripple in Rust master branch, but keeps us prepared for changes surely coming up in stable

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -109,9 +109,9 @@ jobs:
           name: Test
           command: cargo test --release --offline --verbose
 
-  test_nightly:
+  test_beta:
     docker:
-      - image: rustlang/rust:nightly
+      - image: instrumentisto/rust:beta
     environment:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -109,9 +109,9 @@ jobs:
           name: Test
           command: cargo test --release --offline --verbose
 
-  test_nightly:
+  test_beta:
     docker:
-      - image: rustlang/rust:nightly
+      - image: instrumentisto/rust:beta
```
NicolasDP: why not rustlang?

CodeSandwich: `rustlang/rust` has only nightly and official `rust` has only stable. Beta is not that popular.

NicolasDP: ok thanks for providing that to us. We can merge to the IOHK's official rust docker image when available.

Issue #879, opened 2019-10-01
Move the AssertUnwindSafe at the task level when possible


NicolasDP:
Not sure what is required besides, in the worst case, shifting the `AssertUnwindSafe` wrapper to every use site for the few tasks we have, to leave audit markers for re-evaluation later if this causes problems.
But let's move it forward.

_Originally posted by @mzabaluev in https://github.com/input-output-hk/jormungandr/pull/873#discussion_r330228027_

CodeSandwich:
This was discussed in https://github.com/input-output-hk/jormungandr/pull/1141#pullrequestreview-318199670:
> Whenever there's a panic, we catch it and finish execution, but because it's the only future on executor, it stops right away guaranteeing to never touch !UnwindSafe data anymore.

@mzabaluev is it OK to close this issue or is this really a threat that needs addressing?

mzabaluev:
> because it's the only future on executor

What about tasks spawned in various places?

PR #1231, opened 2019-11-28
Task unwind safety

CodeSandwich:
Should fix https://github.com/input-output-hk/jormungandr/issues/879

Moves burden of panic handling to Tokio removing need for custom unroll catching and `AssertUnwindSafe` usage. Bonus: when task spawned inside task panics, it's logged too.

I'm really unsure how Tokio can catch all panics in tasks and NOT require futures to be `UnwindSafe`, but I trust them that it's somehow a sound design.


mzabaluev:
I haven't given it much thought, but my vague idea is that boxed futures unwind inside the all stack under `catch_unwind`, and Tokio passes a minimal, unwind safe context to observe the unwind.

This also can be as shabby as possible inside while working most of the time, because the real solution would only come out of the port to tokio 0.2.

CodeSandwich:
An answer: https://github.com/tokio-rs/tokio/issues/1786#issuecomment-559591967

PR #1259, opened 2019-12-03
change how to set the reward account in the jcli command line parameter

NicolasDP:
set the reward account as a command line parameter, not a positional parameter. This will prevent issues when setting the output file if needed. and makes the CLI more verbose/easy to check


feqifei:
do you think it's possible to add the information related to the reward account into the results of jcli rest v0 stake-pool get command?

Issue #1248, opened 2019-12-02
REST: update /api/v0/settings


NicolasDP:
- [x] expose the per certificate settings;
- [ ] expose the reward settings
- [ ] expose the treasury settings

PR #1268, opened 2019-12-03
Rest tax and reward settings

CodeSandwich:
Fixes https://github.com/input-output-hk/jormungandr/issues/1248

Adds treasury tax and reward calculated for the next epoch. I've figured out, that the front end and the users are interested in concrete value and if we provided actual settings, they would have to mirror the non-trivial calculation logic on their side.

If all the little knobs are interesting for users, I can add them as well, abandoned half-finished work with this approach is on another branch, I can rebase and finish it.


CodeSandwich:
What are these settings going to be used for?

NicolasDP:
They are going to be used by the different tools building on the node's API to plan the reward mechanism, They can see what are the current settings and plan how to set their own.

NicolasDP:
I think we can close that one now.

PR #1278, opened 2019-12-04
incentive model changes

JaredCorduan:
This PR is code from @kevinhammond that updates the incentives model. It depends on https://github.com/input-output-hk/chain-libs/pull/203

PR #1281, opened 2019-12-05
incentive model changes

JaredCorduan:
This PR is code from @kevinhammond that updates the incentives model. It depends on input-output-hk/chain-libs#203 (I did not intend to close #1278)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -459,6 +459,8 @@ dependencies = [
  "quickcheck 0.8.5 (registry+https://github.com/rust-lang/crates.io-index)",
  "rand_core 0.3.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "rand_os 0.1.3 (registry+https://github.com/rust-lang/crates.io-index)",
+ "serde 1.0.103 (registry+https://github.com/rust-lang/crates.io-index)",
+ "serde_derive 1.0.103 (registry+https://github.com/rust-lang/crates.io-index)",
  "sparse-array 0.1.0",
  "strum 0.15.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "strum_macros 0.15.0 (registry+https://github.com/rust-lang/crates.io-index)",
--- a/jormungandr-lib/src/interfaces/block0_configuration/initial_config.rs
+++ b/jormungandr-lib/src/interfaces/block0_configuration/initial_config.rs
@@ -1,8 +1,8 @@
 use crate::{
     interfaces::{
         ActiveSlotCoefficient, BlockContentMaxSize, ConsensusLeaderId, EpochStabilityDepth,
-        FeesGoTo, KESUpdateSpeed, LinearFeeDef, NumberOfSlotsPerEpoch, RewardParams, SlotDuration,
-        TaxType, Value,
+        FeesGoTo, KESUpdateSpeed, LinearFeeDef, NumberOfSlotsPerEpoch, PoolLimit,
+        RewardLimitByStake, RewardParams, SlotDuration, TaxType, Value,
     },
     time::SecondsSinceUnixEpoch,
 };
@@ -121,6 +121,18 @@ pub struct BlockchainConfiguration {
     #[serde(default)]
     #[serde(skip_serializing_if = "Option::is_none")]
     pub reward_parameters: Option<RewardParams>,
+
+    /// The rewards limit as a fraction of the total delegated stake (annualised), 100% means no limit, 0% means no distribution
+    /// If omitted, then 100% of rewards are distributed
+    #[serde(default)]
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub rewards_limit: Option<RewardLimitByStake>,
+
+    /// "Sybil" control through a target number of stake pools: ratio is target number of pools / threshold number of pools
+    /// If omitted, then there is no control
+    #[serde(default)]
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub pool_limit: Option<PoolLimit>,
 }
 
 impl From<BlockchainConfiguration> for ConfigParams {
@@ -172,6 +184,8 @@ impl BlockchainConfiguration {
             treasury_parameters: None,
             total_reward_supply: None,
             reward_parameters: None,
+            rewards_limit: None,
+            pool_limit: None,
         }
     }
 
@@ -197,6 +211,8 @@ impl BlockchainConfiguration {
         let mut reward_parameters = None;
         let mut per_certificate_fees = None;
         let mut fees_go_to = None;
+        let mut rewards_limit = None;
+        let mut pool_limit = None;
 
         for param in params.iter().cloned() {
             match param {
@@ -257,6 +273,12 @@ impl BlockchainConfiguration {
                 ConfigParam::PerCertificateFees(param) => per_certificate_fees
                     .replace(param)
                     .map(|_| "per_certificate_fees"),
+                ConfigParam::RewardLimitByStake(param) => {
+                    rewards_limit.replace(param.into()).map(|_| "rewards_limit")
+                }
+                ConfigParam::PoolLimit(param) => {
+                    pool_limit.replace(param.into()).map(|_| "pool_limit")
+                }
             }
             .map(|name| Err(FromConfigParamsError::InitConfigParamDuplicate { name }))
             .unwrap_or(Ok(()))?;
@@ -290,6 +312,8 @@ impl BlockchainConfiguration {
             treasury_parameters,
             total_reward_supply,
             reward_parameters,
+            rewards_limit,
+            pool_limit,
         })
     }
 
@@ -311,6 +335,8 @@ impl BlockchainConfiguration {
             treasury_parameters,
             total_reward_supply,
             reward_parameters,
+            rewards_limit,
+            pool_limit,
         } = self;
 
         let mut params = ConfigParams::new();
@@ -356,6 +382,14 @@ impl BlockchainConfiguration {
 
         if let Some(reward_parameters) = reward_parameters {
             params.push(ConfigParam::RewardParams(reward_parameters.into()));
+        }
+
+        if let Some(rewards_limit) = rewards_limit {
+            params.push(ConfigParam::RewardLimitByStake(rewards_limit.into()));
+        }
+
+        if let Some(pool_limit) = pool_limit {
+            params.push(ConfigParam::PoolLimit(pool_limit.into()));
         }
 
         consensus_leader_ids
@@ -424,6 +458,8 @@ mod test {
                 treasury_parameters: Arbitrary::arbitrary(g),
                 total_reward_supply: Arbitrary::arbitrary(g),
                 reward_parameters: Arbitrary::arbitrary(g),
+                rewards_limit: Arbitrary::arbitrary(g),
+                pool_limit: Arbitrary::arbitrary(g),
             }
         }
     }
--- a/jormungandr-lib/src/interfaces/mod.rs
+++ b/jormungandr-lib/src/interfaces/mod.rs
@@ -9,7 +9,9 @@ mod fragment_log;
 mod leadership_log;
 mod linear_fee;
 mod old_address;
+mod pool_limit;
 mod ratio;
+mod reward_limit;
 mod reward_parameters;
 mod settings;
 mod stats;
@@ -36,9 +38,14 @@ pub use self::leadership_log::{
 };
 pub use self::linear_fee::LinearFeeDef;
 pub use self::old_address::OldAddress;
+pub use self::pool_limit::PoolLimit;
 pub use self::ratio::{ParseRatioError, Ratio};
+pub use self::reward_limit::RewardLimitByStake;
 pub use self::reward_parameters::RewardParams;
-pub use self::settings::{ParametersDef, RatioDef, SettingsDto, TaxTypeDef, TaxTypeSerde};
+pub use self::settings::{
+    ParametersDef, PoolLimitDef, PoolLimitSerde, RatioDef, RewardLimitByDef, RewardLimitBySerde,
+    SettingsDto, TaxTypeDef, TaxTypeSerde,
+};
 pub use self::stats::{NodeState, Stats};
 pub use self::tax_type::TaxType;
 pub use self::transaction_input::{TransactionInput, TransactionInputType};
--- a/jormungandr-lib/src/interfaces/pool_limit.rs
+++ b/jormungandr-lib/src/interfaces/pool_limit.rs
@@ -1,0 +1,77 @@
+use crate::interfaces::{Ratio, Value};
+use chain_impl_mockchain::rewards;
+use serde::{Deserialize, Serialize};
+use std::num::NonZeroU32;
+use std::num::NonZeroU64;
+
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(deny_unknown_fields, rename_all = "snake_case")]
+pub struct PoolLimit {
+    pub npools: NonZeroU32,
+    pub npools_threshold: NonZeroU32,
+}
+
+#[cfg(test)]
+mod test {
+    use super::*;
+    use quickcheck::{Arbitrary, Gen, TestResult};
+    use std::num::NonZeroU64;
+    impl Arbitrary for PoolLimit {
+        fn arbitrary<G>(g: &mut G) -> Self
+        where
+            G: Gen,
+        {
+            PoolLimit {
+                npools: NonZeroU32::new(Arbitrary::arbitrary(g))
+                    .unwrap_or(NonZeroU32::new(1).unwrap()),
+                npools_threshold: NonZeroU32::new(Arbitrary::arbitrary(g))
+                    .unwrap_or(NonZeroU32::new(1).unwrap()),
+            }
+        }
+    }
+
+    #[test]
+    fn value_serde_yaml() {
+        const NPOOLS: NonZeroU32 = unsafe { NonZeroU32::new_unchecked(123) };
+        const NPOOLS_THRESHOLD: NonZeroU32 = unsafe { NonZeroU32::new_unchecked(456) };
+        let pool_limit = PoolLimit {
+            npools: NPOOLS,
+            npools_threshold: NPOOLS_THRESHOLD,
+        };
+
+        assert_eq!(
+            serde_yaml::to_string(&pool_limit).unwrap(),
+            format!(
+                "---\nnpools: {}\nnpools_threshold: {}",
+                NPOOLS, NPOOLS_THRESHOLD
+            )
+        );
+    }
+}
+
+/* ************** Conversion *********************************** */
+
+impl From<PoolLimit> for rewards::PoolLimit {
+    fn from(rewards_limit: PoolLimit) -> Self {
+        rewards::PoolLimit {
+            npools: rewards_limit.npools.into(),
+            npools_threshold: rewards_limit.npools_threshold.into(),
+        }
+    }
+}
+
+impl From<rewards::PoolLimit> for PoolLimit {
+    fn from(rewards_limit: rewards::PoolLimit) -> Self {
+        PoolLimit {
+            npools: rewards_limit.npools.into(),
+            npools_threshold: rewards_limit.npools_threshold.into(),
+        }
+    }
+}
+
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(deny_unknown_fields, rename_all = "snake_case")]
+pub struct RewardLimitByStake {
+    pub numerator: u32,
+    pub denominator: NonZeroU32,
+}
--- a/jormungandr-lib/src/interfaces/reward_limit.rs
+++ b/jormungandr-lib/src/interfaces/reward_limit.rs
@@ -1,0 +1,70 @@
+use crate::interfaces::{Ratio, Value};
+use chain_impl_mockchain::rewards;
+use serde::{Deserialize, Serialize};
+use std::num::NonZeroU32;
+use std::num::NonZeroU64;
+
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(deny_unknown_fields, rename_all = "snake_case")]
+pub struct RewardLimitByStake {
+    pub numerator: u32,
+    pub denominator: NonZeroU32,
+}
+
+#[cfg(test)]
+mod test {
+    use super::*;
+    use quickcheck::{Arbitrary, Gen, TestResult};
+    use std::num::NonZeroU64;
+
+    impl Arbitrary for RewardLimitByStake {
+        fn arbitrary<G>(g: &mut G) -> Self
+        where
+            G: Gen,
+        {
+            RewardLimitByStake {
+                numerator: Arbitrary::arbitrary(g),
+                denominator: NonZeroU32::new(Arbitrary::arbitrary(g))
+                    .unwrap_or(NonZeroU32::new(1).unwrap()),
+            }
+        }
+    }
+
+    #[test]
+    fn value_serde_yaml() {
+        const NUMERATOR: u32 = 15;
+        const DENOMINATOR: NonZeroU32 = unsafe { NonZeroU32::new_unchecked(100) };
+        let reward_limit = RewardLimitByStake {
+            numerator: NUMERATOR,
+            denominator: DENOMINATOR,
+        };
+
+        assert_eq!(
+            serde_yaml::to_string(&reward_limit).unwrap(),
+            format!(
+                "---\nnumerator: {}\ndenominator: {}",
+                NUMERATOR, DENOMINATOR
+            )
+        );
+    }
+}
+
+/* ************** Conversion *********************************** */
+
+impl From<RewardLimitByStake> for rewards::RewardLimitByStake {
+    fn from(rewards_limit: RewardLimitByStake) -> Self {
+        rewards::RewardLimitByStake {
+            numerator: rewards_limit.numerator,
+            denominator: rewards_limit.denominator.into(),
+        }
+    }
+}
+
+impl From<rewards::RewardLimitByStake> for RewardLimitByStake {
+    fn from(rewards_limit: rewards::RewardLimitByStake) -> Self {
+        RewardLimitByStake {
+            numerator: rewards_limit.numerator,
+            denominator: rewards_limit.denominator.into(),
+        }
+    }
+}
--- a/jormungandr-lib/src/interfaces/settings.rs
+++ b/jormungandr-lib/src/interfaces/settings.rs
@@ -4,7 +4,9 @@ use crate::{
 };
 use chain_impl_mockchain::block::Epoch;
 use chain_impl_mockchain::fee::LinearFee;
-use chain_impl_mockchain::rewards::{CompoundingType, Parameters, Ratio, TaxType};
+use chain_impl_mockchain::rewards::{
+    CompoundingType, Parameters, PoolLimit, Ratio, RewardLimitByStake, TaxType,
+};
 use chain_impl_mockchain::value::Value;
 use serde::{Deserialize, Serialize};
 use std::num::{NonZeroU32, NonZeroU64};
@@ -26,6 +28,10 @@ pub struct SettingsDto {
     pub treasury_tax: TaxType,
     #[serde(with = "ParametersDef")]
     pub reward_params: Parameters,
+    #[serde(with = "RewardLimitByDef")]
+    pub rewards_limit: RewardLimitByStake,
+    #[serde(with = "PoolLimitDef")]
+    pub pool_limit: PoolLimit,
 }
 
 #[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
@@ -83,5 +89,29 @@ impl PartialEq<SettingsDto> for SettingsDto {
             && self.slots_per_epoch == other.slots_per_epoch
             && self.treasury_tax == other.treasury_tax
             && self.reward_params == other.reward_params
+            && self.rewards_limit == other.rewards_limit
+            && self.pool_limit == other.pool_limit
     }
 }
+
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(deny_unknown_fields, remote = "PoolLimit")]
+pub struct PoolLimitDef {
+    pub npools: NonZeroU32,
+    pub npools_threshold: NonZeroU32,
+}
+
+#[derive(Serialize)]
+#[serde(transparent)]
+pub struct PoolLimitSerde(#[serde(with = "PoolLimit")] pub PoolLimit);
+
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(deny_unknown_fields, remote = "RewardLimitByStake")]
+pub struct RewardLimitByDef {
+    pub numerator: u32,
+    pub denominator: NonZeroU32,
+}
+
+#[derive(Serialize)]
+#[serde(transparent)]
+pub struct RewardLimitBySerde(#[serde(with = "RewardLimitByDef")] pub RewardLimitByStake);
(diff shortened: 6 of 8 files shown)


jormungandr-lib/src/interfaces/reward_parameters.rs
```diff
@@ -11,12 +11,18 @@ pub enum RewardParams {
         ratio: Ratio,
         epoch_start: Epoch,
         epoch_rate: NonZeroU32,
+        rewards_limit: Ratio,
+        npools: NonZeroU32,
+        npools_threshold: NonZeroU32,
```
NicolasDP: the parameters are not very explicit. `npools` is very vague.

jormungandr-lib/src/interfaces/block0_configuration/DOCUMENTED_EXAMPLE.yaml
```diff
@@ -118,26 +118,37 @@ blockchain_configuration:
...
       # indicates when this contribution start. note that if the epoch is not
       # the same or after the epoch_start, the overall contribution is zero.
       epoch_start: 1
 
       # the rate at which the contribution is tweaked related to epoch.
-      epoch_rate: 3
+      epoch_rate: 1
+
+      # the limit on the maximum rewards that can be obtained, as a percentage of the delegated stake.
+      # Calculated as an annualised percentage rate (over 365 days).  So 10/100 is maximum rate of 10% etc.
```
NicolasDP: I have no idea what that mean. please document what that means for stake pools and how that will affect the reward calculation (if set).

jormungandr-lib/src/interfaces/block0_configuration/DOCUMENTED_EXAMPLE.yaml
```diff
@@ -118,26 +118,37 @@ blockchain_configuration:
...
+      rewards_limit: "15/100"
+
+      # the limit on the number of pools, used to encourage pool creation and avoid Sybil attacks ("k" in the incentives paper)
+      npools: 1000
```
NicolasDP: This is not a limit on the number of pools. This is a parameter that can be set to encourage limiting the number of pool creation to a certain number. However nothing prevents anyone to create more pools.

jormungandr-lib/src/interfaces/block0_configuration/DOCUMENTED_EXAMPLE.yaml
```diff
@@ -118,26 +118,37 @@ blockchain_configuration:
   # Halving formula: constant * ratio ^ (#epoch after epoch_start / epoch_rate)
   #
   reward_parameters:
-    halving: # or use "linear" for the linear formula
+    linear: # or use "linear" for the linear formula
```
NicolasDP: I don't understand why you needed to change this value?

jormungandr-lib/src/interfaces/block0_configuration/DOCUMENTED_EXAMPLE.yaml
```diff
@@ -107,7 +107,7 @@ blockchain_configuration:
   #
   # if not set there is no monetary creation
   # once emptied, there is no more monetary creation
-  total_reward_supply: 100000000000000
+  total_reward_supply: 701917808520000
```
NicolasDP: I don't understand why you needed to change this value?

jormungandr-lib/src/interfaces/block0_configuration/DOCUMENTED_EXAMPLE.yaml
```diff
@@ -118,26 +118,37 @@ blockchain_configuration:
   # Halving formula: constant * ratio ^ (#epoch after epoch_start / epoch_rate)
   #
   reward_parameters:
-    halving: # or use "linear" for the linear formula
+    linear: # or use "linear" for the linear formula
       # In the linear formula, it represents the starting point of the contribution
       # at #epoch=0, whereas in halving formula is used as starting constant for
       # the calculation.
-      constant: 100
+      constant: 3835616440000
```
NicolasDP: I don't understand why you needed to change this value?

jormungandr-lib/src/interfaces/block0_configuration/DOCUMENTED_EXAMPLE.yaml
```diff
@@ -118,26 +118,37 @@ blockchain_configuration:
...
 
       # In the halving formula, an effective value between 0.0 to 1.0 indicates a
       # reducing contribution, whereas above 1.0 it indicate an acceleration of contribution.
       #
       # However in linear formula the meaning is just a scaling factor for the epoch zone
       # (current_epoch - start_epoch / epoch_rate). Further requirement is that this ratio
       # is expressed in fractional form (e.g. 1/2), which allow calculation in integer form.
-      ratio: "13/19"
+      ratio: "1/1"
```
NicolasDP: I don't understand why you needed to change this value?

jormungandr-lib/src/interfaces/block0_configuration/DOCUMENTED_EXAMPLE.yaml
```diff
@@ -118,26 +118,37 @@ blockchain_configuration:
...
 
       # indicates when this contribution start. note that if the epoch is not
       # the same or after the epoch_start, the overall contribution is zero.
       epoch_start: 1
 
       # the rate at which the contribution is tweaked related to epoch.
-      epoch_rate: 3
+      epoch_rate: 1
```
NicolasDP: I don't understand why you needed to change this value?

jormungandr-lib/src/interfaces/block0_configuration/DOCUMENTED_EXAMPLE.yaml
```diff
@@ -95,7 +95,7 @@ blockchain_configuration:
   # if not set, the treasury will not grow
   treasury_parameters:
     # the fix value the treasury will take from the total reward pot of the epoch
-    fixed: 1000
+    fixed: 0
```
NicolasDP: I don't understand why you needed to change this value?

jormungandr-lib/src/interfaces/block0_configuration/DOCUMENTED_EXAMPLE.yaml
```diff
@@ -84,7 +84,7 @@ blockchain_configuration:
 
   # initial value the treasury will start with, if not set the treasury
   # starts at 0
-  treasury: 1000000000000
+  treasury: 0
```
NicolasDP: I don't understand why you needed to change this value?

jormungandr-lib/src/interfaces/block0_configuration/DOCUMENTED_EXAMPLE.yaml
```diff
@@ -68,10 +68,10 @@ blockchain_configuration:
     per_certificate_fees:
       # (optional) if not specified, the pool registration certificate fee will be
       # the one set by linear_fees.certificate
-      certificate_pool_registration: 5
+      certificate_pool_registration: 150000000
       # (optional) if not specified, the delegation certificate fee will be
       # the one set by linear_fees.certificate
-      certificate_stake_delegation: 2
+      certificate_stake_delegation: 400000
```
NicolasDP: I don't understand why you needed to change this value?

jormungandr-lib/src/interfaces/block0_configuration/DOCUMENTED_EXAMPLE.yaml
```diff
@@ -68,10 +68,10 @@ blockchain_configuration:
     per_certificate_fees:
       # (optional) if not specified, the pool registration certificate fee will be
       # the one set by linear_fees.certificate
-      certificate_pool_registration: 5
+      certificate_pool_registration: 150000000
```
NicolasDP: I don't understand why you needed to change this value?

jormungandr-lib/src/interfaces/block0_configuration/DOCUMENTED_EXAMPLE.yaml
```diff
@@ -55,11 +55,11 @@ blockchain_configuration:
...
   linear_fees:
     # this is the minimum value to pay for every transaction
-    constant: 2
+    constant: 200000
     # the additional fee to pay for every inputs and outputs
-    coefficient: 1
+    coefficient: 100000
     # the additional fee to pay if the transaction embeds a certificate
-    certificate: 4
+    certificate: 400000
```
NicolasDP: I don't understand why you needed to change this value?

jormungandr-lib/src/interfaces/block0_configuration/DOCUMENTED_EXAMPLE.yaml
```diff
@@ -55,11 +55,11 @@ blockchain_configuration:
   # total fees: constant + (num_inputs + num_outputs) * coefficient [+ certificate]
   linear_fees:
     # this is the minimum value to pay for every transaction
-    constant: 2
+    constant: 200000
     # the additional fee to pay for every inputs and outputs
-    coefficient: 1
+    coefficient: 100000
```
NicolasDP: I don't understand why you needed to change this value?

jormungandr-lib/src/interfaces/block0_configuration/DOCUMENTED_EXAMPLE.yaml
```diff
@@ -55,11 +55,11 @@ blockchain_configuration:
   # total fees: constant + (num_inputs + num_outputs) * coefficient [+ certificate]
   linear_fees:
     # this is the minimum value to pay for every transaction
-    constant: 2
+    constant: 200000
```
NicolasDP: I don't understand why you needed to change this value?

jormungandr-lib/src/interfaces/reward_parameters.rs
NicolasDP: These value should be optionals.

jormungandr-lib/src/interfaces/reward_parameters.rs
```diff
@@ -98,6 +128,11 @@ mod test {
                     epoch_start: Epoch::arbitrary(g),
                     epoch_rate: NonZeroU32::new(Arbitrary::arbitrary(g))
                         .unwrap_or(NonZeroU32::new(1).unwrap()),
+                    rewards_limit: Ratio::arbitrary(g),
+                    npools: NonZeroU32::new(Arbitrary::arbitrary(g))
+                        .unwrap_or(NonZeroU32::new(1000).unwrap()),
```
NicolasDP: why `1000` ?

jormungandr-lib/src/interfaces/reward_parameters.rs
```diff
@@ -98,6 +128,11 @@ mod test {
...
+                    npools_threshold: NonZeroU32::new(Arbitrary::arbitrary(g))
+                        .unwrap_or(NonZeroU32::new(100).unwrap()),
```
NicolasDP: why `100` ?

jormungandr-lib/src/interfaces/reward_parameters.rs
```diff
@@ -106,6 +141,11 @@ mod test {
                     epoch_start: Epoch::arbitrary(g),
                     epoch_rate: NonZeroU32::new(Arbitrary::arbitrary(g))
                         .unwrap_or(NonZeroU32::new(1).unwrap()),
+                    rewards_limit: Ratio::arbitrary(g),
+                    npools: NonZeroU32::new(Arbitrary::arbitrary(g))
+                        .unwrap_or(NonZeroU32::new(1000).unwrap()),
```
NicolasDP: why `1000`?

jormungandr-lib/src/interfaces/reward_parameters.rs
```diff
@@ -106,6 +141,11 @@ mod test {
...
+                    npools_threshold: NonZeroU32::new(Arbitrary::arbitrary(g))
+                        .unwrap_or(NonZeroU32::new(100).unwrap()),
```
NicolasDP: why `100` ?

scripts/bootstrap
```diff
@@ -251,6 +251,9 @@ blockchain_configuration:
       ratio: "1/1"
       epoch_start: 0
       epoch_rate: 1
+      rewards_limit: "15/100"
+      npools: 1000
+      npools_threshold: 300
```
NicolasDP: Don't set them in the bootstrap. You should leave the default (no limits, no threashold, no max).

jormungandr-lib/src/interfaces/block0_configuration/DOCUMENTED_EXAMPLE.yaml
kevinhammond: OK

jormungandr-lib/src/interfaces/block0_configuration/DOCUMENTED_EXAMPLE.yaml
kevinhammond: We will be using a constant return for the testnet

jormungandr-lib/src/interfaces/block0_configuration/DOCUMENTED_EXAMPLE.yaml
kevinhammond: This is the limit that will be set for the testnet.  It gives a maximum lifetime of 6 months at the agreed return rate

jormungandr-lib/src/interfaces/block0_configuration/DOCUMENTED_EXAMPLE.yaml
kevinhammond: This is the constant amount that will be returned per epoch, as agreed in September

jormungandr-lib/src/interfaces/block0_configuration/DOCUMENTED_EXAMPLE.yaml
kevinhammond: This ensures that the return does not decline, but remains constant, as agreed in September

jormungandr-lib/src/interfaces/block0_configuration/DOCUMENTED_EXAMPLE.yaml
kevinhammond: The rate is not actually declining, so it makes more sense to set this to 1 than 3

jormungandr-lib/src/interfaces/block0_configuration/DOCUMENTED_EXAMPLE.yaml
kevinhammond: The treasury return does not have a fixed component per epoch.  It is variable based on monetary expansion

jormungandr-lib/src/interfaces/block0_configuration/DOCUMENTED_EXAMPLE.yaml
kevinhammond: The treasury initially starts empty.  There is no reason for it to contain anything.  Value is transferred to it from the reserves through tax

jormungandr-lib/src/interfaces/block0_configuration/DOCUMENTED_EXAMPLE.yaml
kevinhammond: We discussed this with Vincent.  He felt that 2 lovelace was too low.  This is the standard cost for embedding a certificate.

jormungandr-lib/src/interfaces/block0_configuration/DOCUMENTED_EXAMPLE.yaml
kevinhammond: This is consistent with the expected charge on the mainnet (actually about 1/3-2/3 of the charge, which will be 250-500 Ada).  It's reasonable to set this slightly lower on the testnet.

jormungandr-lib/src/interfaces/block0_configuration/DOCUMENTED_EXAMPLE.yaml
kevinhammond: See above.  This is 0.4 Ada.

jormungandr-lib/src/interfaces/block0_configuration/DOCUMENTED_EXAMPLE.yaml
kevinhammond: See above.  This is 0.1 Ada.

jormungandr-lib/src/interfaces/block0_configuration/DOCUMENTED_EXAMPLE.yaml
kevinhammond: See above.  This is 0.2 Ada.

jormungandr-lib/src/interfaces/reward_parameters.rs
kevinhammond: rewards_limit can be set to "100/100" to have no effect, so there is no need to make this optional.
npools and npools threshold can be both set to zero to have no effect.  They should be present in the actual testnet, but could conceivably be made optionals.

jormungandr-lib/src/interfaces/reward_parameters.rs
kevinhammond: This seemed to be the number of pools that were being aimed at by the testnet.  It could be set differently.  In the mainnet, it will vary over time.  Setting it to 1000 as a test case will confirm that the mechanism is working.

jormungandr-lib/src/interfaces/reward_parameters.rs
kevinhammond: This will allow 100 pools before any penalty is applied.  This seemed to be the initial target number of pools.  Again this will confirm that the mechanisms are working.

jormungandr-lib/src/interfaces/reward_parameters.rs
kevinhammond: See above.

jormungandr-lib/src/interfaces/reward_parameters.rs
kevinhammond: See above.

scripts/bootstrap
kevinhammond: These must be set in the bootstrap since there is no way to change them later in the testnet.  They are key parameters to ensure that the system behaves as intended.

I will discuss the actual values to make sure they have the intended effect.  npools = 1000 may be too high and the threshold might be set higher.

jormungandr-lib/src/interfaces/block0_configuration/DOCUMENTED_EXAMPLE.yaml
NicolasDP: this is a `DOCUMENTED_EXAMPLE.md` not the settings for the testnet.

jormungandr-lib/src/interfaces/block0_configuration/DOCUMENTED_EXAMPLE.yaml
NicolasDP: this is a `DOCUMENTED_EXAMPLE.md` not the settings for the testnet.

jormungandr-lib/src/interfaces/reward_parameters.rs
NicolasDP: optional is way better for the genesis file.

jormungandr-lib/src/interfaces/reward_parameters.rs
NicolasDP: this is for the test. `1` is what has been used everywhere else to replace the `0` case.

jormungandr-lib/src/interfaces/reward_parameters.rs
NicolasDP: this is for the test. `1` is what has been used everywhere else to replace the `0` case.

jormungandr-lib/src/interfaces/reward_parameters.rs
NicolasDP: this is for the test. `1` is what has been used everywhere else to replace the `0` case.

scripts/bootstrap
NicolasDP: bootstrap is not the script used to setup the IOHK's incentivised testnet. So no need to add that.

jormungandr-lib/src/interfaces/block0_configuration/DOCUMENTED_EXAMPLE.yaml
NicolasDP: this is a `DOCUMENTED_EXAMPLE.md` not the settings for the testnet.

jormungandr-lib/src/interfaces/block0_configuration/DOCUMENTED_EXAMPLE.yaml
NicolasDP: this is a `DOCUMENTED_EXAMPLE.md` not the settings for the testnet.

jormungandr-lib/src/interfaces/block0_configuration/DOCUMENTED_EXAMPLE.yaml
NicolasDP: this is a `DOCUMENTED_EXAMPLE.md` not the settings for the testnet.

jormungandr-lib/src/interfaces/block0_configuration/DOCUMENTED_EXAMPLE.yaml
NicolasDP: this is a `DOCUMENTED_EXAMPLE.md` not the settings for the testnet.

jormungandr-lib/src/interfaces/block0_configuration/DOCUMENTED_EXAMPLE.yaml
NicolasDP: this is a `DOCUMENTED_EXAMPLE.md` not the settings for the testnet.

jormungandr-lib/src/interfaces/block0_configuration/DOCUMENTED_EXAMPLE.yaml
NicolasDP: this is a `DOCUMENTED_EXAMPLE.md` not the settings for the testnet.

jormungandr-lib/src/interfaces/block0_configuration/DOCUMENTED_EXAMPLE.yaml
NicolasDP: this is a `DOCUMENTED_EXAMPLE.md` not the settings for the testnet.

jormungandr-lib/src/interfaces/block0_configuration/DOCUMENTED_EXAMPLE.yaml
NicolasDP: this is a `DOCUMENTED_EXAMPLE.md` not the settings for the testnet.

jormungandr-lib/src/interfaces/block0_configuration/DOCUMENTED_EXAMPLE.yaml
NicolasDP: this is a `DOCUMENTED_EXAMPLE.md` not the settings for the testnet.

jormungandr-lib/src/interfaces/reward_parameters.rs
kevinhammond: They're described in the comments and will be discussed in blogs etc.  Other suggestions for names are welcome (but be aware that this may affect the code as well, of course).

jormungandr-lib/src/interfaces/block0_configuration/DOCUMENTED_EXAMPLE.yaml
kevinhammond: Will add some additional notes.

kevinhammond: Added some more notes.  This will be fully described in the incentives documentation.

jormungandr-lib/src/interfaces/block0_configuration/DOCUMENTED_EXAMPLE.yaml
kevinhammond: Changed limit to target and added some more description.

jormungandr-lib/src/interfaces/block0_configuration/DOCUMENTED_EXAMPLE.yaml
kevinhammond: That's why it's good to have the same settings where possible (avoids confusion).  You can always add another example for an alternative setup if you like.

jormungandr-lib/src/interfaces/block0_configuration/DOCUMENTED_EXAMPLE.yaml
kevinhammond: See above.

jormungandr-lib/src/interfaces/reward_parameters.rs
kevinhammond: 1 won't have the same effect as a higher number.  If you want to test the code, it needs to be at least 2.

jormungandr-lib/src/interfaces/reward_parameters.rs
kevinhammond: See above.  Needs to be at least 2 and preferably higher to exercise the test.

jormungandr-lib/src/interfaces/reward_parameters.rs
kevinhammond: See above.

jormungandr-lib/src/interfaces/reward_parameters.rs
kevinhammond: See above.

scripts/bootstrap
kevinhammond: But the parameters will also be fixed for the testnet, and we can expect users to pick this up to run their own equivalents for test purposes.  As I said, providing other examples is good, but you do want something that's consistent with the actual testnet settings.

jormungandr-lib/src/interfaces/block0_configuration/DOCUMENTED_EXAMPLE.yaml
kevinhammond: See above.

jormungandr-lib/src/interfaces/block0_configuration/DOCUMENTED_EXAMPLE.yaml
kevinhammond: See above.

jormungandr-lib/src/interfaces/reward_parameters.rs
kevinhammond: OK.

jormungandr-lib/src/interfaces/block0_configuration/DOCUMENTED_EXAMPLE.yaml
kevinhammond: I have created an alternative setup. It's sensible to show the actual testnet parameters in a config file

mark-stopka:
Any calculation with a calendar year should use 365.25 to account for leap-years; I see in a comment

`# Calculated as an annualised percentage rate (over 365 days).  So 10/100 is maximum rate of 10% etc.`

jormungandr-lib/src/interfaces/block0_configuration/DOCUMENTED_EXAMPLE_TESTNET.yaml
```diff
@@ -0,0 +1,181 @@
+# The Blockchain Configuration defines the settings of the blockchain.
```
NicolasDP: delete this file. there is no point into adding this. really

scripts/bootstrap
NicolasDP: there is no need to be consistent, this script is meant to help bootstrapping a simple node and start doing some development tests.

jormungandr-lib/src/interfaces/reward_parameters.rs
```diff
@@ -98,6 +128,15 @@ mod test {
                     epoch_start: Epoch::arbitrary(g),
                     epoch_rate: NonZeroU32::new(Arbitrary::arbitrary(g))
                         .unwrap_or(NonZeroU32::new(1).unwrap()),
+                    rewards_limit: Some(Ratio::arbitrary(g)),
+                    npools: Some(
+                        NonZeroU32::new(Arbitrary::arbitrary(g))
+                            .unwrap_or(NonZeroU32::new(1000).unwrap()),
```
NicolasDP:
```suggestion
                            .unwrap_or(NonZeroU32::new(1).unwrap()),
```

jormungandr-lib/src/interfaces/reward_parameters.rs
```diff
@@ -98,6 +128,15 @@ mod test {
...
+                    ),
+                    npools_threshold: Some(
+                        NonZeroU32::new(Arbitrary::arbitrary(g))
+                            .unwrap_or(NonZeroU32::new(100).unwrap()),
```
NicolasDP:
```suggestion
                            .unwrap_or(NonZeroU32::new(1).unwrap()),
```

jormungandr-lib/src/interfaces/reward_parameters.rs
```diff
@@ -106,6 +145,15 @@ mod test {
                     epoch_start: Epoch::arbitrary(g),
                     epoch_rate: NonZeroU32::new(Arbitrary::arbitrary(g))
                         .unwrap_or(NonZeroU32::new(1).unwrap()),
+                    rewards_limit: Some(Ratio::arbitrary(g)),
+                    npools: Some(
+                        NonZeroU32::new(Arbitrary::arbitrary(g))
+                            .unwrap_or(NonZeroU32::new(1000).unwrap()),
```
NicolasDP:
```suggestion
                            .unwrap_or(NonZeroU32::new(1).unwrap()),
```

jormungandr-lib/src/interfaces/reward_parameters.rs
```diff
@@ -106,6 +145,15 @@ mod test {
...
+                    ),
+                    npools_threshold: Some(
+                        NonZeroU32::new(Arbitrary::arbitrary(g))
+                            .unwrap_or(NonZeroU32::new(100).unwrap()),
```
NicolasDP:
```suggestion
                            .unwrap_or(NonZeroU32::new(1).unwrap()),
```

NicolasDP:
@mark-stopka comments is also of actuality. How did you handle the leap year case?

jormungandr-lib/src/interfaces/reward_parameters.rs
NicolasDP: Here arbitrary is used for validity of the parameter. If you believe `1` is not a valid value, then add a better parsing instead of relying on the default derived one otherwise it means you allow such a value and it then means we are supposed to support that bad value.

jormungandr-lib/src/interfaces/block0_configuration/DOCUMENTED_EXAMPLE.yaml
NicolasDP: devops or comm will do that. No need to show the actual parameters here.

scripts/bootstrap
kevinhammond: OK.  Have created an additional example (and tested our parameters using it).

jormungandr-lib/src/interfaces/reward_parameters.rs
kevinhammond: 1 is a valid value.  I found it useful to differentiate the numbers to detect byte-swapping in the serialisation code.

> Any calculation with a calendar year should use 365.25 to account for leap-years; I see in a comment
> 
> `# Calculated as an annualised percentage rate (over 365 days). So 10/100 is maximum rate of 10% etc.`



> Any calculation with a calendar year should use 365.25 to account for leap-years; I see in a comment
> 
> `# Calculated as an annualised percentage rate (over 365 days). So 10/100 is maximum rate of 10% etc.`

I think this might be an integer calculation.  In this situation, the difference should be essentially noise.

PR #1289, opened 2019-12-05
Fix mismatch step number

Vetro7:
The addition of submodule update command has moved correlating step down by one.


NicolasDP:
thanks

Vetro7:
Np

PR #1291, opened 2019-12-05
Add reward and treasury settings to settings REST

CodeSandwich:
Fixes https://github.com/input-output-hk/jormungandr/issues/1248


jormungandr-lib/src/interfaces/settings.rs
```diff
@@ -14,6 +21,53 @@ pub struct SettingsDto {
...
+    pub numerator: u64,
+    pub denominator: NonZeroU64,
+}
+
+#[derive(Deserialize, Serialize)]
+#[serde(remote = "Parameters", rename_all = "camelCase")]
+pub struct ParametersDef {
+    pub initial_value: u64,
+    #[serde(with = "RatioDef")]
+    pub compounding_ratio: Ratio,
```
NicolasDP: duplicate of https://github.com/input-output-hk/jormungandr/blob/9a10bf7bc211969e8c331532b956f6e4110d9446/jormungandr-lib/src/interfaces/ratio.rs#L12

jormungandr-lib/src/interfaces/value.rs
```diff
@@ -83,6 +83,10 @@ impl<'de> Deserialize<'de> for Value {
     }
 }
 
+#[derive(Deserialize, Serialize)]
+#[serde(transparent, remote = "value::Value")]
+pub struct ValueDef(u64);
```
NicolasDP: Not going to be always true that a `Value` is a `u64`. Use the `Value` type above and pay the cost of writing `.into()`.

jormungandr/src/rest/v0/handlers.rs
```diff
@@ -393,6 +397,7 @@ pub fn get_network_stats(context: State<Context>) -> ActixFuture!() {
                         "lastBlockReceived": stats.last_block_received().map(SystemTime::from),
                         "lastFragmentReceived": stats.last_fragment_received().map(SystemTime::from),
                         "lastGossipReceived": stats.last_gossip_received().map(SystemTime::from),
+
```
NicolasDP: non needed newline

doc/openapi.yaml
```diff
@@ -682,8 +743,26 @@ paths:
                     "constant": 2
                   },
                   "maxTxsPerBlock": 100,
+                  "rewardParams": {
+                      "compoundingRatio": {
+                          "denominator": 1024,
+                          "numerator": 1
+                      },
+                      "compoundingType": "Linear",
+                      "epochRate": 4294967295,
```
NicolasDP: this epoch rate is not likely on normal operation of the blockchain.

```suggestion
                      "epochRate": 100,
```

jormungandr-lib/src/interfaces/settings.rs
```diff
@@ -14,6 +21,53 @@ pub struct SettingsDto {
...
+pub struct TaxTypeDef {
+    #[serde(with = "ValueDef")]
+    pub fixed: Value,
+
+    #[serde(with = "RatioDef")]
+    pub ratio: Ratio,
+
+    #[serde(default, rename = "max", skip_serializing_if = "Option::is_none")]
+    pub max_limit: Option<NonZeroU64>,
+}
```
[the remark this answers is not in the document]
CodeSandwich: It's different. The new version serializes `ratio` as two fields while the existing one as `"X/Y"`. The existing one is fine for humans to write in `config.yaml`, but the new one doesn't require parsing on client. It's also fitting the existing REST API in `stake_pool`.

jormungandr-lib/src/interfaces/value.rs
CodeSandwich: This is guaranteed by Serde's `remote` attribute, see https://serde.rs/remote-derive.html

doc/jcli/rest.md
```diff
@@ -267,13 +267,27 @@ fees:                                           # transaction fee configuration
...
+    certificate_pool_registration: 5            # fee per pool registration, zero if absent (optional)
+    certificate_stake_delegation: 15            # fee per stake delegation, zero if absent (optional)
+    certificate_owner_stake_delegation: 2       # fee per pool owner stake delegation, zero if absent (optional)
 maxTxsPerBlock: 100                             # maximum number of transactions in block
+rewardParams:                                   # parameters for rewards calculation
+  compoundingRatio:                             # speed at which reward is reduced. Expressed as numerator/denominator
+    denominator: 1024
+    numerator: 1
+  compoundingType: Linear                       # reward reduction algorithm. Possible values: "Linear" and "Halvening"
+  epochRate: 4294967295                         # number of epochs between reward reductions
```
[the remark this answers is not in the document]
CodeSandwich: Good one. It's `2^32 - 1` and this was actually returned from my node. Smells like a bug.

jormungandr-lib/src/interfaces/settings.rs
CodeSandwich: Same as above

doc/jcli/rest.md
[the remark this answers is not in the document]
CodeSandwich: Definitely not a bug, just a default value which is supposed to virtually disable reward reduction.

jormungandr-lib/src/interfaces/settings.rs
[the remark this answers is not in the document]
NicolasDP: arf... true that... the duplication is annoying but fair enough

jormungandr-lib/src/interfaces/value.rs
NicolasDP: I still don't see the point in duplicating that `Value`. I see how it makes it _easy_. for you but I'd rather like to see all of this going toward a harmonised interface.

CodeSandwich: All fields in `SettingsDto` are "normal" types from `chain-impl-mockchain`, but with custom serialization (mostly `with` and `remote` types). This make its usage easier, especially in tests. Adding `remote` type for `Value` lets us keep this property of `SettingsDto`.

I also think that its a cleaner solution than conversion from and to the `jormungandr-lib::Value` and it's generally a good idea to start using it in all Serde DTOs.

Issue #1165, opened 2019-11-20
Improve error handling


eugene-babichenko:
Error types are defined in several different ways in `jormungandr` and `chain-libs`: with `custom_error` macro, plain `enum`s or `struct`s, sometimes the required traits are implemented by hand.

It would be nice to have a standardised way of handling errors with something like the [`failure`](https://crates.io/crates/failure) crate which does pretty good job.

PR #1345, opened 2019-12-11
remove custom_error

eugene-babichenko:
Remove `custom_error` in favour of using `thiserror`. This provides several benefits:

- Code becomes easier to navigate because now it uses the native enum syntax
- For the same reason it can be formatted with `rustfmt`
- The `CustomErrorFiller` workaround is no longer necessary
- More consistent error handling (2 libraries instead 3)

Related to #1165


eugene-babichenko:
The question is whether we should rewrite errors like

```rust
BlockFileCorrupted {
    #[source]
    source: std::io::Error,
}
```

this way

```rust
BlockFileCorrupted(#[source] std::io::Error)
```

it should improve the overall readability.

eugene-babichenko:
pushed an update

jormungandr-lib/src/interfaces/account_identifier.rs
```diff
@@ -53,11 +53,7 @@ impl fmt::Display for AccountIdentifier {
 #[derive(Debug, Error)]
 pub enum ParseAccountIdentifierError {
     #[error("Cannot parse account identifier")]
-    CannotParseAddress {
-        #[source]
-        #[from]
-        source: chain_addr::Error,
-    },
+    CannotParseAddress(#[from] chain_addr::Error),
```
NicolasDP: to keep the same behavior this should be a `#[source]` too

jormungandr-lib/src/interfaces/block0_configuration/initial_config.rs
```diff
@@ -135,15 +136,28 @@ impl From<BlockchainConfiguration> for ConfigParams {
...
+    #[error("Invalid consensus leader id")]
+    ConsensusLeaderId(#[from] super::leader_id::TryFromConsensusLeaderIdError),
+    #[error("Invalid active slot coefficient value")]
+    ActiveSlotCoefficient(
+        #[from] super::active_slot_coefficient::TryFromActiveSlotCoefficientError,
+    ),
+    #[error("Invalid KES Update speed value")]
+    KESUpdateSpeed(#[from] super::kes_update_speed::TryFromKESUpdateSpeedError),
+    #[error("Invalid FeesGoTo setting")]
+    FeesGoTo(#[from] super::fees_go_to::TryFromFeesGoToError),
```
NicolasDP: in `custom_error` the field called `source: ` had the `#[source]` implemented. Not having this available will make us lose some info regarding the cause of an error.

eugene-babichenko: The documentation on `thiserror` says that

> The `#[from]` attribute always implies that the same field is `#[source]`, so you don't ever need to specify both attributes.

so I believe this should work as you want it to.

NicolasDP: great, `thiserror` is really the right choice.

PR #1354, opened 2019-12-12
add command line tool to prepare the changelogs

NicolasDP:


eugene-babichenko:
I think adding a download link to the tool somewhere in the docs would be good.

tools/prepare-changelog.sh
```diff
@@ -17,7 +17,14 @@ fi
 GITHUB_TOKEN=${1}
 NEW_VERSION=${2}
 
-github_changelog_generator \
+GITHUB_CHANGELOG_GENERATOR="$(which github_changelog_generator)"
```
eugene-babichenko: I think the check can be something like

```bash
# check that the command exists
command -v github_changelog_generator > /dev/null
# a non-zero code indicates that the command was not found
if [ $? -ne 0 ]; then
    echo "'github_changelog_generator' not installed? see https://github.com/github-changelog-generator/github-changelog-generator" >&2
    exit 1
fi
```

and you just use `github_changelog_generator`

PR #1357, opened 2019-12-13
fix fees application for owner stake delegation

vincenthz:


rinor:
Out of curiosity what would be the side effects of having stake nodes with mixed versions, meaning some with this patch (v0.8.1) and some without (v0.8.0). Is it correct to assume that it depend on which version the node producing the block is?

vincenthz:
there will be discrepencies between the 2 set of nodes (the one running .0 and the one running .1), which would only lead to a temporary hard fork when someone try to withdraw from an account the likely very small missing amount or something similar. It would be a subtle bug to discover down the line as it could happens many many epochs after this reward difference and you'ld have node agreeing on a history and some node not agreeing.

The end result should autocorrect itself by consensus, if the prevaling block creator are running the newest version. the old node <= 0.8.1 should at some point realize they are creating less blocks on their version of the history.

The good news is that it's only when using the owner stake delegation certificate, all the other fragments are properly accounted for.

rinor:
Thank you.

Issue #1349, opened 2019-12-12
Limit concurrent pull requests


mzabaluev:
Currently, processing of `BlockMsg::AnnouncedBlock` will issue a header pull solicitation for every peer's announced block where the parent is missing from local storage. The header streams received in response are largely redundant. It would make sense to register header pulls already in progress by their `to` target block hash, and implement a policy that will cap the number of simultaneous pulls to the same target block with a small limit.

mzabaluev:
Same applies to block solicitation if the parent is present: there should be a list of peers that have announced a block, a solicitation should go out to the first few, and the block received from one peer and validated should cancel all outstanding request state. Dunno if this deserves a separate issue item.

Issue #1034, opened 2019-10-25
Limits on the number of simultaneous gRPC requests


mzabaluev:
It's convenient, sometimes probably necessary, to spawn a Tokio task to process a protocol request or handle a response stream. This, however, means that the number of tasks on the network runtime depends on the number of protocol requests in flight. To control memory usage, we need to ensure that limits on the number of simultaneous requests per gRPC connection are enforced.

PR #1365, opened 2019-12-14
Limit concurrent pull requests

CodeSandwich:
Fixes https://github.com/input-output-hk/jormungandr/issues/1349,
part of https://github.com/input-output-hk/jormungandr/issues/1034

@dkijania I had hard time testing these changes. No tests seem to fail even when block announcements are completely ignored. Some tests that look like they could detect that are feature gated and fail to compile, e.g. soak tests. This branch seems to be working for me, but could you check it with the QA before we merge it? It would be quite a big deal if block announcement and fetching was failing.

### DIFF
--- a/jormungandr/src/blockchain/process.rs
+++ b/jormungandr/src/blockchain/process.rs
@@ -6,6 +6,7 @@ use super::{
 };
 use crate::{
     blockcfg::{Block, FragmentId, Header},
+    blockchain::Checkpoints,
     intercom::{
         self, BlockMsg, ExplorerMsg, NetworkMsg, PropagateMsg, ReplyHandle, TransactionMsg,
     },
@@ -13,8 +14,12 @@ use crate::{
     stats_counter::StatsCounter,
     utils::{
         async_msg::{self, MessageBox, MessageQueue},
+        fire_forget_scheduler::{
+            FireForgetScheduler, FireForgetSchedulerConfig, FireForgetSchedulerFuture,
+        },
         task::TokioServiceInfo,
     },
+    HeaderHash,
 };
 use chain_core::property::{Block as _, Fragment as _, HasHeader as _, Header as _};
 use jormungandr_lib::interfaces::FragmentStatus;
@@ -29,11 +34,19 @@ use tokio::{
 use std::{sync::Arc, time::Duration};
 
 type TimeoutError = timeout::Error<Error>;
+type PullHeadersScheduler = FireForgetScheduler<HeaderHash, NodeId, Checkpoints>;
 
 const DEFAULT_TIMEOUT_PROCESS_LEADERSHIP: u64 = 5;
 const DEFAULT_TIMEOUT_PROCESS_ANNOUNCEMENT: u64 = 5;
 const DEFAULT_TIMEOUT_PROCESS_BLOCKS: u64 = 60;
 const DEFAULT_TIMEOUT_PROCESS_HEADERS: u64 = 60;
+
+const PULL_HEADERS_SCHEDULER_CONFIG: FireForgetSchedulerConfig = FireForgetSchedulerConfig {
+    max_running: 16,
+    max_running_same_task: 2,
+    command_channel_size: 1024,
+    timeout: Duration::from_millis(500),
+};
 
 pub struct Process {
     pub blockchain: Blockchain,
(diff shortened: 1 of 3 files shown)


dkijania:
I'll fix tests on master, after that can you rebase ? I'll also run private network test to see degration/improvement

dkijania:
Can you rebase now? non functional tests are fixed.

CodeSandwich:
Thank you! Unfortunately even on `master` running `cargo test -p jormungandr-integration-tests --all-features` causes
```
failures:
    networking::testnet::e2e_stake_pool
    non_functional::soak::test_100_transaction_is_processed
    non_functional::soak::test_blocks_are_being_created_for_more_than_15_minutes
    non_functional::stress::test_100_transaction_is_processed_in_10_packs

test result: FAILED. 94 passed; 4 failed; 2 ignored; 0 measured; 0 filtered out
```
Running `cargo test -p jormungandr-integration-tests --all-features -- --ignored` fails even faster:
```
failures:
    non_functional::stress::test_100_transaction_is_processed_in_10_packs_to_single_account

test result: FAILED. 1 passed; 1 failed; 0 ignored; 0 measured; 98 filtered out
```

dkijania:
I did another fix and test `non_functional::stress::test_100_transaction_is_processed_in_10_packs_to_single_account` passed on master, can you take a look ?

CodeSandwich:
Are you sure you've pushed the changes? I can't see any new commits from you on master or any other branch and the tests fail exactly the same.

jormungandr/src/utils/fire_forget_scheduler.rs
```diff
@@ -0,0 +1,286 @@
...
+/// Launching is defined by a closure (LAUNCHER), which also accepts additional data (DATA).
+/// Multiple instances of task with same TID, but different WID may be running in parallel.
+/// Launcher must be quick and non-blocking.
+///
+/// When task is scheduled, it's queued. When it's finally launched, a timeout is started.
+/// Until it runs out or task is declared complete it's considered running and it consumes parallel
+/// task execution limits. Finishing with timeout doesn't make task completed and other queued
+/// instances with same TID, but different WID may be run. Timed out instances are considered
+/// failed and aren't rescheduled. Finishing by declaration task completion cancels all task
+/// instances with the same TID.
```
mzabaluev: I'm not sure I get the semantics of this. If we want to run competing pulls against different peers, would that mean the pull tasks receive the same TID, but have different data?

jormungandr/src/utils/fire_forget_scheduler.rs
```diff
@@ -0,0 +1,286 @@
...
+        } else {
+            "unknown channel error"
+        };
+        Error::CommandSend(cause)
+    }
+}
+
+/// Schedule for fire-forget tasks
+///
+/// Each task has an ID (TID) and can be launched onto a worker identified with an ID (WID).
```
mzabaluev: Why do we need to concern ourselves with explicit worker IDs in this API?
It would be much simpler to have a single-dimensional table keyed with the target block ID, with a limited list (or `FuturesUnordered`) of running task objects in each entry, and maybe a list (no matter the order here, could be a `Vec` with LIFO) of pending pull states for tasks that will be launched when running tasks fail.

jormungandr/src/utils/fire_forget_scheduler.rs
CodeSandwich: They all should have same TID, but because they can be run in parallel, they need different WID. The most sensible WID in this case is `NodeID`, because the node will be processing the request. The DATA is not shared between tasks or workers, each scheduling has its own DATA, it can contain anything.

jormungandr/src/utils/fire_forget_scheduler.rs
CodeSandwich: The only reason WID is exposed is to de-duplicate requests and not allow flooding the same (possibly broken) node with copies of the same request while starving others. If that's not a concern, WID is useless and NodeId can be part of Data.

mzabaluev: Right. I still think these abstractions do not apply well to the subject matter, and we should not try to create this as a `utils` thing. Let's try with a purpose-specific `ConcurrentFetch` or something like that, and get rid of the generics where not needed: we only use the block hash for the grouping key (not "ID"), and the node ID as a discriminant for competing fetch jobs.

CodeSandwich:
@NicolasDP The node after joining ITN sends requests for blocks and chains correctly. Other nodes send correct responses, so they must be receiving them.

jormungandr/src/blockchain/process.rs
```diff
@@ -53,13 +74,26 @@ impl Process {
...
         })
     }
 
-    fn handle_input(&mut self, info: &TokioServiceInfo, input: BlockMsg) {
+    fn handle_input(
+        &mut self,
+        info: &TokioServiceInfo,
+        input: BlockMsg,
+        pull_headers_scheduler: &PullHeadersScheduler,
+        get_next_block_scheduler: &GetNextBlockScheduler,
```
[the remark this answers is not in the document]
CodeSandwich: Schedulers require `TokioServiceInfo`'s logger and executor. It's provided in `start`, so the only way to store it in `Process` would be an `Option`, which then would have to be unwrapped.

jormungandr/src/utils/fire_forget_scheduler.rs
CodeSandwich: Should we assume that all `ConcurrentFetch` jobs are always going to be identifiable by a `HeaderHash`-`NodeId` pair? It's easy to specialize `FireForgetSchedule` into `ConcurrentFetch` in current setup with a single line `type`, but the other way is irreversible.

mzabaluev: There may be custom logic involved. For example, we may choose not to start a new pull with a node where a slow pull for an earlier block ID is still in progress, but update the target block ID for the next pending pull job.

PR #1393, opened 2019-12-18
[Tests] Stability fix - Change reward test parameters

dkijania:


jormungandr-integration-tests/src/jormungandr/genesis/rewards.rs
```diff
@@ -11,7 +11,8 @@ pub fn collect_reward() {
     let (jormungandr, stake_pool_id) = startup::start_stake_pool(
         &actor_account,
         ConfigurationBuilder::new()
-            .with_slots_per_epoch(60)
+            .with_slots_per_epoch(20)
+            .with_consensus_genesis_praos_active_slot_coeff("1.0")
```
NicolasDP: Don't set `1.0` this is not a good value, it can creates more problem than anything.

```suggestion
            .with_consensus_genesis_praos_active_slot_coeff("0.999")
```

Issue #1384, opened 2019-12-17
Negative Balance


stakeada:
**Describe the bug**

I've set up a stake pool and now I have negative balance.

**Mandatory Information**
1. `jcli 0.8.2 (heads/v0.8.2-fe327f9a, release, linux [x86_64]) - [rustc 1.38.0 (625451e37 2019-09-23)]`
2. `jormungandr 0.8.2 (heads/v0.8.2-fe327f9a, release, linux [x86_64]) - [rustc 1.38.0 (625451e37 2019-09-23)]"`

**To Reproduce**

1. Fund my wallet:
```
# export PLEDGE_AMOUNT=500700000
# cardano-wallet transaction create $SHELLEY_ID --payment $PLEDGE_AMOUNT@addr1skyy52dhd2nls9sxdg0vtt8dguvacalgcxzhuax9r7y8fslnre702ux9k87
```
2. Create stake pool as is normally the case:

```
# jcli key generate --type=Curve25519_2HashDH > vrf.prv
# jcli key to-public < vrf.prv > vrf.pub
# jcli key generate --type=SumEd25519_12 > kes.prv
# jcli key to-public < kes.prv > kes.pub
# export TAX_LIMIT=1000000
# export TAX_RATIO="1/20"
# export TAX_FIXED=0
# jcli certificate new stake-pool-registration --tax-fixed $TAX_FIXED --tax-ratio $TAX_RATIO --kes-key $(cat kes.pub) --vrf-key $(cat vrf.pub) --owner $(cat owner.pub) --start-validity 0 --management-threshold 1 > stake-pool-registration.cert
# export FEE_CONSTANT=200000
# export FEE_COEFFICIENT=100000
# export FEE_POOL_REGISTRATION=500000000
# export FEE_STAKE_DELEGATION=400000
# export GENESIS_HASH=8e4d2a343f3dcf9330ad9035b3e8d168e6728904262f2c434a4f8f934ec7b676
# jcli transaction new | \
>   jcli transaction add-account $(cat owner.addr) $PLEDGE_AMOUNT  | \
>   jcli transaction add-certificate $(cat stake-pool-registration.cert) | \
>   jcli transaction finalize $(cat owner.addr) \
>     --fee-constant $FEE_CONSTANT \
>     --fee-coefficient $FEE_COEFFICIENT \
>     --fee-pool-registration $FEE_POOL_REGISTRATION \
>   > transaction.tmp
# jcli transaction make-witness \
[...]

PR #1397, opened 2019-12-18
Eliminate duplicates in transactions by address

enzoc4:
Since the transactions by address are recorded in a sequence, and not a set as before (since the pagination was included for this query), it is necessary to eliminate duplicates in inputs and outputs before adding them to the transactions by address index, as an address can be both an input and and output, causing duplicates.

This appears to be the cause of #1384, although it may still be necessary to implement a proper balance method, as the transactions don't include the fee information (I think).

### DIFF
--- a/jormungandr/src/explorer/mod.rs
+++ b/jormungandr/src/explorer/mod.rs
@@ -476,20 +476,20 @@ fn apply_block_to_addresses(mut addresses: Addresses, block: &ExplorerBlock) -> 
 
     for tx in transactions {
         let id = tx.id();
-        for output in tx.outputs() {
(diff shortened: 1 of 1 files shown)


NicolasDP:
> as the transactions don't include the fee information

of course it does. If you do `SUM(INPUTS) - SUM(OUTPUTS)` you have the fees of the transaction.

enzoc4:
> > as the transactions don't include the fee information
> 
> of course it does. If you do `SUM(INPUTS) - SUM(OUTPUTS)` you have the fees of the transaction.

Oh, sorry. That was a mistake. I was thinking about *rewards* and not fees

PR #1419, opened 2019-12-20
Index blocks by chain length and block time

eugene-babichenko:
Implements indexing blocks by chain length and block time which are mapped into block ids.

Index is updated every time the tip is updated and can handle switching between branches when necessary.

### DIFF
--- a/jormungandr/src/blockchain/index.rs
+++ b/jormungandr/src/blockchain/index.rs
@@ -1,0 +1,108 @@
+use crate::blockchain::{
+    chain::MAIN_BRANCH_TAG,
+    storage::{Storage, StorageError},
+};
+use chain_impl_mockchain::header::{Header, HeaderId};
+use std::collections::BTreeMap;
+use thiserror::Error;
+use tokio::{prelude::*, sync::lock::Lock};
+
+type HeaderToKey<K> = fn(Header) -> K;
+
+pub struct Index<K>
+where
+    K: std::cmp::Ord,
+{
+    transform: HeaderToKey<K>,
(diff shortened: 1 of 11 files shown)


eugene-babichenko:
Note that the REST API was added just to check if it works well and probably should not be merged.

jormungandr/src/blockchain/index.rs
```diff
@@ -0,0 +1,131 @@
...
+#[derive(Clone)]
+pub struct Index {
+    inner: Lock<IndexInternal>,
+}
+
+#[derive(Default)]
+struct IndexInternal {
+    chain_length_index: BTreeMap<ChainLength, HeaderId>,
+    block_date_index: BTreeMap<BlockDate, HeaderId>,
+}
```
NicolasDP: The this index is only valid in the context of a given state of the blockchain. If the `tip` changes the state of the blockchain will chain as we are moving to a new branch of the blockchain. This will mean some blocks indexed won't be in the blockchain anymore.

I would suggest using the HAMT instead here, and instead of a `Lock<InternalIndex>` to have a `Multiverse` of Index.

eugene-babichenko: Thanks for the suggestion I will look into that. So the idea basically is to have index for every branch and to select the appropriate index by the tip id?

NicolasDP: no need for the multiverse actually here. Using the HAMT is enough here. Every time you add value you create a new state of the index that you attach in the Branch when the branch is updated.

eugene-babichenko:
closing in favor #1797

PR #1507, opened 2020-01-06
Update chain-deps for rand and network API changes

mzabaluev:


NicolasDP:
adding the label for sub-network because it affects also the network and want to enhance that for later.

PR #1509, opened 2020-01-06
remove KV pairs with duplicate keys from logs

### LINKED ISSUE #1499: Jormungandr - logs - poldercast / policy log task / sub_task
**Describe the bug**
When logging 'move node to quarantine' message, 'task' key is used twice. 

**Mandatory Information**
jcli 0.8.5 (HEAD-3db0680, release, linux [x86_64]) - [rustc 1.40.0 (73528e339 2019-12-16)]
jormungandr 0.8.5 (HEAD-3db0680, release, linux [x86_64]) - [rustc 1.40.0 (73528e339 2019-12-16)]

**To Reproduce**
1. Start a node with debug log level a plain output.
2. At some point the following is logged: 
Jan 04 01:01:06.999 DEBG move node to quarantine, id: 02a29f423fa669a7ada6effc58c0cb383b3395a38cffe429, task: policy, task: poldercast
3. Key 'task' is used twice
4 Restart the node with json output.
5. 'policy' value is missing

**Expected behavior**
One of the value policy or poldercast should have another key

eugene-babichenko:
`slog` do not take care of duplicate keys in KV list which leads to errors
in log parsing, `DedupDrain` deals with such keys.

Fix #1499

### DIFF
--- a/jormungandr/src/settings/logging.rs
+++ b/jormungandr/src/settings/logging.rs
@@ -116,13 +116,147 @@ impl<D: Drain> Drain for DrainMux<D> {
     }
 }
 
+/// slog serializers do not care about duplicates in KV fields that can occur
+/// under certain circumstances. This serializer serves as a wrapper on top of
+/// the actual serializer that takes care of duplicates. First it checks if the
+/// same key was already. If so, this key is skipped during the serialization.
+/// Otherwise the KV pair is passed to the inner serializer.
+struct DedupSerializer<'a> {
+    inner: &'a mut dyn slog::Serializer,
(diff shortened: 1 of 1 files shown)


eugene-babichenko:
@mzabaluev I believe that this is an actual error/missing feature in the logging subsystem. There can be cases when we need to create a new task from the existing one and set an appropriate task name for the logger in this task.

We can use something like `"task" => "task1", "subtask" => "task1.1"` (I see this being done in a couple of places). Which do you thing is better?

eugene-babichenko:
Closing in favor of #1533

PR #1546, opened 2020-01-13
use one Runtime for all the services.

NicolasDP:
so previously we were pushing different runtimes for the different
services. this has created more issues then we thought. One of the
issue was that each runtime was supposed to be configured to have
specific set of parameters depending on the actual needs and we
never had time to do that. Now the new issue is that we are moving
away from the legacy tokio framework for a new async/await model
and this is creating more issues overall.

part of the work for #1547


mark-stopka:
@auskatodd, what does ` ~/logs/node.out` say?

PR #1568, opened 2020-01-15
Switch actix-web to rustls

CodeSandwich:
This is first step to upgrading actix-web. In 1.0 it dropped support for native-tls and kept only openssl and rustls.

### DIFF
--- a/doc/quickstart/02_passive_node.md
+++ b/doc/quickstart/02_passive_node.md
@@ -62,7 +62,9 @@ Description of the fields:
       - `file`: path to the log file.
 - `rest`: (optional) Configuration of the REST endpoint.
     - `listen`: _address_:_port_ to listen for requests
-    - `pkcs12`: (optional) Certificate file
+    - `tls`: (optional) enables TLS and disables plain HTTP if provided
+      - `cert_file`: path to server X.509 certificate chain file, must be PEM-encoded and contain at least 1 item
+      - `priv_key_file`: path to server private key file, must be PKCS8 with single PEM-encoded, unencrypted key
     - `cors`: (optional) CORS configuration, if not provided, CORS is disabled
       - `allowed_origins`: (optional) allowed origins, if none provided, echos request origin
       - `max_age_secs`: (optional) maximum CORS caching time in seconds, if none provided, caching is disabled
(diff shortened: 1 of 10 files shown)


doc/quickstart/02_passive_node.md
```diff
@@ -62,7 +62,9 @@ Description of the fields:
       - `file`: path to the log file.
 - `rest`: (optional) Configuration of the REST endpoint.
     - `listen`: _address_:_port_ to listen for requests
-    - `pkcs12`: (optional) Certificate file
+    - `tls`: (optional) enables TLS and disables plain HTTP if provided
+      - `server_certs_file`: path to server X.509 certificate chain file, must be PEM-encoded and contain at least 1 item
```
mzabaluev: Could do away with the `server_` prefix (we don't have a REST client side in jormungandr), and maybe bikeshed it as `cert_file` if it's going to carry just one self-signed certificate most of the time; that will be more consistent with the `*CertFile*` error variants.

NicolasDP:
Not to merge yet as we are preparing `master` for a release today

jormungandr/src/rest/server/error.rs
```diff
@@ -1,38 +1,28 @@
...
+use thiserror::Error;
 
-#[derive(Debug)]
+#[derive(Debug, Error)]
 pub enum Error {
-    Pkcs12LoadFailed(IoError),
-    Pkcs12Invalid(TlsError),
-    BindFailed(IoError),
+    #[error("failed to open server certificate file")]
+    ServerCertFileOpenFailed(#[source] io::Error),
```
mzabaluev: All these `Server` prefixes just add typing, please remove them.

The existing members (`ServerAlreadyStopped` and below) do make some narrative sense.

jormungandr/src/rest/server/mod.rs
```diff
@@ -37,23 +38,25 @@ pub struct ServerStopper {
 
 impl Server {
     pub fn start<F, H>(
-        pkcs12: Option<PathBuf>,
+        tls_config_opt: Option<TlsConfig>,
```
mzabaluev: Don't need the `_opt` suffix.

PR #1582, opened 2020-01-16
Apply ref on each synchronized block during bootstrap

### LINKED ISSUE #1496: JIP - short-term bootstrap improvement

JIP
---
1. Does NOT introduce [breaking-change](https://github.com/input-output-hk/jormungandr/issues?q=is%3Aissue+is%3Aopen+label%3Abreaking-change)
---

Will analyze only bootstrap from block0 case, but it applies to any bootstrap case.

When we start bootstraping, the node checks the latest block available in storage, and to be more exact it checks the value from `Tags` table:
```
Tags (
  name text primary key,
  hash blob not null,
  foreign key(hash) references BlockInfo(hash)
)
```
If we are starting with empty storage, once created, that table will be initially empty. After we retrieve the block0 (genesis block) from one of the trusted peers, after other tables `Blocks` and `BlockInfo` are populated, that table will be updated with:
`HEAD | block0 hash`

name | hash
-- | --
HEAD | 8e 4d 2a 34 3f 3d cf 93 30 ad 90 35 b3 e8 d1 68 e6 72 89 04 26 2f 2c 43 4a 4f 8f 93 4e c7 b6 76

This is related to this part from the logs:
```
Jan 03 14:38:42.223 DEBG retrieving block0 from network with hash 8e4d2a343f3dcf9330ad9035b3e8d168e6728904262f2c434a4f8f934ec7b676, task: init
Jan 03 14:38:42.250 INFO fetching block 8e4d2a343f3dcf9330ad9035b3e8d168e6728904262f2c434a4f8f934ec7b676, peer_address: 13.114.196.228:3000, block: 8e4d2a343f3dcf9330ad9035b3e8d168e6728904262f2c434a4f8f934ec7b676, task: init
Jan 03 14:44:20.670 INFO genesis block fetched, peer_address: 13.114.196.228:3000, block: 8e4d2a343f3dcf9330ad9035b3e8d168e6728904262f2c434a4f8f934ec7b676, task: init
```
---
Now once that completed the node goes on fetching/pulling other blocks starting for the `HEAD` hash value from `Tags` table, **block0** in this case.

```
Jan 03 14:46:37.072 INFO connecting to bootstrap peer 13.56.0.226:3000, peer_addr: 13.56.0.226:3000, task: bootstrap
Jan 03 14:46:37.272 DEBG pulling blocks starting from 8e4d2a343f3dcf9330ad9035b3e8d168e6728904262f2c434a4f8f934ec7b676, peer_addr: 13.56.0.226:3000, task: bootstrap
Jan 03 14:46:37.502 DEBG validated block, block_date: 0.530, hash: d79f2082ddd12ab881f7786f376ee1dc548774efa4823efe84e028fd7e72c63f, peer_addr: 13.56.0.226:3000, task: bootstrap
Jan 03 14:46:37.547 DEBG validated block, block_date: 0.532, hash: 01b5b36e20528aba650424fc7990736d6b9b2bf44ecc93dd1275eca57a73c026, peer_addr: 13.56.0.226:3000, task: bootstrap
...
```
While pulling and validating blocks the node **populates** `Blocks` and `BlockInfo` tables for each validated block, but leaves `Tags` table **unmodified** that in our example keeps pointing to block0.

- If everything goes as expected, the bootstrap completes and the nodes start receiving blocks from gossiping and when feasible updates the `Tags` table as needed with the last tip. _This case is not of interest for us_.

What is interesting is related to the cases where the first/following bootstrap attempts fail to complete after receiving some blocks. When this happens, no matter how many blocks the node validated, once it fails from one peer and connects to another peer, it ignores the last validated block and starts again from the block available in `Tags` table, block0 in this example.
So for example, if you were to boostrap right now and manage to get the blocks all the way to epoch 19 (as example), but for some reason the remote trusted node you were bootstraping from is loaded and you get disconnected, once your node connects to another trusted node, it starts bootstraping again from the first block (block0 in our example) and not from the last received and validated block from previous attempt (in our example we have epoch 19 blocks).

While there may be legitimate reasons for this implementation, right now (think of daedalus clients) it causes a very load to the bootstraping peers, which may be aleviated a bit if we were to safely change this behaviour.

While there is some work scheduled to improve the bootstrap process in long-term, in short-term maybe there are some benefits if the node were to update also the `Tags` table, so even if disconnected from one peer, upon next peer connection it continues from where the previous one left (always if possible, since it may not be possible in some cases and starting from genesis is legit).

michaeljfazio:
Resolves #1496 by applying each bootstrap block as it is streamed from peer. This allows bootstrap nodes to recover from peer failure and pick-up where it left off with another peer. This should reduce load significantly on bootstrap peers under load from constantly restarting peers.

### DIFF
--- a/jormungandr/src/network/bootstrap.rs
+++ b/jormungandr/src/network/bootstrap.rs
@@ -47,9 +47,6 @@ pub async fn bootstrap_from_peer(
 ) -> Result<Arc<Ref>, Error> {
     info!(logger, "connecting to bootstrap peer {}", peer.connection);
 
-    let blockchain2 = blockchain.clone();
-    let logger2 = logger.clone();
-
     let bootstrap = grpc::connect(peer.address(), None)
         .map_err(|e| Error::Connect { source: e })
         .and_then(|client: Connection<BlockConfig>| {
@@ -64,12 +61,9 @@ pub async fn bootstrap_from_peer(
             client
                 .pull_blocks_to_tip(&[tip_hash])
                 .map_err(|e| Error::PullRequestFailed { source: e })
-                .and_then(move |stream| bootstrap_from_stream(blockchain, tip, stream, logger))
-        })
-        .and_then(move |tip| {
-            blockchain::process_new_ref(logger2, blockchain2, branch, tip.clone())
-                .map_err(|e| Error::ChainSelectionFailed { source: e })
-                .map(|()| tip)
+                .and_then(move |stream| {
+                    bootstrap_from_stream(blockchain, branch, tip, stream, logger)
+                })
         });
 
     bootstrap.compat().await
@@ -77,6 +71,7 @@ pub async fn bootstrap_from_peer(
 
 fn bootstrap_from_stream<S>(
     blockchain: Blockchain,
+    branch: Tip,
     tip: Arc<Ref>,
     stream: S,
     logger: Logger,
@@ -85,16 +80,30 @@ where
     S: Stream<Item = Block, Error = NetworkError>,
     S::Error: Debug,
 {
-    let fold_logger = logger.clone();
-
     let block0 = blockchain.block0().clone();
+    let logger2 = logger.clone();
+    let blockchain2 = blockchain.clone();
 
     stream
         .map_err(|e| Error::PullStreamFailed { source: e })
         .filter(move |block| block.header.hash() != block0)
-        .fold(tip, move |_, block| {
-            handle_block(blockchain.clone(), block, fold_logger.clone())
+        .and_then(move |block| handle_block(blockchain.clone(), block, logger.clone()))
+        .and_then(move |new_tip| {
+            // A further optimization could be to process bootstrap blocks in batches (e.g. of 100)
+            // and then only call process_new_ref on the last value of each processed batch. However,
+            // at this time, it would seem processing_new_ref for every bootstrap block does not
+            // introduce too much overhead. Thus such an optimization may be premature and/or
+            // unnecessary.
+            blockchain::process_new_ref(
+                logger2.clone(),
+                blockchain2.clone(),
+                branch.clone(),
+                new_tip.clone(),
+            )
+            .map_err(|e| Error::ChainSelectionFailed { source: e })
+            .map(|()| new_tip)
         })
+        .fold(tip, |_, block| future::ok(block))
 }
 
 fn handle_block(


jormungandr/src/network/bootstrap.rs
```diff
@@ -64,44 +61,49 @@ pub async fn bootstrap_from_peer(
...
-        .fold(tip, move |_, block| {
-            handle_block(blockchain.clone(), block, fold_logger.clone())
+        .and_then(move |block| handle_block(blockchain.clone(), block, logger.clone()))
+        .and_then(move |new_tip| {
+            // A further optimization could be to process bootstrap blocks in batches (e.g. of 100)
+            // and then only call process_new_ref on the last value of each processed batch. However,
+            // at this time, it would seem processing_new_ref for every bootstrap block does not
+            // introduce too much overhead. Thus such an optimization may be premature and/or
+            // unnecessary.
+            blockchain::process_new_ref(logger2.clone(), blockchain2.clone(), branch.clone(), new_tip.clone())
```
mzabaluev: No need to clone values other than possibly `new_tip`

jormungandr/src/network/bootstrap.rs
```diff
@@ -64,44 +61,49 @@ pub async fn bootstrap_from_peer(
...
         })
+        .fold(tip, |_, block| future::ok(block))
 }
 
 fn handle_block(
     blockchain: Blockchain,
     block: Block,
     logger: Logger,
-) -> impl Future<Item = Arc<Ref>, Error = Error> {
+) -> impl Future<Item=Arc<Ref>, Error=Error> {
```
mzabaluev: This makes rustfmt unhappy, and looks like a gratuitous change.

mzabaluev:
Good idea overall. It could be improved performance-wise by only applying at the end like before, but also doing this in the error handling closure (or in a `.then` closure handling both outcomes).

michaeljfazio:
Formatted code using ```rustfmt 1.4.9-stable (33e36670 2019-10-07)```

jormungandr/src/network/bootstrap.rs
michaeljfazio: I thought so as well. However, the compiler emphatically disagrees. I think the move semantics are different because this is a stream future? Happy to take advice otherwise.

jormungandr/src/network/bootstrap.rs
michaeljfazio: Changed my IDE settings to automatically format on save using ```rustfmt 1.4.9-stable (33e36670 2019-10-07)```

I've force pushed an update to the file with new formatting applied.

PR #1587, opened 2020-01-17
keep more room for nodes that are actually giving us blocks

NicolasDP:
add new `p2p.max_connections_threshold` configuration parameter. Default is set to `64`.

this value is meant to reserve more rooms for block propagating connections.

- [x] is dropping the connections enough to close it?
- [ ] shall we be more **aggressive** with the GCing peer?

### DIFF
--- a/jormungandr/src/network/mod.rs
+++ b/jormungandr/src/network/mod.rs
@@ -156,7 +156,11 @@ impl GlobalState {
                 .into(),
         );
 
-        let peers = Peers::new(config.max_connections, logger.clone());
+        let peers = Peers::new(
+            config.max_connections,
+            config.max_connections_threshold,
+            logger.clone(),
+        );
 
         GlobalState {
             block0_hash,
@@ -278,6 +282,28 @@ pub fn start(
                 .for_each(move |_| Ok(tp2p.force_reset_layers())),
         );
     }
+
+    let peers = global_state.peers.clone();
+    let gc_err_logger = global_state.logger.clone();
+    let gc_info_logger = global_state.logger.clone();
+    global_state.spawn(
+        Interval::new_interval(Duration::from_secs(13))
+            .map_err(move |e| {
+                error!(gc_err_logger, "interval timer error: {:?}", e);
+            })
+            .for_each(move |_| {
+                if let Some(number_gced) = peers.gc() {
+                    warn!(
+                        gc_info_logger,
+                        "Peer GCed {number_gced} nodes",
+                        number_gced = number_gced
+                    );
+                } else {
+                    debug!(gc_info_logger, "Peer GCed nothing");
+                }
+                Ok(())
+            }),
+    );
 
     let gossip = Interval::new_interval(global_state.config.gossip_interval.clone())
         .map_err(move |e| {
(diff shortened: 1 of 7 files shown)


jormungandr/src/network/mod.rs
```diff
@@ -279,6 +283,28 @@ pub fn start(
         );
     }
 
+    let peers = global_state.peers.clone();
+    let gc_err_logger = global_state.logger.clone();
+    let gc_info_logger = global_state.logger.clone();
+    global_state.spawn(
+        Interval::new_interval(Duration::from_secs(13))
```
mzabaluev: I'd like this to be a named const defined at the beginning of the file.

NicolasDP: done

jormungandr/src/network/p2p/comm/peer_map.rs
```diff
@@ -88,6 +88,30 @@ impl PeerMap {
...
+    pub fn clear(&mut self) {
+        self.map.clear()
+    }
+
+    pub fn gc(&mut self, at_most: usize) -> usize {
+        let mut len = 0;
+
+        for entry in self.map.entries() {
+            if let Some(_peer) = entry.get().stats.last_block_received() {
+                // TODO: check for more reasons to evict
```
michaeljfazio: Perhaps the criteria to remove a peer could be provided (i.e. strategy pattern). This would be a bit more modular and allow for more sophisticated logic to be applied without complicating this code.

As a follow up thought. Would it be reasonable to evict peers that forward blocks for a slot that are signed by the same originating node. Such a change would starve adversarial nodes from creating unnecessary forks? Of course, initially one might want to also put logic in to prevent a node from forwarding double-signed single-origin blocks in the first place so as not to create a situation in which honest nodes are dropped for unwittingly participating in fork propagation.

NicolasDP: Thanks, this is interesting. Indeed the criteria can be evolved. Right now this is a bit of an out of the blue enhancement to see how things can be improved that way.

regarding the twin block creation issue this is something linked to incentive model and needs way more work than this.

NicolasDP: thanks. this is interesting. this is still an out of the blue change to explore how we can improve the server side reliability side of things rather than a long term solution (more to come on this by the end of the month).

regarding the sibling blocks issue, this is a flaw deep down in the incentive model. Our team is exploring solutions to propose for this, but it cannot be done by blacklisting nodes as it may trigger a more dangerous domino effect of blacklisting too aggressively nodes.

michaeljfazio: > regarding the sibling blocks issue, this is a flaw deep down in the incentive model. Our team is exploring solutions to propose for this, but it cannot be done by blacklisting nodes as it may trigger a more dangerous domino effect of blacklisting too aggressively nodes.

Understood. I'd be curious to see how the incentive model alone (at least from a delegation/stake perspective) could solve this given that there is negligible cost involved in running a node capable of signing blocks and stake delegators will at the end of the day delegate to those pools that perform the best. It's fairly clear that pools which are double signing blocks are probably doing so because creating a reliable node (at the moment) is difficult given network stability issues and the lack of native support for managing a cluster of nodes in a H/A setup with fault tolerant and rapid leader election. Granted this will be solved in time, but this still leaves this vector open for exploitation by nodes that solely want to disrupt the network.

Then again... I suppose once stake pool pledge comes into effect it would be conceivable that dishonest nodes could have their pledge destroyed if enough nodes observe their behaviour and deem it malicious. Though, this does leave the possibility that a node is not actually dishonest but misconfigured or simply comes to be in a state where it runs multiple leaders by accident (which we have already observed is the case with a few several honest node operators who have implemented faulty leader election mechanisms).

A feel like a double-pronged approach will be required in the future that targets both the incentive model and a technical solution.

Food for thought in any case 👍

Issue #1526, opened 2020-01-09
Jormungandr Integration tests improvements


dkijania:
**Is your feature request related to a problem/context ? Please describe if applicable.**
Epic for jormungandr-integration-tests improvements. All improvements/refactoring are intended to remove duplication against jormungandr-lib and speed up tests mainly but not using jcli which is slow but rather  jormungandr-lib structure (of course apart from situation in which we would like to test jcli)

- [x] move wallet from jormungandr-scenario-tests to jormungandr-lib (required to get rid off AddressDataProvider and its dependencies)
- [x] remove AddressDataProvider and use Wallet struct from jormungandr-lib (dependency to #1 req)
- [x] remove duplication for genesis model between 
- [x] remove duplication for jormungandr configuration #1583
- [x] remove duplication for jormungandr secret 
- [x] migrate form grpc proto
- [x] introduce better interface for Jormugandr struct which will hold its internal away from tests (process,logger,config)
- [x] introduce better interface for JCLI struct which hides it's internals
- [x] move fragment sender to testing utils
- [x] move measurement to testing utils
- [x] move node synchronization to testing utils

PR #1589, opened 2020-01-17
[Tests] Remove genesis model duplication

dkijania:
Removed `BlockchainConfig` structs and use existing one from jormungandr-lib. 
Updated usages.

Part of epic: #1526


NicolasDP:
Tests are not passing anymore though @dkijania

PR #1594, opened 2020-01-17
cleanup the sync Mutex from the PeerMap

### LINKED ISSUE #1591: remove the sync Mutex to an async Mutex in the PeerMap collection
Currently we have still have sync mutex https://github.com/input-output-hk/jormungandr/blob/835e373d7021efb3b0208a4dde7259b1773993af/jormungandr/src/network/p2p/comm.rs#L379-L383

ideally it needs to be replaced to:

```rust
pub struct Peers {
    mutex: Lock<peer_map::PeerMap>,
    logger: Logger,
    capacity_threshold: usize,
}
```

NicolasDP:
resolves #1591

### DIFF
--- a/jormungandr/src/network/mod.rs
+++ b/jormungandr/src/network/mod.rs
@@ -262,8 +262,9 @@ pub fn start(
     let initial_nodes = global_state.topology.view(poldercast::Selection::Any);
     let self_node = global_state.topology.node();
     for node in initial_nodes {
-        connect_and_propagate_with(node, global_state.clone(), channels.clone(), |comms| {
-            let gossip = Gossip::from_nodes(iter::once(self_node.clone().into()));
+        let self_node_copy = self_node.clone();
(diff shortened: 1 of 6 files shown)


jormungandr/src/network/mod.rs
```diff
@@ -323,50 +322,48 @@ fn handle_network_input(
...
         NetworkMsg::PullHeaders { node_id, from, to } => {
-            state.peers.pull_headers(node_id, from.into(), to);
-            Ok(())
+            B(B(A(state.peers.pull_headers(node_id, from.into(), to))))
         }
         NetworkMsg::PeerInfo(reply) => {
-            let infos = state.peers.infos();
-            reply.reply_ok(infos);
-            Ok(())
+            B(B(B(state.peers.infos().map(|infos| reply.reply_ok(infos)))))
```
mzabaluev: This is the nesting at which a dedicated enum begins to look more attractive.

mzabaluev:
Fails the jcli build, though.

PR #1597, opened 2020-01-18
Initial watchdog commit

NicolasDP:
This is the initial ground work for the Jörmungandr service update as part of #1488. This is a PoC on how this can be down and to be used as ground work to prepare the #1488 epic's subtasks.

The idea is to provide multiple enhance services/controls over the service so we can provide:

* dynamic update of a service settings, this will be greatly appreciated by the stake pool operators to dynamically change paramters such as public address, topic of interests, the mempool fragment selections parameters (how to chose transactions to put in the blocks)...
* state saving of a service. this will allow saving a global state and to allow faster reboot. We could then store the poldercast p2p state for example and allow rebooting without having to depend solely on the trusted peers;
* easier to add new services, as they are isolated from each other it will become easier to just start a service to extend new features. This will be greatly appreciated for the hackers in the community with a bit of rust knowledge, facilitate integration in 3rd party ecosystem (we could add an audit to monitor the blockchain transactions, a new set of APIs for the a user with specific needs, etc...)

A simplify schema of the core design could be: 

Service Management_ Yggdrasil

The following example if extracted from one of the integration tests. It shows 2 services establishing an intercom relation between each other and sending messages to each others.

```rust
#[async_trait]
impl Service for Ping {
    // ... missing entry not relevant ...

    async fn start(mut self) {
        // query the intercom connection to the `Pong` service so we can start
        // sending them PongMsg 
        let mut pong = self.state.watchdog_query.intercom::<Pong>();

        while let Some(msg) = self.state.intercom_receiver.recv().await {
            dbg!(msg);
            delay_for(Duration::from_millis(50)).await;
            if let Err(_err) = pong.send(PongMsg).await {
                dbg!(err);
                break;
            }
        }
    }
}

#[async_trait]
impl Service for Pong {
    // ... missing entry not relevant ...

    async fn start(mut self) {
        let mut ping = self.state.watchdog_query.intercom::<Ping>;

        // this assume that the Ping service was started before the Pong service
        ping.send(PingMsg).await.unwrap();

        while let Some(msg) = self.state.intercom_receiver.recv().await {
            dbg!(msg);
            delay_for(Duration::from_millis(50)).await;
            if let Err(err) = ping.send(PingMsg).await {
                dbg!(err);
                break;
            }
        }
    }
}

```

still to do:

- [x] documentation
- [x] more unit tests
- [x] proc-macro with autoderive of the `CoreServices`
- [ ] more tooling and control
    - [ ] update a running service settings live
    - [ ] update a running service state live
    - [ ] watchdog handler on state update (store the state?)
    - [x] query a service health/status
    - [x] query service general health/status from watchdog
- [x] querying the intercom or sending messages via the intercom needs a bit of work
       to provide better guarantee (i.e. query an new intercom handler on failure, await
       a service to finish booting etc...).
 - [x] test/example with a legacy Future (tokio.1-22).


mzabaluev:
Impressive. The name may be a bit confusing, because a watchdog in traditional sense is only a part of what this infrastructure does.

jormungandr-watchdog/src/service/control.rs
```diff
@@ -0,0 +1,82 @@
...
+    Kill,
+}
+
+pub struct Controller {
+    sender: Sender<Control>,
+    receiver: Receiver<Control>,
+}
+
+pub struct ControlReader {
+    receiver: Receiver<Control>,
```
mzabaluev: `watch::Receiver` is only guaranteed to receive the last item. Is this intended for `Control` command channel to be lossy in this way: only the last command matters?

jormungandr-watchdog/src/service/control.rs
```diff
@@ -0,0 +1,82 @@
...
+}
+
+impl ControlReader {
+    pub async fn updated(&mut self) -> Option<Control> {
+        self.receiver.recv().await
+    }
+}
+
+impl Future for ControlReader {
+    type Output = Option<Control>;
```
mzabaluev: What does `None` mean here? If it's not meant to be normal to drop the sender before sending the final command, perhaps make it a `Result` with an informative error?

jormungandr-watchdog/src/service/intercom.rs
```diff
@@ -0,0 +1,268 @@
...
+    received_counter: Arc<AtomicU64>,
+    stats: Arc<Mutex<Stats>>,
+}
+
+impl IntercomMsg for NoIntercom {}
+
+#[derive(Debug, Clone, Copy)]
+pub struct IntercomStatus {
+    /// number of messages that has been sent through the intercom
+    pub number_sent: u64,
```
mzabaluev: This looks like entirely too many counters to manage.

jormungandr-watchdog/src/service/intercom.rs
```diff
@@ -0,0 +1,268 @@
...
+                    ),
+                }
+            }
+            Ok(Err(err)) => {
+                tracing::error!(error = %err, "cannot connect to the service");
+                Err(err)
+            }
+            Err(err) => {
+                // we assume the server will always reply one way or another
+                unreachable!(
```
mzabaluev: If this can occur in real life, better use `panic!`?

jormungandr-watchdog/src/service/mod.rs
```diff
@@ -0,0 +1,374 @@
...
+
+pub type ServiceIdentifier = &'static str;
+
+#[async_trait]
+pub trait Service: Send + Sized + 'static {
+    const SERVICE_IDENTIFIER: ServiceIdentifier;
+
+    type State: State;
+    type Settings: Settings;
+    type Intercom: IntercomMsg;
```
mzabaluev: Perhaps name the associated type `IntercomMsg` to avoid confusion.

jormungandr-watchdog/src/service/mod.rs
```diff
@@ -0,0 +1,374 @@
...
+
+    fn prepare(service_state: ServiceState<Self>) -> Self;
+
+    async fn start(self);
+}
+
+pub trait ServiceManagerTrait {
```
mzabaluev: Gah, please, no `Trait` traits :slightly_frowning_face:

jormungandr-watchdog/src/service/control.rs
NicolasDP: So far this is my finding. Trying to do something else was more complex as it needed to store more state in the service. However this is still open for comment/discussion.

jormungandr-watchdog/src/service/intercom.rs
NicolasDP: it should not and if it does, it is a bug

jormungandr-watchdog/src/service/control.rs
NicolasDP: Also, there is only 2 control commands: `Shutdown` and `Kill`. Usually if you miss the `Shutdown` and it is now time to `Kill` missing the first one is not a problem but actually a welcome simplification of the process of handling the control commands.

jormungandr-watchdog/src/service/control.rs
NicolasDP: This should really be only a stream and not a `Future`. It is used as a `Stream` in the `service/mod.rs`'s `select!`. If you have a suggestion to use the `Stream` instead of the future here that would be welcome?

jormungandr-watchdog/src/service/mod.rs
NicolasDP: What do you suggest? 

* it is not meant to be implemented/used outside of `jormungandr-watchdog-derive`;
* it clashes name with the `struct ServiceManager`

jormungandr-watchdog/src/service/control.rs
NicolasDP: I've removed the future and use `updated` instead. easier and cleaner

jormungandr-watchdog/src/service/mod.rs
mzabaluev: Maybe "verb" the trait, like `ManageService`.

Issue #1610, opened 2020-01-20
Add Peer Stats to Node Stats Output


michaeljfazio:
**Is your feature request related to a problem/context ? Please describe if applicable.**

Many pool operators monitor the number of p2p connections for their nodes using commands such as:

```
watch -n 5 "curl -G -H 'Cache-Control: no-cache' http://localhost:3100/api/v0/network/p2p/available 2>/dev/null | jq '.' | grep id | wc -l"
```
This is problematic because this call is very inefficient as per documentation:

![image](https://user-images.githubusercontent.com/1845178/72749436-01b6b980-3bf5-11ea-83c4-ab4a52167127.png)

Indeed some operators have identified that calling these apis regularly (every ~5s) quickly results in the entire node locking up and crashing.

**Describe the solution you'd like**

A better alternative would just be to show the stat counts in peer stats call, e.g:

```
MBP:jormungandr michaelfazio$ jcli rest v0 node stats get --host http://localhost:3100/api
---
blockRecvCnt: 0
lastBlockContentSize: 315
lastBlockDate: "37.40949"
lastBlockFees: 800000
lastBlockHash: eb322e2f7ac960cfa8b5c2c29afb853366c5092fd35594d1c6624b6c4ce85630
lastBlockHeight: "115750"
lastBlockSum: 93667164000
lastBlockTime: "2020-01-20T17:58:35+00:00"
lastBlockTx: 1
lastReceivedBlockTime: ~
peerAvailableCnt: "2"
peerQuarantinedCnt: "5"
peerUnreachableCnt: "0"
state: Running
txRecvCnt: 0
uptime: 220
version: jormungandr 0.8.6-546497fc+
```

The count values would not need to iterate the entire list of available, quarantined or unreachable nodes, but rather just return the size of their underlying data structures.

michaeljfazio:
Implemented in #1611

PR #1611, opened 2020-01-20
Added p2p stats to node stat output

michaeljfazio:
Adds p2p stats to node stat output (REST API) as requested in #1610

### DIFF
--- a/jormungandr/src/rest/v0/handlers.rs
+++ b/jormungandr/src/rest/v0/handlers.rs
@@ -156,6 +156,8 @@ async fn create_stats(context: &FullContext) -> Result<serde_json::Value, Error>
         .map_err(|e| ErrorInternalServerError(format!("Block value calculation error: {}", e)))?;
     let stats = &context.stats_counter;
     let tip_header = tip.header();
+    let stats = &context.stats_counter;
+    let nodes_count = &context.p2p.nodes_count();
     Ok(json!({
         "txRecvCnt": stats.tx_recv_cnt(),
         "blockRecvCnt": stats.block_recv_cnt(),
@@ -169,6 +171,9 @@ async fn create_stats(context: &FullContext) -> Result<serde_json::Value, Error>
         "lastBlockContentSize": tip_header.block_content_size(),
         "lastBlockSum": block_input_sum.0,
         "lastBlockFees": block_fee_sum.0,
+        "peerAvailableCnt": nodes_count.all_available_nodes_count().to_string(),
+        "peerUnreachableCnt": nodes_count.all_unreachable_nodes_count().to_string(),
+        "peerQuarantinedCnt": nodes_count.all_quarantined_nodes_count().to_string(),
     }))
 }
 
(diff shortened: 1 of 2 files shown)


AndrewWestberg:
@michaeljfazio Another common one is to get connected peers that takes a lot of processing power.
```
jcli rest v0 network stats get -h http://localhost:3100/api | grep addr | wc -l
```
Could this be added to this PR?

michaeljfazio:
> @michaeljfazio Another common one is to get connected peers that takes a lot of processing power.
> 
> ```
> jcli rest v0 network stats get -h http://localhost:3100/api | grep addr | wc -l
> ```
> 
> Could this be added to this PR?

From what I can grok of the network peer handler, the process for getting peer info is a bit less straight forward than it is for p2p topology info. I'm not entirely confident that I can easily obtain the peer count without introducing a new performance problem. @NicolasDP do you have any additional insight on the matter?

In any case we could defer your request to another issue if it means holding up this one.

mzabaluev:
We have updated the REST service to Actix-web 2.0. @michaeljfazio, please resolve the resulting conflicts in the `handlers.rs` file, it should be straightforward. Or comment if you need help with resolving.

michaeljfazio:
> Acquiring the lock three times to compose a single response is excessive. Please change this to a method retrieving all data in a single go in a struct.

Fair point. Changes made 🎉

vincenthz:
Thanks @michaeljfazio this is looking even better now !

jormungandr/src/rest/v0/handlers.rs
```diff
@@ -156,6 +156,8 @@ async fn create_stats(context: &FullContext) -> Result<serde_json::Value, Error>
         .map_err(|e| ErrorInternalServerError(format!("Block value calculation error: {}", e)))?;
     let stats = &context.stats_counter;
     let tip_header = tip.header();
+    let stats = &context.stats_counter;
```
rinor: Delete this please since is already present.

jormungandr/src/rest/v0/handlers.rs
```diff
@@ -169,6 +171,9 @@ async fn create_stats(context: &FullContext) -> Result<serde_json::Value, Error>
         "lastBlockContentSize": tip_header.block_content_size(),
         "lastBlockSum": block_input_sum.0,
         "lastBlockFees": block_fee_sum.0,
+        "peerAvailableCnt": nodes_count.all_available_nodes_count().to_string(),
+        "peerUnreachableCnt": nodes_count.all_unreachable_nodes_count().to_string(),
+        "peerQuarantinedCnt": nodes_count.all_quarantined_nodes_count().to_string(),
```
rinor: I would convert peer -> peers and lose the Cnt suffix.

michaeljfazio: @rinor I tried to follow the convention set forth by the precedent of ```txRecvCnt``` and ```blockRecvCnt```

PR #1620, opened 2020-01-22
[Tests] Move config structs to jormungandr lib

dkijania:
Removed duplication across `jormungandr-integration-tests` and `jormungandr-scenario-tests` (Config structs: NodeConfig/NodeSecret and children)
Updated usages.

Part of epic: #1526


jormungandr-lib/src/interfaces/config/mod.rs
```diff
@@ -1,3 +1,9 @@
+mod log;
 mod mempool;
+mod node;
+mod secret;
 
-pub use mempool::Mempool;
+pub use log::*;
+pub use mempool::*;
+pub use node::*;
+pub use secret::*;
```
mzabaluev: Please list the public exports explicitly here.

jormungandr-lib/src/interfaces/config/node.rs
```diff
@@ -0,0 +1,64 @@
...
+use std::net::SocketAddr;
+use std::path::PathBuf;
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct Rest {
+    pub listen: SocketAddr,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct P2p {
```
mzabaluev: Should we also deduplicate the config structures in jormungandr?

dkijania: I think yes. There were some pr's after which test code had to be updated as well. I will try to do it in other pr

PR #1640, opened 2020-01-24
Changed peers counts from strings to int

michaeljfazio:
Peer stats should have been integer not string values in REST output.


rinor:
Just reminder to myself that this needs docs/api update before next release.

Issue #1637, opened 2020-01-24
Stuck at listening and accepting gRPC connections


skokasik:
I tried running 0.8.7 today after a sucessful 0.8.6 but latest version seem to be problematic for me. I see that not everyone has the same issue i am getting. 

it stuck here for a long time. Never receive any block. 

```
Jan 24 06:36:44.889 INFO update current branch tip, peer_addr: 13.56.0.226:3000, task: bootstrap
Jan 24 06:36:44.906 INFO initial bootstrap completed, peer_addr: 13.56.0.226:3000, task: bootstrap
Jan 24 06:36:44.915 INFO listening and accepting gRPC connections, local_addr: 0.0.0.0:3000, task: network
```

This shows only uptime 81 , but i have run it for much longer without luck. 
```
blockRecvCnt: 0
lastBlockContentSize: 0
lastBlockDate: "41.16510"
lastBlockFees: 0
lastBlockHash: e12773816bb432ad1415ceeb00690e61d09d910ea9ac95acaaa75c7e729c6d31
lastBlockHeight: "126157"
lastBlockSum: 0
lastBlockTime: "2020-01-24T04:23:57+00:00"
lastBlockTx: 0
lastReceivedBlockTime: ~
peerAvailableCnt: "0"
peerQuarantinedCnt: "0"
peerUnreachableCnt: "0"
state: Running
txRecvCnt: 0
uptime: 81
version: jormungandr 0.8.7-364cd84e+


```

I notice the the Recv-Q is being filled to 1025 and then it stops.

```
State     Recv-Q     Send-Q          Local Address:Port           Peer Address:Port     
LISTEN    0          128             127.0.0.53%lo:domain              0.0.0.0:*        
LISTEN    0          128                   0.0.0.0:ssh                 0.0.0.0:*        
LISTEN    647        1024                  0.0.0.0:3000                0.0.0.0:*        
LISTEN    0          1024                127.0.0.1:3100                0.0.0.0:*        
LISTEN    0          2048                127.0.0.1:3101                0.0.0.0:*        
LISTEN    0          1024                  0.0.0.0:3200                0.0.0.0:*
```

TCPdump:
lots of retransmittion
[tcpdump.zip](https://github.com/input-output-hk/jormungandr/files/4107199/tcpdump.zip)

kernel version:
5.0.0-1028-azure

fish2plain:
Same issue on `4.15.0-1057-aws #59-Ubuntu` t2.medium EC2.

[...]

Issue #1635, opened 2020-01-23
Jormungandr v0.8.7 fails to bootstrap


A-Caccese:
**Describe the bug**
After the upgrade to Jorm. 8.7 I'm unable to bootstrap on Ubuntu 18.04. I tried to delete the storage and also to delisting the unreacheble trusted peers but nothing works. Reinstalled Jorm. 8.6, it works without problem. Tried to reinstall Jorm. 8.7 after 8 hours, but I had the same problem.

**Mandatory Information**
1. jcli 0.8.7 (linux [x86_64]) - [rustc 1.40.0 (73528e339 2019-12-16)]
2. jormungandr 0.8.7 (linux [x86_64]) - [rustc 1.40.0 (73528e339 2019-12-16)]

**Additional context**
Logs tail:
Jan 23 19:59:21.464 DEBG validated block, block_date: 41.788, hash: eacdf60c15eda399a898322450b0264c6037c3b08c12d2388cbf3cbb69b75171, peer_addr: 3.125.183.71:3000, task: bootstrap
Jan 23 19:59:21.471 INFO update current branch tip, peer_addr: 3.125.183.71:3000, task: bootstrap
Jan 23 19:59:21.475 DEBG validated block, block_date: 41.790, hash: e4072bf7ef6dca7c3e07ef800e3a12fb28106b0ff80f150d33f41803d2e06371, peer_addr: 3.125.183.71:3000, task: bootstrap
Jan 23 19:59:21.479 INFO update current branch tip, peer_addr: 3.125.183.71:3000, task: bootstrap
Jan 23 19:59:21.482 DEBG validated block, block_date: 41.814, hash: e177586e1c9b52d3caf31f445d1584744f5ec7d6b2646402cde0da9a0f3ef102, peer_addr: 3.125.183.71:3000, task: bootstrap
Jan 23 19:59:21.486 INFO update current branch tip, peer_addr: 3.125.183.71:3000, task: bootstrap
Jan 23 19:59:21.491 DEBG validated block, block_date: 41.832, hash: 929a6d313ac8cc95cb07b1c061f36a554e4888d819ccf68788c6a22bdc9808a0, peer_addr: 3.125.183.71:3000, task: bootstrap
Jan 23 19:59:21.494 INFO update current branch tip, peer_addr: 3.125.183.71:3000, task: bootstrap
[...]

Issue #1617, opened 2020-01-21
Node stuck bootstrapping forever with no warnings or errors in the log


shawnim:
**Describe the bug**

After starting, jormungandr stayed in Bootstrapping state for over 20 hours without writing any warnings or errors to the log file.
Note that at the time this happened there were only 2 to 4 of the trusted peers responding to tcpping and only those were included in the node config.yaml.

**Mandatory Information**

1. `jcli --full-version` output;

jcli 0.8.6 (HEAD-76863977, release, linux [x86_64]) - [rustc 1.40.0 (73528e339 2019-12-16)]

2. `jormungandr --full-version` output;

jormungandr 0.8.6 (HEAD-76863977, release, linux [x86_64]) - [rustc 1.40.0 (73528e339 2019-12-16)]

**To Reproduce**

1. Change node config.yaml to only include 2 to 4 trusted peers.
2. Start jormungandr as a passive node.
3. Watch logs.
4. Wait forever.

**Expected behavior**

Node should write warnings or errors if it is not able to bootstrap and it should have a configurable amount of time after which it would exit if it does not reach a Running state.

**Additional context**

Jormungandr is running on Ubuntu 18.04 on an AWS t3a.small instance with 2 vCPUs, 2G memory, shared network up to 5Gbps.

$ status

---
state: Bootstrapping
version: jormungandr 0.8.6-76863977
nodeCount: 2
date: Mon Jan 20 09:10:07 UTC 2020

$ count-open-peers

9 open peers

$ tail-logs 

Jan 19 12:57:22.083 WARN Node started without path to the stored secret keys (not a stake pool or a BFT leader), task: init

config.yaml

---
log:
- output: stderr
  format: plain
  level: warn
p2p:
  topics_of_interest:
    blocks: normal
    messages: low
  listen_address: "/ip4/0.0.0.0/tcp/20002"
  public_address: "/ip4/52.70.79.184/tcp/20002"
  public_id: 189c346de1704ee6cf54e7fd4de9420638acb82372a76728
  max_connections: 160
  max_unreachable_nodes_to_connect_per_event: 16
  gossip_interval: 10s
  trusted_peers:
[ ...
 Filled in automatically at start based on what peers respond as open to tcpping
 ... ]
rest:
  listen: "127.0.0.1:20001"
storage: /home/ubuntu/chains/jormungandr-itn_rewards_v1
mempool:
    pool_max_entries: 20000
    fragment_ttl: 30m
[...]

Issue #1585, opened 2020-01-17
Bootstrap failed 


BitcoinMicroPaymentNetwork:
**Describe the bug**
Jan 17 09:58:29.675 WARN unable to reach peer for initial bootstrap, reason: HTTP/2.0 connection error, peer_addr: 52.8.15.52:3000, task: bootstrap
Jan 17 09:58:33.135 WARN initial bootstrap failed, error: PullRequestFailed { source: Error { code: Unknown, source: Status { code: Unknown, message: "connection error: broken pipe" } } }, peer_addr: 3.125.75.156:3000, task: bootstrap
Jan 17 09:59:48.507 WARN unable to reach peer for initial bootstrap, reason: HTTP/2.0 connection error, peer_addr: 13.114.196.228:3000, task: bootstrap
Jan 17 09:59:49.364 WARN initial bootstrap failed, error: PullRequestFailed { source: Error { code: Unknown, source: Status { code: Unknown, message: "connection error: broken pipe" } } }, peer_addr: 52.9.132.248:3000, task: bootstrap

**Mandatory Information**
1. 0.8.6 release

dmindfuse:
I also agree with the above difficulties to complete a successful bootstrap in under 10 minutes. I update the node-config.yaml peers at every start of jormungandr from the latest available node-config.yaml example online. Yet, it seems impossible to guarantee a swift bootstrap phase, leading to multiple reboots/restarts (because what other options do we have?).

slackwaree:
Fortunately now that it does not crash anymore we don't need to bootstrap every 30 minutes after automatically restarting my nodes... but yeah it is an issue, an official list of the trusted servers should be provided by the Cardano foundation and jormungandr should download these at start.

michaeljfazio:
See #1582 and #1612

dorin100:
still happening on 0.8.9

- without storage - bootstrapping is imposible
```
Feb 03 08:52:07.731 INFO Starting jormungandr 0.8.9 (HEAD-30d20d2e, release, windows [x86_64]) - [rustc 1.40.0 (73528e339 2019-12-16)], task: init
Feb 03 08:52:07.745 WARN Node started without path to the stored secret keys (not a stake pool or a BFT leader), task: init
jormungandr ---config node_config.yaml --genesis-block-hash 8e4d2a343f3dcf9330ad9035b3e8d168e6728904262f2c434a4f8f934ec7b676
[...]

PR #1646, opened 2020-01-25
Fixes bootstrapping from trusted peers falling through when node is n…

michaeljfazio:
Currently, if Jormungandr fails to bootstrap from any of the trusted peers, it simply falls-through and assumes it should be a "genesis node" (i.e. the first node in a new network). This has been very confusing for pool operators as it seems as though their node is stuck.

This change implements a few things. Firstly, a new configuration setting ```genesis_node```. Which by default is set to ```false```. Setting this value to ```true``` essentially makes Jormungandr behave as is currently does. Setting it to ```false```, or accepting the default of ```false``` ensures that if the node fails to bootstrap from any of the trusted peers, it will just try again until it succeeds or the process is terminated.

Setting ```genesis_node``` to ```true``` will still allow the node to bootstrap from trusted peers (if it is possible). However, if it is not possible (e.g there are none defined or none accept a connection), then it will fall-through and attempt to establish itself as the first node in a new p2p network.

This should also fix the ITN Daedalus version from hanging on startup in some circumstances.

### DIFF
--- a/jormungandr/src/main.rs
+++ b/jormungandr/src/main.rs
@@ -80,6 +80,7 @@ pub struct BootstrappedNode {
 const BLOCK_TASK_QUEUE_LEN: usize = 32;
 const FRAGMENT_TASK_QUEUE_LEN: usize = 1024;
 const NETWORK_TASK_QUEUE_LEN: usize = 32;
+const BOOTSTRAP_RETRY_WAIT: Duration = Duration::from_secs(5);
 
 fn start_services(bootstrapped_node: BootstrappedNode) -> Result<(), start_up::Error> {
     if let Some(context) = bootstrapped_node.rest_context.as_ref() {
@@ -343,12 +344,41 @@ fn bootstrap(initialized_node: InitializedNode) -> Result<BootstrappedNode, star
 
     let (blockchain, blockchain_tip) = start_up::load_blockchain(block0, storage, block_cache_ttl)?;
 
-    let bootstrapped = network::bootstrap(
-        &settings.network,
-        blockchain.clone(),
-        blockchain_tip.clone(),
-        &bootstrap_logger,
-    )?;
+    let mut bootstrap_attempt: usize = 0;
+    loop {
+        bootstrap_attempt += 1;
+
+        // If we have exceeded the maximum number of bootstrap attempts, then we break out of the
+        // bootstrap loop.
+        if let Some(max_bootstrap_attempt) = settings.network.max_bootstrap_attempts {
+            if bootstrap_attempt > max_bootstrap_attempt {
+                warn!(
+                    &bootstrap_logger,
+                    "maximum allowable bootstrap attempts exceeded, continuing..."
+                );
+                break; // maximum bootstrap attempts exceeded, exit loop
+            };
+        }
+
+        // Will return true if we successfully bootstrap or there are no trusted peers defined.
+        if network::bootstrap(
+            &settings.network,
+            blockchain.clone(),
+            blockchain_tip.clone(),
+            &bootstrap_logger,
+        )? {
+            break; // bootstrap succeeded, exit loop
+        }
+
+        info!(
+            &bootstrap_logger,
+            "bootstrap attempt #{} failed, trying again in {} seconds...",
+            bootstrap_attempt,
+            BOOTSTRAP_RETRY_WAIT.as_secs()
+        );
+        // Sleep for a little while before trying again.
+        std::thread::sleep(BOOTSTRAP_RETRY_WAIT);
+    }
 
     let explorer_db = if settings.explorer {
         Some(explorer::ExplorerDB::bootstrap(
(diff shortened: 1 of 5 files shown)


michaeljfazio:
Fixes #1637 #1635 #1617 #1585

mzabaluev:
Some conflicts have been introduced in the test configurations.

thiago-tsnnst:
@michaeljfazio "and attempt to establish itself as the first node in a new p2p network."

This will be a similar behavior when we bootstrap without the trusted peers settings?

rinor:
@michaeljfazio thank you for the contribution. Very good stuff.
I have some concerns related to `genesis_node` config since genesis node/mode is not a recurring config/state, it is just a specific node state in a specific time and introducing a new config geared toward that notion may be a bit confusing.

What I would do is the following:

I would change a bit https://github.com/input-output-hk/jormungandr/blob/d1b0623d72123126d0b0e42ab295bfddf8892b60/jormungandr/src/network/mod.rs#L534-L550
from 
```rust
     if config.trusted_peers.is_empty() { 
         warn!(logger, "No trusted peers joinable to bootstrap the network"); 
     } 
  
     let mut bootstrapped = false; 
```
to
```rust
    let mut bootstrapped = false;

    if config.trusted_peers.is_empty() {
        warn!(logger, "No trusted peers joinable to bootstrap the network");
        bootstrapped = true;
    }
```
so that function returns true if the node has no trusted peers, allowing genesis case scenario to work properly, and also public nodes that some operators are sometime starting without trusted peers.Both cases work in the actual (master) code since even though `network::bootstrap` returns false, no action is taken due to 
```rust
    if !bootstrapped {
        // TODO, the node didn't manage to connect to any other nodes
        // for the initial bootstrap, that may be an error however
        // it is not necessarily an error, especially in the case the node is
        // the first ever to wake
    }
```
You could add a config option smth like `trusted_retry` or `bootstrap_retry` defaulting to `None` (meaning loop until bootstraped), but also giving the users the possibility to define how many times they want to retry bootstrapping.

Let me know what you think.

michaeljfazio:
@rinor I see what you mean about the "recurring state" and agree. If we leave a configuration parameter in to determine if a node should fall through the bootstrap nodes then it would be better to rename it to ```bootstrap_retry``` or similar.

The first iteration of my change had exactly what you proposed, that is, an empty peer list being used as a "just fall through case". But i reasoned that such a change might break existing pool configurations and test cases so I decided better to be on the safe side. Furthermore, I could conceive of a scenario (mostly in automated test setup) where one might want to start a bunch of nodes each referencing the others in trusted peer list and not really caring which one ended up acting as the "genesis node". A race if you will....

So, i'm leaning towards renaming the configuration parameter to ```bootstrap_retry``` with the default true and adding the appropriate warnings regarding a potential empty peer list. The other added benefit of this is that mis-configurations are more easily identified as well. Thoughts?

michaeljfazio:
> @michaeljfazio "and attempt to establish itself as the first node in a new p2p network."
> 
> This will be a similar behavior when we bootstrap without the trusted peers settings?

Yes. The root of this issue however is that even if you have a peer list defined currently, your node will only attempt to bootstrap from that list once. If it fails to bootstrap from any of those nodes first time it just falls through and sits idle acting as the first node in the network (which is obviously problematic).

michaeljfazio:
> Some conflicts have been introduced in the test configurations.

I'll rebase on master after I've worked out some more implementation details (as per @rinor comments).

rinor:
> @rinor I see what you mean about the "recurring state" and agree. If we leave a configuration parameter in to determine if a node should fall through the bootstrap routine and act as a geneis node then it would be better to rename it to `bootstrap_retry` or similar.
> 
> The first iteration of my change had exactly what you proposed, that is, an empty peer list being used as a "just fall through" case. But i reasoned that such a change might break existing pool configurations and test cases so I decided better to be on the safe side. Furthermore, I could conceive of a scenario (mostly in automated test setup) where one might want to start a bunch of nodes each referencing the others in trusted peer list and not really caring which one ended up acting as the "genesis node". A race if you will....
> 
> So, i'm leaning towards renaming the configuration parameter to `bootstrap_retry` with the default true and adding the appropriate warnings regarding a potential empty peer list. The other added benefit of this is that mis-configurations are more easily identified as well. Thoughts?

That is also fine.

michaeljfazio:
@rinor in the end I made the changes you suggested (more or less). Changed the configuration parameter to be a p2p config option named ```max_bootstrap_attempts``` which can be provided as ```None``` or a positive integer. When ```None``` and if there are trusted peers defined, the node will loop forever attempting to bootstrap. Otherwise, the node will loop for a maximum of ```max_bootstrap_attempts```. The value can also be provided as ```0``` in which case the node will not bootstrap from trusted peers regardless of if they are provided. This covers off all possible configuration cases.

rinor:
Will add also the doc once merged.

Linicks:
I would love to see this in 8.8!!  Bootstrapping has been an issue since the launch of the ITN and anything that helps improve this is much appreciated/needed.

PR #1653, opened 2020-01-26
Fixed Deadlocking

michaeljfazio:
This reverts commit e6bfb93, reversing changes made to ac36553. A critical bug
was introduced in this commit (deadlock when using P2PTopology) which cannot 
be easily resolved. 

It think it's probably easiest to revert the change for now and revisit an appropriate 
solution? Other option is to try an resolve the issue which I believe to be caused by replacing the P2PTopology R/W lock with a tokio Lock (non read/write). I know this was done for good reason, but at the moment I don't think there is a workable alternative that is likely to manifest before next release date (prove me wrong!).

Fixes: #1637

### DIFF
--- a/jormungandr/src/main.rs
+++ b/jormungandr/src/main.rs
@@ -100,7 +100,7 @@ fn start_services(bootstrapped_node: BootstrappedNode) -> Result<(), start_up::E
         bootstrapped_node.settings.leadership.log_ttl.into();
 
     let topology = P2pTopology::new(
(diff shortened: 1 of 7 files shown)


mzabaluev:
Without an analysis of where the deadlock occurs, this may just hide the problem.
We are applying changes to separate components' thread pools again (#1658), add more logging to narrow down the locking problems (#1656) and remove some identified problematic cases (to be submitted).

Issue #1263, opened 2019-12-03
Rework locking and connection pooling in storage


mzabaluev:
Currently we have an async global lock for the storage backend, guarding a blocking access API.
This stands in the way of high-performance implementations of streaming towards network peers that balance asynchronous state of the stream with holding the lock only while the thread is not being blocked, to avoid starving other requests from getting the lock.

One possible solution is to implement a connection pool using [bb8](https://github.com/khuey/bb8) and farm storage queries out to a thread pool running synchronous jobs, giving future/stream handles to the async tasks pending on them. This can keep the chain-libs storage API simple and synchronous, as most backends currently are.

Related: https://github.com/input-output-hk/chain-libs/issues/187#issuecomment-561092079

mzabaluev:
Cc @eugene-babichenko

eugene-babichenko:
Thanks for wrapping that up. Basically I'm already working on something like that.

There were problems with sqlite configuration that affected testing, but now I have almost solved them and everything should work fine.

Regarding the implementation on jormungandr side: I think that this is a good idea to use `tokio-threadpool` particularly because it already deals with this part:

> and farm storage queries out to a thread pool running synchronous jobs, giving future/stream handles to the async tasks pending on them

It is already used somewhere in the networking part of the project and we just need to make a global threadpool.

`bb8` seems to be a good choice, I like [this part](https://github.com/khuey/bb8/blob/master/bb8/src/lib.rs#L616).

PR #1655, opened 2020-01-27
Asynchronous storage query execution and connection pool

eugene-babichenko:
Implement #1263
Integrate https://github.com/input-output-hk/chain-libs/pull/239

### DIFF
--- a/jormungandr/src/blockchain/storage.rs
+++ b/jormungandr/src/blockchain/storage.rs
@@ -6,7 +6,7 @@ use async_trait::async_trait;
 use bb8::{ManageConnection, Pool, PooledConnection};
 use chain_storage::store::{for_path_to_nth_ancestor, BlockInfo, BlockStore};
 use futures03::{
-    future,
+    executor::block_on,
     prelude::*,
     sink::{Sink, SinkExt},
     stream::{self, Stream},
@@ -78,16 +78,18 @@ struct BlockIterState {
 }
 
 impl Storage {
-    pub async fn new(storage: NodeStorage) -> Self {
-        let manager = ConnectionManager::new(storage);
-        let pool = Pool::builder().build(manager).await.unwrap();
-        let write_connection_lock =
-            Arc::new(Mutex::new(pool.dedicated_connection().await.unwrap()));
+    pub fn new(storage: NodeStorage) -> Self {
+        block_on(async move {
+            let manager = ConnectionManager::new(storage);
+            let pool = Pool::builder().build(manager).await.unwrap();
+            let write_connection_lock =
+                Arc::new(Mutex::new(pool.dedicated_connection().await.unwrap()));
 
-        Storage {
-            pool,
-            write_connection_lock,
-        }
+            Storage {
+                pool,
+                write_connection_lock,
+            }
+        })
     }
 
     async fn run<F, R>(&self, f: F) -> Result<R, StorageError>
--- a/jormungandr/src/blockchain/storage.rs
+++ b/jormungandr/src/blockchain/storage.rs
@@ -3,18 +3,17 @@ use crate::{
     start_up::{NodeStorage, NodeStorageConnection},
 };
 use async_trait::async_trait;
-use bb8::{ManageConnection, Pool, PooledConnection};
+use bb8::{ManageConnection, Pool, RunError};
 use chain_storage::store::{for_path_to_nth_ancestor, BlockInfo, BlockStore};
 use futures::{Future as Future01, Sink as Sink01, Stream as Stream01};
 use futures03::{
     compat::*,
-    executor::block_on,
     prelude::*,
     sink::{Sink, SinkExt},
     stream::{self, Stream},
 };
 use std::{convert::identity, pin::Pin, sync::Arc};
-use tokio02::sync::Semaphore;
+use tokio02::{sync::Semaphore, task::spawn_blocking};
 use tokio_compat::runtime;
 
 pub use chain_storage::error::Error as StorageError;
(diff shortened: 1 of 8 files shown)


jormungandr/src/start_up/mod.rs
```diff
@@ -63,12 +64,14 @@ pub async fn prepare_block_0(
             })
         }
         Block0Info::Hash(block0_id) => {
-            if storage.block_exists(&block0_id)? {
+            let connection = storage.connect().unwrap();
```
mzabaluev: This is a blocking APl, should not be used in async code without offloading the work to the threadpool scheduler?

eugene-babichenko: This can be problematic. Using the threadpool scheduler requires `'static`, the signature of this function is `storage.connect(&self)` and `storage` is not `'static`. We can wrap the `storage` into `Arc` like we do in `Storage03` and always pass it around like that.

mzabaluev: Let's make sure we are on the the same page with this. I think the ideal solution is to have a thread/DB connection pool specifically for synchronous storage operations, with an API that would allow Tokio 0.2 tasks to submit closures of the form `|connection| { /* do things synchronously with the storage connection */ }` and await the results asynchronously. The connections are created or recycled by the pool implementation and passed to the closure.

jormungandr/src/blockchain/storage.rs
```diff
@@ -37,205 +113,306 @@ struct BlockIterState {
...
+        let mut rt = runtime::Builder::new()
+            .name_prefix("new-storage-worker-")
+            .core_threads(1)
+            .build()
+            .unwrap();
+
+        rt.block_on_std(async move {
+            let manager = ConnectionManager::new(storage);
+            let pool = Pool::builder().build(manager).await.unwrap();
+            let write_lock = Arc::new(Semaphore::new(1));
```
mzabaluev: What's the benefit of semaphore over `Mutex` here?

jormungandr/src/blockchain/storage.rs
```diff
@@ -37,205 +113,306 @@ struct BlockIterState {
...
-            read_connection: storage.clone(),
-            write_connection_lock: Lock::new(storage),
-        }
+        let mut rt = runtime::Builder::new()
+            .name_prefix("new-storage-worker-")
+            .core_threads(1)
+            .build()
+            .unwrap();
+
+        rt.block_on_std(async move {
```
mzabaluev: I guess we only initialize the storage once in the synchronous init code. I vaguely remember there is a one-stop-shop function for this in Tokio, so that you don't have to build a runtime explicitly.

jormungandr/src/blockchain/storage.rs
```diff
@@ -37,205 +113,306 @@ struct BlockIterState {
...
-        let mut write_connection_lock = self.write_connection_lock.clone();
-
-        future::poll_fn(move || Ok(write_connection_lock.poll_lock())).and_then(move |mut guard| {
-            match guard.put_tag(&tag, &header_hash) {
-                Err(error) => future::err(error),
-                Ok(res) => future::ok(res),
-            }
-        })
+    pub async fn put_tag(&self, tag: String, header_hash: HeaderHash) -> Result<(), StorageError> {
+        let _ = self.write_lock.acquire().await;
```
mzabaluev: Is it still necessary to lock sqlite for mutative operations? Is it some sort of a way to avoid transaction conflicts?

jormungandr/src/blockchain/storage.rs
eugene-babichenko: It's about lifetimes. Mutex guards depend on the lifetime of a mutex, which needs to be `'static` in this case (because of the requirements of `spawn_blocking`).

jormungandr/src/blockchain/storage.rs
eugene-babichenko: I didn't find such function in Tokio. There is `block_on` in `futures`, but it's not applicable here, because bb8 depends on Tokio runtime timer.

jormungandr/src/blockchain/storage.rs
eugene-babichenko: We had cases when write operations were done in the wrong order. In particular, a block could've been written before its parent. Introducing this lock helped to resolve the issue. I think the issue appeared due to wrong scheduling of futures involving write operations: operations are not necessarily done in the same order they were requested.

mzabaluev: If consistency is not verified on the database level, I think it's still possible for the client threads to put blocks in the wrong order. The lock just makes it less likely at a cost of extra synchronization.

jormungandr/src/blockchain/storage.rs
mzabaluev: But the result of `Semaphore::acquire` is lifetime-bound too?

jormungandr/src/blockchain/storage.rs
mzabaluev: This will do for now; I have added https://github.com/input-output-hk/chain-libs/issues/276 to resolve the consistency problems in chain-storage-sqlite.

jormungandr/src/blockchain/storage.rs
eugene-babichenko: It is, but is not used by any of closures running in `spawn_blocking`.

jormungandr/src/start_up/mod.rs
```diff
@@ -91,33 +91,40 @@ pub fn load_blockchain(
     logger: &Logger,
 ) -> Result<(Blockchain, Tip), Error> {
     use tokio::prelude::*;
+    use tokio_compat::runtime;
+
+    let mut rt = runtime::Builder::new()
+        .name_prefix("load-blockchain-worker-")
+        .core_threads(1)
+        .build()
+        .unwrap();
```
vincenthz: you can drop this commit (c60c618e6c1bd2b8016fea4d8a6559b7a4efcd55) in favor of:
https://github.com/input-output-hk/jormungandr/pull/1745

jormungandr/src/blockchain/storage.rs
mzabaluev: I still don't get how is this different from replacing all occurrences of `Semaphore` with the `Mutex` equivalent, except semaphores being slower.

eugene-babichenko: Do you mean `Mutex<()>`?

mzabaluev: Yes, as you obviously just need a critical section. Semaphores add an overhead of keeping a counter, which we don't need.

PR #1667, opened 2020-01-29
[Tests] Changed public and listen port to be different 

dkijania:
made listen and public ports different to each other for integration and scenario tests

### DIFF
--- a/jormungandr-integration-tests/src/common/configuration/jormungandr_config.rs
+++ b/jormungandr-integration-tests/src/common/configuration/jormungandr_config.rs
@@ -58,8 +58,8 @@ impl JormungandrConfig {
         .unwrap();
     }
 
-    pub fn get_p2p_port(&self) -> u16 {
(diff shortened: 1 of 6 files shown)


thiago-tsnnst:
this test open the door in the future to a mandatory use of different ports in this 2 parameters?

dkijania:
@thiago-tsnnst, this is not my decision to make but from my experience this is tricky part of settings. Integrations tests uses very simple configuration for single environment, so it's possible to have problem if we use the same port for `public_address` and `listen_address`. However, i can imagine 
 other configuration (some public node using firewall or loadbalancer) which can mitigate port clash problem. 

I've just learnt recently that some tests manifest problems in a non-obvious way when  `public_address` and `listen_address` are  equal, thus this change which in the face of fast pace of development is a safest way to ensure tests stability. In my todo list there is also automation for various `public_address` and `listen_address` values relations, so probably in the future I will have something more to say :)

PR #1671, opened 2020-01-29
Add/Improve some early logging and network related choices

### LINKED ISSUE #1669: bootstrap with large number of blocks looked block
linked to #1664

### LINKED ISSUE #1670: Increase selection information in the existing chain-selection/application
right now the set of information reported is not very precise and doesn't allow to understand very easily
why switch are done and what operation has been done

vincenthz:
* Add some informative logging in bootstrap loading from storage (fix #1669) 
* Add some further debugging related to chain-selection (fix #1670)

### DIFF
--- a/jormungandr/src/blockchain/process.rs
+++ b/jormungandr/src/blockchain/process.rs
@@ -446,16 +446,31 @@ pub fn process_new_ref(
         .get_ref()
         .and_then(move |tip_ref| {
             if tip_ref.hash() == candidate.block_parent_hash() {
-                info!(logger, "update current branch tip");
+                info!(
+                    logger,
+                    "update current branch tip: {} -> {}",
+                    tip_ref.header().description(),
+                    candidate.header().description(),
+                );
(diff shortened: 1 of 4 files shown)


jormungandr/src/blockchain/process.rs
```diff
@@ -446,16 +446,31 @@ pub fn process_new_ref(
...
+                    candidate.header().description(),
+                );
                 A(A(tip.update_ref(candidate).map(|_| true)))
             } else {
                 match chain_selection::compare_against(blockchain.storage(), &tip_ref, &candidate) {
                     ComparisonResult::PreferCurrent => {
-                        info!(logger, "create new branch");
+                        info!(
+                            logger,
+                            "create new branch with tip {} | current-tip {}",
```
mzabaluev: Could be done with KV pairs instead, that would be better for JSON log parsers and the like.

vincenthz: yes. this can be revised on top, but right now doesn't really matter

PR #1690, opened 2020-01-31
Add node culling to Poldercast policy

### LINKED ISSUE #1599: Jormungandr consumes 100% CPU and crashes
**Describe the bug**

After a while running Jormungandr, CPU usage slowly creeps to 100% at which point the node is unable to keep in sync with the network. The only way to recover is to reset the node.

The following diagram illustrates this behaviour:

image

This is a 2 core node, and 50% CPU utilisation represents and entire core being consumed by Jormungandr (I think because the async thread pool being used is singly threaded in the particular path of code highlighted below).

The sawtooth pattern changes depending on max_connections and gossip period. The period between 16:00 and 08:00 shows a steeper climb to total CPU consumption. Later the climb is not so steep because I have changed max_connections and gossip interval from 512 and 5s to 256 and 10s respectively.

Running a 30 minute profile of Jormungandr (post bootstrap) shows that the majority of CPU usage (greater than 50%) occurs in call to [jormungandr::network::send_gossip()](https://github.com/input-output-hk/jormungandr/blob/e4dd3eb0f353198464d66e49240eef4325f0fc89/jormungandr/src/network/mod.rs#L385) function. Within this function, most cycles are consumed deep in the poldercast library during vicinity distance calculation (a merge sort on the total peer list is executed with each iteration).

image

**Mandatory Information**
1. `jcli --full-version` output; ```jcli 0.8.6 (master-64780cf3, release, linux [x86_64]) - [rustc 1.39.0 (4560ea788 2019-11-04)]```
2. `jormungandr --full-version` output: ```jormungandr 0.8.6 (performance-e56aada5, release, linux [x86_64]) - [rustc 1.39.0 (4560ea788 2019-11-04)]```

 **Note:** Above versions correspond to test branches that im working off, but tip of master is affected in the exact same way.

**To Reproduce**
1. Start Jormungandr with public_address specified and max_connections set to 512.
2. Wait a while....

**Expected behavior**
Jormungandr runs without failure.

**Additional context**

It's not clear to me if the root cause here as inefficiencies in the vicinity calculation or a build up of the peer list and subscriptions data structures that are being merge sorted often. Probably a combination of both.

Here is the specific call to [sort() which within poldercast lib](https://github.com/primetype/poldercast/blob/88b89ff304f1d1c5c21693032ebd1f694b95f37b/src/poldercast/vicinity.rs#L78)

One theory is that continuous restarting of nodes currently plaguing the ITN is causing the poldercast node data structures to grow more rapidly than anticipated. Perhaps a bug in the node removal policy is also a factor.

Unfortunately it is quite impossible for me to successfully bootstrap the last few epochs which has not only destroyed sandstone pools profits but also limited my ability to effectively debug this issue further!

michaeljfazio:
Rework poldercast policy to cull stale nodes. The intention of this change is to reduce the total size of available peers and therefore significantly reduce CPU usage processing redundant poldercast nodes.

Hopefully fixes #1599.

### DIFF
--- a/jormungandr/src/network/p2p/policy.rs
+++ b/jormungandr/src/network/p2p/policy.rs
@@ -6,13 +6,16 @@ use slog::Logger;
 /// default quarantine duration is 30min
 const DEFAULT_QUARANTINE_DURATION: std::time::Duration = std::time::Duration::from_secs(1800);
 
+// default stale duration is 5min
+const DEFAULT_STALE_DURATION: std::time::Duration = std::time::Duration::from_secs(300);
+
 /// This is the P2P policy. Right now it is very similar to the default policy
 /// defined in `poldercast` crate.
 ///
 #[derive(Debug, Clone)]
 pub struct Policy {
     quarantine_duration: std::time::Duration,
-
+    stale_duration: std::time::Duration,
     logger: Logger,
 }
 
@@ -20,12 +23,14 @@ pub struct Policy {
 #[serde(deny_unknown_fields, rename_all = "snake_case")]
 pub struct PolicyConfig {
     quarantine_duration: Duration,
+    stale_duration: Duration,
 }
 
 impl Policy {
     pub fn new(pc: PolicyConfig, logger: Logger) -> Self {
         Self {
             quarantine_duration: pc.quarantine_duration.into(),
+            stale_duration: pc.stale_duration.into(),
             logger,
         }
     }
@@ -35,6 +40,7 @@ impl Default for PolicyConfig {
     fn default() -> Self {
         Self {
             quarantine_duration: Duration::from(DEFAULT_QUARANTINE_DURATION),
+            stale_duration: Duration::from(DEFAULT_STALE_DURATION),
         }
     }
 }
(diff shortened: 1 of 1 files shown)


michaeljfazio:
Note that this change introduces a new policy parameter ```stale_duration```. A node that has not been gossiped about, or gossiped with for longer than ```stale_duration``` will be "forgotten about" by poldercast and removed from internal data structures.

michaeljfazio:
Okay. So just had a read of section 5.4 on "churn handling". I see now what you mean @mzabaluev.

Has anyone done the analysis on what aspects of the currentl poldercast implementation does not conform to the poldercast paper? I think for one, gossips are being pushed onto nodes rather than being pull-only as is indicated by the paper (fundamental to the stateless culling mechanism described in churn handling).

michaeljfazio:
Closing this issue out. I now agree that this is probably not the right way to go about solving the issue of excessive node churn causing performance issues. I think the right was moving forward is two fold. Firstly, a rework of how poldercast stores and processes peer information would go a long way to preventing nodes from becoming increasingly bogged down. Second, the introduction of a valance property (i.e. timestamp) to node profiles that is also propagated to peers through gossips. Nodes should then favour contacting peers with the highest timestamp (i.e. oldest first) in order to "flush" the network quicker and prevent "forever-peers" propagating indefinitely.

PR #1696, opened 2020-02-03
Separate limit for client connections

### LINKED ISSUE #1593: Limit the number of client gRPC connections, use GC to close excess
Currently, a client connection stays open until being rotated off the `PeerMap` due to the total number of connections exceeding the `max_connections` limit. This creates an explosion of connections in aggregate saturating some nodes (though trusted peers should be ultimately relieved by #1386) and is bad for nodes behind NAT due to increased pressure on the port mapping in the NAT router which may lead to some connections being evicted or reset outside of node's control.

To take away the pain, introduce another limit in the configuration, `max_client_connections`, with a low default value. With every GC pass on the peer map, remove the least recently active client connections that are in excess of the limit.

mzabaluev:
Resolves #1593

### DIFF
--- a/doc/configuration/network.md
+++ b/doc/configuration/network.md
@@ -74,10 +74,8 @@ rest:
     typical settings for a non mining node: `"normal"`. For a stakepool: `"high"`.
 - `max_connections`: the maximum number of P2P connections this node should
     maintain. If not specified, an internal limit is used by default `[default: 256]`
-- `max_connections_threshold`: this is the number of available connections to keep
-  reserved for actually active connections. The node will apply this GC operation
-  from time to time if the `max_connections - max_connections_threshold` is reached.
-  `[default: 64]`
+- `max_client_connections`: the maximum number of client P2P connections this
+    node should keep open. `[default: 8]`
 - `policy`: (optional) set the setting for the policy module
   - `quarantine_duration` set the time to leave a node in quarantine before allowing
     it back (or not) into the fold.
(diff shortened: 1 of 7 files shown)


rinor:
Not sure if is worth checking if `max_client_connections` > `max_connections` and if so decreasing that value to `max_connections`, smth like this.

```diff
diff --git a/jormungandr/src/settings/start/mod.rs b/jormungandr/src/settings/start/mod.rs
index 0dde4195..8d79bba0 100644
--- a/jormungandr/src/settings/start/mod.rs
+++ b/jormungandr/src/settings/start/mod.rs
@@ -130,7 +130,17 @@ impl RawSettings {
             config,
         } = self;
         let command_arguments = &command_line.start_arguments;
-        let network = generate_network(&command_arguments, &config)?;
+        let mut network = generate_network(&command_arguments, &config)?;
+
+        if network.max_client_connections > network.max_connections {
+            warn!(
+                logger,
+                "Decreasing p2p.max_client_connections value from {} to {}",
+                network.max_client_connections,
+                network.max_connections
+            );
+            network.max_client_connections = network.max_connections;
+        }
 
         let storage = match (
             command_arguments.storage.as_ref(),
```

Furthermore a bit concerned about `max_client_connections` default value 8, since for non public nodes that will be the overall total connections and not sure if it is enough. Not a big deal just smth to keep in mind until users start using this new config.

mzabaluev:
> Furthermore a bit concerned about `max_client_connections` default value 8

My intent is to keep the overall number of connections in the network lower and renewal rate high. These are supposed to be connections to functional, responsive nodes, not counting any possible server-side connections, so I figure it should be enough to ensure network updates are received.

mzabaluev:
> Not sure if is worth checking if `max_client_connections > max_connections` and if so decreasing that value to `max_connections`

The limit on `max_connections` is enforced in any case, but this check will avoid unintended churn on server connections. Thanks, I'm applying this.

PR #1698, opened 2020-02-03
Refactor REST API to remove sync locking and make it more isolated

### LINKED ISSUE #1645: Refactor REST to remove gratuitous synchronous locking
Lots of the stability issues exemplified by #1626 seem to be triggered by issuing REST requests.
The REST service is one of the few remaining components of jormungandr where synchronous locking is used; it also departs from the architecture of other services in that it does not use message passing from other services for state updates, but uses shared state guarded by a lock, that can be modified by other parts of the program. Every request served needs to acquire that lock. It also has shared access to network task's state, being able to lock out network tasks to perform costly statistic data collection requests, all the while a synchronous lock is held blocking actix tasks; all this is now running on the shared tokio executor.

A way to solve this is to convert the REST service to message passing, so that it works on its own representation of the relevant node state and receives updates from other tasks over the intercom channel.

eugene-babichenko:
Closes #1645

### DIFF
--- a/jormungandr/src/rest/v0/handlers.rs
+++ b/jormungandr/src/rest/v0/handlers.rs
@@ -30,7 +30,7 @@ use std::sync::Arc;
 pub use crate::rest::{Context, FullContext};
 
 async fn chain_tip(context: &Data<Context>) -> Result<Arc<Ref>, Error> {
-    chain_tip_from_full(&*context.try_full()?).await
+    chain_tip_from_full(&*context.try_full().await?).await
 }
 
 async fn chain_tip_from_full(context: &FullContext) -> Result<Arc<Ref>, Error> {
@@ -69,7 +69,8 @@ pub async fn get_account_state(
 
 pub async fn get_message_logs(context: Data<Context>) -> Result<impl Responder, Error> {
     context
-        .try_full()?
+        .try_full()
+        .await?
         .logs
         .logs()
         .compat()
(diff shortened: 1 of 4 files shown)


jormungandr/src/rest/v0/handlers.rs
```diff
@@ -180,39 +207,62 @@ pub async fn get_block_id(
...
     let tip = chain_tip_from_full(&full_context).await?;
-    full_context
-        .blockchain
-        .storage()
-        .stream_from_to(block_id, tip.hash())
+    let (reply_handle, reply_stream) =
+        intercom::stream_reply::<_, Error>(1, context.logger().await?);
+
+    context
+        .try_full()
```
mzabaluev: Why do `context.try_full()` twice?

jormungandr/src/rest/v0/handlers.rs
```diff
@@ -103,29 +117,42 @@ struct NodeStatsDto {
...
     context
-        .blockchain
-        .storage()
-        .get(tip.hash())
+        .client_task
+        .clone()
+        .send(ClientMsg::GetBlocks(vec![tip.hash()], reply_handle))
+        .compat()
+        .await
+        .map_err(ErrorInternalServerError)?;
```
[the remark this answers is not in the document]
eugene-babichenko: Then I will revert the first commits.

jormungandr/src/rest/v0/handlers.rs
eugene-babichenko: Thanks for noticing that

PR #1715, opened 2020-02-04
use tokio-compat runtime instead of tokio 0.1

eugene-babichenko:


eugene-babichenko:
Not sure if we should convert all methods in `task.rs` into using std futures.

NicolasDP:
> Not sure if we should convert all methods in `task.rs` into using std futures.

no, not yet. better having the runtime that support both now and have a smooth, step by step transition

PR #1717, opened 2020-02-04
IPv6 dafault to IPv4 methods in gossip is_global 

danielSanchezQ:
I run into this while reading the code. I think we can rely on the IPv4 methods if the IPv6 is transformable. At least until we have the IPv6 ones.

### DIFF
--- a/jormungandr/src/network/p2p/gossip.rs
+++ b/jormungandr/src/network/p2p/gossip.rs
@@ -176,74 +176,74 @@ impl property::Deserialize for Gossip {
 #[cfg(test)]
 mod tests {
     use super::*;
-    use poldercast::{NodeProfile, NodeProfileBuilder, Address};
-    use std::str::FromStr;
+    use poldercast::{Address, NodeProfile, NodeProfileBuilder};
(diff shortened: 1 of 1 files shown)


danielSanchezQ:
I do not know if this is being tested. I'm searching for a place where I can write some tests for it. It would be nice if someone points out for me :)

mzabaluev:
@danielSanchezQ Feel free to add unit tests to the module.

jormungandr/src/network/p2p/gossip.rs
```diff
@@ -72,9 +72,20 @@ impl Gossip {
                 if ip.is_loopback() {
                     return false;
                 }
-                // FIXME: add more tests when Ipv6Addr convenience methods
-                // get stabilized:
+                // Try using same methods by trying to cast address to ipv4
+                // FIXME: use Ipv6 tests when Ipv6Addr convenience methods are get stabilized:
```
NicolasDP:
```suggestion
                // FIXME: use Ipv6 tests when Ipv6Addr convenience methods get stabilized:
```

danielSanchezQ:
Thanks @mzabaluev

PR #1720, opened 2020-02-05
Added node ID to stats output

michaeljfazio:
Adds node ID to node stats output. Requested by BCSH pool operator.


vincenthz:
node-id will be removed so I don't think that's a good idea

PR #1738, opened 2020-02-06
Do not truncate the list of peers for propagation

mzabaluev:
The vector obtained from `Topology::view` is sorted by node ID, so truncations make the network biased toward nodes with low IDs.
Let the client connections rotate "naturally", allowing some churn.
A limit on the view size should be implemented in Poldercast:
https://github.com/primetype/poldercast/issues/6

### DIFF
--- a/jormungandr/src/network/mod.rs
+++ b/jormungandr/src/network/mod.rs
@@ -350,8 +350,7 @@ fn handle_propagation_msg(
     // active subscriptions map, connect to them and deliver
     // the item.
     send_to_peers.then(move |res| {
-        if let Err(mut unreached_nodes) = res {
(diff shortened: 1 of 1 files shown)


jormungandr/src/network/mod.rs
```diff
@@ -418,7 +417,6 @@ fn send_gossip(state: GlobalStateR, channels: Channels) -> impl Future<Item = ()
         .view(poldercast::Selection::Any)
         .and_then(move |view| {
             let mut peers = view.peers;
```
rinor: why mutable

mzabaluev: No need for that anymore, fixed.

Issue #1580, opened 2020-01-16
ERRO Eek... Too late, we missed an event schedule, system time might be off?


adascanBovich:
**Describe the bug**
My nodes keep missing blocks from time to time because of "ERRO Eek... Too late, we missed an event schedule, system time might be off?" error. System clock is in sync, ntp installed. Happens on different nodes, no pattern, one node can lose block for this error, then make 3 in a row just fine.

**Mandatory Information**
1. jcli 0.8.6 (HEAD-7686397, release, linux [x86_64]) - [rustc 1.40.0 (73528e339 2019-12-16)]
2. jormungandr 0.8.6 (HEAD-7686397, release, linux [x86_64]) - [rustc 1.40.0 (73528e339 2019-12-16)]

**To Reproduce**
1. Start a pool.
2. Check logs every time you are missing your blocks.
3. Try no to cry.

**Expected behavior**
Blocks should be creating just fine with system clock synced.

**Additional context**
Couple of examples:

<img width="714" alt="Screenshot 2020-01-16 at 13 43 40" src="https://user-images.githubusercontent.com/58727428/72541306-f416d200-388a-11ea-90e5-23be4710a02d.png">

<img width="706" alt="Screenshot 2020-01-16 at 18 01 49" src="https://user-images.githubusercontent.com/58727428/72541322-faa54980-388a-11ea-8f67-0bd61d078fa3.png">

Proxiweb:
I had this one two time with 0.8.6. I switched back to 0.8.5. Not happening again.

mmahut:
Seeing this as well.

```
Jan 16 15:44:37.326 ERRO Eek... Too late, we missed an event schedule, system time might be off?, event_end: 2020-01-16T14:44:37+00:00, event_start: 2020-01-16T14:44:35+00:00, event_date: 33.35129, leader_id: 1, task: leadership
Jan 16 17:43:52.056 ERRO Eek... Too late, we missed an event schedule, system time might be off?, event_end: 2020-01-16T16:43:51+00:00, event_start: 2020-01-16T16:43:49+00:00, event_date: 33.38706, leader_id: 1, task: leadership
```

However, not all the time:

```
---
- created_at_time: "2020-01-16T07:20:28.087825432+00:00"
  enclave_leader_id: 1
  finished_at_time: "2020-01-16T15:58:29.011607694+00:00"
  scheduled_at_date: "33.37346"
[...]

PR #1739, opened 2020-02-06
Added node limit to P2P topology

### LINKED ISSUE #1599: Jormungandr consumes 100% CPU and crashes

michaeljfazio:
This commit adds limit to the amount of gossips a node will store in memory at any one time. This allows nodes to be resilient against acquiring too many stale node profiles from the network.

This issue definitively fixes #1599 and makes the processing requirements of Jormungandr deterministic in relation to the total number of network profiles being gossiped in the network.

This is what we see before the change. The processing requirements are increasing relative to the number of profiles being processed into p2p overlays.

72666360-5539d380-3a4c-11ea-9904-d18f90c3aec2

Nodes will become so bogged down that they start to miss block production (see #1580). Eventually they fall too far behind the tip of the main chain, resulting in complete de-sync. Longer term, nodes crash completely due to expended resources.

Here is the result after this change with default max peer size set to 10000.

Screen Shot 2020-02-06 at 8 41 58 pm

In this case, my CPU usage never climbs above 2% and the node never de-syncs from the network.

A follow up to this issue will allow the maximum number of stored node profiles to be configured. Though 10000 as a default for the time being seems to be more than sufficient for most nodes.

This solution doesn't solve the underlying issue of nodes forever being gossiped on the network. To solve this will probably require a protocol level enhancement (i.e. including liveness timestamp in propagated gossip). However, the fix also does not detract from the poldercast specification. It simply accepts the resource limitations of the node.

### DIFF
--- a/jormungandr/src/network/p2p/topology.rs
+++ b/jormungandr/src/network/p2p/topology.rs
@@ -15,6 +15,8 @@ use slog::Logger;
 use tokio::prelude::future::{self, Future};
 use tokio::sync::lock::{Lock, LockGuard};
 
+const DEFAULT_MAX_NODES: usize = 10000;
+
 pub struct View {
     pub self_node: NodeProfile,
     pub peers: Vec<Node>,
@@ -23,6 +25,7 @@ pub struct View {
 /// object holding the P2pTopology of the Node
 #[derive(Clone)]
 pub struct P2pTopology {
+    max_nodes: usize,
     lock: Lock<Topology>,
     node_id: Id,
     logger: Logger,
@@ -30,6 +33,7 @@ pub struct P2pTopology {
 
 /// Builder object used to initialize the `P2pTopology`
 struct Builder {
+    max_nodes: usize,
     topology: Topology,
     logger: Logger,
 }
@@ -38,9 +42,15 @@ impl Builder {
     /// Create a new topology for the given node profile
     fn new(node: poldercast::NodeProfile, logger: Logger) -> Self {
         Builder {
+            max_nodes: DEFAULT_MAX_NODES,
             topology: Topology::new(node),
             logger,
         }
+    }
+
+    fn set_max_nodes(mut self, max_nodes: usize) -> Self {
+        self.max_nodes = max_nodes;
+        self
     }
 
     fn set_policy(mut self, policy: PolicyConfig) -> Self {
@@ -73,6 +83,7 @@ impl Builder {
     fn build(self) -> P2pTopology {
         let node_id = self.topology.profile().id().clone();
         P2pTopology {
+            max_nodes: self.max_nodes,
             lock: Lock::new(self.topology),
             node_id: node_id.into(),
             logger: self.logger,
@@ -81,11 +92,13 @@ impl Builder {
 }
 
 impl P2pTopology {
+    // TODO: this should be private to the module with the builder exposed externally
     pub fn new(config: &Configuration, logger: Logger) -> Self {
         Builder::new(config.profile.clone(), logger)
             .set_poldercast_modules()
             .set_custom_modules(&config)
             .set_policy(config.policy.clone())
+            .set_max_nodes(DEFAULT_MAX_NODES)
             .build()
     }
 
@@ -126,8 +139,11 @@ impl P2pTopology {
         from: Id,
         gossips: Gossips,
     ) -> impl Future<Item = (), Error = E> {
-        self.write()
-            .map(move |mut topology| topology.accept_gossips(from.into(), gossips.into()))
+        self.write().map(move |mut topology| {
+            if topology.nodes().node_count().all_count < DEFAULT_MAX_NODES {
+                topology.accept_gossips(from.into(), gossips.into())
+            }
+        })
     }
 
     pub fn exchange_gossips<E>(


jormungandr/src/network/p2p/topology.rs
```diff
@@ -126,8 +139,11 @@ impl P2pTopology {
         from: Id,
         gossips: Gossips,
     ) -> impl Future<Item = (), Error = E> {
-        self.write()
-            .map(move |mut topology| topology.accept_gossips(from.into(), gossips.into()))
+        self.write().map(move |mut topology| {
+            if topology.nodes().node_count().all_count < DEFAULT_MAX_NODES {
+                topology.accept_gossips(from.into(), gossips.into())
+            }
```
mzabaluev: I would rather always admit the newly advertised entries after freeing the requisite number of older ones. It does not matter how the nodes are picked for removal now; might just use the collection iterator.

mzabaluev: This may need changes in `poldercast`. I suggest implementing a configurable size limit and eviction functionality inside `poldercast::Topology`, so that we can fine-tune it later, possibly using the global node timestamps.
Cc @NicolasDP

michaeljfazio: > I would rather always admit the newly advertised entries after freeing the requisite number of older ones. It does not matter how the nodes are picked for removal now; might just use the collection iterator.

It's kind of arbitrary when you think about it. It's just easier to let the natural eviction process work to purge nodes as they are attempted to be gossiped with. That would of course mean a change to the PolderCast library rather than Jormungandr.

> This may need changes in `poldercast`. I suggest implementing a configurable size limit and eviction functionality inside `poldercast::Topology`, so that we can fine-tune it later, possibly using the global node timestamps.

Yep. I think you are right on the money here. That's precisely why I decided to make the change Jormungandr-side. I initially started to implement this in PolderCast create but quickly realised a few things:

1. There really needs to be that valance/timestamp value propagated on the network that you [spoke about here](https://github.com/primetype/poldercast/issues/10). No way that change would make it in for the next release in time.
2. More important that this pool nodes are stabilised. This problem is feeding itself; causing nodes to restart all the time with new IDs, only for those to be propagated forever. We can slow that down by relieving processing on the nodes.
3. I acknowledge this is not a long term solution. But its minimal code change for maximum effect. Low risk, high reward (I hope).

Linicks: While testing this PR, I've also observed a significant reduction in CPU consumption while maintaining a flat utilization curve.  Others have reported the same behavior as well as building and propagating their blocks.  I have observed a slightly lower "Established" TCP connection count in my configuration.

DanielEpstein: As nodes in peerAvailableCnt restart or disconnect they are Quarantined and over time peerQuarantinedCnt might grow to be a larger count than peerAvailableCnt.

I'm not sure if GC is flushing peerQuarantinedCnt fast enough if at all to allow fresh connections when setting a hard limit. Ultimately you may need to restart, though definitely a step in the right direction

NicolasDP: > I would rather always admit the newly advertised entries after freeing the requisite number of older ones. It does not matter how the nodes are picked for removal now; might just use the collection iterator.

I agree with this. I have made some changes in my implementation of _poldercast_. `0.12` adds the limits of the overall peer collections based using an LRU. This allows for our local node to ignore more easily nodes that are not of used or that have been quarantined under long network churn time (which may happen when new versions are being propagated to the network -- every nodes rebooting to newer versions and propagating new gossips).

michaeljfazio: Sounds good. I'll close out this issue in light of the upstream changes in PolderCast.

PR #1744, opened 2020-02-07
Add std::future compliant version of spawn and run_periodic

vincenthz:


jormungandr/src/utils/task.rs
```diff
@@ -199,6 +202,16 @@ impl TokioServiceInfo {
...
+        F: StdFuture<Output = ()> + Send + 'static,
+    {
+        let logger = self.logger.clone();
+        trace!(logger, "spawning {}", name);
+        self.executor.spawn_std(future)
+    }
+
     // Run the closure with the specified period on the executor
     // and execute the resulting closure.
     pub fn run_periodic<F, U>(&self, name: &'static str, period: Duration, mut f: F)
```
eugene-babichenko: maybe add a deprecation notice here too?

eugene-babichenko:
Can you please convert spawning the leadership process to `spawn_std`? It's already using std futures.

vincenthz:
this need `spawn_future` to be converted (or a _std variant added); I would like too, but this is not a one-liner, so I'll delay it to another future PR

PR #1754, opened 2020-02-09
Input process async

vincenthz:
* convert some tip/branches to async
* convert more input processing to async

### DIFF
--- a/jormungandr/src/blockchain/process.rs
+++ b/jormungandr/src/blockchain/process.rs
@@ -99,7 +99,7 @@ impl Process {
         let blockchain_tip = self.blockchain_tip.clone();
         let network_msg_box = self.network_msgbox.clone();
         let explorer_msg_box = self.explorer_msgbox.clone();
-        let mut tx_msg_box = self.fragment_msgbox.clone();
+        let tx_msg_box = self.fragment_msgbox.clone();
         let stats_counter = self.stats_counter.clone();
 
         match input {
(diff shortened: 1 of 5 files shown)


jormungandr/src/blockchain/process.rs
```diff
@@ -401,29 +298,91 @@ fn try_request_fragment_removal(
...
+    let storage = blockchain.storage().clone();
+
+    let tip_ref = tip.get_ref_std().await;
+
+    let tip_updated = if tip_ref.hash() == candidate.block_parent_hash() {
+        info!(
+            logger,
+            "update current branch tip: {} -> {}",
+            tip_ref.header().description(),
+            candidate.header().description(),
```
NicolasDP: you are repeating yourself with the logs here. 

create a log with the new context value instead. I.e:

```rust
let logger = logger.new(o!(
  "current_tip" => tip_ref.header().description(),
  "candidate" => tip.header().description(),
));
``` 

That will make the code shorter too and more easily readable.

jormungandr/src/blockchain/process.rs
```diff
@@ -401,29 +298,91 @@ fn try_request_fragment_removal(
...
+    } else {
+        match chain_selection::compare_against(blockchain.storage(), &tip_ref, &candidate) {
+            ComparisonResult::PreferCurrent => {
+                info!(
+                    logger,
+                    "create new branch with tip {} | current-tip {}",
+                    candidate.header().description(),
+                    tip_ref.header().description(),
+                );
+                false
```
NicolasDP: no new branch is created here

vincenthz: indeed, but that's what aa7834627403e69984a4a1c4c55d33d945d40877 used to do

jormungandr/src/blockchain/process.rs
vincenthz: just converting the old code, not doing any substantial improvement or anything different right now

PR #1760, opened 2020-02-10
Enclave avoid duplicated Leaders

### LINKED ISSUE #1498: REST API - leaders registration - Registering twice the same leader is allowed
**Describe the bug**
Registering twice the same leader with the same parameters through REST endpoing POST /api/v0/leaders is allowed and returns 2 different leader ids.

**Mandatory Information**
1. jcli 0.8.5 (HEAD-3db0680, release, linux [x86_64]) - [rustc 1.40.0 (73528e339 2019-12-16)]
2. jormungandr 0.8.5 (HEAD-3db0680, release, linux [x86_64]) - [rustc 1.40.0 (73528e339 2019-12-16)]

**To Reproduce**
1. Register a leader with REST endpoing POST /api/v0/leaders
2. Register again the same leader. Call does not fail, and a new id is returned

**Expected behavior**
The second call should fail

danielSanchezQ:
Should fix #1498


danielSanchezQ:
It is not tested still. But I wanted to add some considerations here first.

jormungandr/src/secure/enclave.rs
```diff
@@ -4,13 +4,14 @@ use crate::blockcfg::{
...
 use jormungandr_lib::interfaces::EnclaveLeaderId as LeaderId;
-use std::collections::BTreeMap;
+use std::collections::{BTreeMap, HashMap, HashSet};
 use std::sync::Arc;
 use tokio02::sync::RwLock;
 
 #[derive(Clone)]
 pub struct Enclave {
     leaders: Arc<RwLock<BTreeMap<LeaderId, Leader>>>,
+    added_cache: Arc<RwLock<HashMap<String, LeaderId>>>,
```
danielSanchezQ: I used `String` but maybe we could use `PublicKey` instead, I wanted to go ahead and used the easy way first but if `PublicKey` is hashable should be fine also.

jormungandr/src/secure/enclave.rs
```diff
@@ -38,14 +40,32 @@ impl Enclave {
         e
     }
 
+    pub async fn get_leader_id_if_present(&self, leader: &Leader) -> Option<LeaderId> {
+        match &leader.bft_leader {
```
danielSanchezQ: What happens if `bft_leader` is None, should it even be a viable Leader?

jormungandr/src/secure/enclave.rs
```diff
@@ -38,14 +40,32 @@ impl Enclave {
...
+            Some(id) => return id,
+            None => {}
+        }
         let mut leaders = self.leaders.write().await;
         let next_leader_id = get_maximum_id(&leaders).next();
+        let mut cache = self.added_cache.write().await;
+        leader
+            .bft_leader
+            .as_ref()
+            .map(|l| cache.insert(l.sig_key.to_public().to_string(), next_leader_id));
```
danielSanchezQ: Same as [here](https://github.com/input-output-hk/jormungandr/pull/1760/files#r377229048), if it do not have a `bft_leader` from where we can take the public key then we cannot insert it in the already added cache.
I'm missing something here.

danielSanchezQ:
I added some tests to check that the method behaves as expected.

danielSanchezQ:
I need to update the tests for the last commit

jormungandr/src/secure/enclave.rs
```diff
@@ -4,13 +4,14 @@ use crate::blockcfg::{
...
 use jormungandr_lib::interfaces::EnclaveLeaderId as LeaderId;
-use std::collections::BTreeMap;
+use std::collections::{BTreeMap, HashMap, HashSet};
 use std::sync::Arc;
 use tokio02::sync::RwLock;
 
 #[derive(Clone)]
 pub struct Enclave {
     leaders: Arc<RwLock<BTreeMap<LeaderId, Leader>>>,
+    added_leaders_cache: Arc<RwLock<HashMap<String, LeaderId>>>,
```
NicolasDP: clearly we would like only want Locking point

jormungandr/src/secure/enclave.rs
```diff
@@ -152,3 +231,118 @@ impl Enclave {
...
+        assert_eq!(leaders_data.added_leaders_cache.len(), 1);
+    }
+
+    //    #[tokio::test]
+    //    async fn enclave_add_different_genesis_leaders() {
+    //        let mut enclave = Enclave::new();
+    //        let rng = rand_core::OsRng;
+    //        let leader1 = Leader {
+    //            bft_leader: None,
+    //            genesis_leader: Some( GenesisLeader {
```
danielSanchezQ: I'm having some trouble creating a `GenesisLeader`. I couldn't find any reference or example on how to do it for testing. Any help on how to do it is more than welcome. @NicolasDP, any tip?

jormungandr/Cargo.toml
```diff
@@ -41,6 +41,7 @@ network-core    = { path = "../chain-deps/network-core" }
 network-grpc    = { path = "../chain-deps/network-grpc" }
 poldercast = "0.11.2"
 rand = "0.7"
+rand_core = "0.5"
```
NicolasDP: this is only used for the test, move this to the `dev-dependency`

jormungandr/Cargo.toml
```diff
@@ -57,15 +58,22 @@ structopt = "^0.2"
...
+[dev-dependencies]
+chain-core           = { path = "../chain-deps/chain-core", features=["property-test-api"]}
+chain-crypto         = { path = "../chain-deps/chain-crypto", features=["property-test-api"] }
+chain-addr           = { path = "../chain-deps/chain-addr", features=["property-test-api"] }
+chain-impl-mockchain = { path = "../chain-deps/chain-impl-mockchain", features=["property-test-api"] }
+
 [build-dependencies]
 versionisator = "1.0.2"
 
+
```
NicolasDP: non necessary trailing line

jormungandr/src/secure/enclave.rs
```diff
@@ -4,13 +4,20 @@ use crate::blockcfg::{
...
+    leaders: BTreeMap<LeaderId, Leader>,
+    added_leaders_cache: HashMap<String, LeaderId>,
+}
+
 #[derive(Clone)]
 pub struct Enclave {
-    leaders: Arc<RwLock<BTreeMap<LeaderId, Leader>>>,
+    leaders_data: Arc<RwLock<EnclaveLeadersWithCache>>,
+    //    leaders: Arc<RwLock<BTreeMap<LeaderId, Leader>>>,
+    //    added_leaders_cache: Arc<RwLock<HashMap<String, LeaderId>>>,
```
NicolasDP: remove deadcode

danielSanchezQ: Yep, sorry for this.

danielSanchezQ:
This should be ready now @NicolasDP , thanks!

PR #1764, opened 2020-02-11
upgrade bech32 to 0.7

vincenthz:
also remove some warnings

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -452,11 +452,6 @@ dependencies = [
 
 [[package]]
 name = "bech32"
-version = "0.6.0"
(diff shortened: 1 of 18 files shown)


mzabaluev:
> clearly we are doing something wrong with utilising these dependencies. especially these with no stable APIs yet.

Same can be said about tokio 0.2 :)

NicolasDP:
> > clearly we are doing something wrong with utilising these dependencies. especially these with no stable APIs yet.
> 
> Same can be said about tokio 0.2 :)

This is what I meant by **these dependencies**.

PR #1778, opened 2020-02-13
Add (somewhat hardcoded peers queries

vincenthz:
for now we just print the number of peers that we got from a trusted peer,
and do nothing else with the query.

The limit of number of peers is also hardcoded.

### DIFF
--- a/jormungandr/src/client.rs
+++ b/jormungandr/src/client.rs
@@ -1,6 +1,7 @@
 use crate::blockcfg::{Block, Header, HeaderHash};
 use crate::blockchain::{Storage, Tip};
 use crate::intercom::{ClientMsg, Error, ReplySendError, ReplyStreamHandle};
+use crate::network::p2p::{P2pTopology, Peer, PeersResponse};
 use crate::utils::task::{Input, TokioServiceInfo};
 use chain_core::property::HasHeader;
 
@@ -11,6 +12,7 @@ use tokio::timer::Timeout;
 use std::time::Duration;
 
 const PROCESS_TIMEOUT_GET_BLOCK_TIP: u64 = 5;
+const PROCESS_TIMEOUT_GET_PEERS: u64 = 10;
 const PROCESS_TIMEOUT_GET_HEADERS: u64 = 5 * 60;
 const PROCESS_TIMEOUT_GET_HEADERS_RANGE: u64 = 5 * 60;
 const PROCESS_TIMEOUT_GET_BLOCKS: u64 = 10 * 60;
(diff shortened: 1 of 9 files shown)


jormungandr/src/network/p2p/topology.rs
```diff
@@ -166,6 +166,21 @@ impl P2pTopology {
...
+    pub fn list_available_limit<E>(
+        &self,
+        n: usize,
+    ) -> impl Future<Item = Vec<poldercast::Node>, Error = E> {
+        self.read().map(move |topology| {
+            topology
+                .nodes()
+                .all_available_nodes()
+                .into_iter()
+                .take(n)
```
mzabaluev: This will give everybody the same list of peers, no randomization?

jormungandr/src/network/service.rs
```diff
@@ -299,4 +300,12 @@ impl GossipService for NodeService {
             logger,
         )
     }
+
+    fn peers(&mut self) -> Self::PeersFuture {
+        intercom::unary_future(
+            self.channels.client_box.clone(),
+            self.logger().new(o!("request" => "Peers")),
+            ClientMsg::GetPeers,
+        )
```
mzabaluev: Could have just used `global_state.topology`, no need to involve the client task in bouncing messages around and sharing the topology state.

jormungandr/src/network/p2p/topology.rs
vincenthz: no idea but it's indeed quite possible indeed, I didn't really have time to think about how to randomize this access (at it rightly should)

jormungandr/src/network/service.rs
vincenthz: I followed what was done in GetTip, and I agree it seems overkill to do it this way

jormungandr/src/client.rs
```diff
@@ -11,6 +12,7 @@ use tokio::timer::Timeout;
 use std::time::Duration;
 
 const PROCESS_TIMEOUT_GET_BLOCK_TIP: u64 = 5;
+const PROCESS_TIMEOUT_GET_PEERS: u64 = 10;
 const PROCESS_TIMEOUT_GET_HEADERS: u64 = 5 * 60;
 const PROCESS_TIMEOUT_GET_HEADERS_RANGE: u64 = 5 * 60;
 const PROCESS_TIMEOUT_GET_BLOCKS: u64 = 10 * 60;
```
NicolasDP: Different topic... but it seems to be these timeouts are a bit overly large

jormungandr/src/network/p2p/topology.rs
NicolasDP: use `view` instead

Issue #1774, opened 2020-02-13
test jormungandr::mempool::test_log_ttl is failing randomly


eugene-babichenko:
Here is one of the examples, in general we have observed it a couple of times the last week.

https://app.circleci.com/jobs/github/input-output-hk/jormungandr/22636

eugene-babichenko:
```
                               at /rustc/5e1a799842ba6ed4a57e91f7ab9435947482f7d8/src/libstd/panic.rs:318
  29:     0x5640d5cac046 - std::panicking::try::do_call::h3bf71a2dd778c40e
                               at /rustc/5e1a799842ba6ed4a57e91f7ab9435947482f7d8/src/libstd/panicking.rs:292
  30:     0x5640d6728f0a - __rust_maybe_catch_panic
                               at src/libpanic_unwind/lib.rs:78
  31:     0x5640d5caca76 - std::panicking::try::haa1491030708e35a
                               at /rustc/5e1a799842ba6ed4a57e91f7ab9435947482f7d8/src/libstd/panicking.rs:270
  32:     0x5640d5caca76 - std::panic::catch_unwind::h74d1656cfaa8c137
                               at /rustc/5e1a799842ba6ed4a57e91f7ab9435947482f7d8/src/libstd/panic.rs:394
  33:     0x5640d5caca76 - std::thread::Builder::spawn_unchecked::{{closure}}::h1aef0c9f6c50afd6
                               at /rustc/5e1a799842ba6ed4a57e91f7ab9435947482f7d8/src/libstd/thread/mod.rs:468
  34:     0x5640d5caca76 - core::ops::function::FnOnce::call_once{{vtable.shim}}::h27f633ebc324b9d3
                               at /rustc/5e1a799842ba6ed4a57e91f7ab9435947482f7d8/src/libcore/ops/function.rs:232
  35:     0x5640d670d9bf - <alloc::boxed::Box<F> as core::ops::function::FnOnce<A>>::call_once::h7bfd6e98bcf6c67e
                               at /rustc/5e1a799842ba6ed4a57e91f7ab9435947482f7d8/src/liballoc/boxed.rs:1022
  36:     0x5640d6727fd0 - <alloc::boxed::Box<F> as core::ops::function::FnOnce<A>>::call_once::h9979661149bbc525
                               at /rustc/5e1a799842ba6ed4a57e91f7ab9435947482f7d8/src/liballoc/boxed.rs:1022
[...]

PR #1779, opened 2020-02-13
Don't hold the fragment pool lock for too long

mzabaluev:
The scoped lock guard held the lock also while
the poll_purge method is called on logs.

### DIFF
--- a/jormungandr/src/fragment/pool.rs
+++ b/jormungandr/src/fragment/pool.rs
@@ -86,8 +86,10 @@ impl Pool {
     }
 
     pub async fn poll_purge(&mut self) -> Result<(), time::Error> {
-        let mut pool = self.pool.lock().await;
-        future::poll_fn(move |cx| pool.poll_purge(cx)).await?;
+        {
+            let mut pool = self.pool.lock().await;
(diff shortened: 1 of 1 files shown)


jormungandr/src/fragment/pool.rs
```diff
@@ -86,8 +86,10 @@ impl Pool {
     }
 
     pub async fn poll_purge(&mut self) -> Result<(), time::Error> {
-        let mut pool = self.pool.lock().await;
-        future::poll_fn(move |cx| pool.poll_purge(cx)).await?;
+        {
+            let mut pool = self.pool.lock().await;
+            future::poll_fn(move |cx| pool.poll_purge(cx)).await?;
+        }
```
NicolasDP: what about adding a `std::mem::drop(pool)` too to make it more obvious ?

mzabaluev: I prefer scoping since it demarcates the lifetime boundaries better and is less prone to refactoring errors.

mzabaluev:
The over-expansive lock be the root cause of #1580 and #1774.

PR #1794, opened 2020-02-14
Use peers retrieved during bootstrap

mzabaluev:
Make use of the list of peers retrieved from a trusted peers during bootstrap.


vincenthz:
for now just rebased on master

PR #1802, opened 2020-02-18
Improve performance of storage streaming

mzabaluev:
Added 0.3 communication channels to `intercom`, with the old future 0.1 API shunted to it in a backward compatible way.

Add non-blocking API in the intercom channels and use it in n the storage pool access closures to send as many items per go as the sink accepts before becoming full.

`Storage::stream_from_to` is changed to use the intercom streams
under the hood.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,700 +4,761 @@
 name = "actix-codec"
 version = "0.2.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "09e55f0a5c2ca15795035d90c46bd0e73a5123b72f68f12596d6ba5282051380"
 dependencies = [
- "bitflags 1.2.1 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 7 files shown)


eugene-babichenko:
Using the intercom thing for a returned stream looks like a bit of an overkill tbh. Simply a channel should be more than enough. But I'm not sure how much complexity it will add if we make `run_block_iter` generic over `Sink`.

mzabaluev:
> Using the intercom thing for a returned stream looks like a bit of an overkill tbh. Simply a channel should be more than enough.

Perhaps, but this is an abstraction layer on top of Tokio channels that we use to get consistent error handling for fallible streaming, and the debug logging is a nice thing to have too. I didn't feel like duplicating the sending code paths for this specific case.

mzabaluev:
> make run_block_iter generic over `Sink`.

No way, we want a non-blocking API that can be used outside of an async task context. `tokio::sync::mpsc` provides that, as now does our `intercom` adaptation layer on top of it.

PR #1805, opened 2020-02-19
[Test] Benchmark api for tests

dkijania:


jormungandr-integration-tests/src/common/jormungandr/starter.rs
```diff
@@ -181,9 +189,22 @@ impl Starter {
         self.config.as_ref().unwrap().clone()
     }
 
+    pub fn start_benchmark_run(&self) -> Option<SpeedBenchmarkRun> {
```
mzabaluev: No plans to use the `#[bench]` infrastructure?
We could build with nightly for benchmarks, it's not a big problem now.

dkijania: noted down. will explore this possibility one in the nearest free time

PR #1809, opened 2020-02-19
Blockchain tip tracking in stats_counter

danielSanchezQ:
In order to avoid storage access from the REST service, the latest tip block is dumped into the stats_counter each time it is updated. That way is directly available by the REST service when it is needed.

### DIFF
--- a/jormungandr/src/stats_counter.rs
+++ b/jormungandr/src/stats_counter.rs
@@ -4,7 +4,6 @@ use jormungandr_lib::time::SecondsSinceUnixEpoch;
 use std::sync::atomic::{AtomicU64, AtomicUsize, Ordering};
 use std::sync::Arc;
 use std::time::Instant;
-use tokio02::sync::RwLock;
 
 const SLOT_START_TIME_UNDEFINED: u64 = u64::max_value();
 
--- a/jormungandr/src/stats_counter.rs
+++ b/jormungandr/src/stats_counter.rs
@@ -80,6 +80,6 @@ impl StatsCounter {
     }
 
     pub fn get_tip_block(&self) -> Option<Arc<Block>> {
-        self.stats.tip_block.load_full().clone()
+        self.stats.tip_block.load_full()
     }
 }
--- a/jormungandr/src/stats_counter.rs
+++ b/jormungandr/src/stats_counter.rs
@@ -1,7 +1,8 @@
 use chain_impl_mockchain::block::Block;
 use jormungandr_lib::time::SecondsSinceUnixEpoch;
 use std::sync::atomic::{AtomicU64, AtomicUsize, Ordering};
-use std::sync::{Arc, RwLock};
+use std::sync::Arc;
+use tokio02::sync::RwLock;
 use std::time::Instant;
 
 const SLOT_START_TIME_UNDEFINED: u64 = u64::max_value();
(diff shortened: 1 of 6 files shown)


jormungandr/src/stats_counter.rs
```diff
@@ -16,6 +17,7 @@ struct StatsCounterImpl {
     block_recv_cnt: AtomicUsize,
     start_time: Instant,
     slot_start_time: AtomicU64,
+    tip_block: RwLock<Option<Block>>,
```
danielSanchezQ: I don't know for sure if this should be a `tokio::RWLock` instead. It felt confusing the mix we have right now with the futures implementations versions plus the tokio version itself.

vincenthz: I think there's a way to make this lockless, some kind of Cell+Arc ?

danielSanchezQ: Well, since it needs to change and I saw that the other `StatsCounterImpl ` attributes are atomic I decided to use that. But we can use `Cell` if it is only accessed synchronously.

danielSanchezQ:
Tests in master branch are not stable atm

jormungandr/src/stats_counter.rs
```diff
@@ -1,6 +1,7 @@
+use chain_impl_mockchain::block::Block;
 use jormungandr_lib::time::SecondsSinceUnixEpoch;
 use std::sync::atomic::{AtomicU64, AtomicUsize, Ordering};
-use std::sync::Arc;
+use std::sync::{Arc, RwLock};
```
mzabaluev: Wait, why are we adding synchronous locks again?
Everything that is used in async code should use Tokio locks for synchronization.

But since you are cloning this block to set it into that mutable context, a better solution might be as well to send it over an in-memory channel to the REST task, like we currently do for the explorer.

danielSanchezQ: Definetely should be a tokio RwLock.

@mzabaluev Let me check about the channel. Should we go for that @vincenthz ?

jormungandr/src/stats_counter.rs
```diff
@@ -69,4 +73,12 @@ impl StatsCounter {
         }
         .map(SecondsSinceUnixEpoch::from_secs)
     }
+
+    pub async fn set_tip_block(&self, block: Option<Block>) {
+        *self.stats.tip_block.write().await = Arc::new(block);
+    }
+
+    pub async fn get_tip_block(&self) -> Arc<Option<Block>> {
```
mzabaluev: Try to change this to return `Option<&Block>` and see if it passes the borrow checker at the use site.

danielSanchezQ:
Rebased latest changes in master

mzabaluev:
The change appears to break the tests, though I can't see why.

danielSanchezQ:
I'm checking the logs one by one now. I don't see why too....still....

jormungandr/src/stats_counter.rs
vincenthz: I'm getting convinced that it shoudn't be a tokio lock on stuff that do:

```
lock
something that is : not blocking, relatively fast, not taking any other lock, not async
unlock
```

mzabaluev: @vincenthz Trouble is, the Tokio executor also relies on locks to implement a work-stealing scheduler, so any contentious locks in tasks will interfere with that.

danielSanchezQ:
@vincenthz , you were right. I made it non-blocking. Finally, I used the  `arc-swap` crate (that we were using in other places too).

jormungandr/src/stats_counter.rs
```diff
@@ -1,7 +1,10 @@
+use arc_swap::ArcSwapOption;
+use chain_impl_mockchain::block::Block;
 use jormungandr_lib::time::SecondsSinceUnixEpoch;
 use std::sync::atomic::{AtomicU64, AtomicUsize, Ordering};
 use std::sync::Arc;
 use std::time::Instant;
+use tokio02::sync::RwLock;
```
NicolasDP: you don't need this import?

PR #1815, opened 2020-02-20
Use compiler option `target-feature=+crt-static` on Windows MSVC release builds

### LINKED ISSUE #1699: v0.8.9 appears to have a new dependency on Windows
**Describe the bug**
Since `v0.8.9` both `jormungandr.exe` and `jcli.exe` require additional dependencies on Windows (Microsoft Visual C++ Runtime Libraries). This is not the case using `v0.8.8` or earlier versions and is seen with both the msvc as well as gnu binaries. When using this version via Daedalus the error is suppressed and there is nothing in the resulting logs to indicate that this is what is happening.

**Mandatory Information**
1. jcli 0.8.9 (HEAD-30d20d2e, release, windows [x86_64]) - [rustc 1.40.0 (73528e339 2019-12-16)]
2. jormungandr 0.8.9 (HEAD-30d20d2e, release, windows [x86_64]) - [rustc 1.40.0 (73528e339 2019-12-16)]

**To Reproduce**
1. Download either of the windows releases from https://github.com/input-output-hk/jormungandr/releases/tag/v0.8.9 
2. Attempt to launch `jormungandr.exe` or call any command via jcli (e.g. `jcli.exe --version`)

**Expected behavior**
Jcli runs and the correct version information is returned.

**Additional context**
The process does not launch and there is an error message indicating that required dependencies are missing.

**via bash:**
`C:/jm/jcli.exe: error while loading shared libraries: api-ms-win-crt-heap-l1-1-0.dll: cannot open shared object file: No such file or directory`

**via cmd\explorer:**
```
---------------------------
jcli.exe - System Error
---------------------------
The code execution cannot proceed because VCRUNTIME140.dll was not found. Reinstalling the program may fix this problem. 
---------------------------
```

mzabaluev:
Fixes #1699

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -32,24 +32,51 @@ jobs:
       matrix:
         config:
           # Linux
-          - {os: ubuntu-latest, toolchain: stable, target: x86_64-unknown-linux-gnu, cross: false}
-          - {os: ubuntu-latest, toolchain: stable, target: aarch64-unknown-linux-gnu, cross: true}
-          - {os: ubuntu-latest, toolchain: stable, target: arm-unknown-linux-gnueabi, cross: true}
-          - {os: ubuntu-latest, toolchain: stable, target: armv7-unknown-linux-gnueabihf, cross: true}
+          - target: x86_64-unknown-linux-gnu
+            os: ubuntu-latest
+            cross: false
+            rustflags: -C lto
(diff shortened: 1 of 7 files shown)


mzabaluev:
Something goes differently in the build image from the system I tested on locally: https://github.com/input-output-hk/jormungandr/runs/457322020

PR #1816, opened 2020-02-20
Synchronous garbage collection in fragments

eugene-babichenko:


eugene-babichenko:
**TBD** group entries in chunks so that there is one chunk per purge call

jormungandr/src/fragment/expirations.rs
```diff
@@ -0,0 +1,93 @@
...
+use std::{
+    collections::{BTreeMap, HashSet},
+    time::{Duration, Instant},
+};
+
+/// A synchronous alternative to tokio's `DelayQueue` intended to be
+/// periodically polled for expired entries.
+pub struct Expirations<T> {
+    slab: Slab<ExpirationEntry<T>>,
+    sorted: BTreeMap<Instant, HashSet<usize>>,
```
NicolasDP: Use the `Key` to remain consistent

```suggestion
    sorted: BTreeMap<Instant, HashSet<Key>>,
```

NicolasDP:
worried about size of the logs, might want to add some bounds too and circular buffering?

eugene-babichenko:
This is out of scope this pull request. Besides, I think the size of the log is not really critical because of the short TTL of log entries.

NicolasDP:
The log entries don't have short TTL

jormungandr/src/fragment/expirations.rs
```diff
@@ -0,0 +1,104 @@
...
+use std::{
+    collections::{BTreeMap, HashSet},
+    time::{Duration, Instant},
+};
+
+/// A synchronous alternative to tokio's `DelayQueue` intended to be
+/// periodically polled for expired entries.
+pub struct Expirations<T> {
+    slab: Slab<ExpirationEntry<T>>,
+    sorted: BTreeMap<SlotId, HashSet<Key>>,
```
mzabaluev: The B-tree not ideal for popping entries, which is likely the most frequent operation. But since this queue must also support removals and reschedulings, a simple binary heap would be a poor fit too.

jormungandr/src/fragment/expirations.rs
```diff
@@ -0,0 +1,104 @@
...
+        for (expires_at, _) in to_remove.iter() {
+            self.sorted.remove(expires_at);
+        }
+
+        to_remove
+            .into_iter()
+            .map(|(_, slab_keys)| slab_keys.into_iter())
+            .flatten()
+            .map(|key| self.slab.remove(key.0).data)
+            .collect()
```
mzabaluev: Both things could be done in one pass.

jormungandr/src/fragment/expirations.rs
```diff
@@ -0,0 +1,104 @@
...
+        use std::ops::Bound::*;
+
+        let now = self.get_slot_id(Instant::now());
+        // the current slot is excluded because it is still in progress
+        let range = (Unbounded, Excluded(&now));
+        let to_remove = self
+            .sorted
+            .range(range)
+            .map(|(expires_at, entry)| (expires_at.clone(), entry.clone()))
+            .collect::<Vec<(SlotId, HashSet<Key>)>>();
```
mzabaluev: You can use `split_off` method to split off part of the tree as long as you are going to remove the other part. This might be more efficient than using an intermediate vector, and certainly less code.

jormungandr/src/fragment/mod.rs
```diff
@@ -1,4 +1,5 @@
 mod entry;
+mod expirations;
```
mzabaluev: May want to host this module somewhere in `utils`, I think we'll need the same generic data structure elsewhere we do GC.

jormungandr/src/fragment/pool.rs
```diff
@@ -192,59 +190,40 @@ pub(super) mod internal {
...
+                    self.expirations.remove(expiration_key);
+                }
+            }
+            let mut clean_entries_by_time = VecDeque::new();
+            for entry in self.entries_by_time.iter().cloned() {
+                if !to_remove.contains(&entry) {
+                    clean_entries_by_time.push_back(entry);
+                }
             }
+            self.entries_by_time = clean_entries_by_time;
```
mzabaluev: `VecDeque::retain` would be simpler and more efficient.

jormungandr/src/fragment/pool.rs
```diff
@@ -192,59 +190,40 @@ pub(super) mod internal {
...
-
-                    // Here Pending means there are still items in the DelayQueue but
-                    // they are not expired. We don't want this function to wait for these
-                    // ones to expired. We only cared about removing the expired ones.
-                    Poll::Pending => return Poll::Ready(Ok(())),
-                }
-            }
+        pub fn purge(&mut self) {
+            let expired = self.expirations.pop_expired();
+            self.remove_all(expired);
```
mzabaluev: I wonder if making `pop_expired` return an `impl Iterator` and then piping that into `.remove_all` would make it more efficient. But this can wait for another PR.

jormungandr/src/fragment/expirations.rs
```diff
@@ -0,0 +1,104 @@
...
+            None => {
+                let mut slot = HashSet::new();
+                slot.insert(key);
+                self.sorted.insert(expires_at, slot);
+            }
+        }
+        entry.expires_at = expires_at;
+    }
+
+    pub fn pop_expired(&mut self) -> Vec<T> {
```
mzabaluev: As discussed elsewhere, this could return `impl Iterator<Item = T>`.

mzabaluev:
There are more changes to garbage collection here than just the logs, so it should be rebased.

NicolasDP:
> There are more changes to garbage collection here than just the logs, so it should be rebased.

And I am talking about both `fragment pool and fragment logs`.

eugene-babichenko:
closing in favor of #1828

PR #1826, opened 2020-02-21
Remove delay queue from ref cache

NicolasDP:
This continues the work to remove the slow entry points where some locks are blocking un order to perform GC.

### DIFF
--- a/jormungandr/src/main.rs
+++ b/jormungandr/src/main.rs
@@ -337,8 +337,10 @@ fn bootstrap(initialized_node: InitializedNode) -> Result<BootstrappedNode, star
 
     let block0_explorer = block0.clone();
 
+    let cache_capacity = 10_240;
+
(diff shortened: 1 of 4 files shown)


jormungandr/src/main.rs
```diff
@@ -337,11 +337,10 @@ fn bootstrap(initialized_node: InitializedNode) -> Result<BootstrappedNode, star
 
     let block0_explorer = block0.clone();
 
-    // TODO: we should get this value from the configuration
-    let block_cache_ttl: Duration = Duration::from_secs(5 * 24 * 3600);
+    let cache_capacity = 10_240;
```
vincenthz: Should be way higher than this, probably x10

NicolasDP:
```suggestion
    let cache_capacity = 102_400;
```

PR #1828, opened 2020-02-21
Use LRU instead of DelayQueue in fragment module

eugene-babichenko:


jormungandr/src/fragment/pool.rs
```diff
@@ -157,25 +157,13 @@ pub(super) mod internal {
...
         pub fn remove_all(&mut self, fragment_ids: impl IntoIterator<Item = FragmentId>) {
-            // TODO fix terrible performance, entries_by_time are linear searched N times
-            for fragment_id in fragment_ids {
-                self.remove(&fragment_id);
+            use std::collections::HashSet;
+            let fragment_ids: HashSet<_> = fragment_ids.into_iter().collect();
+            for fragment_id in fragment_ids.iter() {
+                self.entries.pop(fragment_id);
             }
+            self.entries_by_time.retain(|id| !fragment_ids.contains(id));
```
NicolasDP: remove the `entries_by_time`. It is no longer needed because they entries are already ordered by time in the pool now that we use the LRU cache

eugene-babichenko: Ah, you're right, they're not modified, so the oldest entry in the cache would be exactly what we need here.

PR #1836, opened 2020-02-21
Time limits on client connections

### LINKED ISSUE #991: Hanging (i.e. no time-out implemented?) during bootstrap phase of the node startup
Using Windows 10, powershell

**Starting jormungandr 0.6.3 (master-cf87d45+, release, windows [x86-64)]  [rust 1.38.0], task:init
Storing blockchain in '"c:/<...>/blocks.sqlite"', task: init
Connecting to bootstrap peer <ip address>:4000, peer_addr: <ip address>:4000, task: bootstrap**

It sat at the above line for 51 minutes before my friend restarted their node (referenced by the ip address above), after which my node logged:

WARN initial bootstrap failed, error: PullRequesteFailed { source: Error { code: Unknown, source: Status { code: Unknown, message: "connection error: broken pipe" } } }, peer_addr: <ip>:4000, task bootstrap

### LINKED ISSUE #1829: grpc connect / shortly after seems to get stuck permanently

mzabaluev:
Limit time it can take to establish a client connection, taking the timeout value in `Peer` to its intended use.
Unfortunately, `hyper` 0.12 does not provide a way to pass it down the TCP stack, ~so we may still end up with sockets hanging in connecting state~. With the port to `hyper` 0.13 and Tokio 0.2, a more organic solution will be possible.

An attempt to fix #1829 (or at least figure out that the executor is stuck).
Fixes #991

### DIFF
--- a/jormungandr/src/settings/start/network.rs
+++ b/jormungandr/src/settings/start/network.rs
@@ -42,7 +42,7 @@ pub const DEFAULT_MAX_CONNECTIONS: usize = 256;
 pub const DEFAULT_MAX_CLIENT_CONNECTIONS: usize = 8;
 
 /// The default timeout for connections
-const DEFAULT_TIMEOUT_MILLISECONDS: u64 = 10_000;
+const DEFAULT_TIMEOUT: Duration = Duration::from_secs(10);
 
 ///
 /// The network static configuration settings
@@ -109,10 +109,7 @@ impl From<super::config::TrustedPeer> for TrustedPeer {
 
 impl Peer {
     pub fn new(connection: SocketAddr) -> Self {
-        Peer::with_timeout(
-            connection,
-            Duration::from_millis(DEFAULT_TIMEOUT_MILLISECONDS),
-        )
+        Peer::with_timeout(connection, DEFAULT_TIMEOUT)
     }
 
     pub fn with_timeout(connection: SocketAddr, timeout: Duration) -> Self {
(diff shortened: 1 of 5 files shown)


jormungandr/src/settings/start/network.rs
```diff
@@ -107,24 +108,32 @@ impl From<super::config::TrustedPeer> for TrustedPeer {
...
 
 impl Listen {
-    pub fn new(connection: SocketAddr, protocol: Protocol) -> Self {
+    pub fn new(connection: SocketAddr) -> Self {
         Listen {
             connection,
-            protocol,
-            timeout: Duration::from_micros(DEFAULT_TIMEOUT_MICROSECONDS),
+            protocol: Protocol::Grpc,
+            timeout: Duration::from_micros(DEFAULT_TIMEOUT_MILLISECONDS),
```
NicolasDP: you are still using `Duration::from_micros`
```suggestion
            timeout: Duration::from_millis(DEFAULT_TIMEOUT_MILLISECONDS),
```

mzabaluev: Thanks for noticing. I've used a const fn to initialize a `Duration` const in one place.

mzabaluev:
> we may still end up with sockets hanging in connecting state

Not an issue, assuming that `std::net::TcpStream` cleans up OS resources when dropped.

NicolasDP:
I am surprised by the CI error 🤔

vincenthz:
the expected error is in the log for `genesis_wrong_hash``, but not in the same format, effectively something that looks like Debug instead of Display.

PR #1441, opened 2019-12-24
enforce the correct order of writes to storage

eugene-babichenko:
Removing the global lock from the storage allowed for better
performance. However, there is an issue of writes possibly done
out of the expected order due to the underlying pool of parallel sqlite
connections or put_block/tag tasks being run in parallel tokio
executors. Now we introduce the global write lock for the block storage
while still allowing parallel reads.

Should fix #1422 (p. 3) #1438

PR #1412, opened 2019-12-20
leverage SQLite multi-threading power by removing the bottleneck lock

NicolasDP:
This PR makes the node more resilient with multiple connections and remove locks that were creating bottle necks and potential deadlocks.

**STILL EXPERIMENTAL**

PR #1840, opened 2020-02-23
A more efficient way to synchronize storage writes

eugene-babichenko:
## Background

We allowed parallel reads/writes from/to the storage without any application-level which was introduced in #1412. While providing a significant performance benefit, it resulted in a bug: storage operations involving write transactions:

1) could've been scheduled to different executor threads
2) almost always used different connections

This led to a situation when a block could've been written to the storage before its parent (causing a panic) or less likely when a tag was written before the corresponding block.

An obvious way to fix that was to synchronize the writes like it was before #1412 but to keep reads parallel. To achieve that, a lock for write operations was introduced (#1441).

## Motivation

The downside of the previous approach was that it involved a number of costly operations, namely: locking on a mutex and multiple operations of acquiring and releasing a DB connection. This what a write operation looked like:

- acquire the mutex
- acquire a connection
- get a connection
- do _one_ operation in a thread pool
- release a connection
- release a lock

## Approach

Instead of locking the database for certain cases we use a queue of writes. A separate process (service) receives requests for write operations. Instead of performing the above sequence of operations for each request, this process reads all of the incoming operations from the incoming stream and executes them in a batch. Thus, we avoid costly locking and repeated connections operations. This may improve the performance in places like network bootstrap.

## Further uses

Similar approach can be applied explicitly for processing large streams of requests.

## TBD in this PR

- [ ] Use `intercom` utils instead of bare `futures` channels.
- [ ] Limit the size of requests batches.
- [ ] Logging in the requests process.

### DIFF
--- a/jormungandr/src/blockchain/storage.rs
+++ b/jormungandr/src/blockchain/storage.rs
@@ -5,11 +5,17 @@ use crate::{
 };
 use chain_storage_sqlite_old::{for_path_to_nth_ancestor, BlockInfo};
 use futures::{Future as Future01, Stream as Stream01};
-use futures03::{compat::Compat, prelude::*, ready, stream::FusedStream};
+use futures03::{
+    channel::{mpsc, oneshot},
+    compat::Compat,
+    prelude::*,
+    ready,
+    stream::FusedStream,
+};
 use pin_utils::{unsafe_pinned, unsafe_unpinned};
 use r2d2::{ManageConnection, Pool};
 use slog::Logger;
-use tokio02::{sync::Mutex, task::spawn_blocking};
+use tokio02::task::spawn_blocking;
 
 use std::convert::identity;
 use std::error::Error;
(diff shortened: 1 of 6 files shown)


jormungandr/src/blockchain/storage.rs
```diff
@@ -156,17 +157,83 @@ enum StreamingError {
...
+    pub async fn start(self) {
+        let StorageSyncQueryExecutor { mut queries, pool } = self;
+
+        let mut queries = stream::poll_fn(move |cx| {
+            let mut queries_vec = Vec::new();
+            loop {
+                let queries_queue = Pin::new(&mut queries);
+                match queries_queue.poll_next(cx) {
+                    Poll::Ready(Some(query)) => queries_vec.push(query),
+                    Poll::Ready(None) => return Poll::Ready(None),
```
eugene-babichenko: potentially broken: there may be some blocks in `queries_queue` which should be returned

mzabaluev:
I don't understand why do we need any external synchronization to SQLite. Confirming writes should be simple: { begin transaction, update tables, commit transaction }, resolve the block writing future. The part in brackets can be performed in the sync closure that gets the pooled storage connection. If the table update is done in one query, the explicit transaction statements are not needed.
This ignores possibility of transactions being rejected, but I'd need to see how it is possible with the append-only storage data model.

If jormungandr bombs blocks from a stream into storage without waiting for their parent to be committed (i.e. the future described above to resolve for the parent block), this should be fixed instead.

eugene-babichenko:
closing in favor of #1852 and #1853

PR #1845, opened 2020-02-24
Structopt to 0.3

danielSanchezQ:
Fixes #1708


jormungandr-integration-tests/src/common/jcli_wrapper/certificate/commands.rs
```diff
@@ -26,6 +26,9 @@ impl CertificateCommands {
             .arg("new")
             .arg("stake-delegation")
             .arg(&delegation_id)
+            // This was added as part of updating structopt to 0.3+, it changed the interface
+            // and required to add `--` before the stake pool ids
+            .arg("--")
```
danielSanchezQ: I had to add this in order to make the command work. I could not find the change in the documentation of `structopt` but it is due to a change in it. It means that now a command should be run with:
```
jcli.exe certificate new stake-delegation delegation_id -- stake_pool_id
```
I'm taking a look to see if I can make it work without the `--` to keep the same behavior as in previous versions

jormungandr-integration-tests/src/jcli/key/from_bytes.rs
```diff
@@ -52,6 +52,6 @@ pub fn test_from_bytes_for_unknown_key() {
     jcli_wrapper::assert_key_from_bytes_fails(
         &byte_key_file,
         "ed25519Exten",
-        "Invalid value for '--type <key_type>'",
+        "Invalid value for '--type <key-type>':",
```
danielSanchezQ: Command renaming behavior [changed here](https://github.com/TeXitoi/structopt/blob/master/CHANGELOG.md#change-default-case-from-verbatim-into-kebab-by-0ndorio-202)

NicolasDP: this one is not a problem, it does not change the user's APIs

jormungandr-integration-tests/src/common/jcli_wrapper/certificate/commands.rs
vincenthz: is that maybe a bug in structopt ?

eugene-babichenko: Strange. I tested on macOS with and without `--` and both work just fine.

danielSanchezQ: @eugene-babichenko , did you check with this version (`0.3^`)?

danielSanchezQ: I made a `cargo clean`, `cargo build -all` and `cargo test`, it is working as expected. I may had a bad build when testing it before.

eugene-babichenko: Great to hear. Yes, I was playing with this branch.

danielSanchezQ:
I removed the `--`, tests are working. The issue was triggered because of a bad local build.

PR #1852, opened 2020-02-25
Avoid inconsistency between block store and state

eugene-babichenko:
`apply_and_store_block` was calling the `apply_block` function before
the actual write to storage with resulted into the inconsistent
behavior. If there was an error when writing a block to the storage, the
function was exiting without retrying to write the block. Thus, the
block was applied to the state but not to the storage.

To avoid that, `apply_block` is now split into two functions:

* `apply_block_dry_run` checks if the provided block can be applied to
  the state and returns the draft state without applying it anywhere.
  This function is called before writing to the storage.
* `apply_block_finalize` updates the state and returns a block
  reference. This is called only after a successful write.

Blockchain tip is updated only after the tag was successfully written to the storage.

### DIFF
--- a/jormungandr/src/blockchain/chain.rs
+++ b/jormungandr/src/blockchain/chain.rs
@@ -482,37 +482,47 @@ impl Blockchain {
         }
     }
 
-    async fn apply_block(
+    fn apply_block_dry_run(
         &self,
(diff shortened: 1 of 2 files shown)


NicolasDP:
Good work

PR #1866, opened 2020-02-27
Warnings cleaning

danielSanchezQ:
Cleaned some warnings about:
* Unused import
* Unused variables
* No captured results

Left:
* Warnings in `chain-deps`
* Deadcode warnings
* Deprecated code warnings (they should be update in forward PRs related to other issues)

### DIFF
--- a/jormungandr-integration-tests/src/common/configuration/node_config_builder.rs
+++ b/jormungandr-integration-tests/src/common/configuration/node_config_builder.rs
@@ -27,7 +27,6 @@ impl NodeConfigBuilder {
     pub fn new() -> NodeConfigBuilder {
         let rest_port = super::get_available_port();
         let public_address_port = super::get_available_port();
-        let _listen_address_port = super::get_available_port();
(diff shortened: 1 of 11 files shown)


jormungandr-integration-tests/src/common/configuration/node_config_builder.rs
```diff
@@ -27,7 +27,7 @@ impl NodeConfigBuilder {
     pub fn new() -> NodeConfigBuilder {
         let rest_port = super::get_available_port();
         let public_address_port = super::get_available_port();
-        let listen_address_port = super::get_available_port();
+        let _listen_address_port = super::get_available_port();
```
eugene-babichenko: Can we just remove that? Seems like this unused statement doesn't have any side-effects.

danielSanchezQ: Yeah, let's do that, I chose the `_` because it was a less aggressive  change

Issue #1495, opened 2020-01-03
Tokio/Future/Async updates


vincenthz:
This issue is to tie with the work on **Jormungandr services and watchdog** (#1488). Especially for the part regarding transition to new async await.

the strategy here is to leverage the work on `tokio-compat` to simplify the transition as much as possible and to isolate the work to rewrite with async/await the different services. In order to do that, the following path is going to be under taken:

1. blocked on the first step of #1488: writing the first PoC watchdog to replace the old task management service.
2. Once #1488 has been started and the first implementation of the watchdog is available we can then transition from the old task.rs model to the new service based model, still utilising the compat layer of the legacy `Future` API.
3. Once every services are wrapped in the compat layer, we will be able to move to new service API and async/await feature, unfortunatelly, not entirely independently: the intercom mechanism are still utilising the legacy `Future`. #1488 will provide a new intercom mechanism and we will need to have a temporary layer of abstraction in order to allow for the smooth compatibility.

## tasks

Following task are to be made in order, each task being dependency of the following task

- [x] use default `Runtime` (#1547)
- [x] use `tokio-compat` (#1548)
- [ ] add compatibility layer with the first implementation of the watchdog;
- [ ] move from the task.rs service to the new watchdog's legacy service runner;
- [x] prepare temporary APIs or strategy for the intercom compatibility;

Once this is done, the following can then start to be done more or less independently. However keep in mind that #1488 will being changes on how the different services are meant to communicate with each other.

- [x] replace the REST service
- [x] update the Network service
- [x] update the MemPool Service
- [x] update the Leadership Service
- [x] update the client service
- [x] update the Blockchain service
- [x] update the (optional) Explorer service

mzabaluev:
Closing the outdated tracking item. The async update is done.

PR #1907, opened 2020-03-12
move different modules to new tokio runtime

NicolasDP:
removing un-needed use of compat layer

part of #1495


jormungandr/src/utils/task.rs
```diff
@@ -118,6 +119,49 @@ impl Services {
         self.services.push(task);
     }
 
+    /// Spawn the given Future in a new dedicated runtime
+    pub fn spawn_try_future_std<F, T>(&mut self, name: &'static str, f: F)
```
eugene-babichenko: Why do that if we already have `spawn_failable_std`?

NicolasDP: because that's for a task, not for a service.

PR #1908, opened 2020-03-12
blockchain: convert internals to futures 0.3

eugene-babichenko:
Everything is async/await now, except multiverse, which is used by explorer, which is futures 0.1 now.


eugene-babichenko:
I think it is ready to merge, since I don't want to do much work on `explorer` in `blockchain`-related pull request.

PR #1912, opened 2020-03-13
Port network to tokio 0.2 and tonic

mzabaluev:


eugene-babichenko:
@mzabaluev pushed all made in #1940

Can you take a look please?

jormungandr/src/network/p2p/comm.rs
```diff
@@ -611,40 +607,38 @@ impl Peers {
...
-                    );
-                    entry.remove();
-                    e.into_item()
-                })
-            } else {
-                Err(gossip)
-            }
-        })
+        self.inner()
+            .and_then(move |mut map| {
```
NicolasDP, line 618: can't we do the other way around:

```rust
   let mut map = self.inner().await?;
   if let Some(mut entry) = map.entry(target) {
      ...
   }
```

NicolasDP: I just find it is easier to read (separate the async logic from the processing logic

jormungandr/src/network/p2p/comm.rs
```diff
@@ -611,40 +607,38 @@ impl Peers {
...
+                        e.into_item()
+                    })
+                } else {
+                    Err(gossip)
+                }
+            })
+            .await
     }
 
     pub fn refresh_peer_on_block<E>(&self, node_id: Id) -> impl Future<Item = bool, Error = E> {
```
NicolasDP, line 644: do we still need these old futures?

jormungandr/src/network/p2p/comm.rs
mzabaluev: You are right, changed in another commit.

jormungandr/src/network/p2p/comm.rs
mzabaluev: I have converted these in another commit.

NicolasDP:
Woohoo!

PR #1917, opened 2020-03-14
Changed default value of max_client_connections

michaeljfazio:
The default value for `max_client_connections` doesn't make sense.

When connections > `max_client_connections` the Jormungandr will initiate a purge of `count - self.config.max_client_connections` connections. With a default value of `8` this means that the node will find it difficult to build up a stable topology because the purge limit is hit far too soon.

### DIFF
--- a/doc/configuration/network.md
+++ b/doc/configuration/network.md
@@ -75,7 +75,7 @@ rest:
 - `max_connections`: the maximum number of P2P connections this node should
     maintain. If not specified, an internal limit is used by default `[default: 256]`
 - `max_client_connections`: the maximum number of client P2P connections this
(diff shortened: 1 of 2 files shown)


Linicks:
This made a huge improvement to my nodes.  Please merge this!

NicolasDP:
thanks, seems right to me. actually this is raising questions about the handling of clients connection. This parameter was meant to control the number of client connections as in the number of non publicly reachable connections. But it seems we are filtering without discrimination the publicly and non publicly reachable.

NicolasDP:
in the spirit, we should rename (and add an alias to it for compat) this value parameter to `max_inboud_connections`.

Issue #1919, opened 2020-03-16
One Stats to rule them all


NicolasDP:
We have a Stats object defined in `jormungandr` and a poor man `Stats` in `jormungandr-lib` that does provide a json deserialization compat layer to the one defined in `jormungandr`.

see comment: https://github.com/input-output-hk/jormungandr/pull/1918#commitcomment-37833164

This would benefit from being properly defined and implemented for `jormungandr-lib` and `jormungandr` avoiding future breakage.

PR #1923, opened 2020-03-16
Rest: node stats - use interface dto

rinor:
fix https://github.com/input-output-hk/jormungandr/issues/1919


dkijania:
can you also build jormungandr-scenario-tests project and check it compiles ?

dkijania:
Those changes contains fix for private network tests, thus merging it now

Issue #1830, opened 2020-02-21
partially received netboot stream should be kept in the blockchain state


vincenthz:
it seems that we don't update any storage state, so cancelling a partial netboot, seems to restart everything

eugene-babichenko:
Currently we only update the main tag in storage so it gets lost if bootstrap is interrupted. I pushed https://github.com/input-output-hk/jormungandr/pull/1940/commits/003b0ea86591529a0bc7aad770029d83010984b0 to address this issue but I'm not sure if this is exactly what you wanted.

If you meant shutting down the node in the midst of bootstrap, then we need to handle `SIGINT` and tell the bootstrap process to stop in some way. Probably by using channels or something like that.

eugene-babichenko:
So, I tried 3 approaches for that:

1. Just update the tip for each new block. about 30% slower than what we have now.
2. Catch `SIGINT`, update the tip and shutdown. No performance drop, works perfectly fine, but the code is a disaster.
3. Update tip every `n` blocks. Looks like a good trade-off. Can be implemented by adding a few lines of code to `bootstrap_from_stream`. If we update tip on every 50-th or 100-th block we miss a reasonable number of block on interruption. At the same time no significant performance drop was observed.

PR #1940, opened 2020-03-19
Convert bootstrap to futures 0.3 + multiple improvements

eugene-babichenko:


mzabaluev:
This will conflict a lot with #1912, where much of the same is also done.

eugene-babichenko:
Contains a solution for #1830

jormungandr/src/main.rs
```diff
@@ -276,6 +278,8 @@ fn start_services(bootstrapped_node: BootstrappedNode) -> Result<(), start_up::E
         });
     }
 
+    services.spawn_try_future_std("sigint_listener", move |_info| ctrl_c().map_err(|_| ()));
+
```
NicolasDP: this is unfortunate to do that... :/

eugene-babichenko: Actually I can check if can live without that. I was a bit confused by this line from the documentation

> Even if this Signal instance is dropped, subsequent SIGINT deliveries will end up captured by Tokio, and the default platform behavior will NOT be reset.

but Tokio may be able to do some implicit handling.

eugene-babichenko: but it doesn't :(

eugene-babichenko:
Closing this pull request as the changes made here were ported to #1912.

PR #1944, opened 2020-03-20
update submodule & deps

vincenthz:
supersede dependabot many PRs whilst also updating chain-libs


NicolasDP:
done in master already

PR #1970, opened 2020-03-26
Initiate gossips fix 1946

### LINKED ISSUE #1946: When starting gossiping, network does not check for already connected node
Indeed, we call `topology.view(Any)` and quickly we already to `connect_and_propagate` with the node without checking if the node is already connected to us or not:

https://github.com/input-output-hk/jormungandr/blob/19f8eb49e15165e4a50a248e6a19221376f07cf6/jormungandr/src/network/mod.rs#L404-L414

---

instead of doing the pre-filtering of the already connected node it might be easier to do the already connected check in the `connect_and_propagate` function.

danielSanchezQ:
Fixes #1946

### DIFF
--- a/jormungandr/src/network/mod.rs
+++ b/jormungandr/src/network/mod.rs
@@ -381,11 +381,13 @@ fn handle_propagation_msg(
 
 fn start_gossiping(state: GlobalStateR, channels: Channels) -> impl Future<Item = (), Error = ()> {
     let config = &state.config;
-    let topology = state.topology.clone();
+    let topology_accept = state.topology.clone();
+    let topology_initiate = state.topology.clone();
     let conn_state = state.clone();
+    let logger = state.logger().new(o!(log::KEY_SUB_TASK => "start_gossip"));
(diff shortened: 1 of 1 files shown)


jormungandr/src/network/mod.rs
```diff
@@ -381,11 +381,13 @@ fn handle_propagation_msg(
 
 fn start_gossiping(state: GlobalStateR, channels: Channels) -> impl Future<Item = (), Error = ()> {
     let config = &state.config;
-    let topology = state.topology.clone();
+    let topology_accept = state.topology.clone();
```
danielSanchezQ: I had to duplicate the topology. I'm not fully aware of what consequences could this has.

danielSanchezQ:
@mzabaluev I created a draft with this so we can discuss about the solution and needs to be done here.

jormungandr/src/network/mod.rs
NicolasDP: It's a `Arc<Mutex<...>>` under the hood

Issue #1969, opened 2020-03-26
Native RustTLS not supported on android build?


NicolasDP:
**Describe the bug**

On the latest release we canno longer build the ARVM android build: https://github.com/input-output-hk/jormungandr/runs/536061370?check_suite_focus=true

The issue is:

```
error[E0432]: unresolved import `platform`
##[error]  --> /cargo/registry/src/github.com-1ecc6299db9ec823/rustls-native-certs-0.1.0/src/lib.rs:22:9
   |
22 | pub use platform::load_native_certs;
   |         ^^^^^^^^ use of undeclared type or module `platform`
```

**Mandatory Information**
1. `jcli --full-version`: v0.8.16;
2. `jormungandr --full-version` v0.8.16;

**To Reproduce**

Steps are described here: https://github.com/input-output-hk/jormungandr/runs/536061370?check_suite_focus=true

**Expected behavior**

To compile

**Additional context**

recent change introduced from #1961

eugene-babichenko:
Oops, seems so. https://github.com/ctz/rustls-native-certs/issues/3

I will at what we can do.

eugene-babichenko:
The fix was merged in reqwest, waiting for a new release.

PR #1976, opened 2020-03-27
update all crates to use the same version of reqwest

eugene-babichenko:


eugene-babichenko:
@dkijania this touches the testing part (incl. scenario tests), can you please check this?

NicolasDP:
If this fixes #1969 please say so in the PR comment

eugene-babichenko:
@NicolasDP it does not. Just a migration to make the dependency version consistent across all crates.

PR #1982, opened 2020-03-27
Added whitelisting to Policy

### LINKED ISSUE #1973: Network whitelist: prevent specific addresses (pool id) to be quarantined forever
Based on the covid-19 events across the world and the quarantines/lockdowns of many countries. it became obvious the poldercast policy module is missing an important aspect: essential workers.

Just like in our real world there are services, people, our society relies on. Doctors, nurses, postmans, power stations, water, ... having a quarantine policy that does not allow to distinguish important agents from the crowed.

the goal is to add a configuration parameter to the policy in order to avoid blacklisting some IPs addresses that are known to be important to us/the network.

---

**consideration**: be ready to also add some extension with the regard to the 4th layer (#1972 ). We will need to also make sure none of the stake pool listed in this 4th layer are ever evicted/blacklisted or quarantined.

danielSanchezQ:
closes #1973

### DIFF
--- a/doc/configuration/network.md
+++ b/doc/configuration/network.md
@@ -81,7 +81,7 @@ rest:
     it back (or not) into the fold.
     It is recommended to leave the default value `[default: 30min]`.
   - `quarantine_whitelist` set a trusted list of peers that will not be quarantined in any circumstance. 
-    It should be a list of valid addresses, for example: `["/ip4/127.0.0.1"]`.
+    It should be a list of valid addresses, for example: `["/ip4/127.0.0.1/tcp/3000"]`.
     By default this list is empty, `[default: []]`.
 - `layers`: (optional) set the settings for some of the poldercast custom layers (see below)
 - `max_unreachable_nodes_to_connect_per_event`: (optional) set the maximum number of unreachable nodes
--- a/doc/configuration/network.md
+++ b/doc/configuration/network.md
@@ -80,6 +80,9 @@ rest:
   - `quarantine_duration` set the time to leave a node in quarantine before allowing
     it back (or not) into the fold.
     It is recommended to leave the default value `[default: 30min]`.
+  - `quarantine_whitelist` set a trusted list of peers that will not be quarantined in any circumstance. 
+    It should be a list of valid addresses, for example: `["/ip4/127.0.0.1"]`.
+    By default this list is empty, `[default: []]`.
 - `layers`: (optional) set the settings for some of the poldercast custom layers (see below)
 - `max_unreachable_nodes_to_connect_per_event`: (optional) set the maximum number of unreachable nodes
   to contact at a time for every new notification.
--- a/jormungandr/src/network/p2p/policy.rs
+++ b/jormungandr/src/network/p2p/policy.rs
@@ -3,6 +3,7 @@ use lru::LruCache;
 use poldercast::{Id, Node, PolicyReport};
 use serde::{Deserialize, Serialize};
 use slog::Logger;
+use std::collections::HashSet;
 use std::time::{Duration as StdDuration, Instant};
 
 /// default quarantine duration is 10min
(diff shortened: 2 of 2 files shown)


danielSanchezQ:
Should we be able to add more addresses to the whitelist after loading the Policy?

jormungandr/src/network/p2p/policy.rs
```diff
@@ -116,7 +121,12 @@ impl poldercast::Policy for Policy {
     fn check(&mut self, node: &mut Node) -> PolicyReport {
         let id = node.id().to_string();
         let logger = self.logger.new(o!("id" => id));
-
+        // if the node is whitelisted
+        if let Some(Some(socket_address)) = node.address().map(|a| a.to_socketaddr()) {
+            if self.quarantine_whitelist.contains(&socket_address) {
+                return PolicyReport::None;
+            }
+        }
```
rinor: If it is all the same I would move this inside quarantine checks https://github.com/input-output-hk/jormungandr/blob/5b6268aea9d9fa9e6662586a3ff6b840123a4835/jormungandr/src/network/p2p/policy.rs#L157-L162 and maybe add a debug log before `return` 
```rust
debug!(
    logger,
    "node is whitelisted, peer_addr: {}",
    socket_address.to_string()
);
```

jormungandr/src/network/p2p/policy.rs
```diff
@@ -40,6 +41,8 @@ pub struct PolicyConfig {
     max_quarantine: Option<Duration>,
     #[serde(default)]
     max_num_quarantine_records: Option<usize>,
+    #[serde(default)]
+    quarantine_whitelist: Option<HashSet<std::net::SocketAddr>>,
```
NicolasDP: use the [`Address`](https://docs.rs/poldercast/0.13.0/poldercast/struct.Address.html) directly here. It supports `serde` already. and also it supports extending the support for FQDN later on if needed.

jormungandr/src/network/p2p/policy.rs
NicolasDP: yes, in the `if else` to make things more consistent:

```
} else if node in the whitelist {
    debug!("should be quarantined, but whitelist so it's cool bro...");
    PolicyReport::None
 } else { 
     // if the record is not `clear` then we quarantine the block for some time 
     debug!(logger, "move node to quarantine"); 
     self.update(node.id().clone()); 
     PolicyReport:: LiftQuarantine; // just in case! 
 } 
```

jormungandr/src/network/p2p/policy.rs
```diff
@@ -40,6 +41,8 @@ pub struct PolicyConfig {
     max_quarantine: Option<Duration>,
     #[serde(default)]
     max_num_quarantine_records: Option<usize>,
+    #[serde(default)]
+    quarantine_whitelist: Option<HashSet<Address>>,
```
NicolasDP: the default of `HashSet` is already an empty `HashSet`:
```suggestion
    #[serde(default)]
    quarantine_whitelist: HashSet<Address>,
```

danielSanchezQ: Ok, it makes sense. Actually I had in mind that in some cases we would not want that whitelist, and my mind pop up `Option` instantly.

NicolasDP:
I was about to approve and merge it, and then I realised: there is no documentation about the new configuration options

danielSanchezQ:
I added the docs for the `quarantine_whitelist` @NicolasDP

doc/configuration/network.md
```diff
@@ -80,6 +80,9 @@ rest:
   - `quarantine_duration` set the time to leave a node in quarantine before allowing
     it back (or not) into the fold.
     It is recommended to leave the default value `[default: 30min]`.
+  - `quarantine_whitelist` set a trusted list of peers that will not be quarantined in any circumstance. 
+    It should be a list of valid addresses, for example: `["/ip4/127.0.0.1"]`.
```
rinor: I don't think it works that way. It should be a multi format address `["/ip4/127.0.0.1/tcp/3000"]` iirc

danielSanchezQ: It comes from the `MultiAddress` example. But I updated it so is a clearer example. Thanks!

PR #1990, opened 2020-03-30
Preferred list for poldercast layer

NicolasDP:
this PR contains a preferred list of nodes that can be used to build
a special layer of extra nodes that we want to keep in constant connections
with.

This is going to be useful for IOHK or operators with multiple nodes to
make sure their nodes are constantly communicating with each others.

- [x] once is updated, make sure the preferred list also maintain the whitelist #1982;
- [x] depends on `poldercast` version `0.13.0` so we don't enter NodeIDs

---

- Regarding the dependency to `0.13.0`: mitigation has been added.
- no need for whitelisting because the entries are added forcibly in the `view`


jormungandr/src/network/p2p/layers/preferred_list.rs
```diff
@@ -0,0 +1,91 @@
...
+
+const DEFAULT_PREFERRED_VIEW_MAX: usize = 20;
+
+#[derive(Debug, Serialize, Deserialize, Default)]
+pub struct PreferredListConfig {
+    #[serde(default)]
+    view_max: PreferredViewMax,
+
+    #[serde(default)]
+    peers: Vec<Address>,
```
danielSanchezQ: I would use a `HashSet` here unless we want the order to matter. 
Duplicates make no sense, or do them?

jormungandr/src/network/p2p/layers/preferred_list.rs
```diff
@@ -0,0 +1,91 @@
...
+
+#[derive(Clone, Copy, Debug, PartialEq, Eq, PartialOrd, Ord, Hash, Serialize, Deserialize)]
+struct PreferredViewMax(usize);
+
+pub struct PreferredListLayer {
+    /// the max number of entries to add in the list of the view
+    view_max: usize,
+
+    /// the buddy list
+    peers: Vec<Address>,
```
danielSanchezQ: Likewise, `HashSet`

danielSanchezQ: Actually, here can be a `Vec` if it is gonna be static from the config.

jormungandr/src/network/p2p/layers/preferred_list.rs
NicolasDP: This is a good idea, we can't do that yet now, as we cannot use just the `Address` here, we need the `Id` too (we are pre-0.12 era of poldercast still). I have updated the comment to remind to use a `HashSet` when we update poldercast

doc/configuration/network.md
```diff
@@ -107,6 +108,46 @@ for now is that this is the first node any node tries to connect in order to mee
...
+**COMPATIBILITY NOTE**: in near future the peer list will be only a list of addresses and the **ID**
+part will not be necessary.
+
+##### Example:
+
+```yaml
+p2p:
+  layers:
+    preferred_list:
+      max_view: 20
```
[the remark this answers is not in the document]
NicolasDP: thanks <3

Issue #1979, opened 2020-03-27
[bottle in the sea]: the layer driver


NicolasDP:
The #1972 describe a new kind of client service. We will receive messages from the network, need to process some blobs data (can we decrypt them?) and then update the poldercast 4th layer (#1978).

## in the context of a passive node (no VRF)

* store received blobs's hash in a LRU Cache (see crate `lru`);
* if the blob is already in the LRU Cache, ignore;
* if not, query the `P2PTopology` for a `view` and broadcast the blob to them;

## in the contect of a stake pool

This module will need to perform the following actions:

### sending bottles

* wait for interval between 2 bottles to yield (**need configuration** option);
* from the `tip` of the blockchain, select the `stake pool` views and select `N` number of them (**need configuration** option);
* prepare `our bottle` and encrypt with shared_asymlock for these pools (input-output-hk/chain-libs#310);
* query the `P2PTopology` for a `view` and broadcast the blob to them;

### receiving bottles

* store the received blob's hash in a LRU Cache (see crate `lru`);
* if the blob is already in the LRU Cache, ignore;
* query the `enclave` to decrypt the blob: if blob is a message for us, send it to the `4th layer` (#1978);
* query the `P2PTopology` for a `view` and broadcast the blob to them;

--- 

### Compatibility considerations

it may be some nodes in our `view` do not support this new network message, do not `strike` them if we cannot send that message otherwise we will quarantine nodes that are perfectly valid.

### Configurations

We will need to be able to configure some parameters:

* send bottle interval: a `jormungandr_lib::time::Duration` provides already the UX to read it from both CLI or `serde` (config file): this is the interval between trying

PR #2001, opened 2020-04-01
Select stake pool ids for sea in the bottle

danielSanchezQ:
This is part of #1979


danielSanchezQ:
This is part of  #1979

PR #2009, opened 2020-04-03
[Tests] transaction propagation speed measurements

dkijania:


NicolasDP:
There is no relay in jormungandr

dkijania:
it's just an informal name for node role in network. Core - it's a node bootstrapped at start,  Relay - a node which is bootstrapped after core and it's used as a trusted peer at beginning , Leader - leaf node.

Issue #1997, opened 2020-03-31
Nodes Compatibility test 


dkijania:
Request for implementing test which verify compatibility between master version and N released older version. Should also be possible to test two arbitrary released version:

Tasks:

- [x] List all releases (https://api.github.com/repos/input-output-hk/jormungandr/releases)
- [x] Take N last releases (N ~ 10). The list is sorted.
- [x] Take windows or linux or mac version depending on underlying system
- [x] Test should have two options. Newest master or specific release
- [x] Add code which bootstrap node from given version
- [x] make some simple operations like transaction passing and block height is the same
- [x] check node are ok (no erros in logs) and transaction is accepted , nodes are in sync

PR #2010, opened 2020-04-03
[Tests] Jts legacy test

dkijania:
Private network tests for compatibility between node from master with node from last release


dkijania:
fixes #1997

PR #2018, opened 2020-04-06
Work to migrate REST to tokio 0.2

eugene-babichenko:


eugene-babichenko:
will also work on `src/rest/server/mod.rs`

PR #2021, opened 2020-04-06
Nightly build recreate fix

danielSanchezQ:
Not properly remove and creates a new release each time, with updated data.

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -14,8 +14,15 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@master
 
-      - name: Remove nightly tag and release
-        run: git push --delete origin nightly
+      - name: Install hub
+        uses: geertvdc/setup-hub@master
+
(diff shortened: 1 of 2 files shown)


danielSanchezQ:
Apparently there are some issues with the deps. Updating it right now.

.github/workflows/nightly.yml
```diff
@@ -12,6 +12,16 @@ jobs:
...
 
+      - name: Install hub
+        uses: geertvdc/setup-hub@master
+
+      - name: Delete nightly with hub
+        env:
+          GITHUB_USER: ${{ secrets.GITHUB_USER }}
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          hub release delete nightly
```
NicolasDP, line 23: what if nightly does not exist?

danielSanchezQ: It will fail but it shouldn't block the job to continue.

PR #2028, opened 2020-04-06
client: migration to tokio 0.2

eugene-babichenko:


jormungandr/src/client.rs
```diff
@@ -57,45 +56,46 @@ pub fn handle_input(
...
+                    );
+                }),
             );
         }
         ClientMsg::GetHeaders(ids, handle) => {
             info.timeout_spawn_failable_std(
                 "GetHeaders",
                 Duration::from_secs(PROCESS_TIMEOUT_GET_HEADERS),
-                handle_get_headers(task_data.clone(), ids, handle),
+                handle_get_headers(task_data.clone(), ids, handle.into_03()),
```
mzabaluev, line 80: This could be a good time to change the handle type to the 0.3 API in the messages, as well.

eugene-babichenko: Did it and also removed 0.2 API `ReplyStreamHandle`.

PR #2037, opened 2020-04-08
Watchdog logger POC

danielSanchezQ:


jormungandr-watchdog/src/service/mod.rs
```diff
@@ -145,8 +145,8 @@ impl<T: Service> ServiceState<T> {
     /// this from there one can "borrow" the settings or clone the reader
     ///
     /// [`SettingsReader`]: ./struct.SettingsReader.html
-    pub fn settings(&self) -> &SettingsReader<T::Settings> {
-        &self.settings
+    pub fn settings(&mut self) -> &mut SettingsReader<T::Settings> {
```
danielSanchezQ, line 148: I had to change this reference. Otherwise, it is impossible to use the `updated` method over the settings.

jormungandr-watchdog/examples/stdin_echo.rs
```diff
@@ -83,20 +90,109 @@ impl Service for StdoutWriter {
     }
 }
 
+#[derive(Debug, Clone, Serialize, Deserialize)]
```
danielSanchezQ, line 93: I made a wrapper so it was easy to implement `Serialize` and `Deserialize`. Couldn't find the **serde** implementation for the `tracing` values.

jormungandr-watchdog/examples/stdin_echo.rs
```diff
@@ -83,20 +90,109 @@ impl Service for StdoutWriter {
...
-
-    tracing::subscriber::set_global_default(subscriber).expect("setting tracing default failed");
+    // let subscriber = fmt::Subscriber::builder()
+    //     .with_env_filter(EnvFilter::from_default_env())
+    //     .finish();
+    //
+    // tracing::subscriber::set_global_default(subscriber).expect("setting tracing default failed");
 
     let app = App::new("stdin_echo");
     let watchdog = WatchdogBuilder::<StdEcho>::new(app).build();
```
danielSanchezQ, line 198: Right now it compiles but it does not run properly. Failing with `'called `Result::unwrap()` on an `Err` value: Message("missing field `stdin`", Some(Pos { marker: Marker { index: 7, line: 1, col: 7 }, path: "." }))'`
I'm checking it.

jormungandr-watchdog/examples/stdin_echo.rs
NicolasDP: have a look at : https://serde.rs/remote-derive.html

jormungandr-watchdog/examples/stdin_echo.rs
```diff
@@ -83,20 +90,109 @@ impl Service for StdoutWriter {
...
+
+impl Settings for LoggerConfig {
+    fn add_cli_args<'a, 'b>() -> Vec<Arg<'a, 'b>> {
+        vec![Arg::with_name("Log level")
+            .short("ll")
+            .long("log_level")
+            .takes_value(true)
+            .default_value("Warning")
+            .value_name("LOG_LEVEL")
+            .help("Services log level: []")]
```
NicolasDP, line 133: add the `possible_values`: https://docs.rs/clap/2.33.0/clap/struct.Arg.html#method.possible_values

jormungandr-watchdog/examples/stdin_echo.rs
```diff
@@ -83,20 +90,109 @@ impl Service for StdoutWriter {
...
+    };
+}
+
+impl Settings for LoggerConfig {
+    fn add_cli_args<'a, 'b>() -> Vec<Arg<'a, 'b>> {
+        vec![Arg::with_name("Log level")
+            .short("ll")
+            .long("log_level")
+            .takes_value(true)
+            .default_value("Warning")
```
NicolasDP, line 131: default is lowercase please

jormungandr-watchdog/examples/stdin_echo.rs
```diff
@@ -83,20 +90,109 @@ impl Service for StdoutWriter {
...
+        LogLevel::ERROR => tracing::Level::ERROR,
+        LogLevel::TRACE => tracing::Level::TRACE,
+    };
+}
+
+impl Settings for LoggerConfig {
+    fn add_cli_args<'a, 'b>() -> Vec<Arg<'a, 'b>> {
+        vec![Arg::with_name("Log level")
+            .short("ll")
+            .long("log_level")
```
NicolasDP, line 129:
```suggestion
            .long("log-level")
```

jormungandr-watchdog/examples/stdin_echo.rs
```diff
@@ -83,20 +90,109 @@ impl Service for StdoutWriter {
...
+        LogLevel::DEBUG => tracing::Level::DEBUG,
+        LogLevel::ERROR => tracing::Level::ERROR,
+        LogLevel::TRACE => tracing::Level::TRACE,
+    };
+}
+
+impl Settings for LoggerConfig {
+    fn add_cli_args<'a, 'b>() -> Vec<Arg<'a, 'b>> {
+        vec![Arg::with_name("Log level")
+            .short("ll")
```
NicolasDP, line 128: no short options

jormungandr-watchdog/examples/stdin_echo.rs
```diff
@@ -83,20 +90,109 @@ impl Service for StdoutWriter {
...
+        LogLevel::WARN => tracing::Level::WARN,
+        LogLevel::DEBUG => tracing::Level::DEBUG,
+        LogLevel::ERROR => tracing::Level::ERROR,
+        LogLevel::TRACE => tracing::Level::TRACE,
+    };
+}
+
+impl Settings for LoggerConfig {
+    fn add_cli_args<'a, 'b>() -> Vec<Arg<'a, 'b>> {
+        vec![Arg::with_name("Log level")
```
NicolasDP, line 127: define a `const` and use it for the `match()` call too

```rust
const LOGGER_CONFIG_LOG_LEVEL: &str = "logger config log level";

Arg::with_name(LOGGER_CONFIG_LOG_LEVEL);
matches.value_of(LOGGER_CONFIG_LOG_LEVEL);
```

jormungandr-watchdog/examples/stdin_echo.rs
NicolasDP: also allow for `env` : https://docs.rs/clap/2.33.0/clap/struct.Arg.html#method.env

jormungandr-watchdog/examples/stdin_echo.rs
```diff
@@ -83,20 +90,109 @@ impl Service for StdoutWriter {
...
+
+    fn matches_cli_args<'a>(&mut self, matches: &ArgMatches<'a>) {
+        if let Some(level) = matches.value_of("cfg") {
+            self.level = match level.to_lowercase().as_str() {
+                "info" => LogLevel::INFO,
+                "warn" => LogLevel::WARN,
+                "debug" => LogLevel::DEBUG,
+                "error" => LogLevel::ERROR,
+                "trace" => LogLevel::TRACE,
+                _ => self.level.clone(),
```
NicolasDP, line 144: This line should not return a default level. If you use clap's `possible_values` you can expect that if this parameter is used then it cannot have invalid values in it.

```suggestion
                _ => unreachable!(),
```

jormungandr-watchdog/src/service/mod.rs
NicolasDP: `SettingsReader` is clonable: https://github.com/input-output-hk/jormungandr/blob/40c120957a00e042060df90071303e024c1da731/jormungandr-watchdog/src/service/settings.rs#L90-L94

jormungandr-watchdog/Cargo.toml
```diff
@@ -16,6 +16,7 @@ Process and service management and API
 jormungandr-watchdog-derive = { path = "../jormungandr-watchdog-derive" }
 tokio = { version = "^0.2.12", features = ["sync", "rt-threaded", "parking_lot", "time"] }
 tracing = "^0.1.13"
+tracing-subscriber = "0.2.4"
```
NicolasDP, line 19: This is a `dev-dependencies`

jormungandr-watchdog/examples/stdin_echo.rs
danielSanchezQ: Aha, I was looking for something like this but I couldn't find it. Thanks!

jormungandr-watchdog/src/service/mod.rs
```diff
@@ -145,7 +145,7 @@ impl<T: Service> ServiceState<T> {
     /// this from there one can "borrow" the settings or clone the reader
     ///
     /// [`SettingsReader`]: ./struct.SettingsReader.html
-    pub fn settings(&self) -> &SettingsReader<T::Settings> {
+    pub fn settings(&mut self) -> &SettingsReader<T::Settings> {
```
NicolasDP, line 148: do we still need the mut here?

danielSanchezQ: Indeed we do not. Missing change from bringing it back to original form.

jormungandr-watchdog/examples/stdin_echo.rs
NicolasDP: still failing here?

danielSanchezQ: Yes, we should open a ticket to check this. I believe is a missing file, although not sure why would it not drop to the default settings anyway.

PR #2072, opened 2020-04-15
Net tokio 0.2 fixes

eugene-babichenko:


jormungandr/src/client.rs
```diff
@@ -19,6 +19,7 @@ const PROCESS_TIMEOUT_GET_HEADERS_RANGE: u64 = 5 * 60;
 const PROCESS_TIMEOUT_GET_BLOCKS: u64 = 10 * 60;
 const PROCESS_TIMEOUT_PULL_BLOCKS_TO_TIP: u64 = 60 * 60;
 
+#[derive(Clone)]
```
mzabaluev, line 22: I'm trying to remove cloning there by making the peers stuff internal to the network task so that only one instance of `P2pTopology` is needed. I should push some more commits to the base branch, at which point this should be rebased and updated.

eugene-babichenko: Got it. Please ping me when it's ready.

mzabaluev: Pushed now, please rebase and remove the derives here and on `P2pTopology` again.

PR #2076, opened 2020-04-16
Track retirement in explorer stake pool data

### LINKED ISSUE #2074: Unregistered pools are still listed in the explorer
Unregistered pools are still visible on the explorer, it seems the certificate is not visible

danielSanchezQ:
Candidate to fix #2074

### DIFF
--- a/jormungandr/src/explorer/graphql/mod.rs
+++ b/jormungandr/src/explorer/graphql/mod.rs
@@ -927,27 +927,16 @@ impl Pool {
         }
     }
 
-    pub fn retirement(&self, context: &Context) -> FieldResult<PoolRetirement> {
+    pub fn retirement(&self, context: &Context) -> FieldResult<Option<PoolRetirement>> {
         match &self.data {
-            Some(data) => data
-                .retirement
-                .clone()
-                .map(|retirement| PoolRetirement::from(retirement))
-                .ok_or(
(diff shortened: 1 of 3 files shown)


jormungandr/src/explorer/graphql/mod.rs
```diff
@@ -926,6 +926,30 @@ impl Pool {
...
     }
+
+    pub fn retirement(&self, context: &Context) -> FieldResult<PoolRetirement> {
+        match &self.data {
+            Some(data) => data
+                .retirement
+                .clone()
+                .map(|retirement| PoolRetirement::from(retirement))
+                .ok_or(
+                    ErrorKind::NotFound("Stake pool has no retirement data associated".to_owned())
```
enzoc4, line 937: I think it's better to return `None` here (null in graphql). And change the return type to `FieldResult<Option<PoolRetirement>>`.
Currently `Error::NotFound` is being used mostly for search-by-id queries.

danielSanchezQ: Makes sense, will do.

danielSanchezQ: Makes sense. I updated it. Thanks

NicolasDP:
Can you review @enzoc4 ?

PR #2077, opened 2020-04-16
doc - api v0, cleaning

rinor:


NicolasDP:
Good work, will this be open for review soon?

rinor:
Finishing up some checks while testing the new warp changes. Will push some more fixes that were found while crosschecking the api definitions with the real responses.

Issue #2033, opened 2020-04-07
remove the use of the mutex


NicolasDP:
https://github.com/input-output-hk/jormungandr/blob/3ed43d274cfd680e7cb8eb05a11afdb282900c2a/jormungandr-watchdog/src/service/intercom.rs#L35

PR #2079, opened 2020-04-16
Atomic operations to Watchdog Stats

danielSanchezQ:
Implements #2033

### DIFF
--- a/jormungandr-watchdog/src/service/stats.rs
+++ b/jormungandr-watchdog/src/service/stats.rs
@@ -1,53 +1,62 @@
+use atomig::Atomic;
+use std::sync::atomic::{AtomicU64, Ordering};
+
 /// keep some stats based on [Welford's online algorithm]
 ///
 /// [Welford's online algorithm]: https://en.wikipedia.org/wiki/Algorithms_for_calculating_variance#Welford's_online_algorithm
 pub struct Stats {
-    count: u64,
-    mean: f64,
-    m2: f64,
+    count: AtomicU64,
+    mean: Atomic<f64>,
+    m2: Atomic<f64>,
 }
 
 impl Stats {
     pub fn new() -> Self {
         Self {
-            count: 0,
-            mean: 0.0,
-            m2: 0.0,
+            count: AtomicU64::new(0),
+            mean: Atomic::new(0.0),
+            m2: Atomic::new(0.0),
         }
     }
 
-    pub fn push(&mut self, entry: f64) {
-        self.count += 1;
-
-        if self.count == 1 {
-            self.mean = entry;
-            self.m2 = 0.0;
+    pub fn push(&self, entry: f64) {
+        self.count.fetch_add(1, Ordering::Relaxed);
+        let self_count = self.count.load(Ordering::Relaxed);
+        if self_count == 1 {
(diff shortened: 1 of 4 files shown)


jormungandr-watchdog/src/service/stats.rs
```diff
@@ -1,53 +1,62 @@
...
 /// keep some stats based on [Welford's online algorithm]
 ///
 /// [Welford's online algorithm]: https://en.wikipedia.org/wiki/Algorithms_for_calculating_variance#Welford's_online_algorithm
 pub struct Stats {
-    count: u64,
-    mean: f64,
-    m2: f64,
+    count: AtomicU64,
+    mean: Atomic<f64>,
+    m2: Atomic<f64>,
```
NicolasDP, line 10: No need to add an extra deps for this. All needed is [f64::to_bits](https://doc.rust-lang.org/stable/std/primitive.f64.html#method.to_bits) and [f64::from_bits](https://doc.rust-lang.org/stable/std/primitive.f64.html#method.from_bits) and you store the value in an `AtomicU64`.

jormungandr-watchdog/src/service/stats.rs
```diff
@@ -1,53 +1,62 @@
...
 
-    pub fn push(&mut self, entry: f64) {
-        self.count += 1;
-
-        if self.count == 1 {
-            self.mean = entry;
-            self.m2 = 0.0;
+    pub fn push(&self, entry: f64) {
+        self.count.fetch_add(1, Ordering::Relaxed);
+        let self_count = self.count.load(Ordering::Relaxed);
```
NicolasDP, line 24:
```suggestion
        let self_count = self.count.fetch_add(1, Ordering::Relaxed) + 1;
```

jormungandr-watchdog/src/service/stats.rs
```diff
@@ -1,53 +1,62 @@
...
+            let delta = entry - self.mean.load(Ordering::Relaxed);
+            self.mean.store(
+                self.mean.load(Ordering::Acquire) + delta / self_count as f64,
+                Ordering::Release,
+            );
+            let delta2 = entry - self.mean.load(Ordering::Relaxed);
+            self.m2.store(
+                self.m2.load(Ordering::Acquire) + delta * delta2,
+                Ordering::Release,
+            )
```
NicolasDP, line 38: too many breaking point... the order of the operation mater. Don't use atomic. use a `std::sync::Mutex` instead here

NicolasDP:
my bad,  I told you to look at the atomic in the issue, actually looking at the code, we should use a std::sync::Mutex instead of a tokio mutex here:

1. there is no need for asynchronicity here;
2. the operation underlying is really fast anyway

https://github.com/input-output-hk/jormungandr/blob/3ed43d274cfd680e7cb8eb05a11afdb282900c2a/jormungandr-watchdog/src/service/intercom.rs#L35 this mutex is a tokio mutex, use a std::sync::Mutex

danielSanchezQ:
Let's close this branch and I'll make the replacement in another PR.

danielSanchezQ:
Moved [here](https://github.com/input-output-hk/jormungandr/pull/2082)

PR #2082, opened 2020-04-17
Change mutex to std instead of tokio in watchdog intercom

danielSanchezQ:


jormungandr-watchdog/src/service/intercom.rs
```diff
@@ -224,7 +224,7 @@ impl<T> IntercomReceiver<T> {
 
 impl IntercomStats {
     pub async fn status(&self) -> IntercomStatus {
-        let stats = self.stats.lock().await;
+        let stats = self.stats.lock().unwrap();
```
NicolasDP, line 227: I suggest you clone it so it does not hold the lock. Right now the other operations are atomic too but it may be there are are more costly operations later. Clone or copy it so we don't hold the lock here in the future.

```suggestion
        let stats = self.stats.lock().unwrap().clone();
```

PR #2083, opened 2020-04-17
REST API: replace actix with warp

eugene-babichenko:


jormungandr/src/main.rs
```diff
@@ -1,6 +1,7 @@
 // Rustc default type_length_limit is too low for complex futures, which generate deeply nested
 // monomorphized structured with long signatures. This value is enough for current project.
-#![type_length_limit = "10000000"]
+// TODO change this back to 10000000 when rust 1.43 is out
```
NicolasDP, line 3: create an issue for this so we can track it down

eugene-babichenko: Also note that the build time has increased 5 times on stable rustc, while beta manages to build in 2 minutes as before.

eugene-babichenko: Done https://github.com/input-output-hk/jormungandr/issues/2094

PR #2091, opened 2020-04-20
Net tokio 0.2 network mod type errors

danielSanchezQ:


danielSanchezQ:
I don't know what happened to commit history. It is probably is better to squash this.

jormungandr/src/network/mod.rs
```diff
@@ -309,7 +322,11 @@ async fn handle_network_input(
             } => {
                 state
                     .peers
-                    .pull_headers(node_address, from.into(), to)
+                    .pull_headers(
+                        node_address,
+                        try_ids_from_iter(from.as_slice()).unwrap(),
+                        BlockId::try_from_network(to).unwrap(),
```
danielSanchezQ, line 328: Is this `unwrap` ok? or should we skip is something goes wrong? I wasn't sure of it.

jormungandr/src/network/convert.rs
```diff
@@ -34,3 +37,35 @@ impl TryFromNetwork<net_data::Fragment> for Fragment {
...
+}
+
+impl TryFromNetwork<Vec<HeaderHash>> for BlockIds {
+    fn try_from_network(block_ids: Vec<HeaderHash>) -> Result<Self, Error> {
+        try_ids_from_iter(block_ids.iter())
+    }
+}
+
+// TODO: Check if this is completly compatible
+impl TryFromNetwork<chain_impl_mockchain::header::Header> for chain_network::data::block::Header {
```
mzabaluev, line 55: `Header` and `Fragment` of `chain_impl_mockchain` are actually imported above via `blockcfg`.

jormungandr/src/network/mod.rs
mzabaluev: Yes, conversion to the network types from internal types should always be correct.

jormungandr/src/network/mod.rs
```diff
@@ -77,7 +77,10 @@ use std::sync::Arc;
 use std::time::Duration;
 
 pub use self::bootstrap::Error as BootstrapError;
+use crate::network::convert::TryFromNetwork;
 use crate::stats_counter::StatsCounter;
+use chain_network::data::block::try_ids_from_iter;
```
mzabaluev, line 82: There will be also `fragment::try_ids_from_iter`, so it's better to import the `block` module and disambiguate with the module name.

jormungandr/src/network/mod.rs
```diff
@@ -331,7 +348,13 @@ async fn handle_propagation_msg(msg: PropagateMsg, state: GlobalStateR, channels
                     topic: p2p::topic::BLOCKS,
                 })
                 .await;
-            prop_state.peers.propagate_block(view.peers, header).await
+            prop_state
+                .peers
+                .propagate_block(
+                    view.peers,
+                    chain_network::data::block::Header::try_from_network(header).unwrap(),
```
mzabaluev, line 355: This looks unwieldy too, import `Header` and `Fragment` from `blockcfg`.

jormungandr/src/network/convert.rs
```diff
@@ -1,7 +1,8 @@
-use crate::blockcfg::{Fragment, Header, HeaderId};
+use crate::blockcfg::{Fragment, Header, HeaderHash, HeaderId};
```
mzabaluev, line 1: `HeaderHash` and `HeaderId` are the same, but I'll fix it later.

jormungandr/src/network/convert.rs
```diff
@@ -34,3 +35,29 @@ impl TryFromNetwork<net_data::Fragment> for Fragment {
         read(&data)
     }
 }
+
+// TODO: Check if this is completly compatible
+impl TryFromNetwork<HeaderHash> for BlockId {
```
mzabaluev, line 40: Oh, it should be the other way around: `BlockId` is a network type.

jormungandr/src/network/convert.rs
```diff
@@ -34,3 +35,29 @@ impl TryFromNetwork<net_data::Fragment> for Fragment {
...
+    fn try_from_network(header_hash: HeaderHash) -> Result<Self, Error> {
+        BlockId::try_from(header_hash.serialize_as_vec().unwrap().as_slice())
+    }
+}
+
+impl TryFromNetwork<Vec<HeaderHash>> for BlockIds {
```
mzabaluev, line 46: Same here, there should be `IntoNetwork<BlockIds> for Vec<HeaderHash>` instead.

jormungandr/src/network/mod.rs
mzabaluev: ... and this is why the `IntoNetwork` trait should be used here instead.

danielSanchezQ:
Cleaned and implemented [here](https://github.com/input-output-hk/jormungandr/pull/2093)

PR #2104, opened 2020-04-22
fix borrow checker errors

eugene-babichenko:


jormungandr/src/rest/v0/logic.rs
```diff
@@ -274,8 +274,8 @@ pub async fn get_stake_distribution(
             unassigned: distribution.unassigned.into(),
             pools: distribution
                 .to_pools
-                .into_iter()
-                .map(|(key, value)| (key.into(), value.stake.total.into()))
+                .iter()
+                .map(|(key, value)| (key.clone().into(), value.stake.total.clone().into()))
```
mzabaluev, line 278: This is small peanuts, but I wonder what caused this to fail.

PR #2111, opened 2020-04-22
Net tolkio 0.2 preferred list

danielSanchezQ:


jormungandr/src/network/p2p/layers/preferred_list.rs
```diff
@@ -61,13 +64,18 @@ impl Layer for PreferredListLayer {
...
-            let info = poldercast::NodeInfo::new(selected.id.clone(), selected.address.clone());
-
-            view.add_info(info);
+    fn view(&mut self, view: &mut ViewBuilder, all_nodes: &mut Nodes) {
+        let selected: HashSet<Address> = self
+            .peers
+            .iter()
+            .choose_multiple(&mut self.prng, self.view_max)
+            .iter()
+            .map(|x| *x.clone())
```
mzabaluev, line 73: I don't understand this: result of `choose_multiple` is an iterator already? Also use `.cloned()`?

danielSanchezQ: The problem is that `choose_multiple` returns a `Vec<&Address>` and with second iter y have to dereference it in order o clone. If using `.cloned()` I just got the references back.

danielSanchezQ: Double cloning does it. But, I'm not sure if we could use just the `&Address` for this. Does the `HahsMap` hashes the `&` or the actual `Address`?

mzabaluev: I think it's possible to have a temporary `HashMap`/`HashSet` with references as values as long as the collection doesn't outlive the borrowed values.

mzabaluev: Removed the cloning in a follow up comment; `HashSet<&Address>` works well here.

PR #2112, opened 2020-04-22
[Tests] Start Node with TLS certificate

dkijania:


jormungandr-integration-tests/src/jormungandr/tls.rs
```diff
@@ -0,0 +1,55 @@
...
+    file_assert::assert_file_exists_and_not_empty(&cert_file);
+
+    let config = ConfigurationBuilder::new()
+        .with_rest_tls_config(Tls {
+            cert_file: cert_file.as_os_str().to_str().unwrap().to_owned(),
+            priv_key_file: pk8_key_file.as_os_str().to_str().unwrap().to_owned(),
+        })
+        .build();
+
+    let jormungandr = Starter::new().config(config).start().unwrap();
```
eugene-babichenko, line 50: I think the problem is this call actually. As I understand, under the hood it tries to verify if the REST API node stats are available, but they aren't because of the long bootstrap phase. Can we just kill the node before the bootstrap is over?

PR #2121, opened 2020-04-23
Net tokio 0.2 graphql

danielSanchezQ:


jormungandr/src/explorer/graphql/mod.rs
```diff
@@ -1046,19 +1047,23 @@ impl Epoch {
...
+        let epoch_lower_bound = block_on(async {
             context
                 .db
                 .get_block(&epoch_data.first_block)
-                .map(|block| u32::from(block.expect("The block to be indexed").chain_length)),
-        );
+                .await
+                .map(|block| u32::from(block.chain_length))
+        })
+        .unwrap();
```
mzabaluev, line 1057: It's useful to retain that panic message from `expect` here instead of uninformative `unwrap`.

jormungandr/src/explorer/graphql/mod.rs
```diff
@@ -1046,19 +1047,23 @@ impl Epoch {
...
+        let epoch_upper_bound = block_on(async {
             context
                 .db
                 .get_block(&epoch_data.last_block)
-                .map(|block| u32::from(block.expect("The block to be indexed").chain_length)),
-        );
+                .await
+                .map(|block| u32::from(block.chain_length))
+        })
+        .unwrap();
```
mzabaluev, line 1066: Same as above.

jormungandr/src/rest/v0/logic.rs
```diff
@@ -12,7 +12,7 @@ use crate::{
     rest::Context,
     secure::NodeSecret,
 };
-use chain_core::property::{Block as _, Deserialize, Serialize, FromStr};
+use chain_core::property::{Block as _, Deserialize, FromStr, Serialize};
```
mzabaluev, line 15: I have already applied this in a rebase, to be pushed soon.

jormungandr/src/explorer/graphql/mod.rs
danielSanchezQ: yep, actually I updated it but the push was rejected....

PR #2122, opened 2020-04-23
Rest - api, /committees - added

rinor:
- [x] node implement /api/v0/committees
- [x] api specification
- [x] expose to jcli
- [x] docs


NicolasDP:
nice, this is going the right direction

PR #2129, opened 2020-04-24
Peer map remove unused id

danielSanchezQ:


jormungandr/src/network/p2p/comm/peer_map.rs
```diff
@@ -150,7 +150,6 @@ impl PeerMap {
         self.map
             .iter()
             .map(|(&id, data)| PeerInfo {
-                id,
```
mzabaluev, line 153: Notice that we also get the address in `id`. Maybe we should remove the address member from `PeerData`.

PR #2135, opened 2020-04-24
Expose VotePlans in rest service

danielSanchezQ:
It needs this chain-libs PR first input-output-hk/chain-libs#352

### DIFF
--- a/jormungandr-lib/src/interfaces/vote/mod.rs
+++ b/jormungandr-lib/src/interfaces/vote/mod.rs
@@ -1,4 +1,5 @@
 use serde::Serialize;
+
 #[derive(Serialize)]
 #[serde(remote = "chain_impl_mockchain::block::BlockDate")]
 struct BlockDate {
@@ -14,18 +15,8 @@ struct ExternalProposalId {
 }
 
 #[derive(Serialize)]
-#[serde(remote = "chain_impl_mockchain::certificate::VoteOptions")]
-struct VoteOptions {
-    #[serde(getter = "chain_impl_mockchain::certificate::VoteOptions::as_byte")]
-    num_choices: u8,
-}
-
-#[derive(Serialize)]
-// #[serde(remote="chain_impl_mockchain::certificate::Proposal")]
 pub struct Proposal {
-    // #[serde(with="ExternalProposalId", getter="chain_impl_mockchain::certificate::Proposal::external_id")]
     pub external_id: String,
-    // #[serde(with="VoteOptions", getter="chain_impl_mockchain::certificate::Proposal::options")]
     pub options: u8,
 }
 
--- a/jormungandr-lib/src/interfaces/vote/mod.rs
+++ b/jormungandr-lib/src/interfaces/vote/mod.rs
@@ -1,69 +1,89 @@
-use serde::Serialize;
+use crate::crypto::hash::Hash;
+use crate::interfaces::blockdate::BlockDateDef;
+use chain_impl_mockchain::certificate::Proposals;
+use serde::ser::SerializeSeq;
+use serde::{Serialize, Serializer};
 
 #[derive(Serialize)]
-#[serde(remote = "chain_impl_mockchain::block::BlockDate")]
-struct BlockDate {
-    pub epoch: u32,
-    pub slot_id: u32,
+pub enum VoteOptions {
+    OneOf { max_value: u8 }, // where max_value is up to 15
+}
+
+fn get_proposal_hash(proposal: &chain_impl_mockchain::certificate::Proposal) -> Hash {
+    Hash::from(proposal.external_id().clone())
+}
+
+fn get_proposal_vote_options(
+    proposal: &chain_impl_mockchain::certificate::Proposal,
+) -> VoteOptions {
+    VoteOptions::OneOf {
+        max_value: proposal.options().as_byte(),
+    }
 }
 
 #[derive(Serialize)]
-#[serde(remote = "chain_impl_mockchain::certificate::ExternalProposalId")]
-struct ExternalProposalId {
-    #[serde(getter = "chain_impl_mockchain::certificate::ExternalProposalId::to_string")]
-    id: String,
-}
-
-#[derive(Serialize)]
+#[serde(remote = "chain_impl_mockchain::certificate::Proposal")]
 pub struct Proposal {
-    pub external_id: String,
-    pub options: u8,
-}
-
-impl Proposal {
-    fn new(p: &chain_impl_mockchain::certificate::Proposal) -> Self {
-        Self {
-            external_id: p.external_id().to_string(),
-            options: p.options().as_byte(),
-        }
-    }
+    #[serde(getter = "get_proposal_hash")]
+    pub external_id: Hash,
+    #[serde(getter = "get_proposal_vote_options")]
+    pub options: VoteOptions,
 }
 
 #[derive(Serialize)]
+pub struct ProposalSerializableHelper<'a>(
+    #[serde(with = "Proposal")] pub &'a chain_impl_mockchain::certificate::Proposal,
+);
+
+fn serialize_proposals<S>(
+    proposals: &chain_impl_mockchain::certificate::Proposals,
+    serializer: S,
+) -> Result<S::Ok, S::Error>
+where
+    S: Serializer,
+{
+    let mut seq = serializer.serialize_seq(Some(proposals.len()))?;
+    for e in proposals.iter() {
+        seq.serialize_element(&ProposalSerializableHelper(e))?;
+    }
+    seq.end()
+}
+
+#[derive(Serialize)]
+#[serde(remote = "chain_impl_mockchain::certificate::VotePlan")]
 pub struct VotePlan {
-    /// the vote start validity
-    #[serde(with = "BlockDate")]
+    #[serde(
+        with = "BlockDateDef",
+        getter = "chain_impl_mockchain::certificate::VotePlan::vote_start"
+    )]
     pub vote_start: chain_impl_mockchain::block::BlockDate,
-    /// the duration within which it is possible to vote for one of the proposals
-    /// of this voting plan.
-    #[serde(with = "BlockDate")]
-    pub vote_end: chain_impl_mockchain::block::BlockDate,
-    /// the committee duration is the time allocated to the committee to open
-    /// the ballots and publish the results on chain
-    #[serde(with = "BlockDate")]
-    pub committee_end: chain_impl_mockchain::block::BlockDate,
-    /// the proposals to vote for
-    pub proposals: Vec<Proposal>,
-}
 
-impl VotePlan {
-    pub fn from_vote_plan(plan: &chain_impl_mockchain::certificate::VotePlan) -> Self {
-        VotePlan {
-            vote_start: plan.vote_start(),
-            vote_end: plan.vote_end(),
-            committee_end: plan.committee_end(),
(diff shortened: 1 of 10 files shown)


danielSanchezQ:
It needs this chain-libs PR first https://github.com/input-output-hk/chain-libs/pull/352

jormungandr/src/rest/v0/handlers.rs
```diff
@@ -274,3 +274,95 @@ pub async fn get_committees(context: ContextLock) -> Result<impl Reply, Rejectio
...
+pub struct VotePlans {
+    plans: Vec<VotePlan>,
+}
+
+impl VotePlans {
+    fn new(plans: Vec<VotePlan>) -> Self {
+        Self { plans }
+    }
+}
+
```
danielSanchezQ, line 355: I don't know where to put these types. I left them as close as possible from where they are used.

jormungandr/src/rest/v0/logic.rs
```diff
@@ -658,3 +659,12 @@ pub async fn get_committees(context: &Context) -> Result<Vec<String>, Error> {
         .map(|cid| cid.to_string())
         .collect())
 }
+
+pub async fn get_active_vote_plans(context: &Context) -> Result<Vec<VotePlan>, Error> {
```
danielSanchezQ, line 663: I'm taking all of them right now. But is it correct though? Which ones do we consider the active ones?

jormungandr/src/rest/v0/handlers.rs
rinor: Probably at jormungandr-lib

jormungandr/src/rest/v0/handlers.rs
```diff
@@ -274,3 +274,95 @@ pub async fn get_committees(context: ContextLock) -> Result<impl Reply, Rejectio
...
+pub async fn get_active_vote_plans(context: ContextLock) -> Result<impl Reply, Rejection> {
+    let context = context.read().await;
+    let vote_plans = logic::get_active_vote_plans(&context).await;
+    vote_plans
+        .map(|v| {
+            v.iter()
+                .map(|p| VotePlan::from_vote_plan(p))
+                .collect::<Vec<VotePlan>>()
+        })
+        .map(VotePlans::new)
```
rinor, line 365: this mey be better pushed on the logic part, sind would like to keep the handlers as small/clean/boring as possible now that @eugene-babichenko separated the logic.

danielSanchezQ: Well, the thing is that this is completely from the outer layer in my opinion. All the types and that transformation is so it can be JSONize. I guess logic should still work with the core types and should know anything about what the output should be.

jormungandr/src/blockchain/reference.rs
```diff
@@ -159,4 +160,8 @@ impl Ref {
     pub fn elapsed(&self) -> Result<Duration, std::time::SystemTimeError> {
         SystemTime::now().duration_since(self.time())
     }
+
+    pub fn active_vote_plans(&self) -> Vec<VotePlan> {
+        self.ledger.state().active_vote_plans()
+    }
```
NicolasDP, line 166: a little comment for this function maybe?

```
/// clone all active vote plans at this given state
///
/// this includes, votes to be voted on, on going votes, votes to be resolved and votes
/// to result into a change on the ledger
```

jormungandr/src/rest/v0/handlers.rs
NicolasDP: all of this to go in `jormungandr-lib/src/interface/vote/*.rs`

jormungandr/src/rest/v0/handlers.rs
NicolasDP: should go on the logic part. The types are not on the right place anyway, they should go to `jormungandr-lib`

jormungandr/src/rest/v0/logic.rs
NicolasDP: all the one returned by the ledger() `active_vote_plans`

jormungandr/src/rest/v0/logic.rs
```diff
@@ -658,3 +659,12 @@ pub async fn get_committees(context: &Context) -> Result<Vec<String>, Error> {
...
+    Ok(context
+        .blockchain_tip()?
+        .get_ref()
+        .await
+        .ledger()
+        .active_vote_plans())
```
NicolasDP, line 669: there is a function `active_vote_plans` from the `Ref`, why using the indirection from the `ledger()`?

jormungandr/src/rest/v0/mod.rs
```diff
@@ -225,6 +225,12 @@ pub fn filter(
         .and_then(handlers::get_committees)
         .boxed();
 
+    let vote_plans = warp::path!("active_vote_plans")
+        .and(warp::get())
+        .and(with_context.clone())
+        .and_then(handlers::get_active_vote_plans)
+        .boxed();
```
NicolasDP, line 232: let's coalesce the `committee` end point in `vote/active/{committee|plans}`.

jormungandr/src/rest/v0/logic.rs
danielSanchezQ: Leftover from a prev version. Thanks for pointing out.

danielSanchezQ:
Maybe something went wrong on some rebase. Can any of you check the `chain-deps` changes here please? @input-output-hk/jormungandr

jormungandr-lib/src/interfaces/vote/mod.rs
```diff
@@ -0,0 +1,69 @@
...
+#[derive(Serialize)]
+#[serde(remote = "chain_impl_mockchain::certificate::ExternalProposalId")]
+struct ExternalProposalId {
+    #[serde(getter = "chain_impl_mockchain::certificate::ExternalProposalId::to_string")]
+    id: String,
+}
+
+#[derive(Serialize)]
+pub struct Proposal {
+    pub external_id: String,
```
NicolasDP, line 19: This is a `Hash` I believe. We already got a type for this

jormungandr-lib/src/interfaces/vote/mod.rs
```diff
@@ -0,0 +1,69 @@
...
+    pub options: u8,
+}
```
NicolasDP, line 21: what prevents us from implementing the Remote here?

jormungandr-lib/src/interfaces/vote/mod.rs
```diff
@@ -0,0 +1,69 @@
...
+struct BlockDate {
+    pub epoch: u32,
+    pub slot_id: u32,
+}
+
+#[derive(Serialize)]
+#[serde(remote = "chain_impl_mockchain::certificate::ExternalProposalId")]
+struct ExternalProposalId {
+    #[serde(getter = "chain_impl_mockchain::certificate::ExternalProposalId::to_string")]
+    id: String,
```
NicolasDP, line 14: There is a type for this, it is a `Hash`

jormungandr-lib/src/interfaces/vote/mod.rs
```diff
@@ -0,0 +1,69 @@
+use serde::Serialize;
+
+#[derive(Serialize)]
+#[serde(remote = "chain_impl_mockchain::block::BlockDate")]
+struct BlockDate {
+    pub epoch: u32,
+    pub slot_id: u32,
+}
```
NicolasDP, line 8: There is a `BlockDate` type here already. Though, this one is better in the sense it uses `serde(remote)`. Add a `BlockDateDef` in this module which is serialize/deserialize for serde.

https://github.com/input-output-hk/jormungandr/blob/7aef084df61a272b56c189322cb37c84aca70b38/jormungandr-lib/src/interfaces/blockdate.rs#L5-L6

jormungandr/src/rest/v0/mod.rs
```diff
@@ -219,12 +219,18 @@ pub fn filter(
...
         .and(with_context.clone())
         .and_then(handlers::get_committees)
         .boxed();
 
+    let vote_plans = warp::path!("vote" / "active" / "plans")
+        .and(warp::get())
+        .and(with_context.clone())
+        .and_then(handlers::get_active_vote_plans)
+        .boxed();
+
```
NicolasDP, line 233: Why not using `warp` grouping mechanism ? as done here https://github.com/input-output-hk/jormungandr/blob/7aef084df61a272b56c189322cb37c84aca70b38/jormungandr/src/rest/v0/mod.rs#L192-L208

jormungandr-lib/src/interfaces/vote/mod.rs
danielSanchezQ: I couldn't chain the base method `external_id` and `to_string` for external_id and the `options` with `as_byte` for options. I'm gonna try making an specific funtion to mine that data.

jormungandr-lib/src/interfaces/vote/mod.rs
danielSanchezQ: But `Hash` doesn't implement `serde::Serialize`, so I wanted to keep it simple instead of making more types.

jormungandr-lib/src/interfaces/vote/mod.rs
danielSanchezQ: That one in that module implements serialize, deserialize already. Gonna use that.

jormungandr-lib/src/interfaces/vote/mod.rs
danielSanchezQ: Ok, was looking at other `Hash`. I found the one I should use. Thanks!

jormungandr-lib/src/interfaces/vote/mod.rs
```diff
@@ -0,0 +1,69 @@
...
+#[serde(remote = "chain_impl_mockchain::certificate::ExternalProposalId")]
+struct ExternalProposalId {
+    #[serde(getter = "chain_impl_mockchain::certificate::ExternalProposalId::to_string")]
+    id: String,
+}
+
+#[derive(Serialize)]
+pub struct Proposal {
+    pub external_id: String,
+    pub options: u8,
```
NicolasDP, line 20: This is a bit more complex than just a `u8`. Can you make it an enum instead? So this can grow with time?

Right now, this byte represent a `bitmask` of the vote choices available, and only the first 4 bits are allowed. So if the option is `0b0011` the options are `0`, `1` or `2`. If the option is `0b0100`, the options are `0`, `1`, `2`, `3`.

NicolasDP: Could you make it an enum with only one entry?

```rust
enum VoteOption {
    OneOf { max_value: u8 } // where max_value is up to 15
}
```

NicolasDP: This will allow us to extend the model in the future without breaking APIs

jormungandr-lib/src/interfaces/vote/mod.rs
```diff
@@ -0,0 +1,89 @@
...
+) -> Result<S::Ok, S::Error>
+where
+    S: Serializer,
+{
+    let mut seq = serializer.serialize_seq(Some(proposals.len()))?;
+    for e in proposals.iter() {
+        seq.serialize_element(&ProposalSerializableHelper(e))?;
+    }
+    seq.end()
+}
```
NicolasDP, line 50: clever

PR #2153, opened 2020-04-29
multiple config files and merging configs

eugene-babichenko:
A user may provide multiple configuration files to jormungandr. When
more than one file was provided, values from the last will override
all values set in previous file.

For example, jormungandr is started with the following line:

	jormungandr --config config.yaml --config config-1.yaml

If the configs value are:

	p2p.public_address = "/ip4/127.0.0.1/tcp/8299" # config.yaml
	p2p.public_address = "/ip6/::1/tcp/9000" # config-1.yaml

then **"/ip6/::1/tcp/9000"** will be used.

### DIFF
--- a/jormungandr-integration-tests/src/jormungandr/bft/start_node.rs
+++ b/jormungandr-integration-tests/src/jormungandr/bft/start_node.rs
@@ -66,7 +66,7 @@ pub fn test_jormungandr_with_wrong_logger_fails_to_start() {
         }]))
         .build();
     Starter::new().config(config).start_fail(
-        r"Error while parsing the node configuration file: log\[0\]\.format: unknown variant",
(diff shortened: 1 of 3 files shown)


rinor:
This is nice, since can also be used to split the configuration sections in multiple files

NicolasDP:
This is a bit too hacky and unfortunatelly not the right time for this kind of change in the node.

* remember, we are planing on moving to the `jormungandr-watchdog` and it has its own way of dealing with configuration;
* configuration will be dynamic at that point and users will be able to dynamically change the settings

also this is adding a bit of a complexity that we don't want to deal with at the moment. Adding this now will mean we will need to keep this in future and I am not sure it is something we can commit to right now (especially that we are going to change how things work soon).

PR #2166, opened 2020-05-04
JCLI - properly expose vote commands

rinor:
- expose `jcli rest v0 vote active plans`
- expose `jcli rest v0 vote active committees` (instead of jcli rest v0 committess)
- fix api spec for active plans and added a temporary example
- fixed the endpoints votes -> vote
- all docs updated


doc/api/v0.yaml
```diff
@@ -2316,3 +2323,39 @@ paths:
                     - vote_end
                     - committee_end
                     - proposals
+              examples:
```
danielSanchezQ, line 2326: Cool. Did not know about this!

PR #2176, opened 2020-05-05
Log unused results

danielSanchezQ:


jormungandr/src/network/subscription.rs
```diff
@@ -259,7 +274,7 @@ impl Sink<net_data::Fragment> for FragmentProcessor {
 
     fn poll_ready(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Result<(), Error>> {
         if self.buffered_fragments.len() >= buffer_sizes::inbound::FRAGMENTS {
-            ready!(self.poll_send_fragments(cx));
+            ready!(self.poll_send_fragments(cx))?;
```
danielSanchezQ, line 277: I'm not sure about this. Since we were not checking them before I did not know if it was the expected behavior. I can `map_err` as the others and just `debug!` the error there instead of lifting.

jormungandr/src/network/subscription.rs
```diff
@@ -288,7 +303,7 @@ impl Sink<net_data::Fragment> for FragmentProcessor {
                     Error::new(Code::Internal, e)
                 });
             } else {
-                ready!(self.poll_send_fragments(cx));
+                ready!(self.poll_send_fragments(cx))?;
```
danielSanchezQ, line 306: Same as above for this one.

jormungandr/src/network/subscription.rs
```diff
@@ -306,15 +321,18 @@ impl Sink<net_data::Fragment> for FragmentProcessor {
...
```
danielSanchezQ, line 324: Same as above for this one.

jormungandr/src/network/client/mod.rs
```diff
@@ -158,16 +158,17 @@ impl Progress {
 impl Client {
     fn process_block_event(&mut self, cx: &mut Context<'_>) -> Poll<Result<ProcessingOutcome, ()>> {
         use self::ProcessingOutcome::*;
-
         // Drive sending of a message to block task to clear the buffered
         // announcement before polling more events from the block subscription
         // stream.
+        let logger = self.logger().clone();
```
mzabaluev, line 164: Why `clone` here, a reference would work just as well?

jormungandr/src/network/service.rs
```diff
@@ -274,7 +274,7 @@ impl GossipService for NodeService {
     }
 
     async fn peers(&self, limit: u32) -> Result<Peers, Error> {
-        let logger = self.logger().new(o!("request" => "Peers"));
+        let _logger = self.logger().new(o!("request" => "Peers"));
```
mzabaluev, line 277: Just remove it, it's not used anywhere.

jormungandr/src/network/subscription.rs
mzabaluev: Good catch, the sending errors are fatal and must be propagated.

jormungandr/src/network/subscription.rs
```diff
@@ -306,15 +321,18 @@ impl Sink<net_data::Fragment> for FragmentProcessor {
...
     }
 }
 
 impl FragmentProcessor {
     fn poll_send_fragments(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), Error>> {
-        ready!(self.mbox.poll_ready(cx));
+        let logger = &self.logger;
+        ready!(self.mbox.poll_ready(cx))
+            .map_err(|e| debug!(logger, "error sending fragment"; "reason" => %e))
+            .ok();
```
mzabaluev, line 335: Bubble up this one as well. It's good to do logging here in a centralized place. Clarify the message as "error sending fragments for processing" because it's an internal channel.

jormungandr/src/rest/v0/logic.rs
```diff
@@ -105,9 +105,12 @@ pub async fn get_account_state(
...
     let logger = context.logger()?.new(o!("request" => "message_logs"));
-    let (reply_handle, reply_future) = intercom::unary_reply(logger);
+    let (reply_handle, reply_future) = intercom::unary_reply(logger.clone());
     let mut mbox = context.try_full()?.transaction_task.clone();
-    mbox.send(TransactionMsg::GetLogs(reply_handle)).await;
+    mbox.send(TransactionMsg::GetLogs(reply_handle))
+        .await
+        .map_err(|e| debug!(&logger, "error getting message logs"; "reason" => %e))
+        .ok();
     reply_future.await.map_err(Into::into)
```
mzabaluev, line 114: If we failed to send the message to the task, the reply will never get received.
Better return the error after logging if `mbox.send` fails; I think it's OK to convert it to status 500 for the response.

jormungandr/src/rest/v0/logic.rs
```diff
@@ -419,9 +422,12 @@ pub async fn get_network_stats(context: &Context) -> Result<Vec<PeerStats>, Erro
...
 
     let logger = context.logger()?.new(o!("request" => "network_stats"));
-    let (reply_handle, reply_future) = intercom::unary_reply(logger);
+    let (reply_handle, reply_future) = intercom::unary_reply(logger.clone());
     let mut mbox = full_context.network_task.clone();
-    mbox.send(NetworkMsg::PeerInfo(reply_handle)).await;
+    mbox.send(NetworkMsg::PeerInfo(reply_handle))
+        .await
+        .map_err(|e| debug!(&logger, "error getting network stats"; "reason" => %e))
+        .ok();
```
mzabaluev, line 430: Same as above.

jormungandr/src/network/client/mod.rs
danielSanchezQ: If I recall well in the `let mut block_sink = Pin::new(&mut self.block_sink);` call we take a mut ref to self which invalidates the others.

jormungandr/src/rest/v0/logic.rs
danielSanchezQ: @mzabaluev , how to wrap or transform the Error here?

mzabaluev: There is `TxMsgSendError` already, it just does not seem to be used.

jormungandr/src/network/client/mod.rs
mzabaluev: Hmm; I'd expect the borrow checker to consider the member borrows separately.

jormungandr/src/rest/v0/logic.rs
danielSanchezQ: It is being used actually. In the calls to `try_send`. Which differs in returning type with `send`.

jormungandr/src/network/client/mod.rs
```diff
@@ -189,11 +190,13 @@ impl Client {
         // the buffered solicitation before polling more events from the
         // block subscription stream.
         let mut client_box = Pin::new(&mut self.client_box);
-        ready!(client_box.as_mut().poll_ready(cx));
+        let logger = &self.logger;
+        ready!(client_box.as_mut().poll_ready(cx))
+            .unwrap_or_else(|e| debug!(logger, "error processing block event"; "reason" => %e));
```
mzabaluev, line 195: In fact, why should this sink error be handled differently than `start_send` below? Use `map_err` and `?` here as well?

danielSanchezQ: Yep, I missed this one from the last fix. Sorry.

PR #2179, opened 2020-05-06
Sundry improvements and code cleanup in network

mzabaluev:
Improvements made while debugging the functionality of the new network implementation.

- Clearer error reporting and a better named enum variant for `client::connect` errors.
- Pretty-format the gRPC code generated by tonic-build by default, but disable formatting in CI
  to avoid an unnecessary rustfmt dependency.

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -106,7 +106,7 @@ jobs:
         with:
           use-cross: ${{ matrix.config.cross }}
           command: rustc
-          args: --release --target ${{ matrix.config.target }} -p jormungandr --bin jormungandr -- -C target-cpu=${{ matrix.config.target_cpu }} -C lto
+          args: --release --target ${{ matrix.config.target }} -p jormungandr --bin jormungandr --no-default-features -- -C target-cpu=${{ matrix.config.target_cpu }} -C lto
 
       - name: Build jcli
         uses: actions-rs/cargo@v1
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -76,7 +76,7 @@ jobs:
         with:
           use-cross: ${{ matrix.config.cross }}
           command: rustc
-          args: --release --target ${{ matrix.config.target }} -p jormungandr --bin jormungandr -- -C target-cpu=${{ matrix.config.target_cpu }} -C lto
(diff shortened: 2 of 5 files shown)


mzabaluev:
I should have properly split it into two PRs; this is quickly dumped to commit some preparatory stuff before fixing the issue with missing metadata.

jormungandr/Cargo.toml
```diff
@@ -52,7 +52,7 @@ slog-json = "2.3.0"
 slog-scope = "4.1"
 slog-term = "2.4.0"
 structopt = "^0.3"
-thiserror = "1.0"
+thiserror = "1.0.16"
```
danielSanchezQ, line 55: Doesn't `1.0` already use the latest path version? Or do we need the specific `0.16` one?

mzabaluev: This change uses `transparent` in the macro, which was introduced post 1.0 (not a good practice, API additions properly should be associated minor version increments). So I took the latest patch release as the current base version.

.github/workflows/nightly.yml
```diff
@@ -106,7 +106,7 @@ jobs:
         with:
           use-cross: ${{ matrix.config.cross }}
           command: rustc
-          args: --release --target ${{ matrix.config.target }} -p jormungandr --bin jormungandr -- -C target-cpu=${{ matrix.config.target_cpu }} -C lto
```
NicolasDP, line 109: you can't use this in the root workspace, you need to be in the directory `cd jormungandr && rustc`...

.github/workflows/release.yml
```diff
@@ -76,7 +76,7 @@ jobs:
         with:
           use-cross: ${{ matrix.config.cross }}
           command: rustc
-          args: --release --target ${{ matrix.config.target }} -p jormungandr --bin jormungandr -- -C target-cpu=${{ matrix.config.target_cpu }} -C lto
```
NicolasDP, line 79: same thing here, this won't work

.github/workflows/nightly.yml
NicolasDP:
```Run actions-rs/cargo@v1
/usr/share/rust/.cargo/bin/cargo rustc --release --target x86_64-unknown-linux-gnu -p jormungandr --bin jormungandr --no-default-features -- -C target-cpu=generic -C lto
error: --no-default-features is not allowed in the root of a virtual workspace```

mzabaluev: Sorry, fixed in #2186.

PR #2181, opened 2020-05-06
Remove unused polls results warnings

danielSanchezQ:


jormungandr/src/network/client/mod.rs
```diff
@@ -177,13 +177,14 @@ impl Client {
             // Ignoring possible Pending return here: due to the following
             // ready!() invocations, this function cannot return Continue
             // while no progress has been made.
-            block_sink.as_mut().poll_flush(cx).map_err(|e| {
-                error!(
+            match block_sink.as_mut().poll_flush(cx) {
```
eugene-babichenko, line 180: Looking at the above comment I think that it _might_ be a good idea just to disable this particular warning here.

danielSanchezQ: I'm down for it if you think it is clearer. But in the end I'm just taking the errors results here and discarding whatever else. Should be the same behavior as before right?

jormungandr/src/network/client/mod.rs
```diff
@@ -177,13 +177,14 @@ impl Client {
...
+                Poll::Ready(Err(e)) => error!(
                     logger,
                     "processing of incoming block messages failed";
                     "reason" => %e,
-                );
-            })?;
```
mzabaluev, line 186: The error is bubbled up with `?` to terminate the client connection. In the proposed change it would be ignored after logging.

mzabaluev: You can return `Err` at the end of the `Ready(Err(_))` arm, and add a match arm for `Poll::Pending` with a do-nothing body, to where you can move the comment above.

jormungandr/src/network/client/mod.rs
```diff
@@ -207,13 +208,16 @@ impl Client {
             // Ignoring possible Pending return here: due to the following
             // ready!() invocation, this function cannot return Continue
             // while no progress has been made.
-            client_box.as_mut().poll_flush(cx).map_err(|e| {
-                error!(
-                    self.logger,
-                    "processing of incoming client requests failed";
-                    "reason" => %e,
-                );
-            })?;
```
mzabaluev, line 216: Same as above.

jormungandr/src/network/subscription.rs
```diff
@@ -290,7 +299,9 @@ impl Sink<net_data::Fragment> for FragmentProcessor {
     fn poll_flush(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Result<(), Error>> {
         loop {
             if self.buffered_fragments.is_empty() {
-                self.poll_complete_refresh_stat(cx);
+                match self.poll_complete_refresh_stat(cx) {
+                    _ => (),
+                };
```
mzabaluev, line 304: I think now that this is slightly wrong. If the stats refresh future is not run to completion in some hypotetical task setup that does not `poll_flush` all the time regardless of input, the statistic may fall out of date. Wrap this call into `ready!`, it won't hurt.

jormungandr/src/network/client/mod.rs
danielSanchezQ: Thanks, I misread the code intentions there.

PR #2182, opened 2020-05-06
Fix type aliases bound warnings

danielSanchezQ:


mzabaluev:
Today I learned. I thought I would need to rewrite this with manual wrapper types, but apparently you can just use `as` on parameters in type aliases.

PR #2183, opened 2020-05-06
CORS origin validation

eugene-babichenko:
Check the value in the configuration before it gets to warp, which panics on invalid values.


NicolasDP:
@eugene-babichenko , can you rebase it, the tests should pass now

eugene-babichenko:
done

PR #2184, opened 2020-05-06
ci/release: remove unnecessary checkout

eugene-babichenko:


eugene-babichenko:
@danielSanchezQ Is it safe to have a single "Publish release" workflow instead of two separate workflows? Same question for nightly also.

danielSanchezQ:
@eugene-babichenko you mean to merge both nightly and release ones?

eugene-babichenko:
@danielSanchezQ No. In `release.yml` there are two jobs: `initial_release` and `release_assets`. Can we make just a single `release` job which creates a release (the `Create Release` step) and then performs a build and uploads assets?

danielSanchezQ:
No, I don't think we can. The reason is that the second one is a matrix, so the release will be remake for each target then.

eugene-babichenko:
Ah, thanks, got it.

PR #2185, opened 2020-05-06
Set peer-address metadata header on the client side

mzabaluev:
Set the header value with the address in the `public_address` setting, if it is provided in the configuration.


mzabaluev:
A different approach is implemented in https://github.com/input-output-hk/chain-libs/pull/374

PR #2186, opened 2020-05-07
GitHub CI: build binaries from subproject manifests

mzabaluev:
This fixes the release and nightly builds while allowing to use `--no-default-features`.

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -106,7 +106,7 @@ jobs:
         with:
           use-cross: ${{ matrix.config.cross }}
           command: rustc
-          args: --release --target ${{ matrix.config.target }} -p jormungandr --bin jormungandr --no-default-features -- -C target-cpu=${{ matrix.config.target_cpu }} -C lto
+          args:  --manifest-path jormungandr/Cargo.toml --release --target ${{ matrix.config.target }} --bin jormungandr --no-default-features -- -C target-cpu=${{ matrix.config.target_cpu }} -C lto
 
       - name: Build jcli
         uses: actions-rs/cargo@v1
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -76,14 +76,14 @@ jobs:
...
-          args: --release --target ${{ matrix.config.target }} -p jormungandr --bin jormungandr --no-default-features -- -C target-cpu=${{ matrix.config.target_cpu }} -C lto
+          args: --manifest-path jormungandr/Cargo.toml --release --target ${{ matrix.config.target }} --bin jormungandr --no-default-features -- -C target-cpu=${{ matrix.config.target_cpu }} -C lto
 
       - name: Build jcli
         uses: actions-rs/cargo@v1
         with:
           use-cross: ${{ matrix.config.cross }}
           command: rustc
-          args: --release --target ${{ matrix.config.target }} -p jcli --bin jcli -- -C target-cpu=${{ matrix.config.target_cpu }} -C lto
+          args: --manifest-path jcli/Cargo.toml --release --target ${{ matrix.config.target }} --bin jcli -- -C target-cpu=${{ matrix.config.target_cpu }} -C lto
```
NicolasDP, line 86: I note that there is no `no-default-features` for `jcli` ?

.github/workflows/nightly.yml
```diff
@@ -115,7 +115,7 @@ jobs:
         with:
           use-cross: ${{ matrix.config.cross }}
           command: rustc
-          args: --release --target ${{ matrix.config.target }} -p jcli --bin jcli -- -C target-cpu=${{ matrix.config.target_cpu }} -C lto
+          args: --manifest-path jcli/Cargo.toml --release --target ${{ matrix.config.target }} --bin jcli -- -C target-cpu=${{ matrix.config.target_cpu }} -C lto
```
NicolasDP, line 118: I note that there is no no-default-features for jcli ?

.github/workflows/release.yml
mzabaluev: Yeah, this is just for conformity so that we don't have a zoo of slightly different command lines.

mzabaluev:
> not sure the `no-default-features` is going to change on the deps side of thing

Currently it disables the `rustfmt` option in `tonic-build`, which I'd like to have enabled by default in developer builds, so that the generated code is browsable. It looks impossible to tweak subproject options in VSCode builds currently.

PR #2191, opened 2020-05-07
[Test] Backward compatibility for nodes in jormungandr integration tests

dkijania:
Test code tries to use as much production code as possible with focus on configuration model and rest api. However, from time to time there are some breaking changes introduced. It was not a problem before when there was a lack of compatibility tests. Right we have them and need to ensure that node before breaking changes are working and we can interact with them in limited way (bootstrap, get some stats, check logs). 

This commit introduced split between test implementation for newest node which is in line with dev code and legacy node, which should be more elastic in sense of how to start node and how to e.g. last block height. 

Legacy Api will have separate configuration which can be formatted for particular version, also rest api is more fault tolerant and can interact with legacy version (like retrieve stats for version before 08.19)


dkijania:
above issues are connected due to bug connected to subscriptions

PR #2193, opened 2020-05-07
DNS for trusted peers

eugene-babichenko:


jormungandr-lib/src/multiaddr.rs
```diff
@@ -0,0 +1,50 @@
...
+
+    let port = if let Some(AddrComponent::TCP(port)) = components.next() {
+        port
+    } else {
+        return Ok(None);
+    };
+
+    let maybe_socket_addr = match ip_or_fqdn {
+        AddrComponent::DNS4(fqdn) => (fqdn.as_ref(), port)
+            .to_socket_addrs()?
```
mzabaluev: This is blocking for the duration of the DNS roundtrip, right?

jormungandr-lib/src/multiaddr.rs
```diff
@@ -0,0 +1,50 @@
...
+            .find(|addr| matches!(addr, SocketAddr::V4(_))),
```
mzabaluev: Another problem is that it will only take the first IP address, so this will be useless for load-balancing setups.

jormungandr-lib/src/multiaddr.rs
```diff
@@ -0,0 +1,50 @@
...
+        AddrComponent::DNS6(fqdn) => (fqdn.as_ref(), port)
+            .to_socket_addrs()?
+            .find(|addr| matches!(addr, SocketAddr::V6(_))),
```
mzabaluev: This is a bit silly part of the multiaddr, too. We are supposed to only use IPv4 or IPv6 results, but we can't try both, which is a very common use case?

jormungandr/src/settings/start/mod.rs
```diff
@@ -216,6 +218,30 @@ fn generate_network(
...
+                Ok(Some(addr)) => {
+                    info!(logger, "DNS resolved"; "fqdn" => peer.address.multi_address().to_string(), "resolved" => addr.to_string());
+                    addr
+                }
+                Ok(None) => {
+                    warn!(logger, "failed to resolve DNS"; "fqdn" => peer.address.multi_address().to_string());
+                    return None;
+                },
+                Err(e) => {
+                    warn!(logger, "failed to resolve dns"; "fqdn" => peer.address.multi_address().to_string(), "error" => e.to_string());
```
mzabaluev, line 236: Nitpick: use "DNS" capitalization for consistency.

mzabaluev:
Some past discussion in the other issues:
https://github.com/input-output-hk/jormungandr/issues/785#issuecomment-532311751
https://github.com/input-output-hk/jormungandr/pull/1164#discussion_r349575632

jormungandr/src/settings/start/mod.rs
NicolasDP: same thing for `FQDN`

mzabaluev: Ah, the "fqdn" is a structural logging keyword, which I think should be lowercase conventionally (and will be uppercased in backends like systemd)

jormungandr-lib/src/multiaddr.rs
eugene-babichenko: Exactly that :(

jormungandr-lib/src/multiaddr.rs
eugene-babichenko: IIRC DNS may be resolved in a "round-robin" fashion. This really depends on the server implementation. For example, Cloudflare states that

> the server hands out a different address each time, operating on a rotation

eugene-babichenko:
@mzabaluev is that necessary right now? It is not used in an asynchronous context.

mzabaluev:
I dislike the FQDN implementation of `ToSocketAddrs`, the unfortunate offspring of the synchronous resolve APIs that every internet stack has dragged since the 1970s, and would prefer it not being used anywhere, especially hidden under public API.

This may do for now, but alternatively we should convert an FQDN multiaddr to an FQDN `http::Uri` and an IP address family preference flag to feed as connection parameters to `tonic` and let the HTTP stack asynchronously resolve the IP addresses and work through possible round-robin alternatives, instead of doing a poor job of it ourselves.

Also, rust-multiaddr should be made to support `dns` (accepted to the "spec" from the back door in https://github.com/multiformats/multiaddr/pull/100), and maybe `http` once the multiaddr people decide how it should actually be used (https://github.com/multiformats/multiaddr/issues/63)

jormungandr-lib/src/multiaddr.rs
mzabaluev: Yes, simplistic clients have been known to only try the first IP address provided by the server.

However, we use a fully capable HTTP stack, it's only an internal information loss problem if we can't pass it an FQDN as such from the configuration.

eugene-babichenko:
resolved merge conflicts

jormungandr-lib/src/multiaddr.rs
JamesRobertKelley: > This is a bit silly part of the multiaddr, too. We are supposed to only use IPv4 or IPv6 results, but we can't try both, which is a very common use case?

DNS uses different records for IPv6, so you can query either or both. 'A' records for IPv4 and "AAAA" records for IPv6.

eugene-babichenko: This can be done I think, but will require introducing a separate library for DNS resolution.

mzabaluev: `hyper` has an async resolver abstraction and a default resolver pooling `getaddrinfo` calls in threads.
I think this can be used to plug in a customized resolver that can limit the DNS queries to `A`, `AAAA`, or do both. The main point is, converting multiaddr to a single IP address here is a lossy transformation. We want proper HTTP behavior, Happy Eyeballs and all.

JamesRobertKelley: I may have missed it, but are you considering how to pass the port number through DNS?

There are two scenarios depending on whether you want the genesis hash to be involved in the lookup. Including a genesis hash will allow you to use the same config but switch testnets or the mainnet by specifying a different genesis hash. 

The genesis hash changes the approach because you cannot specify it as part of the DNS name. The maximum length for any part of a DNS name is 63 characters, but the genesis hash is 64 characters.

## With a Genesis Hash

You will want to use TXT records similarly to how antispam SPF records are used. TXT records will allow 255 characters, allowing you to specify the hash, multiple host's DNS names, and the port all in one record.

Example TXT record:
`cardano=8e4d2a343f3dcf9330ad9035b3e8d168e6728904262f2c434a4f8f934ec7b676 node.example.com:3000 node2.example.com:3000`

Multiple TXT records may be created, one for each genesis hash you want to maintain.

Since the TXT record can include any domain name, the usual rules geo-based or load balancing would still apply when looking up the IP for the domain name. Notice that it can include ANY domain name, so simply inputting iohk.com in the config file would allow hosts with any other domain name to be returned.

Just keep in mind that the limitation is the length of the TXT record, which is 255 characters.

## No Genesis Hash

If considerations regarding the genesis hash are not needed or desired, then SRV records are ideal because they specify port numbers and priority, similarly to MX records:

```
# _service._proto.name.      TTL   class SRV priority weight port target.
_cardano._tcp.example.com.   300   IN    SRV 10       60     3000 node.example.com.
_cardano._tcp.example.com.   300   IN    SRV 20       20     3000 node2.example.com.
```
The lookup will return a number of hosts, including the priority, to which the node should connect. This is a clean method of configuring which hosts to connect to because you can specify the port, protocol (TCP or UDP), and TTL, all in these records.

Each SRV record returns a DNS name which can benefit from geo-based or load balancing when looking up the IP address. You would also be able to specify many different hosts all with the same priority, and the cardano-node could continue to look up and connect to hosts as necessary. 

## Possible Combo of Genesis Hash and SRV Records

Just to throw this out there. A hash of the genesis hash could be used to shorten it for DNS use and then it could be used in the SRV records. Below is an example using two fictional hashes of  a testnet and mainnet:

```
# _service._proto.name.      TTL   class SRV priority weight port target.
_cardano-cf9330ad90._tcp.example.com.   300   IN    SRV 10       60     3000 mainnet1.example.com.
_cardano-cf9330ad90._tcp.example.com.   300   IN    SRV 20       20     3000 mainnet2.example.com.
_cardano-e672890426._tcp.example.com.   300   IN    SRV 10       60     3000 testnet1.example.com.
_cardano-e672890426._tcp.example.com.   300   IN    SRV 20       20     3000 testnet2.example.com.
```

NicolasDP:
@eugene-babichenko can you rebase, the tests should pass now

NicolasDP:
Also updating the documentation would be handy

eugene-babichenko:
done

doc/configuration/network.md
```diff
@@ -57,8 +57,9 @@ rest:
 ## P2P configuration
 
 - `trusted_peers`: (optional) the list of nodes' [multiaddr][multiaddr] to connect to in order to
-    bootstrap the p2p topology (and bootstrap our local blockchain) with the associated `id` (24 bytes
-    in hexadecimal given by the trusted peers to allow initial connection to it).
+    bootstrap the p2p topology (and bootstrap our local blockchain). Note that you can use a DNS
+    name in the following format: `/dns4/node.example.com/tcp/3000`. Use `dns6` instead of `dns4`
+    if you want the peer to connect with IPv6.
```
NicolasDP, line 62: nice, now it is more obvious for the user how to use this feature

jormungandr-lib/src/multiaddr.rs
NicolasDP: Thanks @JamesRobertKelley. I like the approach with the SRV and I can see how that could be useful on the long run. Would you mind converting this to a [`JIP`](https://github.com/input-output-hk/jormungandr/issues/new?template=jip.md)? It will be easier to track down the discussion around this in a separate issue.

PR #2199, opened 2020-05-08
Cleanup futures 0.1 from jormungandr

eugene-babichenko:


jormungandr/Cargo.toml
```diff
@@ -52,7 +52,7 @@ slog-scope = "4.1"
 slog-term = "2.4.0"
 structopt = "^0.3"
 thiserror = "1.0.16"
-tokio = { version = "0.2.12", features = ["signal"] }
+tokio = { version = "0.2.12", features = ["rt-threaded", "time", "sync", "stream", "signal", "test-util", "blocking"] }
```
mzabaluev, line 55: So many features are added with this small change, are these needed for other things to work?

mzabaluev: I guess it's leaner than "full" in any case...

jormungandr/src/blockchain/process.rs
```diff
@@ -104,7 +104,7 @@ impl Process {
 
                 info!(logger, "receiving block from leadership service");
 
-                info.timeout_spawn_failable_std(
+                info.timeout_spawn_failable(
```
mzabaluev, line 107: Might as well rename to `*_fallible` which is the preferred Rust lingo ;)

jormungandr/src/log/stream.rs
```diff
@@ -76,24 +82,25 @@ impl<'a, S: Debug, M: Debug> Debug for LoggingStream<'a, S, M> {
...
-    M: Display,
+    S: Stream + Unpin,
+    S::Item: Debug + Unpin,
+    M: Display + Unpin,
 {
     type Item = S::Item;
-    type Error = S::Error;
 
-    fn poll(&mut self) -> Poll<Option<Self::Item>, Self::Error> {
-        match try_ready!(self.stream.poll()) {
```
mzabaluev, line 87: How did this even work?

jormungandr/src/log/stream.rs
```diff
@@ -76,24 +82,25 @@ impl<'a, S: Debug, M: Debug> Debug for LoggingStream<'a, S, M> {
 
 impl<'a, S, M> Stream for LoggingStream<'a, S, M>
 where
-    S: Stream,
-    S::Item: Debug,
-    M: Display,
+    S: Stream + Unpin,
+    S::Item: Debug + Unpin,
+    M: Display + Unpin,
```
mzabaluev, line 87: I hope this is acceptable sacrifice for being able to use `Pin::into_inner` for convenience. Not every stream is `Unpin`. On the other hand, we hardly use this adapter and we should transition to `tracing` as our logging framework in the future, so it's OK to just remove the whole thing.

Look into [pin-project](https://crates.io/crates/pin-project) as a nice way to do `Pin` projections.

jormungandr/src/utils/task.rs
```diff
@@ -194,10 +194,10 @@ impl TokioServiceInfo {
         self.name
     }
 
-    /// Access the service's executor
+    /// Access the service's handle
     #[inline]
-    pub fn executor(&self) -> &TaskExecutor {
-        &self.executor
+    pub fn handle(&self) -> &Handle {
```
mzabaluev, line 199: Might name the member more descriptively, like `runtime_handle`.

jormungandr/src/log/stream.rs
eugene-babichenko: black magic

jormungandr/src/log/stream.rs
eugene-babichenko: Thanks, will try it!

eugene-babichenko: Worked well for logs. Not sure if will work at all for `FireForgetSchedulerFuture`, but will look how to simplify that one too.

Issue #2216, opened 2020-05-12
[VIT]  jcli certificate get-vote-plan-id 


rinor:
helpful to get the voteplan id from a voteplan certificate, is the same like `jcli certificate get-stake-pool-id`

PR #2220, opened 2020-05-12
jcli - certificate get-vote-plan-id, implemented

rinor:
resolve https://github.com/input-output-hk/jormungandr/issues/2216


jcli/src/jcli_app/certificate/get_stake_pool_id.rs
```diff
@@ -9,7 +9,7 @@ use structopt::StructOpt;
 pub struct GetStakePoolId {
     /// read the certificate from
     pub input: Option<PathBuf>,
-    /// write the certificate too
+    /// write the id to
```
NicolasDP, line 12: while I appreciate the fix, it would be better split into a separate commit. Otherwise the git history will show that this was changed for adding the `get-vote-plan-id` while it is actually a wording change in the doc

NicolasDP:
@rinor , if you rebase your PR the test should pass now

rinor:
done

PR #2221, opened 2020-05-12
[Test] Reuse rest across test projects

dkijania:


testing/jormungandr-integration-tests/src/common/jormungandr/rest.rs
```diff
@@ -10,18 +16,24 @@ pub enum RestError {
     CannotDeserialize(#[from] serde_json::Error),
     #[error("could not send reqeuest")]
     RequestError(#[from] reqwest::Error),
+    #[error("hash parse error")]
+    HashParseError(#[from] chain_crypto::hash::Error)
 }
 
 /// Specialized rest api
-#[derive(Debug)]
+#[derive(Debug,Clone)]
```
NicolasDP, line 24: add some spacing in your derive please
```suggestion
#[derive(Debug, Clone)]
```

testing/jormungandr-integration-tests/src/common/jormungandr/mod.rs
```diff
@@ -8,7 +8,7 @@ pub use benchmark::storage_loading_benchmark_from_log;
 pub use configuration_builder::ConfigurationBuilder;
 pub use logger::{JormungandrLogger, LogEntry};
 pub use process::*;
-pub use rest::JormungandrRest;
+pub use rest::{JormungandrRest,RestError};
```
NicolasDP, line 11: aerate your imports, a space between `,` and the following character
```suggestion
pub use rest::{JormungandrRest, RestError};
```

jormungandr-lib/src/interfaces/peer_stats.rs
```diff
@@ -29,12 +29,7 @@ pub struct Profile {
     pub subscriptions: Vec<Subscription>,
 }
 
-#[derive(Debug, Clone, Eq, PartialEq, Serialize, Deserialize)]
-#[serde(deny_unknown_fields)]
-pub struct Info {
-    pub address: String,
-    pub id: String,
-}
+pub type Info = String;
```
mzabaluev, line 32: If you want to keep the meaning as "peer info" I'd leave it as a one-field struct for future proofing.

testing/jormungandr-scenario-tests/src/node.rs
```diff
@@ -7,7 +7,7 @@ use chain_impl_mockchain::{
 };
 use indicatif::ProgressBar;
 use jormungandr_integration_tests::{
-    common::jormungandr::logger::JormungandrLogger,
+    common::jormungandr::{logger::JormungandrLogger, JormungandrRest, RestError},
```
mzabaluev, line 10: Shouldn't we move all reusable stuff to `jormungandr-testing-utils` to reduce the cross-crate dependency spaghetti?

NicolasDP: agreed

dkijania: also agreed, but test project refactoring is not a highest priority right now. I want to do that in many small steps. this is intermediate step which moves some of rest methods to integration test and then move it to testing-utils.

testing/jormungandr-integration-tests/src/common/jormungandr/mod.rs
dkijania: Apologies for that and other fmt issues. i was running `cargo fmt` in testing directory not in the root folder

PR #2226, opened 2020-05-13
rest - /api/v0/block/block0 - endpoint added

rinor:
Fetch block0 bytes without the need to pass block0 hash.
Added for convenience to be used directly from remote http clients that may not be aware of the block0 hash and have to go through
- /api/v0/settings
- parse block0Hash
- /api/v0/block/abcdef0123456789

### DIFF
--- a/doc/api/v0.yaml
+++ b/doc/api/v0.yaml
@@ -124,6 +124,24 @@ paths:
         '404':
           description: Account with given ID was not found
 
+  '/api/v0/block/block0':
+    get:
+      description: Gets block0
+      operationId: Block0
+      tags:
+        - block
+      responses:
+        '200':
+          description: Success
+          content:
+            application/octet-stream:
+              schema:
+                description: Binary blob with block0
+                type: string
+                format: binary
+        '400':
+          description: Block was not found
+
   '/api/v0/block/{block_id}':
     get:
       description: Gets block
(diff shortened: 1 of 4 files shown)


rinor:
jcli was left out since since may not be needed much there. Let me know if should be added also there

jormungandr/src/rest/v0/mod.rs
```diff
@@ -39,7 +45,7 @@ pub fn filter(
             .and_then(handlers::get_block_next_id)
             .boxed();
 
-        root.and(get.or(get_next)).boxed()
+        root.and(get_block0.or(get.or(get_next))).boxed()
```
eugene-babichenko, line 48:
```suggestion
        root.and(get_block0.or(get).or(get_next)).boxed()
```

doc/api/v0.yaml
```diff
@@ -124,6 +124,24 @@ paths:
...
+          description: Success
+          content:
+            application/octet-stream:
+              schema:
+                description: Binary blob with block0
+                type: string
+                format: binary
+        '400':
+          description: Block was not found
+
```
NicolasDP, line 144: how is it different from the `/api/v0/block/{block_id}` ? also the `block_id`  of the block0 is already available in the settings: `/api/v0/settings`.

rinor: > how is it different from the /api/v0/block/{block_id} ?

The result is the same. It just fetches directly the block0 data without the need to through the settings.

> also the block_id of the block0 is already available in the settings: /api/v0/settings.

Yes, also mentioned in the comment.

rinor:
closing as not useful.

PR #2227, opened 2020-05-13
Performance improvements and clean up in polling inbound subscription streams

mzabaluev:


jormungandr/src/network/subscription.rs
```diff
@@ -360,7 +350,7 @@ impl FragmentProcessor {
             debug!(logger, "error sending fragments for processing"; "reason" => %e);
             Error::new(Code::Internal, e)
         })?;
-        let fragments = self.buffered_fragments.split_off(0);
+        let fragments = mem::replace(&mut self.buffered_fragments, Vec::new());
```
NicolasDP, line 353: I would put an initial capacity of `buffer_sizes::inbound::FRAGMENTS`.

mzabaluev: Will do, but as the chunking is lazy the yielded vectors may be underfilled.

PR #2241, opened 2020-05-18
Rest - /api/v0/message, add framentdID to response

rinor:
Ex: helps to track the status of wallet recovery transactions


rinor:
> I am not sure, this ID can be computed on the client side actually, if you have the fragment, you can have the ID

- Yes, we would need to expose that functionality on the chain-wallet-libs.

Still would be nice to have for other kind of clients (http batch,...), but not a big deal. Feel free to close.

PR #2247, opened 2020-05-18
make `GET /shutdown` API available during bootstrap

eugene-babichenko:


mzabaluev:
Should it properly be `POST /shutdown`? Shutting the service down is not an idempotent request :nerd_face:

eugene-babichenko:
@mzabaluev I do not want to deal with breaking API changes :P

NicolasDP:
we could add a redirection to the right one, deprecating end points is not something unknown in the REST API world

eugene-babichenko:
Is there a way to force a client to use POST instead of GET with a redirect? There is 303 for forcing GET, but I could not find anything like that for POST.

eugene-babichenko:
jcli contains a breaking change now, should I keep `shutdown get` for compatibility?

mzabaluev:
> jcli contains a breaking change now, should I keep `shutdown get` for compatibility?

I would support that. You can put in an `X-Warning:` header in the response that nobody will ever see :trollface:

doc/api/v0.yaml
```diff
@@ -1032,6 +1032,14 @@ paths:
 
   /api/v0/shutdown:
     get:
+      deprecated: true
+      description: Starts node shutdown procedure
+      tags:
+        - utils
+      responses:
+        '200':
+          description: Success
```
NicolasDP, line 1041: Seems like something is missing:

> Operation should have an `operationId`.

eugene-babichenko: done

PR #2252, opened 2020-05-19
Port integration tests to grpc 0.8

mzabaluev:


mzabaluev:
The test failures are blocked on https://github.com/stepancheg/grpc-rust/issues/184.

Suggesting a crash program of porting the tests to tonic. Interop testing on a dissimilar gRPC stack was nice, but it's becoming a drag. Also the API was never great and did not improve much in 0.8. The issues with having to nail down specific versions of protobuf (https://github.com/stepancheg/grpc-rust/issues/172) and protoc-rust (https://github.com/stepancheg/grpc-rust/issues/183) are not likely to go away, either.

dkijania:
Migration to tonic is already on my long todo list. Maybe this problem will move it on top. My suggestion is to ignore test at this moment and I'll prioritize migration to tonic

PR #2270, opened 2020-05-25
Integration test improvements

mzabaluev:
- Replace `mktemp` and home-grown file condition tests with `assert_fs` and `predicates`.
- Gather temporary files created for a node in one directory rather than a hodgepodge.
- Reuse configuration and starter infrastructure between current and legacy configs.


dkijania:
Great! Much cleaner solution than my initial one.

dkijania:
I have only one doubt regarding `assert_fs` crate. Sometime it is useful not to delete any temporary files (like logs or configs) because it allows some debugging session or reproducing problem manually. However, maybe this could be an explicit mechanism to dump all config files and logs into specified location as we are doing it in `jormungandr-scenario-tests` project

mzabaluev:
> I have only one doubt regarding `assert_fs` crate. Sometime it is useful not to delete any temporary files (like logs or configs) because it allows some debugging session or reproducing problem manually.

`TempDir` has methods `into_persistent` and `into_persistent_if`. Hopefully these can be useful to debug on a case-by-case basis, or maybe even triggered in a test fixture if a certain environment variable is set.

PR #2315, opened 2020-06-04
mips and powerpc builds still broken

eugene-babichenko:


NicolasDP:
Shame 😭

mzabaluev:
It's `ring` (we'll be all dead in a week :scream:):
https://github.com/input-output-hk/jormungandr/runs/737008454?check_suite_focus=true

```
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', /cargo/registry/src/github.com-1ecc6299db9ec823/ring-0.16.14/build.rs:371:34
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

@eugene-babichenko It would be nice to put this information when filing, for later archeology.

eugene-babichenko:
It looks like they have an error in their build script. Probably already fixed in the master but not yet released. The dev claims it should work on both architectures now.

PR #2306, opened 2020-06-03
ci: add missing parameters to release pipelines

eugene-babichenko:

PR #2319, opened 2020-06-04
CI Matrix: Reloaded

### LINKED ISSUE #2310: factor out `toolchain` from this CI table and add the few outliers with `include`:
It might be more maintainable if we factor out `toolchain` from this table and add the few outliers with `include`:

```yaml
      matrix:
        config:
          # Linux
          - {os: ubuntu-latest, target: x86_64-unknown-linux-gnu, target_cpu: generic, cross: false}
          - {os: ubuntu-latest, target: x86_64-unknown-linux-gnu, target_cpu: broadwell, cross: false}
          # ...
        toolchain: stable
        include:
          # Windows
          - config: {os: windows-latest, target: x86_64-pc-windows-gnu, target_cpu: generic, cross: false}
            toolchain: stable-x86_64-pc-windows-gnu
          - config: {os: windows-latest, target: x86_64-pc-windows-msvc, target_cpu: generic, cross: false}
            toolchain: stable-x86_64-pc-windows-msvc
          # ...
```

Then, replace every occurrence of `${{ matrix.config.toolchain }}` with `${{ matrix.toolchain }}`.

_Originally posted by @mzabaluev in https://github.com/input-output-hk/jormungandr/pull/2306_

mzabaluev:
I meant to fix #2310, but then I got a little carried away:

- Factor out repetitive parameters from the CI build matrix.
- Disable nightly runs on forked repositories.
- Remove the cross workarounds and the last vestiges of OpenSSL from our dependency graph.
- Add caching of cargo dependencies.
- Use `actions/checkout@v2`
- Replace third party actions with the official GitHub action and the official GitHub script.

(P.S. I know that the Matrix sequels were crap, just wanted to give this PR a cool name)

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -22,6 +22,7 @@ jobs:
           echo ::set-output name=version::$VERSION
 
       - name: Delete nightly with hub
+        continue-on-error: true
         env:
           GITHUB_USER: ${{ secrets.GITHUB_USER }}
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -213,7 +213,15 @@ jobs:
         with:
           use-cross: ${{ matrix.cross }}
           command: rustc
-          args: --manifest-path jormungandr/Cargo.toml --verbose --locked --release --target ${{ matrix.config.target }} --bin jormungandr --no-default-features -- -C target-cpu=${{ matrix.target_cpu }} -C lto
+          args: |
+            --manifest-path jormungandr/Cargo.toml
+            --bin jormungandr
+            --no-default-features
+            --verbose
+            --locked
+            --release
+            --target ${{ matrix.config.target }}
+            -- -C target-cpu=${{ matrix.target_cpu }} -C lto
 
       - name: Build jcli
         uses: actions-rs/cargo@v1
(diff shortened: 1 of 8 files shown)


.github/workflows/nightly.yml
```diff
@@ -39,55 +44,137 @@ jobs:
...
+            cross: true
+          # - config: {os: ubuntu-latest, target: mips64el-unknown-linux-gnuabi64}
+          #   target_cpu: generic
+          #   toolchain: stable
+          #   cross: true
+          # - config: {os: ubuntu-latest, target: powerpc64le-unknown-linux-gnu}
+          #   target_cpu: generic
+          #   toolchain: stable
+          #   cross: true
+
```
danielSanchezQ, line 161: I do not see the improvement here. With this change we are using a matrix but then specifying for each (or some) config pair extra features. I really think that by using the old style is clearer what configuration is run for each item even if it is redundant on some of its attributes.

danielSanchezQ: In fact we end up with the same redundant attributes:
```
 - config: {os: ubuntu-latest,  target: x86_64-unknown-linux-musl}
    target_cpu: generic
    toolchain: stable
    cross: true
```

mzabaluev: I wanted to get away from the giant horizontally extended table where omissions are easy to make, like the one that necessitated #2306.

Now, each entry has the same four attributes in a vertical layout that is easy to review, and the main matrix is used for its intended purpose: create an M x N set product of targets and CPUs.
Unfortunately, it's not possible to do more neat combinatorics with the current `matrix` features,
so the Windows build matrix had to be unrolled in the `include` list.

PR #2322, opened 2020-06-05
Consistently use optimization flags when building releases

mzabaluev:
`cargo rustc` only applies the compiler flags passed as arguments to the end target build.
Use `cargo build` and override the target compiler flags via `.cargo/config.toml`, this way the flags get down to dependency library compilations and LTO can be applied across the board.
This has a dramatic impact on the executable sizes.

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -209,9 +209,10 @@ jobs:
         uses: actions-rs/cargo@v1
         env:
           DATE: ${{ env.DATE }}
+          RUSTFLAGS: -C target-cpu=${{ matrix.target_cpu }} -C lto
         with:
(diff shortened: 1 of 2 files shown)


NicolasDP:
Excellent, how much does this reduce the size?

mzabaluev:
> Excellent, how much does this reduce the size?

Compare the size of cross-compiled assets [here](https://github.com/mzabaluev/jormungandr/releases/tag/nightly) to their size with default release flags [here](https://github.com/mzabaluev/jormungandr/releases/tag/v0.9.0-test.this-cake-is-a-lie).

Can't compare with the official `nightly` release in the main repo because I have accidentally deleted it by playing with `hub` :blush:

PR #2330, opened 2020-06-08
[Tests] Pool update changes keys

dkijania:
Test which send pool update certificate with new vrf and kes key and then using demote/promote changes those key and verify if node can create blocks anymore


minikin:
@dkijania Closing it. Feel free to reopen if we really need it.

PR #2333, opened 2020-06-08
[Tests] unignore topics of interests test

dkijania:
Change assertion for topics of interests tests


mzabaluev:
Still does not seem to work reliably.

PR #2339, opened 2020-06-09
Rest bulk tx actions

eugene-babichenko:


jormungandr/src/rest/v0/mod.rs
```diff
@@ -42,11 +42,24 @@ pub fn filter(
...
+        let root = warp::path!("fragment" / ..).boxed();
+
+        let logs = warp::path!("logs")
+            .and(warp::get())
+            .and(with_context.clone())
+            .and_then(handlers::get_message_logs)
+            .boxed();
+
+        let statuses = warp::path!("statuses")
+            .and(warp::post())
```
rinor, line 55: **GET** would be better instead of **POST** method

rinor:
Tested it and it works perfectly. However I would like to make some changes before merging this.
Right now there is confusion between `fragment` and `message` endpoints. I would like to collapse everything under `fragment` so we would have:
- POST: `fragment` - to send a single trx (right now is `message` - will be kept as deprecated)
- POST: `fragment/bulk` - send bulk trx (implemented here as `messages`)
- GET: `fragment/logs` - all logs (no changes here as it is right now)
- GET: `fragment/status` - get status log from a bulk id list (implemented here as `fragment/statuses`)

doc/api/v0.yaml
```diff
@@ -626,6 +691,37 @@ paths:
...
+              items:
+                type: string
+                pattern: '[0-9a-f]+'
+      responses:
+        '200':
+          description: Success
+          content:
+            application/json:
+              schema:
+                description: A list of fragment IDs
```
NicolasDP, line 716: ordered in the same order as in the input?

doc/api/v0.yaml
```diff
@@ -337,6 +337,71 @@ paths:
...
+          application/json:
+            schema:
+              type: array
+              items:
+                type: string
+                pattern: '[0-9a-f]+'
+            example: ["68dcc12fe0dfe5e7b66ca6f8c959f9aa43b273e120a77fc3e4e2f04f1ecd7968"]
+          
+      responses:
+        '200':
```
NicolasDP, line 359: what if a fragment is not in the log (GCed or ...) ? What will the user gets?

jormungandr/src/intercom.rs
```diff
@@ -458,6 +459,10 @@ pub enum TransactionMsg {
     SendTransaction(FragmentOrigin, Vec<Fragment>),
     RemoveTransactions(Vec<FragmentId>, FragmentStatus),
     GetLogs(ReplyHandle<Vec<FragmentLog>>),
+    GetStatuses(
+        Vec<FragmentId>,
+        ReplyHandle<HashMap<FragmentId, FragmentStatus>>,
```
NicolasDP, line 464: Glad you have been using a message passing for this

jormungandr/src/rest/v0/mod.rs
NicolasDP: I agree with that

doc/api/v0.yaml
eugene-babichenko: Yes.

jormungandr/src/rest/v0/mod.rs
eugene-babichenko: We can do that. I made it `POST` because the HTTP standard does not recommend parsing `GET` bodies. But I agree `GET` is semantically better here.

doc/api/v0.yaml
eugene-babichenko: For now here is no 404, so the value will be just missing from the dictionary. Not very good behavior, but I didn't find a way to make fast fragment lookup in the blockchain.

eugene-babichenko:
@rinor I agree about changing the API like that. However, I don't like this part

> `fragment` endpoint to provide also the bulk functionality, by using the `Content-Type` to decide which functionality to provide, so we **don't need** to have also `fragment/bulk` endpoint

Why not use the bulk API for a single fragment too? This is just more straightforward.

cc @NicolasDP

doc/api/v0.yaml
NicolasDP: document it then, please

doc/api/v0.yaml
NicolasDP: document that if the fragment is no longer in the log it does not mean it was not accepted or accepted and that it is now better to query the explorer API

jormungandr/src/rest/v0/mod.rs
NicolasDP: Good point, I've missed that part. POST is fine.

NicolasDP:
> Tested it and it works perfectly. However would like to make some changes before merging this.
> Right now there is confusion between `fragment` and `message` endpoints. I would like to collapse everything under `fragment` so we would have:
> 
> * POST: `fragment` - to send a single trx (right now is `message` - will be kept as deprecated)
> * POST: `fragment/bulk` - send bulk trx (implemented here as `messages`)
> * GET: `fragment/logs` - all logs (no changes here as it is right now)
> * GET: `fragment/status` - get status log from a bulk id list (implemented here as `fragment/statuses`)
> 
> Edit: Ideally would like POST: `fragment` endpoint to provide also the bulk functionality, by using the `Content-Type` to decide which functionality to provide, so we **don't need** to have also `fragment/bulk` endpoint

I aggree with that. Actually what about making it a `v1` API instead?

rinor:
> @rinor I agree about changing the API like that. However, I don't like this part
> 
> > `fragment` endpoint to provide also the bulk functionality, by using the `Content-Type` to decide which functionality to provide, so we **don't need** to have also `fragment/bulk` endpoint
> 
> Why not use the bulk API for a single fragment too? This is just more straightforward.
> 
@eugene-babichenko was just to keep v0 compatibility, but if we are going v1 it doesn't matter.

jormungandr/src/rest/v0/mod.rs
eugene-babichenko: By the way, why not use `GET` queries? In a form if `?ids=id1,id2,id3` or `?id=id1&id=id2&id=id3`.

eugene-babichenko: So they are going to be comma-separated

rinor: You could do them both actually, **post** for json and **get** `?ids=id1,id2,id3`.
What would be the response type/fromat of that get option?

eugene-babichenko:
So this is the start of v1 API. I decided to stick with the following:

* `POST /fragments` for sending transactions (same data structure as before)
* `GET /fragments/statuses?fragment_ids=id1,id2,id3` for getting statuses
* `GET /fragments/logs` is just a copy of v0 API

jormungandr/src/rest/v0/mod.rs
eugene-babichenko: the same

PR #2341, opened 2020-06-09
Actively refresh cargo index cache

mzabaluev:
Use of the `--locked` option with `cargo fetch` caused the index to be stuck at an old checkout that's sufficient for this `Cargo.lock`, increasing resync times. Better to save the fresh index every time.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -96,18 +96,12 @@ jobs:
       - checkout
       - run: git submodule sync
       - run: git submodule update --init --depth=1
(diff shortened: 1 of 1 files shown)


mzabaluev:
> Use of the `--locked` option with `cargo fetch` caused the index to be stuck at an old checkout that's sufficient for this `Cargo.lock`.

This theory was wrong, `--locked` does not affect this behavior. So we'll just have to live with this index until the world is new again for `cargo fetch`.

PR #2345, opened 2020-06-10
decode the vote plan with the associated data from transaction

### LINKED ISSUE #2068: [VIT] genesis block0: initial voting plans
allow for initial voting plans to be added in the initial block0 certificates

depends on input-output-hk/chain-libs#322

NicolasDP:
fix #2068

### DIFF
--- a/jormungandr-lib/src/interfaces/block0_configuration/initial_fragment.rs
+++ b/jormungandr-lib/src/interfaces/block0_configuration/initial_fragment.rs
@@ -72,6 +72,13 @@ pub fn try_initials_vec_from_messages<'a>(
                 let cert = certificate::SignedCertificate::StakeDelegation(cert, auth);
                 inits.push(Initial::Cert(cert.into()))
             }
+            Fragment::VotePlan(tx) => {
+                let tx = tx.as_slice();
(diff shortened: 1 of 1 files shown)


jormungandr-lib/src/interfaces/block0_configuration/initial_fragment.rs
```diff
@@ -72,6 +72,13 @@ pub fn try_initials_vec_from_messages<'a>(
                 let cert = certificate::SignedCertificate::StakeDelegation(cert, auth);
                 inits.push(Initial::Cert(cert.into()))
             }
+            Fragment::VotePlan(tx) => {
+                let tx = tx.as_slice();
+                let cert = tx.payload().into_payload();
+                let auth = tx.payload_auth().into_payload_auth();
+                let cert = certificate::SignedCertificate::VotePlan(cert, auth);
```
rinor, line 79: the auth here will be empty, right?

NicolasDP: yes. it "future" proof it so if we go for something else we can still change our mind :)

PR #2353, opened 2020-06-11
End binary/library duality in scenario tests

mzabaluev:
The `lib.rs` in the `jormungandr-scenario-tests` project makes it a weird dual crate where the binary links to the library. This only complicates and slows down building.


mzabaluev:
There are so many functions that are not used anywhere.
If they are meant for the future, it might make sense to annotate them with `#[allow(dead_code)]`.
If not, should we just delete them?

dkijania:
Removed some of unused functions added allow dead code for some methods to be used in future

PR #2360, opened 2020-06-15
explorer: get vote plan info by id

eugene-babichenko:
TODO:

- [x] ~Add account values~
- [x] Implement vote tally (depends on account values)
- [x] See what fields are missing and add them
- [x] Add fetching multiple vote plans at once (paginated)
- [x] Have some clue about reward distribution

### DIFF
--- a/jormungandr/src/explorer/graphql/scalars.rs
+++ b/jormungandr/src/explorer/graphql/scalars.rs
@@ -52,9 +52,6 @@ pub enum PayloadType {
 }
 
 #[derive(juniper::GraphQLScalarValue)]
-pub struct Address(pub String);
-
-#[derive(juniper::GraphQLScalarValue)]
 pub struct Weight(pub String);
 
 /// Vote option range
--- a/jormungandr/src/explorer/graphql/scalars.rs
+++ b/jormungandr/src/explorer/graphql/scalars.rs
@@ -51,6 +51,9 @@ pub enum PayloadType {
     Public,
 }
 
+#[derive(juniper::GraphQLScalarValue)]
+pub struct Address(pub String);
+
 /// Vote option range
 ///
 /// provide a range of available choices for a given proposal. Usual value would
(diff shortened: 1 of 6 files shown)


jormungandr/src/explorer/indexing.rs
```diff
@@ -89,6 +94,30 @@ pub enum ExplorerAddress {
...
+    pub payload_type: PayloadType,
+    pub proposals: Vec<ExplorerVoteProposal>,
+}
+
+#[derive(Clone)]
+pub struct ExplorerVoteProposal {
+    pub proposal_id: ExternalProposalId,
+    pub options: Options,
+    pub tally: Option<ExplorerVoteTally>,
+    pub votes: std::collections::HashMap<ExplorerAddress, u8>,
```
NicolasDP, line 112: what is this `u8` ?

eugene-babichenko: Huh, you are right. This should be `Choice`.

jormungandr/src/explorer/graphql/scalars.rs
```diff
@@ -51,6 +51,12 @@ pub enum PayloadType {
     Public,
 }
 
+#[derive(juniper::GraphQLScalarValue)]
+pub struct Address(pub String);
```
enzoc4, line 55: why do we need a new scalar type and not the existing Address type?

jormungandr/src/explorer/indexing.rs
```diff
@@ -89,6 +94,33 @@ pub enum ExplorerAddress {
...
+    pub payload_type: PayloadType,
+    pub proposals: Vec<ExplorerVoteProposal>,
+}
+
+#[derive(Clone)]
+pub struct ExplorerVoteProposal {
+    pub proposal_id: ExternalProposalId,
+    pub options: Options,
+    pub tally: Option<ExplorerVoteTally>,
+    pub votes: std::collections::HashMap<ExplorerAddress, Choice>,
```
enzoc4, line 112: I'm curious about this `HashMap` . Is it not another `Hamt` for a reason? Is it fine to clone it on every vote?

Issue #2354, opened 2020-06-11
Add a publish/subscribe api for blockchain events


enzoc4:
# Add a publish/subscribe api for blockchain events
## Idea
Have a node service/module that allows to subscribe to blockchain events
that can be used for things like an external explorer/blockchain indexer, wallet backend, etc.

### Interesting Events

- New block
- Tip changed

## Steps

1. Create a new message box in the node just like in the explorer, and
   propagate the necessary events (one of those is basically what's done for
   the explorer right now).

2. Add a new warp handler for **websockets** at `ws://host/notifier`
    ### Basic idea
    - the handler will listen for new connections
    - at each local event will send a message to all the
        connected clients

    The messages will be (for now)

    - `NewBlock(blockid)`: Passing id's here means that the actual block will
        need to be obtained from somewhere else, but that's already needed
        for bootstrapping (getting blocks from 0..tip), so it may be
        enough to solve that problem there
    - `NewTip(blockid)`

    All messages are one-directional. Node -> Connected clients. Clients can't
    publish things.

3. Add new settings to configuration.

`max_connections: maximum number of subscribers at the same time`

## Questions
Do we want the clients to be able to subscribe to specific
messages/topics/channels?
How should we encode the messages? Using a binary format? text? If we only send
ids it's not much of a difference, though.

gufmar:
I have no experience with as and so not sure if this makes sense, but reading about 'able to subscribe to specific messages/topics/channels' let me think about MQTT. Their clients and brokers already implement topics and subscriptions. There are plenty of client implementations. Lightweight. Standardized.

gufmar:
Looks like websockets and mqtt are can be combined: http://www.steves-internet-guide.com/mqtt-websockets/

PR #2378, opened 2020-06-16
Add blockchain event notifier service

enzoc4:
# Add a `Notifier` service

## relates to #2354 

- [X] Add websocket endpoint
- [X] Process block events and send message
- [X] Process tip change event and send message
- [x] Add automated test (technically done, but still needs some though)
- [X] Add setting to limit the number of connections (and set a default)
- [X] Define the message format, right now they are ~~just strings manually crafted (`New Block: hash` `New Tip: hash`)~~ json strings with serde. Maybe we can provide binary too depending on the client
- [ ] Add this to documentation.
- [X] Introduce some idea of versioning the API? Something like
    - [X] Add a version in the endpoint path, maybe the same that the rest api.
    - [ ] Use websocket subprotocols
- [ ] Send connection close event when shutting down

The max_connection setting is under the `Rest` settings, for example:

```yaml
rest:
  listen: "127.0.0.1:8443"
  notifier:
      max_connections: 5
```
### Questions

`NewBlock` events are always sent even when `NewTip`  has the same hash. I'm not sure if we want to filter these or not.

It can be manually quick-tested by opening a websocket from a browser, for example:

```html
<!DOCTYPE html>
<html>
    <head>
        <title>Notifier reader</title>
    </head>
    <body>
        <h1>new blocks</h1>
        <div id="status">
            <p><em>Connecting...</em></p>
        </div>
        <script type="text/javascript">
        var uri = 'ws://localhost:8443/notifier';
        var ws = new WebSocket(uri);
        function message(data) {
            var line = document.createElement('p');
            line.innerText = data;
            var status = document.getElementById('status');
            status.appendChild(line);
        }
        ws.onopen = function() {
            var status = document.getElementById('status');
            status.innerHTML = "<p><em>Connected!</em></p>";
        }
        ws.onmessage = function(msg) {
            message(msg.data);
        };
        </script>
    </body>
</html>

```


Cargo.lock
```diff
@@ -2110,6 +2160,39 @@ dependencies = [
  "serde_yaml",
 ]
 
+[[package]]
+name = "openssl"
```
NicolasDP, line 2164: we will have problems on some architecture with this

jormungandr/src/rest/mod.rs
```diff
@@ -32,11 +33,13 @@ pub async fn start_rest_server(config: Rest, explorer_enabled: bool, context: Co
...
+    let notifier = notifier::filter(context.clone());
     if explorer_enabled {
         let explorer = explorer::filter(context);
-        setup_cors(api.or(explorer), config, stopper_rx).await;
+        setup_cors(api.or(notifier).or(explorer), config, stopper_rx).await;
     } else {
-        setup_cors(api, config, stopper_rx).await;
+        setup_cors(api.or(notifier), config, stopper_rx).await;
     }
 }
```
NicolasDP, line 44: the API is going to be off. the notifier is part of the `v1` API.

testing/jormungandr-integration-tests/Cargo.toml
```diff
@@ -40,6 +40,7 @@ url = "2.1.1"
 yaml-rust = "0.4.4"
 indicatif = "0.15.0"
 fs_extra = "1.1.0"
+tungstenite = "0.11.0"
```
NicolasDP, line 43: hmm okay, so this is what brings in all the dependencies

jormungandr/src/notifier/mod.rs
```diff
@@ -0,0 +1,180 @@
...
+    max_connections: usize,
+}
+
+#[derive(Serialize)]
+enum JsonMessage {
+    #[serde(serialize_with = "to_hex")]
+    NewBlock(HeaderId),
+    #[serde(serialize_with = "to_hex")]
+    NewTip(HeaderId),
+}
```
NicolasDP, line 32: This would live in better in jormungandr-lib

jormungandr/src/notifier/mod.rs
```diff
@@ -0,0 +1,180 @@
...
+
+async fn process_message(
+    logger: Logger,
+    clients: Arc<tokio::sync::Mutex<Clients>>,
+    msg: Message,
+    mut disconnected: MessageBox<usize>,
+) {
+    let warp_msg = JsonMessage::from(msg).into();
+    let dead = async move { notify_all(clients, warp_msg).await };
```
NicolasDP, line 111: what is wrong with simply using:

```suggestion
    let dead = notify_all(clients, warp_msg).await;
```

jormungandr/src/notifier/mod.rs
```diff
@@ -0,0 +1,180 @@
...
+// error codes in 4000-4999 are reserved for private use.
+// I couldn't find an error code for max connections, so I'll use the first one for now
+// maybe using the standard error code for Again is the right thing to do
+const MAX_CONNECTIONS_ERROR_CLOSE_CODE: u16 = 4000;
+const MAX_CONNECTIONS_ERROR_REASON: &str = "MAX CONNECTIONS reached";
+
+#[derive(Clone)]
+pub struct Notifier {
+    next_user_id: Arc<AtomicUsize>,
+    clients: Arc<tokio::sync::Mutex<Clients>>,
```
NicolasDP, line 22: hmm... not going to scale well.

Cargo.lock
enzoc4: good point, tungstenite is only used in the tests, so I think turning off the tls feature may be fine.

jormungandr/src/notifier/mod.rs
enzoc4: nothing, just a leftover of a little refactor

jormungandr/src/notifier/mod.rs
enzoc4: moved

jormungandr/src/rest/mod.rs
enzoc4: actually, it was under `/notifier` and not `/v0/notifier`. But I moved it to `v1/notifier`.

eugene-babichenko:
The overall idea is good but I'd go for a different approach to send messages to clients. Namely, having an actor that collects all events from other actors, builds WS messages right away and just broadcast these messages to client connection handlers. As an optimization you could skip serialization when there is no listeners.

Such approach would allow to perform the message serialization only once per each message. Now it is done `n` times where `n` is the number of subscriptions. This will also simplify the WS handling code. In fact, this code will become universal, because all handlers will receive the same type of messages.

enzoc4:
@eugene-babichenko I'm not sure I get your point. The difference I see between the current code and what you are describing is the 

> builds WS messages right away

part. Which I think I get, we can move the serialization before/up in the flow, right now we are just serializing ids, so it's probably not that expensive, but makes sense.

Maybe you are referring to something different, though.

eugene-babichenko:
@enzoc4 

I mean the following data flow:

```
+----------------+                                                       +-----------+
| block events   +--------------+                      +---------------->+ ws 1      |
+----------------+      +-------v-------+      +---------------+         +-----------+
                        | make message  +----->+ serialize msg |
+----------------+      +-------^-------+      +---------------+         +-----------+
| tip   events   +--------------+                      +---------------->+ ws 2      |
+----------------+                                                       +-----------+
```

enzoc4:
> @enzoc4
> 
> I mean the following data flow:
> 
> ```
> +----------------+                                                       +-----------+
> | block events   +--------------+                      +---------------->+ ws 1      |
> +----------------+      +-------v-------+      +---------------+         +-----------+
>                         | make message  +----->+ serialize msg |
> +----------------+      +-------^-------+      +---------------+         +-----------+
> | tip   events   +--------------+                      +---------------->+ ws 2      |
> +----------------+                                                       +-----------+
> ```

I think now the data flow is just like that, but I don't see how does it simplify any code tbh. The websocket handlers were always receiving text anyway. It still makes sense to serialize at once, probably.

PR #2380, opened 2020-06-16
Serialize fragments just like we did last summer

mzabaluev:
Messy and length-prefixed.


mzabaluev:
@NicolasDP 
Well, I sort of wanted it to be backward compatible protocol wise, just never realized the pre-0.9 fragment encoding was length-prefixed for no protobuf-worthy reason.

PR #2396, opened 2020-06-19
update the error type of blockchain's chain

NicolasDP:
remove one more dependency to error_chain crate

### DIFF
--- a/jormungandr/src/blockchain/chain.rs
+++ b/jormungandr/src/blockchain/chain.rs
@@ -66,54 +66,94 @@ use tokio::stream::StreamExt;
 // derive
 use thiserror::Error;
 
-error_chain! {
-    foreign_links {
-        Storage(StorageError);
-        Ledger(ledger::Error);
-        Block0(Block0Error);
+pub type Result<A> = std::result::Result<A, Error>;
+
+#[derive(Error, Debug)]
+pub enum Error {
+    #[error("")]
+    Storage {
+        #[from]
+        #[source]
+        source: StorageError,
+    },
+
+    #[error("")]
+    Ledger {
+        #[from]
+        #[source]
+        source: ledger::Error,
+    },
+
+    #[error("")]
+    Block0 {
+        #[from]
+        #[source]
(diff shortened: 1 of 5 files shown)


jormungandr/src/blockchain/chain.rs
```diff
@@ -66,57 +66,97 @@ use tokio::stream::StreamExt;
...
-    }
+pub type Result<A> = std::result::Result<A, Error>;
 
-    errors {
-        Block0InitialLedgerError {
-            description("Error while creating the initial ledger out of the block0")
-        }
+#[derive(Error, Debug)]
+pub enum Error {
+    #[error("")]
```
mzabaluev, line 73: You can use `#[error(transparent)]` on these

jormungandr/src/blockchain/chain.rs
```diff
@@ -66,57 +66,97 @@ use tokio::stream::StreamExt;
...
+            message: message.to_owned(),
         }
     }
 }
 
+impl From<String> for Error {
+    fn from(message: String) -> Self {
+        Self::Custom { message }
+    }
+}
```
mzabaluev, line 158: I think these two could be unified into `impl<T: Into<String>> From<T> for Error`, but this is finessing it.

Maybe even better to replace these with an `Error::custom` constructor to be able to audit uses, this looks like a stopgap when someone could not be bothered to create another error variant.

jormungandr/src/blockchain/chain.rs
```diff
@@ -66,57 +66,97 @@ use tokio::stream::StreamExt;
...
 
-    errors {
-        Block0InitialLedgerError {
-            description("Error while creating the initial ledger out of the block0")
-        }
+#[derive(Error, Debug)]
+pub enum Error {
+    #[error("")]
+    Storage {
+        #[from]
```
eugene-babichenko, line 75: `from` already implies `source`

eugene-babichenko:
Superseded by #3310.

PR #2398, opened 2020-06-19
implement `jcli certificate new vote-cast`

eugene-babichenko:


jcli/src/jcli_app/certificate/new_vote_cast.rs
```diff
@@ -0,0 +1,38 @@
...
+}
+
+impl VoteCastCmd {
+    pub fn exec(self) -> Result<(), Error> {
+        let vote_cast = VoteCast::new(
+            self.vote_plan_id,
+            self.proposal_index,
+            Payload::Public {
+                choice: Choice::new(self.choice),
+            },
```
NicolasDP, line 33: add an option `--public` so we already enforce the users to be aware that the vote is disclosed on chain

NicolasDP: if the `--public` is not set, then the command fail, because the crypto for the privacy preserving part is not finalised yet.

eugene-babichenko: done

jcli/src/jcli_app/certificate/new_vote_cast.rs
```diff
@@ -0,0 +1,44 @@
...
+}
+
+impl VoteCastCmd {
+    pub fn exec(self) -> Result<(), Error> {
+        let payload = if self.public {
+            Payload::Public {
+                choice: Choice::new(self.choice),
+            }
+        } else {
+            unimplemented!("private votes are not implemented yet");
```
NicolasDP, line 37:
```suggestion
            unimplemented!("private votes are not supported yet");
```

PR #2400, opened 2020-06-19
Protocol update at epoch transition

NicolasDP:
@enzoc4 do we need something in the explorer too to account for any protocol changes?


rinor:
Do we need to handle anything here, or `parent_ledger_state` is fine ? https://github.com/input-output-hk/jormungandr/blob/eb01e6ccef7002f6cb9b1c886fdc9b6eddc99421/jormungandr/src/blockchain/chain.rs#L865-L871

PR #2403, opened 2020-06-20
[Tests] Public id revert

dkijania:
Quick and not very elegant way to bring back public_id into test framework again.
Implemented new behavior which assign public_id  only for trusted peers, as well as remove `id`  from trusted peer if node those not support it

### DIFF
--- a/jormungandr-lib/src/interfaces/config/node.rs
+++ b/jormungandr-lib/src/interfaces/config/node.rs
@@ -16,7 +16,8 @@ pub struct Rest {
 pub struct P2p {
     /// The public address to which other peers may connect to
     pub public_address: poldercast::Address,
-
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub public_id: Option<poldercast::Id>,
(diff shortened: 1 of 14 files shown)


dkijania:
@mzabaluev , strange i had to remove serde_skip attr from public_id because otherwise got :


```
  |-> Error while parsing the node configuration file: p2p: unknown field `public_id`, expected one of `public_address`, `listen_address`, `trusted_peers`, `topics_of_interest`, `max_connections`, `max_inbound_connections`, `max_connections_threshold`, `allow_private_addresses`, `policy`, `layers`, `max_unreachable_nodes_to_connect_per_event`, `gossip_interval`, `topology_force_reset_interval`, `max_bootstrap_attempts` at line 6 column 3
[INFO ][Leader@127.0.0.1:8000]:  |-> p2p: unknown field `public_id`, expected one of `public_address`, `l
```

mzabaluev:
Removing `serde_skip` was necessary otherwise the field is not supported by deserialization at all.

PR #2409, opened 2020-06-22
don't add the trusted peers in the topology

NicolasDP:


NicolasDP:
@mzabaluev can you look at it again, I've made a change to keep the previous behaviour because no time to check in more what this happen like this in the first place. We will look at it further more down the line

mark-stopka:
If a node is both a trusted peer and also in:

```
  layers:
    preferred_list:
      view_max: 20
      peers:
```

will he be still kept in the topology?

NicolasDP:
wrong title.

PR #2415, opened 2020-06-24
Add serialize deserialize quickcheck

NicolasDP:
The Arbitrary for the TaxType in jormungandr-lib was invalid. Fortunatelly on the blockchain/production there's nothing invalid.


NicolasDP:
@dkijania we will need these test fixed or `ignored` for now then

PR #2436, opened 2020-06-30
generate leadership events only per request

eugene-babichenko:
This approach is now adopted as a way to eliminate delays on start and
on an epoch boundary.

### DIFF
--- a/jormungandr/src/secure/enclave.rs
+++ b/jormungandr/src/secure/enclave.rs
@@ -25,6 +25,14 @@ pub struct LeaderEvent {
     pub output: LeaderOutput,
 }
 
+pub struct Schedule {
+    enclave: Arc<Enclave>,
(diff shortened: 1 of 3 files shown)


jormungandr/src/secure/enclave.rs
```diff
@@ -202,6 +150,59 @@ impl Enclave {
...
+            for (id, leader) in leaders {
+                match self.leadership.is_leader_for_date(&leader, date) {
+                    Ok(LeaderOutput::None) => (),
+                    Ok(leader_output) => self.current_slot_data.push(LeaderEvent {
+                        id: *id,
+                        date,
+                        output: leader_output,
+                    }),
+                    Err(_) => {
+                        // For now silently ignore error
```
mzabaluev, line 186: Ehh... OK, it was the same before the change.

PR #2467, opened 2020-07-23
add separate explorer service

enzoc4:
# Add separate explorer service (wip)

**NOTE:** this depends on #2378 (note the base branch)

## Start moving the *explorer* outside of the node into a subcrate.

There are still a lot of questions of what's the best way to move forward with this, so this is mostly to start somewhere. The idea was to just copy-paste the the whole API but there are some part of the indexing process that depend on the ledger/blockchain object (particularly all the vote counting). Shouldn't be that hard to extract, but maybe it's better to do it after starting with the storage.

The logging part is fairly simple to get started, but ideally we should reuse the same setup (configuration) that the node uses.
The warp/api setup should be the same too, not sure how worth it is to share it.

There are more details, but may be better to create an issue to keep track of everything (todo)


enzoc4:
I'll open a new PR more according with the new plans. (and with the current graphql api)

PR #2474, opened 2020-07-27
add a new, bare module, that uses `sled` for the persistent storage of settings

NicolasDP:
Moving toward a set of modules like this will allow us to isolate the different components of each modules from each others. But also to be able to update the internals without suffering upon more complex major migration in the future.

### DIFF
--- a/modules/settings/Cargo.toml
+++ b/modules/settings/Cargo.toml
@@ -1,0 +1,11 @@
+[package]
+name = "settings"
+version = "0.1.0"
+authors = ["Nicolas Di Prima <nicolas.diprima@iohk.io>"]
+edition = "2018"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+sled = "0.34"
+thiserror = "1.0"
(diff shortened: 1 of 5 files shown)


modules/settings/Cargo.toml
```diff
@@ -0,0 +1,11 @@
...
+name = "settings"
+version = "0.1.0"
+authors = ["Nicolas Di Prima <nicolas.diprima@iohk.io>"]
+edition = "2018"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+sled = "0.34"
+thiserror = "1.0"
```
eugene-babichenko, line 11: missing newline

modules/settings/src/lib.rs
```diff
@@ -0,0 +1,171 @@
...
+}
+
+/// The settings domain. This object allows keeping track of the
+/// encapsulation of the settings parameters and also to keep
+/// separated the different operations on different domains
+///
+/// For example, the network module will only be interested about
+/// the network settings.
+#[derive(Clone)]
+pub struct Domain {
```
eugene-babichenko, line 22: So there is domain-service correspondence, right?

NicolasDP:
I don't see how this is not possible to do such construct in here. You can still set a value for `log.terminal.level` and only set this value without touching the `log.terminal.output` (which could here be `stdin` or `stdout`)

PR #2483, opened 2020-08-01
Modules block creation

NicolasDP:
Introduce 3 new modules to expose the APIs needed for the block scheduling:

* `clock`: handy module to do all that is time management with the appropriate `tick` functions (one sync, and one async) to await for the next block schedule;
* the `kms`: to keep the keys in a separate module. Here the difference with Jörmungandr legacy's enclave is that the keys are to work in isolation from the block creation module. I.e. we can have keys in the KMS that are not used for the leadership here yet.
* the `leadership`: it is rather incomplete for now, but the idea is to have the necessary tooling for the management of "calendar" (LeadershipPlan in the code) for any new Leadership's Id (the Id used in the `kms`). It also provides a way to define how much forward in the future we want to plan to "discover" block creation. This will allow to know how much forward ti is worth to lookup, knowing that the further we will the longer it may take to explore all the schedules.

Missing:

- [ ] actually creating blocks, this is going to be yet another module from the `leadership` because it will be tied to the `mempool`. This will define how to create blocks (to prepare them for block creation)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1814,11 +1814,12 @@ dependencies = [
 name = "kms"
 version = "0.1.0"
 dependencies = [
- "chain-crypto",
+ "blockchain",
  "chain-impl-mockchain",
  "hex",
  "rand 0.7.3",
  "thiserror",
+ "tokio",
 ]
 
 [[package]]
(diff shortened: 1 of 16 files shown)


minikin:
Closing it. Feel free to reopen if we really need it.

PR #2518, opened 2020-08-31
Migrate graphql test api to graphql_client

dkijania:
So far tests were using in house custom solution for serializing and deserializing graphql request/response. 
Now we are using graphql_client.


testing/jormungandr-testing-utils/resources/explorer/graphql/lastblock.graphql
```diff
@@ -0,0 +1,14 @@
+query LastBlock{
```
mzabaluev, line 1: Are the schemas here only used for testing? Or should we try to use single graphql source?

dkijania: It's a good questions, from one side it would remove duplication and avoid keeping schema.graphql file in test resources, but i can't find any tool which allow to generate schema from sources. However getting schema out of running grapgh ql service looks like common thing (i was using: https://github.com/prisma-labs/get-graphql-schema / https://github.com/graphql-rust/graphql-client/tree/master/graphql_client_cli) and thanks to attempt to generate schema from running service we've found a bug in some vit internals.

mzabaluev: What I was really concerned about is, this is the snapshot of our actual GraphQL schema, which is only defined implicitly in the implementation sources.

It may be great to have this also as a reference to what we expect to have and test for in the schema, so good stuff then. But there is an added burden of keeping it in sync; hopefully the tests will cover any divergences that matter.
Please add the script you have used to obtain the schema from GraphQL introspection (I assume you did?).

mzabaluev (approved):

testing/jormungandr-testing-utils/resources/explorer/graphql/lastblock.graphql
dkijania: Sure, I will add generation script and test which check if schema schema is up to date

PR #2525, opened 2020-09-07
New storage integration

eugene-babichenko:
Integrate input-output-hk/chain-libs#385
Replace #2411


jormungandr-lib/src/interfaces/config/node.rs
```diff
@@ -175,8 +175,7 @@ pub struct TrustedPeer {
 
 #[derive(Debug, Clone, Deserialize, Serialize)]
 pub struct NodeConfig {
-    #[serde(skip_serializing_if = "Option::is_none")]
-    pub storage: Option<PathBuf>,
```
NicolasDP, line 179: This is a breaking change btw. Needs to label it as such and comment that this is going to change (cannot *not* have a storage)

jormungandr/src/blockchain/storage.rs
```diff
@@ -180,105 +60,109 @@ enum StreamingError {
...
     /// - `Err(CannotIterate)` - `from` is not ancestor of `to`
     /// - `Err(BlockNotFound)` - `from` or `to` was not found
     /// - `Err(_)` - some other storage error
-    pub async fn stream_from_to(
+    pub fn stream_from_to(
         &self,
-        from: HeaderHash,
-        to: HeaderHash,
+        from: &HeaderHash,
+        to: &HeaderHash,
```
mzabaluev, line 150: `HeaderHash` is `Copy` and a small fixed array, it's not necessary to pass these values by reference.

jormungandr/src/blockchain/storage.rs
```diff
@@ -180,105 +60,109 @@ enum StreamingError {
...
-                })
-            }),
-        ))
+        let distance = self
+            .storage
+            .is_ancestor(from.as_bytes(), to.as_bytes())?
+            .ok_or(Error::CannotIterate)?;
+
+        let stream = futures::stream::iter(self.storage.iter(to.as_bytes(), distance)?)
+            .fuse()
```
mzabaluev, line 158: Why do you need to fuse it? The function signature only promises an `impl Stream`.

mzabaluev (dismissed):

jormungandr/src/blockchain/storage.rs
```diff
@@ -310,246 +193,70 @@ impl Storage {
...
-                        return Ok(BlockIteration::Continue(self));
+        let mut ancestor = None;
+        let mut closest_found = std::u32::MAX;
+
+        for checkpoint in checkpoints {
+            // Checkpoints sent by a peer may not
+            // be present locally, so we need to ignore certain errors
+            match self
+                .storage
+                .is_ancestor(checkpoint.as_bytes(), descendant.as_bytes())
```
mzabaluev, line 233: Now that it is a blocking operation, what's the worst case performance for it?

I think this whole operation should be lowered into the storage implementation and optimized: can we limit it to one walk through descendant's parents in the volatile store, comparing each level's block hash to all checkpoints (possibly collected in a `HashSet`) at once?

jormungandr/src/start_up/mod.rs
```diff
@@ -7,42 +7,23 @@ use crate::{
...
-
-    Ok(Storage::new(
-        raw_block_store,
-        logger.new(o!(log::KEY_SUB_TASK => "storage")),
-    ))
+    let logger = logger.new(o!(log::KEY_SUB_TASK => "storage"));
+    if let Some(dir) = &setting.storage {
+        std::fs::create_dir_all(dir).map_err(|err| Error::IO {
+            source: err,
+            reason: ErrorKind::SQLite,
```
mzabaluev, line 18: SQLite?

eugene-babichenko: Oops

jormungandr/src/blockchain/storage.rs
eugene-babichenko: The worst performance will be O(n) where n is the number of blocks in the branch of the volatile storage. But this will depend on how often we remove older branches, because if there is no branching at the length of the given block, the performance will be O(1).

Doing what you suggested makes sense in general, but how often will we hit case with the O(n) performance?

mzabaluev: No idea, but the number of checkpoints that a node sends in a pull request is supposed to grow logarithmically with the length of the current branch in volatile store. Let's optimize this later.

mzabaluev (approved):

PR #2526, opened 2020-09-07
Use only persistent storage

eugene-babichenko:


jormungandr-lib/src/interfaces/config/node.rs
```diff
@@ -175,8 +175,7 @@ pub struct TrustedPeer {
 
 #[derive(Debug, Clone, Deserialize, Serialize)]
 pub struct NodeConfig {
-    #[serde(skip_serializing_if = "Option::is_none")]
-    pub storage: Option<PathBuf>,
+    pub storage: PathBuf,
```
eugene-babichenko, line 178: I doubt a bit about this change. Technically, specifying a storage path is not _required_ in the config file, you need to specify just somewhere.

eugene-babichenko:
No longer needed because now we can do in-memory sled storage

PR #2529, opened 2020-09-08
Add mempool notifier sub

enzoc4:
base branch is #2378 to show changes more clearly (and it would have to be merged first anyway).

### Motivation
To be able to know about fragment rejection or acceptance asynchronously.

### Summary
This basically sends a message each time a fragment status is modified in the logs, to anyone with an open connection on the `api/v1/notifier/mempool` endpoint. For now rejection reason is not added, only the id.

This also moves the previous endpoint from `api/v1/notifer` to `api/v1/notifier/blocks`.

### Other
TODO: update the tests
_To think about_: graphql has subscriptions which can be used with websockets, so we could have a little schema instead of multiple endpoints. I'm not sure if it's worth it, especially considering that juniper doesn't have them finished.


jormungandr/src/notifier/mod.rs
```diff
@@ -72,22 +89,60 @@ impl Notifier {
...
                     }
-                    Message::NewConnection(ws) => {
-                        trace!(logger, "processing notifier new connection");
+                    Message::FragmentRejected(fragment_id) => {
+                        info.spawn("notifier broadcast fragment rejection", async move {
+                            if let Err(_err) =
+                                mempool_sender.send(MemPoolMessage::FragmentRejected(fragment_id))
+                            {
+                                ()
+                            }
```
danielSanchezQ, line 98: I don't really understand the purpose of the `if let` block here. Can't we use `map_err` or something else?

PR #2541, opened 2020-09-16
Fragments pool module

eugene-babichenko:
This module is responsible only for managing the pool of fragments and
doing fragments selection.

Fragments log is not included here because given its current use it
should be an integral part of the REST API module.


modules/pool/src/lib.rs
```diff
@@ -0,0 +1,155 @@
...
+        while let Some((hash, fragment)) = self.pool.pop_lru() {
+            match ledger.apply_fragment(&ledger_params, &fragment, block_date) {
+                Ok(new_ledger) => {
+                    ledger = new_ledger;
+                    selected.push(fragment);
+                }
+                Err(reason) => rejected.push(RejectedFragmentSelection { id: hash, reason }),
+            }
+
+            if selected.len() as u32 >= fragments_number_max {
```
enzoc4, line 119: is there a reason this is `>=` instead of `==`?

enzoc4 (commented):
just out of curiosity, why do we want this to be a separate crate?

eugene-babichenko:
@enzoc4 

> just out of curiosity, why do we want this to be a separate crate?

The basic idea is draw very precise boundaries between modules so that we are not tempted to use shared variables in the future :) Basically, we want to reach the maximum possible isolation of actors in the future and to stop having too much async code with mutexes.

mzabaluev:
Crate separation should be done for purposes of reuse in different end executables and managing outlier dependencies that other functionality does not require. We definitely need this if we want to build multiple binaries in the same project that incorporate different subsets of node functionality. Otherwise, it could be internal modules selected at compile time with feature flags. API boundaries could be defined conservatively with both approaches.

modules/pool/src/lib.rs
mzabaluev: Just a belt-and-suspenders approach :grinning: I also do it instinctively as a safeguard that would prevent the code from continued damage on counter overruns and doesn't make a difference if the functionality is correct.

enzoc4:
I can't think of a different kind of binary that could use a fragment pool, unless we need it for testing or something? And to be honest I think modules with adequate visibility qualifiers should be enough for the stated purpose, but maybe I'm missing something.

mzabaluev:
We could audit for misuse of `pub(crate)`, but there are legitimate uses for it, so it won't be done rigorously.

minikin:

PR #2546, opened 2020-09-21
fix schedule computation

eugene-babichenko:


mzabaluev (approved):
Looks good, at least if what the `Schedule::new` argument `nb_slots` is meant to be the total number of slots.

There should have been a test to validate this when slot_start is not 0.
Also many tests are failing in CI, please check if this is the cause.

eugene-babichenko:
Fixed that.

PR #2548, opened 2020-09-22
Node ID handshake

### LINKED ISSUE #2517: Node identification in handshake to deduplicate propagation of messages
To prevent redundant ~p2p connections~ propagation of messages between nodes, a form of node identification is needed.

Each node generates a key pair and uses the public key as the node ID.

After the HTTP connection is established,

1. The gRPC client node sends a randomly generated nonce in the `Handshake` request.
2. The server node's response includes its node ID (i.e. the public key) and a cryptographic signature of the nonce using the private key. The server also sends its own generated nonce with the response.
3. Optionally, the client sends a `ClientAuth` request with its node ID (public key) and a signature of the server's nonce with the client's private key. Non-public nodes do not need this step.
4. After these requests and responses above are completed, the nodes have proof of peer's identity associated with connection addresses. Propagation of messages can avoid redundant connections with the same node ID. Unused connections get rotated out as usual, however, there is no requirement to close duplicate connections established in opposite directions, as the result will often be that the peers will drop both connections and there is no fair deterministic way to choose which one of the connections to drop on both sides.

This scheme does not rely on encryption of the HTTP connection, which can be added independently of this feature. As such, this handshake-based identification scheme is only meant to detect redundant connections, while preventing trivial spoof attacks.

mzabaluev:
Resolves #2517 
Requires: https://github.com/input-output-hk/chain-libs/pull/445


dkijania (approved):

jormungandr/src/network/mod.rs
```diff
@@ -161,7 +167,9 @@ impl GlobalState {
 
         let mut rng_seed = [0; 32];
         rand::thread_rng().fill(&mut rng_seed);
-        let prng = ChaChaRng::from_seed(rng_seed);
+        let mut prng = ChaChaRng::from_seed(rng_seed);
+
+        let keypair = NodeKeyPair::generate(&mut prng);
```
danielSanchezQ, line 172: Why do we use a seed?

mzabaluev: This is probably overkill (`ThreadRng` is a `CryptoRng` anyway), but the initialization of a secure RNG from seed was there for Poldercast. This key generation call just reuses the same RNG.

mzabaluev: I wrote it too fast: `CryptoRng` is not the trait that ensures unpredictability, but `ThreadRng` delivers that as well.

It may be useful, however, to have a single seeding point that can be overridden from the environment for test reproducibility purposes.

danielSanchezQ: I'm not sure I get it, but having a fixated seed pings something in my mind. I understand that for testing would be interesting. But maybe it would be safer to use `ChaChaRng::from_entropy`?

mzabaluev: I believe this is what the `ThreadRng` does under the hood when initializing. I can change this to streamline the code, `from_entropy` should be searchable enough to change to optionally deterministic seeding later.

PR #2552, opened 2020-09-24
[Tests] Vote e2e test

dkijania:
Added two tests for vote e2e flow, for bft and genesis_praos.


danielSanchezQ:
Awesome!
Will we be able to add the private voting payload and features soon? It would be awesome for checking that everything works as expected.

dkijania:
Yes, it will be tested in similar manner.

PR #2554, opened 2020-09-24
Jcli encrypt vote tools

danielSanchezQ:
Added 3 commands to jcli to generate vote committee related keys.

### DIFF
--- a/jcli/src/jcli_app/mod.rs
+++ b/jcli/src/jcli_app/mod.rs
@@ -6,6 +6,7 @@ mod debug;
 mod key;
 mod rest;
 mod transaction;
+mod vote;
 
 pub mod utils;
 
@@ -52,6 +53,8 @@ pub enum JCliCommand {
     AutoCompletion(auto_completion::AutoCompletion),
     /// Utilities that perform specialized tasks
     Utils(utils::Utils),
+    /// Vote related operations
+    Votes(vote::Vote),
 }
 
 impl JCli {
@@ -83,6 +86,7 @@ impl JCliCommand {
             Certificate(certificate) => certificate.exec()?,
             AutoCompletion(auto_completion) => auto_completion.exec::<Self>()?,
             Utils(utils) => utils.exec()?,
+            Votes(vote) => vote.exec().map_err(Box::new)?,
         };
         Ok(())
     }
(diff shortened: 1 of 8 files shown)


jcli/src/jcli_app/vote/committee_communication_key.rs
```diff
@@ -0,0 +1,26 @@
...
+
+#[derive(StructOpt)]
+#[structopt(rename_all = "kebab-case")]
+pub struct CommitteeCommunicationKey {}
+
+impl CommitteeCommunicationKey {
+    pub fn exec(&self) -> Result<(), super::Error> {
+        let key = generate_committee_member_keys();
+        println!("Private: {}", hex::encode(key.to_bytes()));
+        println!("Public: {}", hex::encode(key.to_public().to_bytes()));
```
danielSanchezQ, line 18: I went for the shortest path here. But I'm not sure if what is the standard we use for outputting data. Will check it too.

jcli/src/jcli_app/vote/committee_member_key.rs
```diff
@@ -0,0 +1,68 @@
...
+
+    #[test]
+    fn test_keys_dance() {
+        let mut rng = ChaCha20Rng::from_seed([0u8; 32]);
+        let mck = MemberCommunicationKey::new(&mut rng);
+        let pk_bytes = mck.to_public().to_bytes();
+        let pk = PublicKey::from_bytes(&pk_bytes).unwrap();
+        MemberCommunicationPublicKey::from_public_key(pk);
+    }
+}
```
danielSanchezQ, line 68: Ups, this should be removed...

jcli/src/jcli_app/mod.rs
```diff
@@ -52,6 +53,8 @@ pub enum JCliCommand {
     AutoCompletion(auto_completion::AutoCompletion),
     /// Utilities that perform specialized tasks
     Utils(utils::Utils),
+    /// Vote related operations
+    Votes(vote::Vote),
```
enzoc4, line 57: why not put the key generation under `Key`?

danielSanchezQ: Actually yes, I didn't even think about it, I just wanted to keep all voting-related tools together. But I'm down for the key option too.

mzabaluev:
Superseded by #2603.

PR #2556, opened 2020-09-24
Storage GC

eugene-babichenko:
Needs https://github.com/input-output-hk/chain-libs/pull/453

### DIFF
--- a/jormungandr/src/blockchain/chain.rs
+++ b/jormungandr/src/blockchain/chain.rs
@@ -299,12 +299,9 @@ impl Blockchain {
     }
 
     pub async fn gc(&self, tip: Arc<Ref>) -> Result<()> {
-        self.ledgers.gc().await;
-        self.storage.gc(
-            tip.epoch_ledger_parameters().epoch_stability_depth,
-            tip.hash().as_ref(),
-        )?;
-
+        let depth = tip.epoch_ledger_parameters().epoch_stability_depth;
+        self.ledgers.gc(depth).await;
+        self.storage.gc(depth, tip.hash().as_ref())?;
         Ok(())
     }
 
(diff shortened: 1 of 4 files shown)


eugene-babichenko:
Need to merge https://github.com/input-output-hk/jormungandr/pull/2548 first. It contains integration of the latest chain-libs master.

jormungandr/src/blockchain/chain.rs
```diff
@@ -298,6 +298,13 @@ impl Blockchain {
         &mut self.branches
     }
 
+    pub async fn gc(&self, tip: Arc<Ref>) -> Result<()> {
+        let depth = tip.epoch_ledger_parameters().epoch_stability_depth;
+        self.ledgers.gc(depth).await;
+        self.storage.gc(depth, tip.hash().as_ref())?;
```
mzabaluev, line 304: What's the effect if the storage GC operation fails? Is it bad to have storage branches that are not attested in the ledger multiverse? May need to swap these two.

jormungandr/src/blockchain/storage.rs
```diff
@@ -259,4 +261,53 @@ impl Storage {
...
+            .flush_to_permanent_store(to_block_info.id().as_ref(), MINIMUM_BLOCKS_TO_FLUSH)?;
+
+        debug!(
+            self.logger,
+            "flushed all blocks ({}) up to {} to the permanent store",
+            blocks_flushed,
+            HeaderHash::hash_bytes(to_block_info.id().as_ref())
+        );
+
+        Ok(())
```
mzabaluev, line 311: This is beginning to look like quite a lot to do while blocking an async task. We should be careful to not lock the executor while waiting for something.

Issue #2590, opened 2020-10-12
Disable legacy compatibility scenario tests involving pre-0.10 nodes


mzabaluev:
As network protocol compatibility was broken with #2548 and we don't commit to supporting node versions < 0.10 and >= 0.10 in the same network, the legacy compatibility tests should be disabled by default in jormungandr-scenario-tests.

PR #2597, opened 2020-10-13
[Tests] Move soak tests and legacy connection tests to unstable

dkijania:
- [x] move all legacy connection tests from scenario tests to unstable
- [x] remove all legacy nodes from scenario soak tests to unstable


dkijania:
fixes #2590

mzabaluev (approved):

PR #2591, opened 2020-10-12
jcli: vote decryption shares

eugene-babichenko:
needs https://github.com/input-output-hk/chain-libs/pull/465

PR #2608, opened 2020-10-17
jcli - add votes tally-decryption-share

rinor:
https://github.com/input-output-hk/jormungandr/pull/2591 cleaned and updated

### DIFF
--- a/jcli/src/jcli_app/vote/mod.rs
+++ b/jcli/src/jcli_app/vote/mod.rs
@@ -1,6 +1,7 @@
 use crate::jcli_app::utils::output_file::{self, OutputFile};
 mod committee;
 mod common_reference_string;
+mod decryption_share;
 mod encrypting_vote_key;
 
 use structopt::StructOpt;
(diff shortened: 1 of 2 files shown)


jcli/src/jcli_app/vote/mod.rs
```diff
@@ -31,6 +32,14 @@ pub enum Error {
...
     #[error("invalid committee member index")]
     InvalidCommitteMemberIndex,
+    #[error("failed to read encrypted tally bytes")]
+    EncryptedTallyRead,
+    #[error("failed to read decryption key bytes")]
+    DecryptionKeyRead,
+    #[error(transparent)]
+    FormatError(#[from] crate::jcli_app::utils::output_format::Error),
+    #[error(transparent)]
+    JsonError(#[from] serde_json::Error),
```
mzabaluev, line 42: This and other variants like `Io`, and `Hex` make the code easier to write, but will probably provide poor UX because they don't really tell which argument is at fault or where the problem has occurred.

mzabaluev (approved):

PR #2616, opened 2020-10-19
[Tests] Rest cors tests

dkijania:
Tests for CORS setting in jormungandr node


dkijania:
Automation for issue: #2618

eugene-babichenko:
@dkijania since the issue was fixed could you try rebasing this? I think this one should be good to merge.

PR #2624, opened 2020-10-21
Jcli tally decrypt with shares tool

danielSanchezQ:


mzabaluev (commented):
How about shortening these and unifying under a common subcommand:

```
jcli vote tally decryption-share ...
jcli vote tally decrypt ...
```

jcli/src/jcli_app/vote/tally/decryption_tally.rs
```diff
@@ -9,7 +9,7 @@ use structopt::StructOpt;
 /// The outputs are provided as hex-encoded byte sequences.
 #[derive(StructOpt)]
 #[structopt(rename_all = "kebab-case")]
```
mzabaluev, line 11: Is this `StructOpt` impl now unnecessary?

danielSanchezQ: It is necessary for the composition of the commands otherwise it won't compile. Or I misunderstand want you meant...

mzabaluev: My bad, I thought `kebab-case` concerns the command names themselves, but this is for options/subcommands.

mzabaluev (approved):

PR #2627, opened 2020-10-22
fix encrypted tally certs and stats

rinor:


jormungandr-lib/src/interfaces/certificate.rs
```diff
@@ -141,6 +141,12 @@ impl Readable for Certificate {
                 let cert = certificate::VoteTally::read(buf)?;
                 Ok(Certificate(certificate::Certificate::VoteTally(cert)))
             }
+            9 => {
+                let cert = certificate::EncryptedVoteTally::read(buf)?;
+                Ok(Certificate(certificate::Certificate::EncryptedVoteTally(
+                    cert,
+                )))
+            }
```
danielSanchezQ, line 149: And this is why those ids are bad. Rust does not complain if it doesn't find them.

danielSanchezQ (approved):
LGTM
Let's merge when tests show green

PR #2630, opened 2020-10-22
Serde encrypted tally as hex string

danielSanchezQ:


jormungandr-lib/src/interfaces/vote/mod.rs
```diff
@@ -330,7 +330,49 @@ impl TallyResult {
 }
 
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
-pub struct EncryptedTally(Vec<u8>);
+pub struct EncryptedTally(#[serde(with = "serde_bytes")] Vec<u8>);
+
+mod serde_hex_bytes {
```
mzabaluev: I don't get it. The attribute says use `serde_bytes` module, what is `mod serde_hex_bytes` for?

Also, this does not distinguish between human-readable and non-human-readable serde backends. For the latter, the serialization should still use a byte vector. We probably don't use any at the moment, but it will be embarrassing if we do and then find out that it uses a wasteful hex encoding.

danielSanchezQ: I did not know it can be discriminated. I'll check about it.

jormungandr-lib/src/interfaces/vote/mod.rs
```diff
@@ -330,7 +330,67 @@ impl TallyResult {
...
+    where
+        D: Deserializer<'de>,
+    {
+        struct ByteStringVisitor;
+        impl<'de> Visitor<'de> for ByteStringVisitor {
+            type Value = Vec<u8>;
+
+            fn expecting(&self, formatter: &mut std::fmt::Formatter) -> std::fmt::Result {
+                formatter.write_str("Invalid hex encoded encrypted tally")
+            }
```
mzabaluev, line 350: The message will be misleading for binary backends. It should also read like a part of an externally formatted sentence, see [the doc](https://docs.rs/serde/1.0.117/serde/de/trait.Visitor.html#tymethod.expecting)

mzabaluev: I'd split this trivially into a binary and string visitor implementing only the respective methods, like in my example.

mzabaluev (changes requested):

jormungandr-lib/src/interfaces/vote/mod.rs
mzabaluev: Perhaps we should make this serde helper module reused throughout the lib, if we have other byte vector fields that are better represented in hex.

danielSanchezQ: Yeah, it would be nice to have a serde sub_module, or sub_crate with this kind of stuff.

mzabaluev (approved):

PR #2639, opened 2020-10-27
Expose total votes in voteplanstatus

danielSanchezQ:


jormungandr-lib/src/interfaces/vote/mod.rs
```diff
@@ -523,6 +524,7 @@ impl From<vote::VoteProposalStatus> for VoteProposalStatus {
             proposal_id: this.proposal_id.into(),
             options: this.options.choice_range().clone(),
             tally: this.tally.map(|t| t.into()),
+            total_votes: this.votes.size(),
             votes: this
                 .votes
                 .iter()
```
mzabaluev, line 530: Do we need to add a field for something that can be derived from the size of the `votes` table?

danielSanchezQ: Yeah, I thought about it. But It felt easier and straight forward to use from a field.
Let's close the PR otherwise :)

mzabaluev: In fact, why do we expose the `votes` table here at all? It's not a scalable API if we expect hundreds or thousands of voters.

mzabaluev (approved):

PR #2649, opened 2020-10-29
Implemented private vote cast jcli command

danielSanchezQ:


jcli/src/jcli_app/certificate/new_vote_cast.rs
```diff
@@ -3,42 +3,94 @@ use chain_impl_mockchain::{
...
     /// the vote plan identified on the blockchain
-    pub vote_plan_id: VotePlanId,
+    vote_plan_id: VotePlanId,
 
     /// the number of proposal in the vote plan you vote for
-    pub proposal_index: u8,
+    proposal_index: u8,
+
+    /// size of voting options
+    options: usize,
```
mzabaluev, line 35: Should this be retrievable from the proposal status endpoint in REST?

jcli/src/jcli_app/certificate/new_vote_cast.rs
```diff
@@ -3,42 +3,94 @@ use chain_impl_mockchain::{
...
         let cert = Certificate::VoteCast(vote_cast);
         write_cert(self.output.as_deref(), cert.into())
     }
 }
+
+impl PrivateVoteCast {
+    pub fn exec(self) -> Result<(), Error> {
+        let mut rng = rand_chacha::ChaChaRng::from_entropy();
+
+        let key_bin = base64::decode(self.encrypting_key).map_err(Error::Base64)?;
```
mzabaluev, line 71: Is there a convention on which format these keys should be encoded with?
We use bech32 for stake keys, maybe the voting keys should follow suit?

jcli/src/jcli_app/certificate/new_vote_cast.rs
danielSanchezQ: It is though

jcli/src/jcli_app/certificate/new_vote_cast.rs
danielSanchezQ: Not sure about compatibilities nor conventions though. I'll take a look.

mzabaluev: base64 should be good for the command line on Unix, but `+` and `/` have special meaning on Windows.
Do we already use it anywhere else?

mzabaluev: If in doubt, use hex.

danielSanchezQ: We can use the urlsafe mode for that I think. But if we are using the other format (bech32) in other keys lets try to keep it simple and use that format for all of them.

danielSanchezQ:
Waiting for #2655 for updating.

jcli/src/jcli_app/certificate/new_vote_cast.rs
```diff
@@ -1,44 +1,106 @@
...
-    pub vote_plan_id: VotePlanId,
+    vote_plan_id: VotePlanId,
 
     /// the number of proposal in the vote plan you vote for
-    pub proposal_index: u8,
+    proposal_index: u8,
 
     /// the number of choice within the proposal you vote for
-    pub choice: u8,
+    choice: u8,
```
mzabaluev, line 21: Let's give these `structopt` annotation to make them options instead of positional parameters?
Or maybe just the first: `--vote-plan-id=foo 5 1` is mnemonic enough.

jcli/src/jcli_app/certificate/new_vote_cast.rs
```diff
@@ -1,44 +1,106 @@
...
+    proposal_index: u8,
+
+    /// size of voting options
+    options: usize,
+
+    /// the number of choice within the proposal you vote for
+    choice: u8,
+
+    /// key to encrypt the vote with
+    encrypting_key_path: Option<PathBuf>,
```
mzabaluev, line 43: Here too, I think, all fields should be options except the `<proposal_index> <choice>` 2D mnemonic.

We can leave these for a later bikeshedding PR if it's too much to do here; I've got some ideas :smiling_imp:

mzabaluev (approved):
Functionality looks good. I will submit a reorg of commands and options in another PR.

mzabaluev (approved):

PR #2650, opened 2020-10-29
Handle SIGTERM on Unix

mzabaluev:


jormungandr/src/main.rs
```diff
@@ -279,7 +281,23 @@ fn start_services(bootstrapped_node: BootstrappedNode) -> Result<(), start_up::E
...
+
+        let signal = match signal::unix::signal(SignalKind::terminate()) {
+            Ok(signal) => signal,
+            Err(e) => {
+                warn!(info.logger(), "failed to install handler for SIGTERM"; "reason" => %e);
+                return future::ready(()).left_future();
+            }
+        };
+        signal.into_future().map(|_| ()).right_future()
+    });
```
danielSanchezQ, line 300: Just out of curiosity. What happens with other platforms?

mzabaluev: With what's available in tokio, "the other platforms" means Windows, and there we can only additionally handle Ctrl+Break. I'm not sure this is actually needed by anyone.

`SIGTERM` is the conventional way to gracefully stop a daemon on Unix, used by IOHK infrastructure as well.

danielSanchezQ (approved):
LGTM

PR #2651, opened 2020-10-29
Add missing error handling operator in serialization of member public keys

mzabaluev:


jormungandr-lib/src/interfaces/vote/mod.rs
```diff
@@ -209,7 +209,7 @@ mod serde_committee_member_public_keys {
     {
         let mut seq = serializer.serialize_seq(Some(keys.len()))?;
         for key in keys {
-            seq.serialize_element(&SerdeMemberPublicKey(key.clone()));
+            seq.serialize_element(&SerdeMemberPublicKey(key.clone()))?;
```
danielSanchezQ, line 212: This is one of the problems with warnings. That they are lost in the sea.

danielSanchezQ (approved):
Good catch!

PR #2655, opened 2020-10-30
Bech32 vote related keys

danielSanchezQ:


jormungandr-lib/src/interfaces/mod.rs
```diff
@@ -63,4 +63,5 @@ pub use self::utxo_info::{UTxOInfo, UTxOOutputInfo};
 pub use self::value::{Value, ValueDef};
 pub use self::vote::{
     Payload, Tally, TallyResult, VotePlanDef, VotePlanStatus, VoteProposalStatus,
+    MEMBER_PUBLIC_KEY_BECH32_HRP,
```
mzabaluev, line 66: I think this name is less ambiguous if namespaced under `interfaces::vote` without the reexports.

But then, the same applies to `Payload` and maybe `Tally`.

jormungandr-lib/src/interfaces/vote/mod.rs
```diff
@@ -31,18 +34,20 @@ enum PayloadTypeDef {
...
-        struct PublicKeyVisitor;
-        impl<'de> Visitor<'de> for PublicKeyVisitor {
+        struct Bech32Visitor;
+        impl<'de> Visitor<'de> for Bech32Visitor {
             type Value = SerdeMemberPublicKey;
 
             fn expecting(&self, formatter: &mut std::fmt::Formatter) -> std::fmt::Result {
                 formatter
-                    .write_str("Expected a compatible hex representation of required public key")
+                    .write_str("Expected a compatible bech32 representation of required public key")
```
mzabaluev, line 50: Remove "Expected ", it will result in a bad error message. Serde API wants this string to complete the phrase "Expecting... "

jcli/src/jcli_app/vote/committee/communication_key.rs
```diff
@@ -9,6 +10,9 @@ use std::{
 };
 use structopt::StructOpt;
 
+pub const COMMUNICATION_SK_HRP: &str = "p256k1_communicationsk";
+pub const COMMUNICATION_PK_HRP: &str = "p256k1_communicationpk";
```
mzabaluev, line 14: Let's make these prefixes shorter, maybe `p256k1_vcommpk`/`p256k1_vcommsk` (as in voting committee/communication).

jormungandr-lib/src/interfaces/mod.rs
mzabaluev: Another style idea is to collect all HRPs under the same mod namespace structure with clear naming logic, like

```
crate::bech32::vote::MEMBER_PUBLIC_KEY_HRP
crate::bech32::vote::COMMUNICATION_PUBLIC_KEY_HRP
```

danielSanchezQ: Yeah, I like that.

mzabaluev (approved):

PR #2664, opened 2020-11-02
Jormungandr settings module cleanup

eugene-babichenko:


eugene-babichenko:
On top of that I am thinking of a more complicated thing: replace optional fields that are later unwrapped with non-optional fields and `#[serde(default = "default_field_value")]`. This can be nice when the default value of a field is a constant.

mzabaluev (approved):

mzabaluev:
> On top of that I am thinking of a more complicated thing: replace optional fields that are later unwrapped with non-optional fields and `#[serde(default = "default_field_value")]`.

Makes sense.

eugene-babichenko:
I will do that separately from cleaning up the simpler stuff then. It might grow big since the `default` attribute requires a function instead of a value.

PR #2695, opened 2020-11-09
Logging improvements for network client

mzabaluev:
Logging improvements:

* Deduplicate peer information, now should only log peer multiaddr.
* Log block solicitations at info level including the hash of the first block.

One behavioral change is that an empty block solicitation is now logged and then the client gets disconnected, quarantining the server.


danielSanchezQ (approved):
LGTM
We can merge both Clippy and the failing test doesn't deal with this and are being investigated.

PR #2709, opened 2020-11-12
Add block downloader service implementation

eugene-babichenko:
Added the service to download blocks from multiple peers simultaneously. This
service is designed for faster syncronization after the network connection
loss or during the bootstrap.

The tests currently run for a very long time (hours) because the `Arbitrary`
implementation for `Block` fills it with an arbitrary number of fragments. This
implementation was tested before on a dummy type, but it envolved too many
practically useless generalization.

Things remaining to be done:

- Faster testing.
- Proper error processing in `block_downloader_task`.

Depends on https://github.com/input-output-hk/chain-libs/pull/485 for sensible test times.

This service will be used as a single point that would:

- Download blocks during bootstrap.
- Do the same thing during the node run.
- Use the existing Jormungandr service module to spawn tasks.
- Configure backpressure handling.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -207,7 +207,7 @@ dependencies = [
  "proc-macro-crate",
  "proc-macro2 1.0.24",
  "quote 1.0.7",
- "syn 1.0.51",
+ "syn 1.0.53",
 ]
 
 [[package]]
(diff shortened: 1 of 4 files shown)


minikin:

PR #2710, opened 2020-11-13
Update to parity-multiaddr

mzabaluev:
Temporarily use an unmerged branch of poldercast that integrates parity-multiaddr instead of multiaddr.
Also replace the jormungandr dependency the same way and update to the API changes.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -308,12 +308,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "base-x"
-version = "0.2.7"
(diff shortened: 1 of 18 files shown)


danielSanchezQ (approved):
Nice update. Everything green again ;)
Thanks!

mzabaluev:
We're effectively forking poldercast for our purposes, though.
If @NicolasDP will not roll it up into a release, we'll pirate the crate :pirate_flag:

PR #2733, opened 2020-11-19
jcli: allow to provide a custom TLS root certificate for REST API

eugene-babichenko:

PR #2734, opened 2020-11-19
doc: update TLS docs with X509 v3 and CA usage

eugene-babichenko:


mzabaluev (approved):

mzabaluev:
Any idea what was missing in the `openssl` commands to get a correct X509 v3 certificate?

eugene-babichenko:
`openssl` uses v1 unless you explicitly specify v3 and all required extensions. Actually, you need to that twice: for a CSR and for an actual certificate (this is considered to be a bug by the docs BTW). Plus we need to make a root certificate + a server certificate signed by the root key.
I decided to go with `easyrsa` because using bare `openssl` commands produces a lot of boilerplate code which I think is not needed for the demonstration purposes.

dkijania (changes requested):

eugene-babichenko:
Waiting for #2733. These docs contain a piece of documentation that specifically refers new features.

PR #2738, opened 2020-11-20
gracefully stop bootstrap on sigterm and support SIGINT

### LINKED ISSUE #2654: Cancellable bootstrap
With #2650 we intercept graceful shutdown signals before bootstrap starts, but the handlers defer shutdown until the node is fully bootstrapped.

It would be useful to pass a [cancellation handle](https://docs.rs/tokio-util/0.4.0/tokio_util/sync/struct.CancellationToken.html) and intersperse the bootstrap pull with checks for cancellation that may be triggered by the termination signal watcher.

To be deferred until #2620 is done.

eugene-babichenko:
Resolves #2654

### DIFF
--- a/jormungandr/src/main.rs
+++ b/jormungandr/src/main.rs
@@ -32,6 +32,7 @@ use jormungandr_lib::interfaces::NodeState;
 use settings::{start::RawSettings, CommandLine};
 use slog::Logger;
 use tokio::signal;
+use tokio_util::sync::CancellationToken;
 
 use std::sync::Arc;
 use std::time::Duration;
@@ -318,6 +319,7 @@ fn bootstrap(initialized_node: InitializedNode) -> Result<BootstrappedNode, star
         logger,
         rest_context,
         mut services,
+        cancellation_token,
     } = initialized_node;
 
     let BootstrapData {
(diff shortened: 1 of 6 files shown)


jormungandr/src/main.rs
```diff
@@ -544,10 +537,8 @@ fn initialize_node() -> Result<InitializedNode, start_up::Error> {
     let settings = raw_settings.try_into_settings(&init_logger)?;
     let mut services = Services::new(logger.clone());
 
-    services.spawn_try_future("sigint_watcher", move |_info| {
-        signal::ctrl_c().map_err(|_| ())
-    });
```
mzabaluev, line 549: We're losing Ctrl+Break handling on Windows here. I think we should handle it too, but move this watcher into the non-Unix implementation of `init_os_signal_watchers`.

mzabaluev (commented):
Looks great, one minor thing to do about Ctrl-C on Windows.

jormungandr/src/main.rs
eugene-babichenko: By the way, it is starting to look like a good time to migrate to GitHub Actions entirely and do testing on Windows too.

mzabaluev (approved):

PR #2739, opened 2020-11-20
error immediately on empty trusted_peers list

### LINKED ISSUE #2662: Meaningless bootstrap retries when trusted peers is empty
```
Nov 02 14:48:41.686 ERRO trusted-peers cannot be empty. to avoid bootstrap use 'skip_bootstrap: true', task: bootstrap
Nov 02 14:48:41.686 INFO bootstrap attempt #1 failed, trying again in 5 seconds..., task: bootstrap
```

This retry attempt doesn't make any sense because it will fail for all retries anyway.

eugene-babichenko:
Resolves #2662

### DIFF
--- a/jormungandr/src/main.rs
+++ b/jormungandr/src/main.rs
@@ -542,6 +542,11 @@ fn initialize_node() -> Result<InitializedNode, start_up::Error> {
     debug!(init_logger, "system settings are: {}", diagnostic);
 
     let settings = raw_settings.try_into_settings(&init_logger)?;
+
+    if settings.network.trusted_peers.is_empty() && !settings.network.skip_bootstrap {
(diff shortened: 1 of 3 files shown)


eugene-babichenko:
Side note: this will load the blockchain from the storage first and then will error. Is this fine as is, or should we detect this error right after we load the settings?

mzabaluev (approved):

mzabaluev:
> Is this fine as is, or should we detect this error right after we load the settings?

It's an improvement already, but if it's not hard to move this check to before the storage is loaded, feel free to do so.
On the other hand, storage init is supposed to be fast now, so I'm not sure there will be much practical difference.

Issue #2623, opened 2020-10-21
Implement explorer for private votes


mzabaluev:
If we need to retain an explorer subsystem in jormungandr, the code retrieving information about private vote payloads needs to be completed.

_Originally posted by @mzabaluev in https://github.com/input-output-hk/jormungandr/pull/2571#r509194847_

PR #2746, opened 2020-11-23
Explorer private vote

danielSanchezQ:
Part of #2623 
Needs https://github.com/input-output-hk/chain-libs/pull/493

### DIFF
--- a/jormungandr/src/explorer/indexing.rs
+++ b/jormungandr/src/explorer/indexing.rs
@@ -14,7 +14,9 @@ use chain_impl_mockchain::certificate::{
 use chain_impl_mockchain::key::BftLeaderId;
 use chain_impl_mockchain::transaction::{InputEnum, TransactionSlice, Witness};
 use chain_impl_mockchain::value::Value;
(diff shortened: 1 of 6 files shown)


jormungandr/src/explorer/indexing.rs
```diff
@@ -104,12 +106,21 @@ pub struct ExplorerVotePlan {
...
+        encrypted_vote: EncryptedVote,
+    },
+}
+
 #[derive(Clone)]
 pub struct ExplorerVoteProposal {
     pub proposal_id: ExternalProposalId,
     pub options: Options,
     pub tally: Option<ExplorerVoteTally>,
-    pub votes: Hamt<ExplorerAddress, Choice>,
```
mzabaluev, line 112: This is yet another non-scalable votes member, should be converted to a paginated connection.

danielSanchezQ: Makes sense. Will do.

mzabaluev: Filed as #2747.

mzabaluev (approved):

PR #2752, opened 2020-11-24
Convert from chain-deps submodule to git dependencies on chain-libs crates

### LINKED ISSUE #2659: Convert chain-deps to git dependencies
Managing chain-libs as a submodule does not scale: now that both jormungandr and chain-wallet-libs do it, any crate that uses both, such as `iapyx`, make a mess of the dependency tree with duplicate instances of chain-libs crates.

The solution is to convert chain-libs dependencies from path dependencies from a submodule to git dependencies pointing at the chain-libs git repository, and removing the submodule.

mzabaluev:
Deduplicate chain-deps in the dependency graph, reduce `Cargo.lock` churn.

Needs https://github.com/input-output-hk/chain-wallet-libs/pull/125

Resolves #2659

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -114,8 +114,10 @@ jobs:
         name: Cache cargo dependencies
         uses: actions/cache@v2
         with:
-          path: ~/.cargo/registry/cache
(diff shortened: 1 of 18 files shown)


eugene-babichenko (approved):
Reducing the of `Cargo.toml` is also great :)

PR #2760, opened 2020-11-27
Rework the bootstrap procedure

eugene-babichenko:
- Use `Tip` combined with `PullBlocks` instead of using `PullBlocksToTip`.
- Repeat the bootstrap procedure until catching with the current chain.
- Run the GC procedure after bootstrapping so that older blocks immediately appear in the permanent storage instead of waiting for the next "regular" GC run.

This establishes the basis for further improvements. In particular, integrating parallel block pulls logic (#2709).

### DIFF
--- a/jormungandr/src/network/mod.rs
+++ b/jormungandr/src/network/mod.rs
@@ -706,6 +706,11 @@ pub async fn bootstrap(
         }
     }
 
+    blockchain
+        .gc(branch.get_ref().await)
+        .await
+        .map_err(bootstrap::Error::GcFailed)?;
+
     Ok(bootstrapped)
 }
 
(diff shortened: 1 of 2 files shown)


jormungandr/src/network/mod.rs
```diff
@@ -706,6 +706,11 @@ pub async fn bootstrap(
         }
     }
 
+    blockchain
+        .gc(branch.get_ref().await)
+        .await
```
mzabaluev, line 711: Is this meant to get the storage into steady state, so that periodic GC runs only do a little work with small depth to prune?

mzabaluev (approved):

jormungandr/src/network/mod.rs
eugene-babichenko: Exactly. I thought this is better to do before we actually start the node operation.

PR #2763, opened 2020-11-30
Document vote tallying

mzabaluev:


danielSanchezQ (approved):
We can fill the gaps from here.
Failing tests are completely non related so it should be safe to merge.

PR #2766, opened 2020-11-30
run tests in GitHub actions

eugene-babichenko:


.github/workflows/test.yml
```diff
@@ -0,0 +1,185 @@
+name: CI
+
+on:
+  push:
+    branches:
+      - master
+      - ci/test/nightly
```
mzabaluev, line 7: No need for this match, that branch is used to test release builds.

.github/workflows/test.yml
```diff
@@ -0,0 +1,185 @@
...
+    needs: cache_info
+    # Caches on Windows and Unix do not interop:
+    # https://github.com/actions/cache/issues/330#issuecomment-637701649
+    strategy:
+      matrix:
+        os: [ubuntu-latest, windows-latest]
+    runs-on: ${{ matrix.os }}
+    env:
+      RUSTFLAGS: -D warnings
+      RUST_BACKTRACE: 1
```
mzabaluev, line 53: These environment variables look out of place here. I'd expect them to be in the `tests` job, however that one does not set them.

.github/workflows/test.yml
```diff
@@ -0,0 +1,185 @@
...
+        os: [windows-latest, ubuntu-latest, macos-latest]
+        flags: ['', '--release']
+        toolchain: [stable, nightly]
+    steps:
+      - uses: actions-rs/toolchain@v1
+        with:
+          toolchain: ${{ matrix.toolchain }}
+          override: true
+          default: true
+          components: rustfmt
```
mzabaluev, line 96: rustfmt shouldn't be needed if you build jormungandr with `--no-default-features`, but then you can only use the `--manifest-path` option, so maybe it's not worth bothering with.

.github/workflows/test.yml
```diff
@@ -0,0 +1,185 @@
...
+  tests:
+    name: Tests
+    needs: [cache_info, update_deps]
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [windows-latest, ubuntu-latest, macos-latest]
+        flags: ['', '--release']
+        toolchain: [stable, nightly]
```
mzabaluev, line 89: To reduce the number of jobs, maybe only test with `nightly` toolchain on `ubuntu-latest` in debug profile.
We need it as a canary to catch unlikely future regressions.

.github/workflows/test.yml
eugene-babichenko: Ouch, you're right, thanks for noticing that!

mzabaluev (approved):

mzabaluev:
Looks good. The lints job invokes a newer clippy than the CircleCI check, so we've got a few more failures.

Issue #2770, opened 2020-12-01
Implement private vote tallying in Jcli


dkijania:
**Is your feature request related to a problem/context ? Please describe if applicable.**
Please implement private vote tallying from jcli. Right now only public vote tally is supported:

`https://github.com/input-output-hk/jormungandr/blob/master/jcli/src/jcli_app/certificate/new_vote_tally.rs#L25`

**Describe the solution you'd like**
Private vote tally from jcli

PR #2772, opened 2020-12-01
Implement private tally certificate in jcli

danielSanchezQ:
Implements #2770

### DIFF
--- a/jcli/src/jcli_app/certificate/mod.rs
+++ b/jcli/src/jcli_app/certificate/mod.rs
@@ -98,9 +98,9 @@ pub enum Error {
     VoteEncryptingKey,
     #[error("invalid bech32 public key, expected {expected} hrp got {actual}")]
     InvalidBech32Key { expected: String, actual: String },
-    #[error("Invalid shares JSON representation")]
+    #[error("invalid shares JSON representation")]
     InvalidJson(#[from] serde_json::Error),
-    #[error("Invalid binary share")]
+    #[error("invalid binary share")]
     InvalidBinaryShare,
 }
 
--- a/jcli/src/jcli_app/certificate/mod.rs
+++ b/jcli/src/jcli_app/certificate/mod.rs
@@ -98,8 +98,10 @@ pub enum Error {
     VoteEncryptingKey,
     #[error("invalid bech32 public key, expected {expected} hrp got {actual}")]
     InvalidBech32Key { expected: String, actual: String },
-    #[error("Invalid json representation")]
+    #[error("Invalid shares JSON representation")]
     InvalidJson(#[from] serde_json::Error),
+    #[error("Invalid binary share")]
+    InvalidBinaryShare,
 }
 
 #[allow(clippy::large_enum_variant)]
--- a/jcli/src/jcli_app/certificate/mod.rs
+++ b/jcli/src/jcli_app/certificate/mod.rs
@@ -98,6 +98,8 @@ pub enum Error {
     VoteEncryptingKey,
     #[error("invalid bech32 public key, expected {expected} hrp got {actual}")]
     InvalidBech32Key { expected: String, actual: String },
+    #[error("Invalid json representation")]
+    InvalidJson(#[from] serde_json::Error),
 }
 
 #[allow(clippy::large_enum_variant)]
--- a/jcli/src/jcli_app/certificate/mod.rs
+++ b/jcli/src/jcli_app/certificate/mod.rs
@@ -100,7 +100,7 @@ pub enum Error {
     InvalidBech32Key { expected: String, actual: String },
     #[error("invalid shares JSON representation")]
     InvalidJson(#[from] serde_json::Error),
-    #[error("invalid binary share")]
+    #[error("invalid binary share data")]
     InvalidBinaryShare,
 }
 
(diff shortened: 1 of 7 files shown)


jormungandr-lib/src/interfaces/vote.rs
```diff
@@ -471,6 +471,23 @@ mod serde_base64_bytes {
     }
 }
 
+#[derive(Debug, Serialize, Deserialize)]
+pub struct TallyDecryptShare(#[serde(with = "serde_base64_bytes")] Vec<u8>);
```
mzabaluev, line 475: This structure is not used in the API. Maybe make it internal for jcli for the time being?..

jormungandr-lib/src/interfaces/vote.rs
```diff
@@ -471,6 +471,23 @@ mod serde_base64_bytes {
...
+
+impl TryFrom<TallyDecryptShare> for chain_vote::TallyDecryptShare {
+    // TODO: Maybe change this error to a custom type?
+    type Error = std::io::Error;
```
mzabaluev, line 479: ... then you can use `certificate::Error` for the command implementation.

This impl looks atypical for what the rest of jormungandr-lib does, let's not add it here until it's clear that we also need the serde impls for REST.

jcli/src/jcli_app/certificate/mod.rs
```diff
@@ -98,6 +98,8 @@ pub enum Error {
     VoteEncryptingKey,
     #[error("invalid bech32 public key, expected {expected} hrp got {actual}")]
     InvalidBech32Key { expected: String, actual: String },
+    #[error("Invalid json representation")]
```
eugene-babichenko, line 101: Maybe something like "Invalid shares JSON" to make it clear?

jormungandr-lib/src/interfaces/vote.rs
```diff
@@ -405,7 +405,7 @@ impl TallyResult {
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
 pub struct EncryptedTally(#[serde(with = "serde_base64_bytes")] Vec<u8>);
 
-mod serde_base64_bytes {
+pub mod serde_base64_bytes {
```
danielSanchezQ, line 408: I think we should start to move this kind of utilities to some other shared module ('serde_utils' or whatever). We can discuss it and I'll do it in a further PR.

testing/jormungandr-integration-tests/src/common/jcli/api/certificate.rs
```diff
@@ -27,9 +27,9 @@ impl Certificate {
             .as_single_line()
     }
 
-    pub fn new_vote_tally<S: Into<String>>(self, vote_plan_id: S) -> String {
+    pub fn new_vote_tally<S: Into<String>>(self, vote_plan_id: S, tally_type: S) -> String {
```
danielSanchezQ, line 30: @dkijania are these changes ok? or do you think it should have a custom type somewhere?

jcli/src/jcli_app/certificate/mod.rs
```diff
@@ -98,6 +98,10 @@ pub enum Error {
     VoteEncryptingKey,
     #[error("invalid bech32 public key, expected {expected} hrp got {actual}")]
     InvalidBech32Key { expected: String, actual: String },
+    #[error("Invalid shares JSON representation")]
+    InvalidJson(#[from] serde_json::Error),
+    #[error("Invalid binary share")]
```
mzabaluev, line 103: This should be clearer worded, I think: "invalid binary share data"?
Capitalization should be more consistent across the enum. Let's go with small letters everywhere.

jormungandr-lib/src/interfaces/vote.rs
mzabaluev: Yeah, but check also https://crates.io/crates/base64-serde

testing/jormungandr-integration-tests/src/common/jcli/api/certificate.rs
mzabaluev: A custom type would be more disciplined. Could reuse from jormungandr-lib?

jcli/src/jcli_app/certificate/mod.rs
danielSanchezQ: my bad should have realized it :/

jormungandr-lib/src/interfaces/vote.rs
danielSanchezQ: I knew about that library. But, without having checked it properly maybe. I would say it doesn't support the check we do for the human-readable serializer in there.

testing/jormungandr-integration-tests/src/common/jcli/api/certificate.rs
danielSanchezQ: We could use `PayloadTypeDef`..

testing/jormungandr-integration-tests/src/common/jcli/command/certificate.rs
```diff
@@ -163,3 +168,11 @@ impl CertificateCommand {
         self.command
     }
 }
+
+fn payload_type_to_string(payload_type: PayloadType) -> String {
```
mzabaluev, line 172: No need to allocate, just return `&'static str`.

mzabaluev (commented):
The rest looks OK, just fix the tests that are failing to build.

mzabaluev (approved):

PR #2775, opened 2020-12-02
correctly interrupt the bootstrap process when the remote peer is not responding

eugene-babichenko:


mzabaluev (approved):
Looks good.

But could we instead wrap the whole bootstrap future into a `select`, rather than passing the cancellation token down to check with every intermediate step?
Also, why did it busy loop when the connection was broken, instead of failing?

eugene-babichenko:
> But could we instead wrap the whole bootstrap future into a `select`?

No. The whole point of passing it down to `bootstrap_from_stream` is to update the blockchain tip after the interruption.

> Also, why did it busy loop when the connection was broken, instead of failing?

Not sure about that really, must have something to do with the networking stack. Everything apart from the newly wrapped function is cancellable.

mzabaluev:
Let's merge this now because it's annoying to test the peer hangups without it.

PR #2803, opened 2020-12-08
[Test] test which sends transaction with more than 97 outputs

dkijania:


mzabaluev:
The PR fails the build.

PR #2807, opened 2020-12-09
explorer: load all available branches during bootstrap

### LINKED ISSUE #2790: Explorer spawned with passive node cannot find transaction
**Describe the bug**
having network topology like below:

image

Where leader3 is a passive node with explorer

I'm sending transaction to node 1, with expectation that explorer setup on passive node would show it when sending query:
```
query:
{
     transaction(id: "bfa87e6841b2fabda40502efebeb129f1b2f09a2ca231193bb129ceb81699595") {
        id
    }
}", against: http://127.0.0.1:8002/explorer/graphql
```

Unfortunately i'm receiving an error:
```
Response: Response { data: None, errors: Some([Error { message: "transaction not found: bfa87e6841b2fabda40502efebeb129f1b2f09a2ca231193bb129ceb81699595", locations: Some([Location { line: 2, column: 5 }]), path: Some([Key("transaction")]), extensions: None }]) }
```
**Mandatory Information**
1. `jcli --full-version` output: `newest master`
2. `jormungandr --full-version` output: `newest master`

**To Reproduce**
`cargo run -- --scenario passive_node_explorer`

**Expected behavior**
transaction should be available in explorer

eugene-babichenko:
resolves #2790

### DIFF
--- a/jormungandr/src/explorer/mod.rs
+++ b/jormungandr/src/explorer/mod.rs
@@ -241,8 +241,7 @@ impl ExplorerDB {
                 hash = block.header.block_parent_hash();
                 blocks.push(block);
             }
-            blocks.reverse();
-            for block in blocks {
+            while let Some(block) = blocks.pop() {
                 db.apply_block(block).await?;
             }
         }
--- a/jormungandr/src/explorer/mod.rs
+++ b/jormungandr/src/explorer/mod.rs
@@ -216,13 +216,38 @@ impl ExplorerDB {
                 )))
             }
         };
-        stream
+
+        let mut db = stream
             .map_err(Error::from)
             .try_fold(bootstraped_db, |mut db, block| async move {
                 db.apply_block(block).await?;
                 Ok(db)
             })
-            .await
+            .await?;
+
+        for branch in blockchain.branches().branches().await.iter() {
+            let mut hash = branch.hash();
+            let mut blocks = vec![];
+            loop {
+                if db.get_block(&hash).await.is_some() {
+                    break;
+                }
+                let block = blockchain.storage().get(hash)?.ok_or_else(|| {
+                    Error::from(ErrorKind::BootstrapError(format!(
+                        "couldn't get block {} from the storage",
+                        hash
+                    )))
+                })?;
+                hash = block.header.block_parent_hash();
+                blocks.push(block);
+            }
+            blocks.reverse();
+            for block in blocks {
+                db.apply_block(block).await?;
+            }
+        }
+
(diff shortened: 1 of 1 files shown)


jormungandr/src/explorer/mod.rs
```diff
@@ -216,13 +216,38 @@ impl ExplorerDB {
...
+                let block = blockchain.storage().get(hash)?.ok_or_else(|| {
+                    Error::from(ErrorKind::BootstrapError(format!(
+                        "couldn't get block {} from the storage",
+                        hash
+                    )))
+                })?;
+                hash = block.header.block_parent_hash();
+                blocks.push(block);
+            }
+            blocks.reverse();
```
enzoc4, line 244: I would use `pop` as this is kind of recursive/stack-like, but probably doesn't make any difference.

jormungandr/src/explorer/mod.rs
```diff
@@ -216,13 +216,38 @@ impl ExplorerDB {
...
                 Ok(db)
             })
-            .await
+            .await?;
+
+        for branch in blockchain.branches().branches().await.iter() {
+            let mut hash = branch.hash();
+            let mut blocks = vec![];
+            loop {
+                if db.get_block(&hash).await.is_some() {
```
enzoc4, line 232: I would just use `apply_block` and see if it returns the `AncestorNotFound` error. Otherwise I would use directly the multiverse lookup, as `get_block` only looks in the current explorer branch (although that will change with #2801), which I think may lead to applying a block twice.

jormungandr/src/explorer/mod.rs
eugene-babichenko: Hm, it actually could make difference. At least this would be a single vector pass instead of two.

jormungandr/src/explorer/mod.rs
eugene-babichenko: > as `get_block` only looks in the current explorer branch

Thanks for noting that. Yeah, for some branching cases this may result into applying a block twice. But will it break anything apart from the performance? Anyway, these branches are rather short (less than 10 blocks) and I wouldn't care much for the performance here. This is a procedure that is ran onle once and we are going to drop it soon.

enzoc4: Hmm, for some reason I actually thought `apply_block` was going to throw an error, but now I think you are right, it will only affect performance. So yeah, it's fine to leave it as it is I guess.

enzoc4 (approved):

PR #2812, opened 2020-12-11
process services panics into errors and exit

eugene-babichenko:


jormungandr/src/utils/task.rs
```diff
@@ -132,31 +143,48 @@ impl Services {
...
+            Some(Ok(Err(service_error))) => Err(ServiceError::Service(service_error)),
+            // A service panicked or was cancelled by the environment
+            Some(Err(join_error)) => {
+                if join_error.is_cancelled() {
+                    Err(ServiceError::Cancelled)
+                } else if join_error.is_panic() {
+                    let desc = join_error
+                        .into_panic()
+                        .downcast_ref::<String>()
+                        .expect("a panic must be convertible to String")
```
mzabaluev, line 180: This is not necessarily true: `panic!` can box anything as the value.
I think it's better to make the `ServiceError::Panic` variant type `Option<String>`.

mzabaluev (changes requested):
Looks good. Handling of panic values should not rely on assumptions even if they are almost always correct, though.

Issue #2811, opened 2020-12-11
Develop Encode/Decode Qr code method in Rust


dkijania:
It would be helpful regarding vote end to end flow testing if we have a very simple method that is able to decode/encode method like below:

https://github.com/input-output-hk/vit-kedqr/blob/master/internal/ked/ked.go

mzabaluev:
`cryptoxide` would be ideal for this if not for https://github.com/typed-io/cryptoxide/issues/13.

enzoc4:
forgot to link here from #2856. In any case, we should probably close this, right?

mzabaluev:
All seems done, please reopen if something is missing.

PR #2814, opened 2020-12-12
jormungandr-testing-utils: add QR code generator

mzabaluev:
The encode part for #2811.


dkijania:
Working good. was able to scan qr using catalyst app

dkijania (approved):

enzoc4:
shouldn't we use the same code that's in `chain-wallet-libs`? The [encryption](https://github.com/input-output-hk/chain-wallet-libs/blob/47bca1e53f93351514c132db950ceb291c3fdf64/symmetric-cipher/src/lib.rs#L32) and [decryption](https://github.com/input-output-hk/chain-wallet-libs/blob/47bca1e53f93351514c132db950ceb291c3fdf64/symmetric-cipher/src/lib.rs#L88) functions that are used for voting are already written in rust, so unless the purpose is to explicitly double check the implementation it seems like we should reuse that part.

dkijania:
@enzoc4 thanks for being watchful. I didn't realize that there is already a method in rust for this (now it's obvious that there should be one).  Yes, we should use as much `chain-wallet-lib` code as possible.

eugene-babichenko:
@darek have you tested it after Enzo's change?

dkijania:
will test it today or tommorow

dkijania:
working good. Thanks @mzabaluev and @enzoc4

Issue #2147, opened 2020-04-27
Panic in jormungandr after stake pool retiremenet


dkijania:
**Describe the bug**
There is an ugly panic in jormungandr after stake pool (the only one in blockchain) was retired:

```
"thread \'tokio-runtime-worker\' panicked at \'Distribution of rewards will not overflow: Delegation(NotFound($ hash_ty(0x38d18bcafb8ba65673dc00eb8ce7150ea3a7e61c24153dafa23cb437caa9b970)))\', jormungandr/src/blockchain/chain.rs:852:46"
```

**Mandatory Information**
1. `jcli --full-version`: newest master
2. `jormungandr --full-version` newest master

**To Reproduce**
1. start blockchain with single stake pool
2. retire it after some time
3. Observe issue

**Expected behavior**
node stats should be updated 

**Additional context**
There is PR with automated test for this issue.  https://github.com/input-output-hk/jormungandr/pull/2145. Run test  by:
```
 cd jormungandr-integration-tests
`cargo test test_pool_update`
```

dkijania:
Panic is produced when we retire the only one stake pool in blockchain. Therefore it's only matter or better error reporting

mzabaluev:
Is this reproducible after #2815?

PR #2815, opened 2020-12-12
blockchain: check rewards distribution before applying a block

### LINKED ISSUE #2800: Unexpected shut down of retired stake pool
**Describe the bug**
For topology like below:
image

when send `retire stake pool` certificate for leader3, node unexpectedly is shutdown.
Logs: 
[retire_stake_pool_explorer.zip](https://github.com/input-output-hk/jormungandr/files/5642195/retire_stake_pool_explorer.zip)

**Mandatory Information**
1. `jcli --full-version` output;
2. `jormungandr --full-version` output;

**To Reproduce**
`cargo run -- --scenario retire_stake_pool_explorer --root-dir retire_stake --document`

eugene-babichenko:
Block rewards distribution was checked only in specific cases and there
was no universal check for that before applying a block. This could have
resulted into a situation, when a block with a rewards distribution that
cannot be calculated could have been applied. Thus, a node might have
been stuck in an invalid state.

A particular example of that is a block signed by the same node that is
being retired by a transaction in this block. Because a node is already
retired when this block is applied, the rewards for this block cannot be
distributed.

Resolves #2800.

### DIFF
--- a/jormungandr/src/blockchain/chain.rs
+++ b/jormungandr/src/blockchain/chain.rs
@@ -476,9 +476,31 @@ impl Blockchain {
 
         let metadata = header.to_content_eval_context();
 
-        ledger
+        let ledger = ledger
             .apply_block(epoch_ledger_parameters, &block.contents, &metadata)
-            .chain_err(|| ErrorKind::CannotApplyBlock)
+            .chain_err(|| ErrorKind::CannotApplyBlock)?;
+
+        // Check if rewards for this block can be distributed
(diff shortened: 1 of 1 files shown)


eugene-babichenko:
@dkijania where is the test mentioned in #2147? I am pretty sure that with this an attempt to retire the last pool will result into a loop where a node cannot produce a new block, but still need to check,

dkijania (approved):

PR #2841, opened 2020-12-18
Private tally docs

danielSanchezQ:


danielSanchezQ:
Python scripts would be: https://github.com/input-output-hk/vit-ops/pull/3

doc/jcli/vote.md
```diff
@@ -7,7 +7,45 @@ to decrypt and certify the tally.
...
+We can get its public representation with:
+
+```shell
+jcli votes committee communication-key to-public --input ./comm.sk > ./comm.pk
+```
+
+#### Committee member key
+
+```shell
+crs=$(jcli vote crs generate)
```
eugene-babichenko, line 32: Is crs generated per committee member? AFAIK it should be the same for all members.

doc/jcli/vote.md
```diff
@@ -7,7 +7,45 @@ to decrypt and certify the tally.
...
+Where `pkX` are each of the committee communication public keys.
+We can also easily get its public representation as before:
+
+```shell
+jcli votes committee member-key to-public --input ./member.sk ./member.pk
+```
+
+
+#### Vote encrypting key
+This key (*public*) is the key **every vote** should be encrypted with.
```
eugene-babichenko, line 44: Please include how to obtain these from the blockchain.

doc/jcli/vote.md
```diff
@@ -39,4 +77,54 @@ jcli transaction seal --staging vote-tally.staging
 jcli transaction auth --staging vote-tally.staging --key committee.sk
 jcli transaction to-message --staging vote-tally.staging > vote-tally.fragment
 jcli rest v0 message post --file vote-tally.fragment
-```
\ No newline at end of file
+```
+
+### Private
+To tally private votes, all committee members are needed.
```
eugene-babichenko, line 83: Apart from creating the tally we need to document the creation of a vote plan.

doc/jcli/vote.md
danielSanchezQ: It should be the same though.

eugene-babichenko: Please add a note that specifically says all members should use the same crs string

doc/jcli/vote.md
eugene-babichenko: Another note on that: not "all committee members" but the number of them specified in the threshold.

danielSanchezQ: Actually is all committee members for now, since the threshold is not still implemented.

doc/jcli/vote.md
eugene-babichenko: typo - votes

doc/jcli/vote.md
```diff
@@ -7,7 +7,98 @@ to decrypt and certify the tally.
...
+
+
+
+## Creating a vote plan
+
+We need to provide a vote plan definition file to generate a new voteplan certificate.
+That file should be a `yaml` (or json) with the following format:
+```yaml
+{
+  "payload_type": "public",
```
eugene-babichenko, line 67: should be private

danielSanchezQ: indeed

doc/jcli/vote.md
```diff
@@ -7,7 +7,98 @@ to decrypt and certify the tally.
...
+  },
+  "proposals": [
+    {
+      "external_id": "d7fa4e00e408751319c3bdb84e95fd0dcffb81107a2561e691c33c1ae635c2cd",
+      "options": 3,
+      "action": "off_chain"
+    },
+    ...
+  ],
+  "commitee_public_keys": [
```
eugene-babichenko, line 88: `committee_member_public_keys` is expected here

doc/jcli/vote.md
```diff
@@ -39,4 +130,55 @@ jcli transaction seal --staging vote-tally.staging
...
+--output-format json > result.json
+```
+
+Notes:
+
+The table size is a cache parameter, any value greater than one would be enough but the best approximated values would be
+`table_size = votes_cast / vote_options`. So, if we had `1000` votes and `2` options (*yes*, *no*), and optimum table size value
+would be `1000/2 = 500`
+
+It may be ***cumbersome*** to do this process manually. So we can use a `python` script for processing the tallying in 3 simple steps:
```
eugene-babichenko, line 180: Maybe move this part to #2844 since there's no actual scripts yet?

danielSanchezQ: yeah, Ill add it to that pr after this is merged.

eugene-babichenko (approved):
The vote plan creation part is looking good to me.

danielSanchezQ:
Failing tests are completely unrelated since this is just documentation.

PR #2856, opened 2020-12-21
add QR decoder and to_img method

enzoc4:
Complement to #2814 , add the decode part.

Also add a way to generate an image (because an svg is not really an image).


dkijania (approved):

dkijania:
thanks!

PR #2858, opened 2020-12-21
fix praos vote test, remove second pool in node

dkijania:


eugene-babichenko:
This is the closest to "all tests green" in the last few month :)

PR #2861, opened 2020-12-22
[Test] Private vote test

dkijania:
Adjust private vote test after docs update


testing/jormungandr-integration-tests/src/jormungandr/vit/private.rs
```diff
@@ -182,46 +194,46 @@ pub fn jcli_e2e_flow_private_vote() {
...
+        .unwrap()
+        .tally
+        .as_ref()
+        .unwrap()
+    {
+        jormungandr_lib::interfaces::Tally::Private { state } => match state {
+            jormungandr_lib::interfaces::PrivateTallyState::Encrypted {
+                encrypted_tally,
+                total_stake: _,
+            } => serde_json::to_string(&encrypted_tally).unwrap(),
```
eugene-babichenko, line 214: This adds unnecessary quotes around the output. `jcli` expects plain `base64` here.

eugene-babichenko (changes requested):
`decrypt_with_shares` implementation is wrong. The actual name of the command is `decrypt` and it expects `--table_size`, not `--tablesize`. The latter will change to `--table-size` once I push the set of changes I made to `jcli` to make this test work.

PR #2900, opened 2021-01-11
Fix YAML in CI workflow

mzabaluev:


.github/workflows/test.yml
```diff
@@ -12,7 +12,7 @@ on:
       - modules/**
       - testing/**
       - Cargo.*
-      - !**.md
```
eugene-babichenko, line 15: Why remove this line?

mzabaluev: It seemingly prevents any other changes from being tested. Once I removed it, the CI workflow started running again.

The whole set of paths is a bit fragile, it will need to be adjusted with any changes in the tree and there are likely things not accounted yet. I think the mdBook workflow filter is a better use of this feature: it's simple and removes the large amount of redundant documentation rebuilds. I'm not sure this path filter helps a lot, because source changes dominate anyway.

mzabaluev: > It seemingly prevents any other changes from being tested. Once I removed it, the CI workflow started running again.

Actually, it's the missing `.github` change filter that probably blocked it; I'll recheck.

PR #2903, opened 2021-01-11
send new-tip message to explorer

enzoc4:
remove the branch selection logic from the explorer by sending a message through the intercom from the branch selection function

this is similar to #2378 (there the new tip is sent through the websocket interface), so this helps a bit with the explorer extraction too.

### DIFF
--- a/jormungandr/src/blockchain/process.rs
+++ b/jormungandr/src/blockchain/process.rs
@@ -188,12 +188,20 @@ impl Process {
     fn start_branch_reprocessing(&self, info: &TokioServiceInfo) {
         let tip = self.blockchain_tip.clone();
         let blockchain = self.blockchain.clone();
+        let explorer = self.explorer_msgbox.clone();
         let logger = info.logger().clone();
 
         info.run_periodic_fallible(
             "branch reprocessing",
             BRANCH_REPROCESSING_INTERVAL,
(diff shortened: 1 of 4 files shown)


eugene-babichenko (approved):
LGTM. BTW do we have any tests that verify the correctness of branch switching in the explorer?

enzoc4:
> LGTM. BTW do we have any tests that verify the correctness of branch switching in the explorer?

@dkijania it's probably better suited to answer that. As far as I remember, there was a non-functional test somewhere that checked the explorer tip against rest for 15 minutes, but I don't know if there's anything involving multiple nodes (for testing forks).

the `explorer_sanity_test` is probably enough to test that this works, though, as otherwise the transaction wouldn't be there, because the state would remain just as it was in the genesis.

dkijania:
@enzoc4 @eugene-babichenko we have test in `jormungandr-scenario-tests` project: `passive_node_explorer` which basically check if single transaction send to leader node  is reflected in explorer setup on passive node :

https://github.com/input-output-hk/jormungandr/blob/master/testing/jormungandr-scenario-tests/src/test/features/explorer.rs#L15

Please tell me if we need some another test

enzoc4:
thanks @dkijania 
I don't think we need anything else at least for this one

PR #2911, opened 2021-01-12
fix cargo fmt errors

eugene-babichenko:


mzabaluev:
Was it not caught by the lints job in CI?

mzabaluev (approved):

PR #2912, opened 2021-01-12
Private vote e2e flow test fix

danielSanchezQ:


Zeegomo:
Can we add a check to ensure that the outcome of the voting is as expected?

mzabaluev (approved):
Looks good for fixing the existing issue. Additional checks can be added in follow-on PRs.

dkijania (approved):

PR #2915, opened 2021-01-13
update vit-servicing-station and move migrations folder

cleverca22:
update to include https://github.com/input-output-hk/vit-servicing-station/pull/112


cleverca22:
conflict is due to another update of the lock file, which is new enough

PR #2916, opened 2021-01-13
[Tests] fix recovery test

dkijania:


mzabaluev:
Duplicate of #2921, see the discussion there.

mzabaluev:
Let's go with #2921, I like the 1 second sleep better and it should be good enough for the flush timeout.

Issue #2592, opened 2020-10-12
Node cannot recover from unexpected shutdown


dkijania:
**Describe the bug**
In rare occasions we can observe that node looses data on abrupt restart. In simple scenario we bootstrap node with some genesis block, send one simple transaction and then node is killed.

From time to time we observe that node after restart does not have in history mentioned transaction.

**Mandatory Information**
1. `jcli --full-version`: newest master
2. `jormungandr --full-version` newest master

**To Reproduce**
1. Go to jormungandr_integration_tests
2. Run test `jormungandr::recovery::test_node_recovers_kill_signal`
3. Observe error

**Expected behavior**
Node should recover from kill signal with 100% correctness

**Additional context**
All test data: [tmpvZ8H5X.zip](https://github.com/input-output-hk/jormungandr/files/5367007/tmpvZ8H5X.zip)
command to run node
```
jormungandr.exe  --secret  node_secret.yaml --config node_config.yaml --genesis-block block-0.bin


```

eugene-babichenko:
This is a storage problem. After the restart the tip can be one block behind. In the logs you provided:

- Here is the message before the restart: `{"msg":"update current branch tip: 454c129a-00000004-0.4 -> 97db7c29-00000005-0.5","level":"INFO","ts":"2020-10-12T17:11:05.017689+02:00","date":"0.5","parent":"454c129a3628d063df58218fdd79963129c19e15032381c69223587f3b45f0d1","hash":"97db7c297e763904031922bda86467ac27a2a15c3da8e1f93bab942924d0e384","task":"block"}`
- But after the restart we have: `{"msg":"Loaded from storage tip is : 454c129a-00000004-0.4","level":"INFO","ts":"2020-10-12T17:11:10.805142+02:00","task":"bootstrap"}`

This is because sled does only periodic flushes triggered by its internal timer. This is why this bug can be hard to reproduce: it is very dependent on the timing which is actually a bit of a randomized factor.

The solution could be to forcibly flush the storage after updating the tag, but this needs to be investigated. I am not sure if the `fsync` system call (or its Windows equivalent) would finish upon receiving `SIGKILL`. If not, this can lead to new problems.

eugene-babichenko:
Since `sled` does its own caching, there is still a possibility that the `.flush()` call to the database will be interrupted by `SIGKILL`. Thus, even if we force flushing, the test may still fail on rare occasions. Although doing more frequent flushes reduces the possibility of such failures.

[...]

PR #2921, opened 2021-01-14
Wait before stopping the node so transactions are flushed to disk

Zeegomo:
Tested locally this change seems to remove the spurious failures of issue #2592 (*) and hopefully fix CI too. The idea is to wait so that transactions are automatically flushed to disk (sled does this by default every 500ms), but I'm not sure if this relax some capability that this test aims to check. 

(*) I had a slightly different issue with the rest request failing because the account account did not exist, but I guess the root cause is the same.

### DIFF
--- a/testing/jormungandr-integration-tests/src/jormungandr/recovery.rs
+++ b/testing/jormungandr-integration-tests/src/jormungandr/recovery.rs
@@ -157,9 +157,9 @@ pub fn test_node_recovers_kill_signal() {
         &jormungandr,
     );
     let snapshot_before = take_snapshot(&account_receiver, &jormungandr, new_utxo.clone());
+    // Wait before stopping so transactions are flushed to disk
+    std::thread::sleep(std::time::Duration::from_secs(1));
     jormungandr.stop();
-
-    std::thread::sleep(std::time::Duration::from_secs(2));
 
     let jormungandr = Starter::new()
         .config(config)


danielSanchezQ:
As we talked about when debugging this. This should be enough for this test. If we want to test the other scenario where some parts of the storage are missing we should set it up with more nodes so it can actually catch up (as described in #2592 ).

danielSanchezQ (approved):

danielSanchezQ:
I approved it but let's wait until @dkijania and @mzabaluev check it out.

testing/jormungandr-integration-tests/src/jormungandr/recovery.rs
```diff
@@ -157,6 +157,8 @@ pub fn test_node_recovers_kill_signal() {
         &jormungandr,
     );
     let snapshot_before = take_snapshot(&account_receiver, &jormungandr, new_utxo.clone());
+    // Wait before stopping so transactions are flushed to disk
+    std::thread::sleep(std::time::Duration::from_secs(1));
```
mzabaluev, line 161: There is still a marginal race condition because the flush interval is exactly one second (correct me if I'm wrong @eugene-babichenko). Wait two seconds to be sure.

testing/jormungandr-integration-tests/src/jormungandr/recovery.rs
```diff
@@ -157,6 +157,8 @@ pub fn test_node_recovers_kill_signal() {
...
     jormungandr.stop();
 
     std::thread::sleep(std::time::Duration::from_secs(2));
```
mzabaluev, line 164: This should probably be removed altogether, there should be nothing to wait for after the node was killed. Cc @dkijania

mzabaluev (commented):
Other than the remarks below, I think it's a reasonable fix.

testing/jormungandr-integration-tests/src/jormungandr/recovery.rs
danielSanchezQ: We checked and by default is every 15ms.

mzabaluev: Oh, I missed part of the description and you have researched the interval better. Ignore this.

mzabaluev (approved):

PR #2923, opened 2021-01-14
use a copy of fragment pool per leader

### LINKED ISSUE #2834: On a node started with multiple stake pools they produce blocks with different sets of transactions
**Describe the bug**
When the node is started with multiple leaders, both of them are creating new blocks. Due to the current implementation of the fragments pool, where the selected blocks are immediately removed from the fragment pool, the two stake pools always end up with two entirely different sets of transactions in their blocks.

This caused random failures in `test_vote_flow_praos` due to the fact that some blocks were not included into the main chain and thus the transaction history in the main chain was different from what the test expected. In the case of this test the failure manifested itself as:

- An "invalid signature" message due to mismatch of the spending counter used for generating the witness (it was smaller in the ledger than in the test client due to a transaction not being in the main branch). Example: https://github.com/input-output-hk/jormungandr/runs/1562468481
- A failure to get the expected voting result due to vote casts missing from the main branch. Example: https://github.com/input-output-hk/jormungandr/runs/1556761271?

Note that this is the state at the time writing this issue and the test may change.

**Mandatory Information**
1. `jcli --full-version` output: `jcli 0.10.0-alpha.2 (master-fc517e40, debug, linux [x86_64]) - [rustc 1.48.0 (7eac88abb 2020-11-16)]`
2. `jormungandr --full-version` output: `jormungandr 0.10.0-alpha.2 (master-fc517e40, debug, linux [x86_64]) - [rustc 1.48.0 (7eac88abb 2020-11-16)]`

**To Reproduce**
1. `cargo build --locked -p jormungandr -p jcli`
2. `cargo test test_vote_flow_praos`

**Expected behavior**
We should either provide a way for multiple pools to work on the same set of transactions or forbid such configuration.

eugene-babichenko:
Fix #2834.

### DIFF
--- a/jormungandr-lib/src/interfaces/fragment_log.rs
+++ b/jormungandr-lib/src/interfaces/fragment_log.rs
@@ -87,6 +87,10 @@ impl FragmentLog {
     /// set the new status
     #[inline]
     pub fn modify(&mut self, new_status: FragmentStatus) {
+        // we must not be able to transition from InABlock to Pending or Rejected
+        if self.status.is_in_a_block() && !new_status.is_in_a_block() {
+            return;
+        }
         self.status = new_status;
         self.last_updated_at = SystemTime::now();
     }
--- a/jormungandr-lib/src/interfaces/fragment_log.rs
+++ b/jormungandr-lib/src/interfaces/fragment_log.rs
@@ -84,15 +84,20 @@ impl FragmentLog {
         self.status().is_in_a_block()
     }
 
-    /// set the new status
+    /// Set the new status
+    ///
+    /// # Returns
+    ///
+    /// `true` if the value was updated. `false` if the upadte was refused.
     #[inline]
-    pub fn modify(&mut self, new_status: FragmentStatus) {
+    pub fn modify(&mut self, new_status: FragmentStatus) -> bool {
         // we must not be able to transition from InABlock to Pending or Rejected
         if self.status.is_in_a_block() && !new_status.is_in_a_block() {
-            return;
+            return false;
         }
         self.status = new_status;
         self.last_updated_at = SystemTime::now();
+        true
     }
 
     #[inline]
--- a/jormungandr/src/fragment/pool.rs
+++ b/jormungandr/src/fragment/pool.rs
@@ -85,18 +85,20 @@ impl Pool {
                 .iter()
                 .map(move |fragment| FragmentLog::new(fragment.id(), origin))
                 .collect::<Vec<_>>();
-            for fragment in new_fragments.into_iter() {
-                let fragment_msg = NetworkMsg::Propagate(PropagateMsg::Fragment(fragment));
-                network_msg_box
-                    .send(fragment_msg)
-                    .await
-                    .map_err(Error::CannotPropagate)?;
-            }
             self.logs.insert_all(fragment_logs);
             if count > max_added {
                 max_added = count;
             }
         }
+
+        for fragment in new_fragments.into_iter() {
+            let fragment_msg = NetworkMsg::Propagate(PropagateMsg::Fragment(fragment));
+            network_msg_box
+                .send(fragment_msg)
+                .await
+                .map_err(Error::CannotPropagate)?;
+        }
+
         Ok(max_added)
     }
 
--- a/jormungandr/src/fragment/pool.rs
+++ b/jormungandr/src/fragment/pool.rs
@@ -17,7 +17,7 @@ use thiserror::Error;
 
 pub struct Pool {
     logs: Logs,
-    pool: internal::Pool,
+    pools: Vec<internal::Pool>,
     network_msg_box: MessageBox<NetworkMsg>,
     logger: Logger,
 }
@@ -31,13 +31,17 @@ pub enum Error {
 impl Pool {
     pub fn new(
         max_entries: usize,
+        n_pools: usize,
         logs: Logs,
         network_msg_box: MessageBox<NetworkMsg>,
         logger: Logger,
     ) -> Self {
+        let pools = (0..n_pools)
+            .map(|_| internal::Pool::new(max_entries))
+            .collect();
         Pool {
             logs,
-            pool: internal::Pool::new(max_entries),
+            pools,
             network_msg_box,
             logger,
         }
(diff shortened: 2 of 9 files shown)


eugene-babichenko:
Tested it with #2858 reverted, seems to work.

eugene-babichenko:
There are compilation issues on nightly rustc, although the stable one seems good.

Zeegomo:
I have some spider-sense tingling on having shared logs for different pools that could have different state. What would happen for example if `pool.select` is first called on pool1 and succeed while a second call on the second pools fails due to (for instance) changes in the ledger? What status should `get_fragment_logs` return? That being said, in a first look such scenarios do not seem highly probable.

Not related to this pr, I think it is a problem if `logs_max_entries` < `pool_max_entries`, because for example some of the logs would be thrown away when inserting a chunk of transactions of size `pool_max_entries`, and an attempt to modify the logs (e.g. a call to select) would flag those fragments as originating from the network while that may not be the case

jormungandr/src/fragment/pool.rs
```diff
@@ -67,40 +71,52 @@ impl Pool {
...
+            let fragment_logs = new_fragments
+                .iter()
+                .map(move |fragment| FragmentLog::new(fragment.id(), origin))
+                .collect::<Vec<_>>();
+            for fragment in new_fragments.into_iter() {
+                let fragment_msg = NetworkMsg::Propagate(PropagateMsg::Fragment(fragment));
+                network_msg_box
+                    .send(fragment_msg)
+                    .await
+                    .map_err(Error::CannotPropagate)?;
```
mzabaluev, line 93: Let's give some thought to not propagating fragments redundantly. This should be outside of the loop over the pools.

jormungandr/src/fragment/pool.rs
```diff
@@ -17,7 +17,7 @@ use thiserror::Error;
 
 pub struct Pool {
     logs: Logs,
-    pool: internal::Pool,
+    pools: Vec<internal::Pool>,
```
mzabaluev, line 20: Nitpick: rename to `Pools` now that it can house multiple pools?

mzabaluev (changes requested):

Zeegomo:
I have another idea on how to deal with multiple leaders in a single node that does not involve having multiple pools:
AFAIU The problem is when 2 or more leaders in the praos settings are leaders for the same slot (that is not possible in bft), and that causes the chain to diverge and some of the transaction are "wasted" because one of the blocks will not be part of the main chain for sure. While this is not direct violation of the protocol (at least in my eyes) this is certainly inefficient. What if, instead, we inhibit one (or more) leaders from issuing blocks for a slot if we know they will produce forks? This can be easily done in the schedule (and also randomly, so that the single leader selection procedure is fair), and from a high enough perspective the node can be seen as a single leader (produces at most one block per slot following normal rules) acting on behalf of both the two leaders (should be similar enough to delegating stake I think). This will remove the need to have multiple pools and shared logs.

mzabaluev:
> What if, instead, we inhibit one (or more) leaders from issuing blocks for a slot if we know they will produce forks?
> This can be easily done in the schedule (and also randomly, so that the single leader selection procedure is fair),

Consider the blockchain-as-a-service scenario when a node operator runs co-located consensus for several stakeholders in exchange for some off-chain compensation. I don't think they will consider it fair unless it's also protocol-fair. That is, if they have a chance to produce a block and have it eventually accepted by the consensus, we should not deny them this.

mzabaluev:
> Consider the blockchain-as-a-service scenario when a node operator runs co-located consensus for several stakeholders in exchange for some off-chain compensation.

On the other hand, we have delegation as a proper, on-chain way to arrange this. Still, we should not penalize leaders for being co-hosted as long as we support the feature.

Zeegomo:
On one side, yes, this is a deviation from the protocol. 
On the other hand, if the leaders produce multiple blocks for the same slot, only one of them will be accepted by the chain, and, presuming the operator of the pools does not impair some, will likely reach outer nodes at the same time. Once the blocks are out on the network what is the probability that a block is picked up instead of the others? I would say they are all equally probable, and as such by making a fair selection at the creation we are not taking things away from the pool owners. 

Actually, this is not true if there are also other blocks for that slot, because then the probability that one of the blocks produced by the multi-pools node will be retained on the main chain is less than if all of the pools produced a block. 

To be precise, with the current implementation there is a consistent order of pool leadership evaluation, so I would say that the first pool, having its blocks always propagated before the other pools blocks, may have an unfair advantage (although probably limited)

eugene-babichenko:
Given all that concerns, probably it would be better to encapsulate pools and logs into leaders and request logs on per-leader basis?

jormungandr-lib/src/interfaces/fragment_log.rs
```diff
@@ -87,6 +87,10 @@ impl FragmentLog {
     /// set the new status
     #[inline]
     pub fn modify(&mut self, new_status: FragmentStatus) {
+        // we must not be able to transition from InABlock to Pending or Rejected
+        if self.status.is_in_a_block() && !new_status.is_in_a_block() {
+            return;
```
mzabaluev, line 92: Um, just that? No error, no warning message?

jormungandr/src/fragment/pool.rs
```diff
@@ -67,40 +71,54 @@ impl Pool {
...
-        self.logs.insert_all(fragment_logs);
-        Ok(count)
+
+        Ok(max_added)
     }
 
     pub fn remove_added_to_block(&mut self, fragment_ids: Vec<FragmentId>, status: FragmentStatus) {
-        self.pool.remove_all(fragment_ids.iter().cloned());
+        for pool in &mut self.pools {
+            pool.remove_all(fragment_ids.iter().cloned());
```
mzabaluev, line 107: Hmm, why would we need to take ownership over the IDs in that iterator argument...

mzabaluev (approved):

jormungandr/src/fragment/pool.rs
eugene-babichenko: No point in that actually.

jormungandr-lib/src/interfaces/fragment_log.rs
eugene-babichenko: Not really sure if this is worth even a warning level message. This is a normal scenario in the multi-leader operation mode.

mzabaluev:
> To be precise, with the current implementation there is a consistent order of pool leadership evaluation, so I would say that the first pool, having its blocks always propagated before the other pools blocks, may have an unfair advantage (although probably limited)

This is true, but another branch may still be selected and we can't replace the network-wide settlement process with arbitrary decisions on the tech level that would draw the accidents of node co-hosting into how the network behaves (even more than we do now by serialized leadership evaluation and possible other unintended effects).

mzabaluev (approved):

PR #2925, opened 2021-01-15
add flake.nix

cleverca22:


flake.nix
```diff
@@ -0,0 +1,34 @@
...
+          buildInputs = [ openssl ];
+          configurePhase =''
+            cc=$CC
+          '';
+          doCheck = false;
+          doInstallCheck = false;
+          PROTOC="${protobuf}/bin/protoc";
+          PROTOC_INCLUDE="${protobuf}/include";
+        }
+      ) {};
```
jonringer, line 26: I would move this into a jormungandr.nix to keep the flake clean

flake.nix
```diff
@@ -0,0 +1,34 @@
...
+    };
+    pkgs = import nixpkgs { inherit system; overlays = [ overlay ]; };
+  in {
+    packages.jormungandr = pkgs.jormungandr;
+    defaultPackage = pkgs.jormungandr;
+    inherit overlay;
+  });
```
jonringer, line 33: I would export this outside of the platform tags.
```suggestion
  }) // {
    inherit overlay;
  };
```

jormungandr.nix
```diff
@@ -0,0 +1,19 @@
+{ lib, rustPlatform, fetchFromGitHub, pkg-config, openssl, protobuf, rustfmt }:
+let
+  cargoPackage = (builtins.fromTOML (builtins.readFile ./jormungandr/Cargo.toml)).package;
+in
+rustPlatform.buildRustPackage rec {
+  pname = cargoPackage.name;
+  version = cargoPackage.version;
+  src = ./.;
+  cargoSha256 = "sha256-rfgmtKweYxk8BELv2YoF3BsGfdGlQGP8HNLE4xklDHs=";
+  nativeBuildInputs = [ pkg-config protobuf rustfmt ];
```
mzabaluev, line 10: A hint on possible optimization: you can omit `rustfmt` if you build jormungandr without default features
(like `cargo build --manifest-path jormungandr/Cargo.toml --no-default-features` does). `rustfmt` is used by default to format some generated code so that developers can browse it.

PR #2929, opened 2021-01-19
Slog to Tracing migration

danielSanchezQ:
Remove Slog usage in favor of Tracing. 
Implements #997


jormungandr/src/log/asyncable_drain.rs
```diff
@@ -1,21 +1,21 @@
...
+// const EVENT_BUFFER_SIZE: usize = 1024;
+//
+// impl<D: Drain + Send + 'static> AsyncableDrain for D
+// where
+//     D::Err: Debug,
+// {
+//     fn into_async(self) -> Async {
+//         Async::new(self.fuse()).chan_size(EVENT_BUFFER_SIZE).build()
+//     }
+// }
```
danielSanchezQ, line 21: Leftovers I need to remove.

jormungandr/src/log/stream.rs
```diff
@@ -1,108 +1,108 @@
...
+//                     &inner.record_static,
+//                     &format_args!("{}", inner.message),
+//                     b!("item" => ?item),
+//                 ));
+//                 Poll::Ready(Some(item))
+//             }
+//             None => Poll::Ready(None),
+//         }
+//     }
+// }
```
danielSanchezQ, line 108: More leftovers

mzabaluev: Just kill it.

danielSanchezQ:
Looks like failing tests are due to the grpc socket bug

testing/jormungandr-integration-tests/src/jormungandr/grpc/client_tests.rs
```diff
@@ -370,6 +370,11 @@ pub async fn pull_blocks_correct_hashes_all_blocks() {
     let blocks_hashes: Vec<Hash> = blocks.iter().map(|x| x.header.hash()).collect();
 
     let block_hashes_from_logs = server.logger.get_created_blocks_hashes();
+    std::fs::copy(
```
Zeegomo, line 373: testing leftover?

danielSanchezQ: Indeed.

jormungandr/src/main.rs
```diff
@@ -9,14 +9,10 @@ extern crate lazy_static;
...
-#[macro_use]
-extern crate slog;
 #[cfg(feature = "gelf")]
-extern crate slog_gelf;
+extern crate tracing_gelf;
 #[cfg(feature = "systemd")]
-extern crate slog_journald;
-#[cfg(unix)]
-extern crate slog_syslog;
+extern crate tracing_journald;
```
mzabaluev, line 15: Can we just drop this pre-2018 stuff?

jormungandr/src/settings/logging.rs
```diff
@@ -83,125 +75,138 @@ impl FromStr for LogOutput {
...
-
-                            break Err(Error::SyslogAccessFailed(e));
-                        }
-                    }
+            LogOutput::Stderr => {
+                let (subscriber, guard) = tracing_appender::non_blocking(std::io::stderr());
+                let builder = builder.with_writer(subscriber).with_max_level(*level);
+                match format {
+                    LogFormat::Plain => Ok((Box::new(builder.finish()), Some(guard))),
+                    LogFormat::Json => Ok((Box::new(builder.json().finish()), Some(guard))),
```
mzabaluev, line 183: This logic could be unified.

jormungandr/src/settings/logging.rs
```diff
@@ -83,125 +75,138 @@ impl FromStr for LogOutput {
...
+                log_id: _graylog_source,
             } => {
                 // Both currently recognized formats can be understood to apply:
                 // GELF formats payloads in JSON so 'json' is redundant,
                 // and plain messages are worked into JSON just the same.
                 // Match them irrefutably so that any new format will need to
                 // be addressed here when added.
-                match format {
-                    LogFormat::Plain | LogFormat::Json => {}
-                };
```
mzabaluev, line 201: The comment applied to the irrefutable match statement that was a guard against adding a format in the future that is not already supported by graylog. Please restore this.

mzabaluev (changes requested):

jormungandr/src/blockchain/candidate.rs
```diff
@@ -141,11 +139,10 @@ where
         // when that can pre-validate headers without
         // up-to-date ledger.
         chain::pre_verify_link(&header, &self.parent_header)?;
-        debug!(
-            self.logger,
-            "adding block to fetch";
-            "hash" => %block_hash,
-            "parent" => %parent_hash,
+        tracing::debug!(
```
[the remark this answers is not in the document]
danielSanchezQ: I did it on purpose. It may be less stylish but I think it is easy to track where those logs are and what we are using for them.

jormungandr/src/main.rs
danielSanchezQ: Definitely

jormungandr/src/settings/logging.rs
danielSanchezQ: What do you mean here?

jormungandr/src/settings/logging.rs
danielSanchezQ: I wanted to ask about this when I first saw it but then forgot to did it in the middle of mayhem. 
If gelf already use its own format which we cannot change, why bother with the format anyway? Just in case we want to assert that configuration is "correct", otherwise ignoring it should be good enough.

mzabaluev: It's just that we have this format option that is not entirely orthogonal to the log backend.

Maybe it would be better to just ignore the format and warn that it has no effect when the backend is anything other than stdout/stderr/file.

jormungandr/src/settings/logging.rs
mzabaluev: Handling of stdout and stderr is pretty much the same code, so maybe the match arm should be `LogOutput::Stdout | LogOutput::Stderr` and then a common body. But it can stay as it is, not much of a concern.

jormungandr/src/settings/logging.rs
mzabaluev: Added a "default" format that is chosen by default (naturally), means the same as "plain" on file descriptor appender backends, and is the only format allowed on journald and gelf.

jormungandr/src/blockchain/candidate.rs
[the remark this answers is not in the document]
mzabaluev: It must be _the_ logging framework within the node, I don't think we have any other uses of the log macros.
Maybe it was good for interim editing, so OK.

mzabaluev (approved):

.github/workflows/test.yml
```diff
@@ -100,23 +99,35 @@ jobs:
...
+            echo "CARGO_FLAGS=$CARGO_FLAGS --release" >> $GITHUB_ENV
+          fi
+          case '${{ matrix.os }}' in
+            ubuntu-latest)  features=systemd,gelf ;;
+            *)              features=gelf ;;
+          esac
+          echo "JORMUNGANDR_FEATURES=$features" >> $GITHUB_ENV
+
+          mkdir .cargo
+          cat > .cargo/config.toml <<EOF
```
eugene-babichenko, line 127: We may just fill in `RUSTFLAGS` instead of putting everything into a file.

eugene-babichenko (commented):
Apart from some nits in CI this is looking good. Although I doubt somebody would review it fully because it is huge :)

.github/workflows/test.yml
mzabaluev: The failed build on 6469cf99538b119eb1767114cd75237b2cf39955 is a demonstration why we can't: `RUSTFLAGS` are also used when building things like proc macros and build scripts, and some of these targets fail to compile when told to use LTO.

A Cargo config file is also a neat way to replace conditionals in scripts or step descriptors: the `[profile.release]` section only applies to release builds. The same trick is used in the release workflow to enable LTO precisely for the target binaries.

jormungandr/src/settings/logging.rs
```diff
@@ -151,37 +152,51 @@ impl LogSettings {
...
+            format: LogFormat,
+        ) -> (Box<dyn Subscriber + Send + Sync>, Option<WorkerGuard>) {
+            let (subscriber, guard) = tracing_appender::non_blocking(writer);
+            let builder = builder.with_writer(subscriber).with_max_level(level);
+            let subscriber: Box<dyn Subscriber + Send + Sync> = match format {
+                LogFormat::Plain => Box::new(builder.finish()),
+                LogFormat::Json => Box::new(builder.json().finish()),
+            };
+            (subscriber, Some(guard))
+        }
```
danielSanchezQ: Lovely. Everything was separated at some point in the refactor. But the problems I had making the dynamic dispatching work made me use excessive duplicated code indeed. Thanks!

jormungandr/src/fragment/process.rs
```diff
@@ -55,61 +55,62 @@ impl Process {
...
-                    reply_handle.reply_ok(contents);
+                        pool.insert_and_propagate_all(origin, txs)
+                            .await
+                            .map(move |count| stats_counter.add_tx_recv_cnt(count))?;
+                    }
+                    TransactionMsg::RemoveTransactions(fragment_ids, status) => {
+                        tracing::debug!(
+                            "removing fragments added to block {:?}: {:?}",
+                            status,
+                            fragment_ids
```
danielSanchezQ, line 83: I guess this would work. But for what I check on the library the way of formatting is using [`sigils`](https://tracing.rs/tracing/#recording-fields), at least for fields. It should be ok in this case.

danielSanchezQ (commented):
Cleanup, rebase and CI looks good. Thanks!

Issue #2926, opened 2021-01-15
Multiple gRPC failures on MacOS tests


Zeegomo:
**Describe the bug**
Multiple failures on grpc client tests with the same error:
```tonic::transport::Error(Transport, hyper::Error(Connect, ConnectError("tcp connect error", Os { code: 49, kind: AddrNotAvailable, message: "Can\'t assign requested address" })))'```

**Mandatory Information**
1. `jcli --full-version` master;
2. `jormungandr --full-version` master;

**To Reproduce**
I cannot reproduce this since it seems to happen only on MacOS and not even all the times. [example run](https://github.com/input-output-hk/jormungandr/pull/2924/checks?check_run_id=1708637013).

**Expected behavior**
Those tests should not fail spuriously

**Additional context**
Some other occurrences of this error code seems to be related to either a large amount of connections (https://github.com/hyperium/hyper/issues/1757), or a bug in the Os ([openvpn faq](https://www.ovpn.com/en/faq/troubleshooting/mac-cant-assign-requested-address-code49)). The solution proposed by OpenVPN is to reboot or flush the routing table. 

A not very precise measurement (`ss --all | grep TIME-WAIT | wc -l`) suggests we are using a peak of about 2.5k ports during testing (the peak seems also to be related to grpc test execution, during other phases it's more like a couple hundred ports).

PR #2930, opened 2021-01-19
Reduce the number of ephemeral ports used during testing

Zeegomo:
This PR reduce the number of ephemeral ports used in testing by reusing the same grpc client instead of creating a new one for each request. This hopefully helps for #2926


mzabaluev (approved):

mzabaluev:
Looks good. Can't get down to the test failure causes due to problems with logging. The failures seem to be random and are probably unrelated.

mzabaluev (approved):

PR #2935, opened 2021-01-20
Improve diagnostics of fragment rejection errors

mzabaluev:
That rabbit hole goes deep, we need to drill down to a few layers of errors to se what went wrong with voting.
Luckily, they are already represented in the source chain.


dkijania (approved):

dkijania:
working as a charm. thank you!

eugene-babichenko (approved):

PR #2939, opened 2021-01-21
Fix documentation and help for the jcli votes subcommand

mzabaluev:
Take out from a hands-on session with devops.


danielSanchezQ (approved):
Thanks for the clarifications and fixes.

PR #2946, opened 2021-01-25
[Test] fix non-functional voting test

dkijania:


testing/jormungandr-scenario-tests/src/node.rs
```diff
@@ -531,8 +532,9 @@ impl Node {
 
         NodeController {
             alias: self.alias().clone(),
-            grpc_client: JormungandrClient::from_address(&p2p_address)
-                .expect("cannot setup grpc client"),
+            grpc_client: runtime::Runtime::new().unwrap().block_on(async {
```
Zeegomo, line 535: I would put this wrapper in `passive_node_explorer`, since this will only fix the creation of the client but all other function of the client will fail

dkijania: We have Context struct which i believe can holds runtime object. However we are lucky because grpc_client is not used at this moment. I need this interim fix to be merged, not to block ~25 scenario tests. I would create follow up issue for fixing it in a proper way

dkijania:
This is interim solution. Complementary fix will be provided under story: #2948

PR #2960, opened 2021-01-27
[Test] Fragment load test maintenance

dkijania:
Attempt to fix :

`test_legacy_node_all_fragments`
`send_all_fragments`

tests by adjusting voteplan/votetally certs timing


testing/jormungandr-integration-tests/src/common/jcli/services/fragments_check.rs
```diff
@@ -5,6 +5,7 @@ use crate::common::{
 };
 use jormungandr_lib::interfaces::FragmentLog;
 use jortestkit::prelude::ProcessOutput;
+use jortestkit::prelude::Wait;
```
mzabaluev, line 8: Why have a prelude if you import the names explicitly anyway?

mzabaluev (commented):
Build fails, no red flags otherwise.

PR #2961, opened 2021-01-27
[Tests] ignore failing tests until storage fix would be applied

dkijania:


mzabaluev:
Is there a bug issue for storage that this could be linked to?

dkijania:
I Will add it

Issue #2950, opened 2021-01-26
Explorer panic when sending `status` graphql request


dkijania:
**Describe the bug**
During test `non_functional::voting::tally_vote_load_test` we are retrieving current block date from explorer. After newest changed there is a panic during request handling in explorer. 
Request:

```
query Status{ 
    status{
        latestBlock {  
            id,  
            chainLength,  
            date {  
                epoch {  
                    id  
                },
                slot  
            },    
            treasury {
                rewards,
                treasury
            } 
        },
        feeSettings {
            constant,
            coefficient,
            certificate,
            certificatePoolRegistration,
            certificateStakeDelegation,
            certificateOwnerStakeDelegation,
            certificateVotePlan,
            certificateVoteCast
        }
    }
}
```

Result:

```
"query Status{ \n    status{\n        latestBlock {  \n            id,  \n            chainLength,  \n            date {  \n                epoch {  \n                    id  \n                },\n                slot  \n            },    \n            treasury {\n                rewards,\n                treasury\n            } \n        },\n        feeSettings {\n            constant,\n            coefficient,\n            certificate,\n            certificatePoolRegistration,\n            certificateStakeDelegation,\n            certificateOwnerStakeDelegation,\n            certificateVotePlan,\n            certificateVoteCast\n        }\n    }\n}", against: http://127.0.0.1:10739/explorer/graphql
thread 'tokio-runtime-worker' panicked at 'the longest chain to be indexed', jormungandr/src/explorer/mod.rs:462:33
stack backtrace:
   0:     0x55655f5cc5b0 - std::backtrace_rs::backtrace::libunwind::trace::h04d12fdcddff82aa
                               at /rustc/e1884a8e3c3e813aada8254edfa120e85bf5ffca/library/std/src/../../backtrace/src/backtrace/libunwind.rs:100:5
   1:     0x55655f5cc5b0 - std::backtrace_rs::backtrace::trace_unsynchronized::h1459b974b6fbe5e1
                               at /rustc/e1884a8e3c3e813aada8254edfa120e85bf5ffca/library/std/src/../../backtrace/src/backtrace/mod.rs:66:5
   2:     0x55655f5cc5b0 - std::sys_common::backtrace::_print_fmt::h9b8396a669123d95
                               at /rustc/e1884a8e3c3e813aada8254edfa120e85bf5ffca/library/std/src/sys_common/backtrace.rs:67:5
[...]

PR #2962, opened 2021-01-27
Explorer add gc and confirmed block notion

enzoc4:
This adds the notion of `epoch_stability_depth` to the explorer. It's used both for multiverse garbage collection and to know if a block is confirmed.

Also, when working on this I found a problem with the way tip changes were handled in #2903

The thing is, tip changes are being sent to the explorer before having the block. This means, that we can't get the State at that particular hash, so we don't have the block and we can't know it's chain_length to find out which one is confirmed.

Also means that a query may come in the middle and try to use a branch not fully indexed and break things like @dkijania also found in #2950

This PR fixes it in a way that may be a bit hacky. It works by saving the tip update for the time the block comes, if it's not already there, so it doesn't care which one comes first.

A proper solution could be to send blocks to the explorer after knowing the branch solution result, but the problem is that by that point the node only has a `Ref` and not a `Block`, so that would require either carrying the block there, or sending the branch comparison result back (up in the stack trace).

An alternative could be to only send hashes, and get blocks from the storage, but I think this is blocking? and maybe not fast.

### DIFF
--- a/jormungandr/src/blockchain/multiverse.rs
+++ b/jormungandr/src/blockchain/multiverse.rs
@@ -1,4 +1,4 @@
-use crate::blockcfg::{ChainLength, HeaderHash, Ledger, Multiverse as MultiverseData};
+use crate::blockcfg::{ChainLength, HeaderHash, Multiverse as MultiverseData};
 use chain_impl_mockchain::multiverse;
 use std::sync::Arc;
 use tokio::sync::RwLock;
@@ -43,12 +43,9 @@ impl<T: Clone> Multiverse<T> {
     }
 }
 
-impl Multiverse<Ledger> {
+impl<T> Multiverse<T> {
     /// run the garbage collection of the multiverse
     ///
-    /// TODO: this function is only working for the `Ledger` at the moment
-    ///       we need to generalize the `chain_impl_mockchain` to handle
-    ///       the garbage collection for any `T`
     pub async fn gc(&self, depth: u32) {
         let mut guard = self.inner.write().await;
         guard.gc(depth)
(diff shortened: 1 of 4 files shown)


jormungandr/src/blockchain/multiverse.rs
```diff
@@ -43,12 +43,9 @@ impl<T: Clone> Multiverse<T> {
     }
 }
 
-impl Multiverse<Ledger> {
+impl<T> Multiverse<T> {
```
eugene-babichenko, line 46: Huh? I was pretty sure I already did that.

jormungandr/src/explorer/mod.rs
```diff
@@ -328,6 +343,45 @@ impl ExplorerDB {
...
+            Arc::clone(state.blocks.lookup(&hash).unwrap())
+        } else {
+            return false;
+        };
+
+        if let Some(confirmed_block_chain_length) = block
+            .chain_length()
+            .nth_ancestor(self.blockchain_config.epoch_stability_depth)
+        {
+            debug_assert!(
```
eugene-babichenko, line 361: Why not just `assert!`? Might be difficult to track problems in production.

eugene-babichenko (approved):

jormungandr/src/blockchain/multiverse.rs
enzoc4: Hmmm, probably you did it the same with the actual multiverse in chain-libs

jormungandr/src/explorer/mod.rs
enzoc4: Well, I just thought debug_assert was paranoid enough, and I wanted to write the invariant in some way at least. I don't really mind using `assert` anyway, but if this actually happens it is most likely a programmer error.

PR #2970, opened 2021-01-29
Start tokio runtime for scenario tests

Zeegomo:
Since some parts used in testing require a tokio runtime, it seems convenient to just start it at the beginning.

Also need to update tokio used in scenario tests to 1.1,  I see https://github.com/input-output-hk/jormungandr/pull/2965 is already doing that, so we can wait for it to be merged.


dkijania:
please fix `cargo clippy` issue. apart from that, everything is ok

dkijania (approved):

Issue #2972, opened 2021-01-29
Offline tallying in CLI


mzabaluev:
Implement offline computation of a private tally in jcli and add the results to the tally transaction payload modified as per #2971.

PR #2980, opened 2021-02-01
Add jcli command to tally all proposals in a vote plan

Zeegomo:
Here I am keeping the old commands for compatibility reasons and I would like to remove them if not needed, but I think they are used in tests at the moment.

Intermediate steps outputs (generate shares, tally private vote) are only needed for inter-command communication, it probably makes sense if the format is fixed (i.e. always in json)

Working on decrypt shares I got a bit confused by `Vec<Vec<Share>>` or `Vec<Share>` since it's not obvious what the structure of that is, and they actually mean different things in different contexts (e.g. is `Vec<Share>` all the shares for a vote plan of a committee member or is it all the shares of multiple committee members for a single proposal), so I added some more types to clarify that.

This pr is becoming quite big, I don't know what the policy is, if I should split it up or navigating diffs by commits is enough.

First step for https://github.com/input-output-hk/jormungandr/issues/2972

### DIFF
--- a/jcli/src/jcli_app/vote/mod.rs
+++ b/jcli/src/jcli_app/vote/mod.rs
@@ -40,10 +40,20 @@ pub enum Error {
     InvalidCommitteMemberIndex,
     #[error("failed to read encrypted tally bytes")]
     EncryptedTallyRead,
+    #[error("failed to read vote plans")]
+    VotePlansRead,
+    #[error("no vote plan found with specified id")]
+    VotePlanIdNotFound,
+    #[error("please specify a vote plan id if multiple vote plans are provided")]
+    UnclearVotePlan,
     #[error("failed to read decryption key bytes")]
     DecryptionKeyRead,
     #[error("failed to read share bytes")]
     DecryptionShareRead,
+    #[error("missing shares for a proposal")]
+    MissingShares,
+    #[error(transparent)]
+    TallyError(#[from] chain_vote::TallyError),
     #[error(transparent)]
     FormatError(#[from] crate::jcli_app::utils::output_format::Error),
     #[error(transparent)]
--- a/jcli/src/jcli_app/vote/mod.rs
+++ b/jcli/src/jcli_app/vote/mod.rs
@@ -1,4 +1,5 @@
 use crate::jcli_app::utils::output_file::{self, OutputFile};
+use jormungandr_lib::interfaces::Tally;
 
 pub mod bech32_constants;
 mod committee;
@@ -50,6 +51,8 @@ pub enum Error {
     DecryptionKeyRead,
     #[error("failed to read share bytes")]
     DecryptionShareRead,
+    #[error("expected encrypted private tally, found {found:?}")]
+    PrivateTallyExpected { found: Option<Tally> },
     #[error("missing shares for a proposal")]
     MissingShares,
     #[error(transparent)]
--- a/jcli/src/jcli_app/vote/mod.rs
+++ b/jcli/src/jcli_app/vote/mod.rs
@@ -1,4 +1,5 @@
 use crate::jcli_app::utils::output_file::{self, OutputFile};
+use crate::jcli_app::utils::vote::{SharesError, VotePlanError};
 use jormungandr_lib::interfaces::Tally;
 
 pub mod bech32_constants;
@@ -41,26 +42,20 @@ pub enum Error {
     InvalidCommitteMemberIndex,
     #[error("failed to read encrypted tally bytes")]
     EncryptedTallyRead,
-    #[error("failed to read vote plans")]
-    VotePlansRead,
-    #[error("no vote plan found with specified id")]
-    VotePlanIdNotFound,
-    #[error("please specify a vote plan id if multiple vote plans are provided")]
-    UnclearVotePlan,
     #[error("failed to read decryption key bytes")]
     DecryptionKeyRead,
-    #[error("failed to read share bytes")]
-    DecryptionShareRead,
     #[error("expected encrypted private tally, found {found:?}")]
     PrivateTallyExpected { found: Option<Tally> },
-    #[error("missing shares for a proposal")]
-    MissingShares,
     #[error(transparent)]
     TallyError(#[from] chain_vote::TallyError),
     #[error(transparent)]
     FormatError(#[from] crate::jcli_app::utils::output_format::Error),
     #[error(transparent)]
     JsonError(#[from] serde_json::Error),
+    #[error(transparent)]
+    VotePlanError(#[from] VotePlanError),
+    #[error(transparent)]
+    SharesError(#[from] SharesError),
 }
 
 #[derive(StructOpt)]
--- a/jcli/src/jcli_app/vote/mod.rs
+++ b/jcli/src/jcli_app/vote/mod.rs
@@ -1,6 +1,5 @@
 use crate::jcli_app::utils::output_file::{self, OutputFile};
 use crate::jcli_app::utils::vote::{SharesError, VotePlanError};
-use jormungandr_lib::interfaces::Tally;
 
 pub mod bech32_constants;
 mod committee;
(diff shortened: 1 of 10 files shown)


jcli/src/jcli_app/vote/tally/decrypt_shares.rs
```diff
@@ -31,9 +30,54 @@ pub struct TallyDecryptWithAllShares {
     output_format: OutputFormat,
 }
 
+#[derive(StructOpt)]
+#[structopt(rename_all = "kebab-case")]
+pub struct TallyVotePlanWithAllShares {
+    /// The path to json-encoded vote plan to decrypt. If this parameter is not
+    /// specified, the vote plan will be read from the standard
+    /// input.
+    #[structopt(long = "vote-plan")]
```
eugene-babichenko, line 39: You don't actually need to do that if you use `rename_all`.

jcli/src/jcli_app/vote/tally/decrypt_shares.rs
```diff
@@ -31,9 +30,54 @@ pub struct TallyDecryptWithAllShares {
...
+    result: Vec<u64>,
+}
+
+fn read_shares_from_file<P: AsRef<Path>>(
+    share_path: &Option<P>,
+    threshold: usize,
+    proposals: usize,
+) -> Result<Vec<Vec<chain_vote::TallyDecryptShare>>, Error> {
+    let shares: Vec<Vec<String>> = serde_json::from_reader(io::open_file_read(share_path)?)?;
+    if shares[0].len() < threshold || shares.len() != proposals {
```
eugene-babichenko, line 67: There is no guarantee that all underlying vectors are of correct length if `shares[0].len() >= threshold` is satisfied. IMO it is better to move that check to the `.map()` call below. I guess this pattern might be repeated in a few other places.

jcli/src/jcli_app/vote/tally/decrypt_shares.rs
```diff
@@ -31,9 +30,54 @@ pub struct TallyDecryptWithAllShares {
...
+                .map(|share| {
+                    chain_vote::TallyDecryptShare::from_bytes(&base64::decode(share)?)
+                        .ok_or(Error::DecryptionShareRead)
+                })
+                .collect::<Result<Vec<_>, Error>>()
+        })
+        .collect::<Result<Vec<_>, Error>>()
 }
 
 impl TallyDecryptWithAllShares {
```
eugene-babichenko, line 83: This command decrypts only a single proposal, right? If so, please document it and note that we should remove that later.

jcli/src/jcli_app/vote/tally/decryption_tally.rs
```diff
@@ -20,31 +22,116 @@ pub struct TallyGenerateDecryptionShare {
     decryption_key: PathBuf,
 }
 
+/// Create decryption shares for all proposals in a vote plan.
+///
+/// The decryption share data will be printed in hexadecimal encoding
+/// on standard output.
+#[derive(StructOpt)]
+#[structopt(rename_all = "kebab-case")]
+pub struct TallyGenerateVotePlanDecryptionShares {
```
eugene-babichenko, line 31: Same `rename_all` comment applies here.

eugene-babichenko (commented):
Apart from some nitpicks this is a great job!

eugene-babichenko (approved):

jcli/src/jcli_app/vote/mod.rs
```diff
@@ -42,12 +44,18 @@ pub enum Error {
     EncryptedTallyRead,
     #[error("failed to read decryption key bytes")]
     DecryptionKeyRead,
-    #[error("failed to read share bytes")]
-    DecryptionShareRead,
+    #[error("expected encrypted private tally, found {found:?}")]
```
mzabaluev, line 47: As a general rule, error messages should not use `Debug` impls. Also including the whole `Tally` enum into the error seems wasteful, especially as its longest variant should never occur here. Maybe cut the `found` type down to what can be displayed in a simple way, use an expression to conditionally format the message, or restructure the cases for "expected encrypted tally, found none" and "expected encrypted tally, found public tally".

jcli/src/jcli_app/vote/tally/decrypt_shares.rs
```diff
@@ -31,9 +34,32 @@ pub struct TallyDecryptWithAllShares {
...
+    #[structopt(long)]
+    vote_plan: Option<PathBuf>,
+    /// The id of the vote plan to decrypt.
+    /// Can be left unspecified if there is only one vote plan in the input
+    #[structopt(long)]
+    vote_plan_id: Option<Hash>,
+    /// The minimum number of shares needed for decryption
+    #[structopt(long, default_value = "3")]
+    threshold: usize,
+    /// The path to json-encoded necessary base64 shares. If this parameter is not
```
mzabaluev, line 52:
```suggestion
    /// The path to a JSON file containing decryption shares necessary to decrypt the vote plan.
    /// If this parameter is not
```

jcli/src/jcli_app/vote/tally/mod.rs
```diff
@@ -12,18 +12,34 @@ pub enum Tally {
     /// The decryption share data will be printed in hexadecimal encoding
     /// on standard output.
     DecryptionShare(decryption_tally::TallyGenerateDecryptionShare),
+    /// Create a decryption share for private voting tally.
+    ///
+    /// The decryption share data will be printed in hexadecimal encoding
+    /// on standard output.
+    VotePlanDecryptionShares(decryption_tally::TallyGenerateVotePlanDecryptionShares),
```
mzabaluev, line 19: If we're going to deprecate `decryption-share` above, just name the subcommand `decryption-shares`.
Nobody needs the carpal tunnel syndrome typing these commands.

jcli/src/jcli_app/vote/tally/mod.rs
```diff
@@ -12,18 +12,34 @@ pub enum Tally {
...
     /// Decrypt a tally with decryption shares.
     ///
     /// The decrypted tally data will be printed in hexadecimal encoding
     /// on standard output.
     Decrypt(decrypt_shares::TallyDecryptWithAllShares),
+    /// Decrypt all proposals in a vote plan.
+    ///
+    /// The decrypted tally data will be printed in hexadecimal encoding
+    /// on standard output.
+    DecryptVotePlan(decrypt_shares::TallyVotePlanWithAllShares),
```
mzabaluev, line 32: The name is a bit misleading: we're decrypting tally results, not the vote plan.
`tally decrypt-results` and deprecate `tally decrypt`?

mzabaluev (changes requested):

jcli/src/jcli_app/vote/tally/mod.rs
```diff
@@ -29,16 +29,16 @@ pub enum Tally {
...
 
 impl Tally {
     pub fn exec(self) -> Result<(), Error> {
         match self {
             Tally::DecryptionShare(cmd) => cmd.exec(),
-            Tally::VotePlanDecryptionShares(cmd) => cmd.exec(),
+            Tally::DecryptionShares(cmd) => cmd.exec(),
             Tally::Decrypt(cmd) => cmd.exec(),
-            Tally::DecryptVotePlan(cmd) => cmd.exec(),
+            Tally::DecryptResults(cmd) => cmd.exec(),
```
mzabaluev: I wonder if we could hide the deprecated commands from help, but still have them supported in the CLI.

mzabaluev (approved):
Good to go.

jcli/src/jcli_app/vote/tally/mod.rs
Zeegomo: Can we remove them altogether once we have tested the new ones? I do not see much value in keeping those, and we would have to maintain more functionalities.

PR #2981, opened 2021-02-01
Add tracking tips explorer multiverse

enzoc4:
Intermediate step, see #2963 

This essentially changes the explorer backend/indexing, but without adding the breaking changes to graphql.

This may conflict with #2962, it's not really dependent, but it touches the same things. I would merge #2962 first anyway, so maybe I'll rebase this one on top.

# Backend

- Add tip tracking multiverse wrapper (proxy)
- Move branch dependent methods/queries to `State`, so the `ExplorerDB` can be used like `get_branch(hash)` or `get_main_branch(hash)`. Also, change the graphql queries so they use `get_main_branch`, so this PR still retains most of the current behaviour.

# GraphQL API

The changes to the `graphql/mod.rs` are only to make things still compile, most queries keep the current behavior except for

- block by id searches in all candidate branches.
- transaction by id searches in all candidate branches.
- vote plan by id searches in all candidate branches.

breaking changes will come in the next PR.

Issue #2963, opened 2021-01-27
Make explorer API reflect the multiverse better


enzoc4:
#2801 was getting too large/ambitious, so I'm closing it and organizing things a bit more in a few PR's and here.

# Motivation

Have the explorer API reflect the multiverse better. This requires keeping track of active branches, be able to search for things in multiple branches, and I'd say adding garbage collection because it also affects the API a bit, among other things.

# Plans

## In GraphQL API

- Change block by id query so it searches in all candidate branches.
- Change transaction by id query so it searches in all candidate branches.
- Change vote plan by id query so it searches in all candidate branches.
- Add a `Branch` graphql type. A Branch is identified by a `Block` hash, but not all blocks are branches, particularly because for old stable/confirmed blocks we may not have the whole State anymore.
- Add query for current tip (this exists already, but to return `Branch`).
- Add query for all tips (or candidate tips).
- Move queries to Branch: all_blocks (paginated block history), transactions_by_address, blocks per epoch (maybe just as a filter in all_blocks?)
- Change `Transaction::block` to `Transaction::blocks` (in graphql, rust syntax is just illustrative)
- Add `Block::branches` to kwow which branches is it in, and by having the chain_length of the tips, an idea of how _confirmed_ it is.
- Change block by chain length query to return a list of blocks when needed.
- Expose epoch stability depth in graphql.
- Add `Block::is_confirmed` (#2962)
- Add query for last confirmed block (#2962)

## In explorer backend

- Add generalized multiverse gc and call it (#2962) .
- Add tip tracking multiverse wrapper (proxy).
- Move branch dependent methods to `State`, so the `ExplorerDB` can be used like `get_branch(hash)` or `get_main_branch(hash)`. Also, just in case, we need to make sure to keep the `Ref` alive during the query, although it would only have strange behavior in edge cases anyway.

# PR's

Explorer add gc and confirmed block notion #2962

PR #2987, opened 2021-02-02
Add graphql branch and breaking changes

enzoc4:
see #2963

This branches from #2981 instead of master because that contains the core logic needed for this to work, but I think the separation is helpful because:

a) This is the PR that involves the graphql breaking changes, so merging the previous one by itself is a bit safer.
b) This can be reviewed independently in general, as just looking at the `schema.graphql` diff gives an idea of what this does. The `jormungandr/explorer/graphql/mod.rs` serves the same purpose, although it's a bit more verbose as I had to move entire functions to the `Branch` type, and that causes a big diff even if the implementation is basically the same.

# Idea

With these changes it should be possible to:

Know which block is the current tip.
Know which blocks are considered tips (which means, which blocks are the branches).
Make queries contextualized to a branch, either just whatever the current one is, or by providing a hash.

## Breaking changes

- `allBlocks` (now just `blocks,` probably should change the other ones too), `allStakePools`, `allVotePlans` are now in `Branch` instead of `Query` (the graphql root)
- `Address { transactions }` is now in `Branch` too, under `transactionsByAddress`.
- Same thing for `Epoch { blocks }`, under `blocksByEpoch`
- `Transaction { block }` is now `Transaction { blocks }` (in plural), so it returns a list instead.
- `blockByChainLength` is now `blocksByChainLength` (in plural), so it returns [Block!]! instead of Block.
- `status` is now `settings` and `Status` is now `Settings`, as it was a weird type anyway. It also doesn't contain the latest block, as that can be queried through `Query { mainTip { block { .. } } }` now.

## Non-breaking changes

- `mainTip` and `tips`. These allow to retain the current behavior of most queries by using `mainTip` to scope the results.
- `branch` can be used to get a branch by it's hash. This will return a not found error if the branch is not in memory anymore.
- Add `Address { confirmed_transactions }`. Although, this is a bit extra.

## Conceptual

This introduces a `Branch` graphql type. It's identified by the hash of a `Block`. The difference between the two is that the `Block` is the block by itself, let's say, a collection of transactions with a parent pointer. The `Branch` represents the multiverse state produced by the application of that block. 
The distinction is mostly because the idea is to garbage collect the multiverse, but that also includes states that are in the mainchain, so you may have a `Block` (on disk, when we implement that) that has no `Branch` (in memory), which means we can't answer queries that depend on the state at that point (tx's by address, delegation, let's say). Although, it's probably possible to design a db (using chain length to timestamp/filter) schema that can account for this, I'm not sure if it is an actual use-case.


jormungandr/src/explorer/graphql/mod.rs
```diff
@@ -19,39 +19,391 @@ use super::indexing::{
...
+    }
+
+    fn from_contents(block: Arc<ExplorerBlock>) -> Block {
+        Block {
+            hash: block.id(),
+            contents: tokio::sync::Mutex::new(Some(block)),
+        }
     }
 
     async fn get_explorer_block(&self, db: &ExplorerDB) -> FieldResult<Arc<ExplorerBlock>> {
```
mzabaluev, line 376: Could rename to `fetch_explorer_block` to reflect that it caches the retrieved block now.

jormungandr/src/explorer/graphql/mod.rs
```diff
@@ -255,72 +613,75 @@ impl From<blockcfg::BlockDate> for BlockDate {
...
+        Ok(result)
     }
 
     async fn get_contents(&self, context: &Context) -> FieldResult<ExplorerTransaction> {
         if let Some(c) = &self.contents {
             Ok(c.clone())
         } else {
-            let block = self.get_block(context).await?;
+            //TODO: maybe store transactions outside blocks? as Arc, as doing it this way is pretty wasty
+            let block = &self.get_blocks(context).await?[0];
```
mzabaluev, line 684: Can there be a `get_block` operation if you already know you can use a block with the specific hash?

jormungandr/src/explorer/graphql/mod.rs
```diff
@@ -1328,6 +1520,30 @@ impl Query {
         Transaction::from_id(id, context).await
     }
 
+    /// get all current tips, sorted (descending) by their length
+    async fn tips(&self, context: &Context) -> Vec<Branch> {
```
mzabaluev, line 1524: We use `tip` in the node protocol for the currently selected branch.
Maybe use `branch_heads` or just `branches`, considering the return type is `Vec<Branch>`?

jormungandr/src/explorer/graphql/mod.rs
```diff
@@ -1328,6 +1520,30 @@ impl Query {
...
+            .await
+            .iter()
+            .cloned()
+            .map(|(id, state_ref)| Branch::from_id_and_state(id, state_ref))
+            .collect()
+    }
+
+    /// get the block that the ledger currently considers as the main branch's
+    /// tip
+    async fn main_tip(&self, context: &Context) -> Branch {
```
mzabaluev, line 1537: So this is just `tip` in the node protocol, let's be consistent.

testing/jormungandr-testing-utils/resources/explorer/graphql/schema.graphql
```diff
@@ -188,25 +202,35 @@ scalar PublicKey
...
   block(id: String!): Block!
-  blockByChainLength(length: ChainLength!): Block
-
-  """query all the blocks in a paginated view"""
-  allBlocks(first: Int, last: Int, before: IndexCursor, after: IndexCursor): BlockConnection!
+  blocksByChainLength(length: ChainLength!): [Block!]!
   transaction(id: String!): Transaction!
+
+  """get all current tips, sorted (descending) by their length"""
+  tips: [Branch!]!
```
mzabaluev, line 209: So this should be `branches` to be consistent with the node protocol's terminology...

testing/jormungandr-testing-utils/resources/explorer/graphql/schema.graphql
```diff
@@ -188,25 +202,35 @@ scalar PublicKey
...
+
+  """
+  get the block that the ledger currently considers as the main branch's
+  tip
+  """
+  mainTip: Branch!
```
mzabaluev, line 215: And this is `tip`.

testing/jormungandr-testing-utils/resources/explorer/graphql/schema.graphql
```diff
@@ -313,11 +336,16 @@ type VoteOptionRange {
...
+  proof: String!
+  encryptedVote: String!
+}
+
 type VotePayloadPublicStatus {
   choice: Int!
 }
 
-union VotePayloadStatus = VotePayloadPublicStatus
+union VotePayloadStatus = VotePayloadPublicStatus | VotePayloadPrivateStatus
```
mzabaluev, line 349: Good, we missed that.

jormungandr/src/explorer/graphql/mod.rs
enzoc4: I'm not sure if I understand. There will always be at least a block there, you mean a function that gets that one?

To be honest, this query is not designed really well, getting the block to get the transaction body is kinda bad, because it also means transactions are duplicated if they are in multiple blocks, which is kinda a waste of memory.

Fetching the entire block to get only one transaction it's probably no ideal either.

I didn't want to beat around the bush too much in this PR, so I just left what has equivalent behavior to the previous thing. (Also, extracting transaction bodies from the blocks is not that trivial anyway, because of gc and other things).

jormungandr/src/explorer/graphql/mod.rs
enzoc4: So, now that you bring that up... I guess what I'm calling `Branch` in the explorer/api is more like what's called `Ref` in the ledger. 

I'm not entirely convinced that name makes sense for the public api, though.

And also I don't think that queries should be limited to heads.

But maybe it makes sense to call the thing `Ref`, `StateRef`, or something like that (not *commit* unfortunately).

mzabaluev: `Ref` is OK in internal code, but for the public explorer API we should stick to the nomenclature used elsewhere. I think "braches" are close in meaning to the concept used in git (and git has a `HEAD` alias for what we mean by the "tip").

jormungandr/src/explorer/graphql/mod.rs
mzabaluev: I mean it's already wasteful to retrieve several blocks if you only need one, and you can just pick the first of `self.block_hashes` and query just that block by its hash.

Optimizing to be able to query individual fragments straight from the store/db would be good (and might be possible with some zero-copy block storage access and iteration on raw fragments in a raw block), but this is a topic for a more substantial rework that would involve the storage layer, so not for this PR. Cc @eugene-babichenko

enzoc4: oh, that... Yeah, sure, makes sense.

Well, considering that the explorer doesn't even use the storage, yeah, that's a big change. I'm not sure we want to do that, though? And using raw fragments wouldn't play that nice with the way we show utxo inputs now, I guess.

jormungandr/src/explorer/graphql/mod.rs
```diff
@@ -681,7 +681,17 @@ impl Transaction {
...
             //TODO: maybe store transactions outside blocks? as Arc, as doing it this way is pretty wasty
-            let block = &self.get_blocks(context).await?[0];
+
+            let block = context
+                .db
+                .get_block(&self.block_hashes[0])
+                .await
+                .ok_or_else(|| {
+                    FieldError::from(ErrorKind::InternalError(
+                        "missing transaction's block body".to_owned(),
```
mzabaluev, line 691:
```suggestion
                        "failed to fetch block containing the transaction".to_owned(),
```

mzabaluev (approved):

Issue #2989, opened 2021-02-03
Implement VotePlanDef serialization 


dkijania:
**Is your feature request related to a problem/context ? Please describe if applicable.**
During tests there is a need to dump vote plan that was put into block0. structure: `jormungandr-lib/interfaces/vote/VotePlanDef` currently can be only deserialized. I would like to have an option to serialize it

dkijania:
@Zeegomo thank you!

PR #2991, opened 2021-02-04
Add serialization to VotePlanDef

Zeegomo:
Does not directly implement `Serialize` but add an associated function with the same signature (because of how serde remote types work).

I think some other ser/de functions can be refactored but It's probably a good idea to delay that until after the freeze if we think it's valuable.

Closes https://github.com/input-output-hk/jormungandr/issues/2989

### DIFF
--- a/jormungandr-lib/src/interfaces/vote.rs
+++ b/jormungandr-lib/src/interfaces/vote.rs
@@ -19,9 +19,7 @@ use std::convert::TryInto;
 use std::fmt::Formatter;
 use std::str;
 
-#[derive(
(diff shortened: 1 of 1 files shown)


eugene-babichenko (approved):

Zeegomo:
I saw some (inconsistent) failures of `jormungandr::vit::public::test_vote_flow_bft` also in my local env, I want to investigate if its related (and how).

mzabaluev:
`test_vote_flow_*` failures are a longstanding issue, most likely #2834.

PR #2992, opened 2021-02-04
Check the correctness of archives and attach checksums to releases

eugene-babichenko:


.github/workflows/release.yml
```diff
@@ -284,12 +299,26 @@ jobs:
...
         run: |
           $archive = "jormungandr-${{ needs.create_release.outputs.version }}-${{ matrix.config.target }}-${{ matrix.target_cpu }}.zip"
+
+          (Get-FileHash -Path "./target/${{ matrix.config.target }}/release/jormungandr.exe" -Algorithm SHA256).Hash.toLower() | Out-File -FilePath "./jormungandr.exe.sha256"
+          (Get-FileHash -Path "./target/${{ matrix.config.target }}/release/jcli.exe" -Algorithm SHA256).Hash.toLower() | Out-File -FilePath "./jcli.exe.sha256"
+
           $args = @{
             Path  = "./target/${{ matrix.config.target }}/release/jormungandr.exe",
-                    "./target/${{ matrix.config.target }}/release/jcli.exe"
+                    "./jormungandr.exe.sha256",
```
mzabaluev, line 308: I don't think it's worth putting these hash files into the archive, as you hash the whole archive and publish the checksum file as well, that covers the need for an integrity check. Just write them into `${{ runner.temp }}` for the sanity check below.

mzabaluev (commented):
Good work!

.github/workflows/release.yml
```diff
@@ -273,8 +273,23 @@ jobs:
       - name: Pack binaries (Unix)
         if: matrix.config.os != 'windows-latest'
         run: |
+          set -e
+
           archive=jormungandr-${{ needs.create_release.outputs.version }}-${{ matrix.config.target }}-${{ matrix.target_cpu }}.tar.gz
+
+          sha256sum ./target/${{ matrix.config.target }}/release/jormungandr | awk '{print $1}' > ./jormungandr.sha256
```
mzabaluev, line 280: Unfortunately, sha256sum is not available out of the box on macOS. Either [install coreutils with brew](https://unix.stackexchange.com/a/484429/415360) or do it with `shasum -a 256` specifically on macOS.

.github/workflows/release.yml
```diff
@@ -273,8 +273,23 @@ jobs:
...
+          sha256sum ./target/${{ matrix.config.target }}/release/jcli | awk '{print $1}' > ./jcli.sha256
+
           tar -C ./target/${{ matrix.config.target }}/release -czvf $archive jormungandr jcli
+
+          mkdir ./archive-test
+          tar -C ./archive-test -xzvf $archive
+          echo "$(cat ./jormungandr.sha256) ./archive-test/jormungandr" | sha256sum --check
+          echo "$(cat ./jcli.sha256) ./archive-test/jcli" | sha256sum --check
```
mzabaluev, line 288: You can check them in batch:

```sh
# $SHA256SUM is set to 'sha256sum' on Linux and to 'shasum -a 256' on macOS

cd target/${{ matrix.config.target }}/release/
$SHA256SUM jormungandr jcli > ${{ runner.temp }}/archive-checksum
...
cd ${{ runner.temp }}/archive-test
$SHA256SUM -c ${{ runner.temp }}/archive-checksum
```

mzabaluev (changes requested):
To check how it works, you can test the workflow in your fork (or just here, no biggie) by pushing to the `ci/test/nightly` branch.

.github/workflows/release.yml
mzabaluev: Conditionalizing to do these sanity checks only on macOS is simplest and the insanity is only lurking there.

eugene-babichenko: Weird enough, but I also have the `shasum` command on my machine. It is a Perl script,  trying to understand where is it coming from.

ci/pack-assets.py
```diff
@@ -0,0 +1,78 @@
...
+if jcli1_checksum != jcli_checksum:
+    print('jcli checksum mismarch: before {} != after {}'.format(
+        jcli_checksum, jcli1_checksum))
+    exit(1)
+
+# save archive checksum
+archive_checksum = sha256sum(archive_name)
+checksum_filename = '{}.sha256'.format(archive_name)
+with open(checksum_filename, 'x') as f:
+    f.write(archive_checksum)
```
mzabaluev, line 74: It could be good to save in the checksum file format:

```suggestion
    f.write('{}  {}'.format(archive_checksum, archive_name))
```

mzabaluev (approved):
Looks good! Have you tested it with a push to `ci/test/nightly`?

ci/pack-assets.py
```diff
@@ -0,0 +1,78 @@
...
+    exit(1)
+if jcli1_checksum != jcli_checksum:
+    print('jcli checksum mismarch: before {} != after {}'.format(
+        jcli_checksum, jcli1_checksum))
+    exit(1)
+
+# save archive checksum
+archive_checksum = sha256sum(archive_name)
+checksum_filename = '{}.sha256'.format(archive_name)
+with open(checksum_filename, 'x') as f:
```
mzabaluev, line 73: Does this open in text mode on Windows?

mzabaluev: Yep, weirdly it's the default in Python and I keep forgetting that.

eugene-babichenko:
> Looks good! Have you tested it with a push to `ci/test/nightly`?

will do

ci/pack-assets.py
eugene-babichenko: `print(archive_checksum, archive_name, file=f)` to make it nicer, but yeah. Should bring the compatibility with coreutils. I am not sure about Windows though, they tend to ignore such conventions, same goes for macOS. Unless we decide to treat them as second-class citizens :trollface:.

eugene-babichenko:
The test build is looking good.

ci/pack-assets.py
mzabaluev: `shasum` works fine with the `-c` option on macOS?
Windows does not have a built-in tool to parse a checksum file and verify the checksums out of the box, but I think there's a reasonably short Powershell command line to do it.

PR #2996, opened 2021-02-08
expose discrimination settings in blockchain template

dkijania:


testing/jormungandr-testing-utils/src/testing/network_builder/wallet/template/mod.rs
```diff
@@ -13,30 +14,49 @@ pub struct WalletTemplate {
...
     }
 
     #[inline]
-    fn new<S: Into<WalletAlias>>(alias: S, value: Value, wallet_type: WalletType) -> Self {
+    fn new<S: Into<WalletAlias>>(
+        alias: S,
+        value: Value,
+        wallet_type: WalletType,
+        discrimination: Discrimination,
+    ) -> Self {
```
mzabaluev, line 42: Looks like another case for a builder API, especially if most tests would like the default discrimination (`Test`)?

mzabaluev (commented):
Generally OK, but the build and the lints need to be fixed.

PR #3003, opened 2021-02-09
ci: use thin LTO in the release workflow and disable LTO in the test workflow

eugene-babichenko:
This should make our builds considerably faster.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -227,7 +227,7 @@ jobs:
           mkdir .cargo
           cat > .cargo/config.toml <<EOF
           [target.${{ matrix.config.target }}]
-          rustflags = ["-C", "target-cpu=${{ matrix.target_cpu }}", "-C", "lto", "-C", "embed-bitcode=yes"]
+          rustflags = ["-C", "target-cpu=${{ matrix.target_cpu }}", "-C", "lto=thin", "-C", "embed-bitcode=yes"]
           EOF
 
       - if: ${{ matrix.cross }}
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -100,17 +100,26 @@ jobs:
       fail-fast: false
       matrix:
         os: [windows-latest, ubuntu-latest, macos-latest]
-        flags: ['', '--release']
+        release: [true, false]
         toolchain: [stable]
         include:
           - os: ubuntu-latest
-            flags: ''
+            release: false
             toolchain: nightly
     env:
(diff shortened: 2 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -227,7 +227,7 @@ jobs:
           mkdir .cargo
           cat > .cargo/config.toml <<EOF
           [target.${{ matrix.config.target }}]
-          rustflags = ["-C", "target-cpu=${{ matrix.target_cpu }}", "-C", "lto", "-C", "embed-bitcode=yes"]
+          rustflags = ["-C", "target-cpu=${{ matrix.target_cpu }}", "-C", "lto=thin", "-C", "embed-bitcode=yes"]
```
Zeegomo, line 230: Should we use `linker-plugin-lto` as well instead of `embed-bitcode`?

eugene-babichenko: The use of plugin-based LTO is not that straightforward because we have some non-Rust code being linked to our Rust code by dependencies. "By the book" approach requires quite a bit of manual work without any obvious performance benefit. 10 seconds on a single run do not really count. Although there's still a room for improvement with `CFLAGS`.

I am leaving the commands and their execution times below for a reference:

* Naive LTO approach (`4m19s`): `RUSTFLAGS='-C lto=thin -C embed-bitcode=yes' cargo build --release --bin jormungandr --manifest-path jormungandr/Cargo.toml --target x86_64-unknown-linux-gnu --no-default-features`
* Plugin-based LTO (`4m29s`): `CC='clang' CFLAGS='-flto=thin -fembed-bitcode' RUSTFLAGS='-Clinker-plugin-lto -Clinker=clang -Clink-arg=-fuse-ld=lld' cargo build --release --bin jormungandr --manifest-path jormungandr/Cargo.toml --target x86_64-unknown-linux-gnu --no-default-features`

Zeegomo (approved):

.github/workflows/release.yml
eugene-babichenko: Hm, actually, the whole point is not the compilation time improvements but making better LTO when building from multiple languages with `clang`. Which is not exactly the case, but we can switch C code build to `clang` easily. Might be interesting to actually measure the final performance improvement on some example. But this should definitely be done as a separate task.

Zeegomo: I was actually referring to the compilation speedup as suggested by the rustc reference, but your point is indeed interesting.

PR #3016, opened 2021-02-11
fix grpc tests

Zeegomo:
When testing for expected results in grpc tests, we should not require client and server blocks to be exactly the same, because the server could have produced one or more blocks in the meanwhile, but rather check that the client has a long enough prefix of the server blocks. This happened for example in https://github.com/danielSanchezQ/jormungandr/pull/7/checks?check_run_id=1878326379

While at it, most of the functionalities were tested with just one block in the chain so I though of waiting a bit in order to make tests more realistic. `pull_blocks_to_tip_incorrect_hash` and `pull_headers_empty_start_hash` are failing after this change and may need investigation


mzabaluev (approved):
Looks great.

PR #3018, opened 2021-02-11
update all dependencies to use cryptoxide 0.3.1

eugene-babichenko:
this prevents compilation errors with future versions of rustc

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -407,7 +407,7 @@ source = "git+https://github.com/input-output-hk/chain-libs.git?branch=master#d0
 dependencies = [
  "cbor_event",
  "chain-ser",
- "cryptoxide 0.3.0",
+ "cryptoxide",
  "ed25519-bip32",
 ]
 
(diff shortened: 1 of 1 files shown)


Zeegomo:
I'm a little confused on what happens if the version in Cargo.lock is more recent than the version in Cargo.toml (chain-crypto is on 0.2.1)? I think this should be handled by updating the version used in dependencies,  since it's not used directly in jormungandr.

 Or was this a temporary fix?

eugene-babichenko:
IIRC `cargo update` automatically decides on updates compatibility based on semver numbers if you do not specify more strict constraints along with the version number.

eugene-babichenko:
But that dependency in chain-libs still needs an update.

Zeegomo (approved):
It looks like I was wrong, the only dependency that still used < 0.3 cryptoxide was symmetric-cipher, that you updated as well in this pr. Good to go then!

PR #3019, opened 2021-02-11
fix new clippy 1.50 lints

Zeegomo:


jormungandr/src/explorer/mod.rs
```diff
@@ -483,7 +483,7 @@ fn apply_block_to_blocks(blocks: Blocks, block: &ExplorerBlock) -> Result<Blocks
         .map_err(|_| Error::BlockAlreadyExists(block_id))
 }
 
-fn apply_block_to_addresses(mut addresses: Addresses, block: &ExplorerBlock) -> Result<Addresses> {
+fn apply_block_to_addresses(mut addresses: Addresses, block: &ExplorerBlock) -> Addresses {
```
mzabaluev, line 486: Odd that this is infallible, but it's an internal method, so OK.

mzabaluev (approved):

PR #3022, opened 2021-02-13
Migrate juniper to async graphql

### LINKED ISSUE #2945: Migrate from juniper to async-graphql and update warp to 0.3
juniper is holding back #2621 and is behind in features we need (such as built-in support for pagination).

enzoc4:
closes #2945 

branches from #2987 to make this simpler. ~Either merge that one first or merge this and close that one.~

There a few minor differences in the schema for some scalar types, but the overall structure is the same.

# Notes

## Playground

I replaced graphiql with graphql playground because judging from the async-graphql examples, that's what they favor. Although, I think graphiql is bundled too, so maybe we should expose both?

## Tip subscription

I wanted to test the subscription integration too, so I just went and implemented a very simple one for the `main_tip` changes. It's relatively minor, and in a self-contained commit, so it's easy to take it off and add it later if someone wants too.

## Context

The way the context is handled is different in async-graphql, previously the resolvers received context as it is declared in `explorer/graphql/mod.rs` because the lock was acquired at the warp filter level. I didn't find a way to do the same with the way `async-graphql-warp` expects to be integrated, mostly because warp filters cannot be 'flatmapped'. This means all resolvers received a wrapped **context** and they need to lock it themselves.

# TODO

- There is still code repetition for the custom scalars (mostly are newtypes). I'm still not sure what's the best way of removing that.
- ~It would be neat to reduce the code repetition for context management, but I'm not sure if it can be done.~


mzabaluev:
Please rebase once #3123 is merged.

jormungandr/src/explorer/graphql/certificates.rs
```diff
@@ -1,14 +1,14 @@
 use super::error::ErrorKind;
+use async_graphql::*;
```
mzabaluev, line 2: Please don't use glob imports.

jormungandr/src/explorer/graphql/certificates.rs
```diff
@@ -1,14 +1,14 @@
...
 
 use super::scalars::{PayloadType, PoolId, PublicKey, TimeOffsetSeconds, VotePlanId};
-use super::{Address, BlockDate, Context, ExplorerAddress, Pool, Proposal, TaxType};
-use juniper::FieldResult;
+use super::{Address, BlockDate, ExplorerAddress, Pool, Proposal, TaxType};
+use crate::rest::explorer::EContext as RestContext;
 
 // interface for grouping certificates as a graphl union
-#[derive(GraphQLUnion)]
+#[derive(Union)]
```
mzabaluev, line 11:
```suggestion
#[derive(async_graphql::Union)]
```
This would look much more descriptive. However, this is a graphql service module, so it's OK just to
import the names explicitly in the top-level `use` statement so that the origin of these derive macros can be seen in the source file.

jormungandr/src/explorer/graphql/mod.rs
```diff
@@ -131,99 +152,142 @@ impl Branch {
...
     pub async fn all_stake_pools(
         &self,
         first: Option<i32>,
         last: Option<i32>,
-        before: Option<IndexCursor>,
-        after: Option<IndexCursor>,
-    ) -> FieldResult<PoolConnection> {
+        before: Option<String>,
+        after: Option<String>,
+    ) -> FieldResult<Connection<IndexCursor, Pool, ConnectionFields<PoolCount>, EmptyFields>> {
```
mzabaluev, line 290: So the `before`/`after` parameters are not typed now because of the proc macro, but there is still an `IndexCursor` type in the return?
Not going into details, I assume this is how async-graphql restricts these pagination APIs.

testing/jormungandr-testing-utils/resources/explorer/graphql/epoch.graphql
```diff
@@ -1,4 +1,4 @@
-query Epoch($id: EpochNumber!, $blocks_limit: Int!) {
+query Epoch($id: Int!, $blocks_limit: Int!) {
```
mzabaluev, line 1: No way to continue representing this as a custom scalar type?

testing/jormungandr-testing-utils/resources/explorer/graphql/schema.graphql
```diff
@@ -17,7 +20,7 @@ type Block {
   date: BlockDate!
 
   """The transactions contained in the block"""
-  transactions(first: Int, last: Int, before: IndexCursor, after: IndexCursor): TransactionConnection!
+  transactions(first: Int, last: Int, before: String, after: String): TransactionConnection!
```
mzabaluev, line 23: Is this within conventions on the paginated connection queries?
I guess the custom `IndexCursor` scalar was easy to do with juniper, but if a custom type is not required, leave it like this.

testing/jormungandr-testing-utils/resources/explorer/graphql/schema.graphql
```diff
@@ -98,9 +102,6 @@ type FeeSettings {
   certificateVoteCast: Value!
 }
 
-"""Non-opaque cursor that can be used for offset-based pagination"""
-scalar IndexCursor
-
```
mzabaluev, line 103: Interesting why these scalar types needed to be dropped, but other types remain.

testing/jormungandr-testing-utils/resources/explorer/graphql/schema.graphql
```diff
@@ -109,11 +110,19 @@ type OwnerStakeDelegation {
   pools: [Pool!]!
 }
 
+"""Information about pagination in a connection"""
 type PageInfo {
-  hasNextPage: Boolean!
+  """When paginating backwards, are there more items?"""
```
mzabaluev, line 115: Yay, docs! This is a test schema, or can this also be normative and used as a doc?

testing/jormungandr-testing-utils/resources/explorer/graphql/schema.graphql
```diff
@@ -239,6 +247,10 @@ type StakeDistribution {
   pools: [PoolStakeDistribution!]!
 }
 
+type Subscription {
+  mainTip: Branch!
```
mzabaluev, line 251: `mainTip` -> `tip` like elsewhere?

mzabaluev (changes requested):
Great work. Glob imports are considered harmful, and there are a few small issues with the schema. Removal of some scalar types is a bit unfortunate, but not a show-stopper.

Also, this needs a rebase again :unamused:

testing/jormungandr-testing-utils/resources/explorer/graphql/schema.graphql
enzoc4: Well, docs were always a thing, even with juniper. This file is actually generated with the introspection system (I mean, we launch a jormungandr instance with the explorer enabled, and ask it for the schema), so it's no different than the schema that's actually served, and this docs can be seen from the playground, for example

testing/jormungandr-testing-utils/resources/explorer/graphql/schema.graphql
enzoc4: Well, to be honest I'm thinking that maybe I should retain the custom scalar, although maybe just as a `String` wrapper, as the semantic here is a bit important.
So, the graphql connection documentation recommends to use opaque cursors, they even recommend base64 encode whatever you use for pagination, so using `String` is actually fine. But when we were designing this originally, we wanted the ability to implement page-based pagination too, and skip to intermediate pages instead of only up and down. So I just named the thing `IndexCursor` and put the comment saying that it was actually an index (or offset) just in case.

testing/jormungandr-testing-utils/resources/explorer/graphql/schema.graphql
enzoc4: nice catch

testing/jormungandr-testing-utils/resources/explorer/graphql/epoch.graphql
enzoc4: yeah, I was experimenting with newtypes to see if they were mapped as custom scalars and then I forgot, but I prefer the custom scalar too I guess

jormungandr/src/explorer/graphql/mod.rs
enzoc4: The `query` macro expects strings, that's why I changed the user-facing api too, because it will perform validation twice, otherwise. 

I'm not sure why couldn't the macro accept anything that implements `CursorType`, to be honest, as it lets you return it.

Now I'm thinking that it is a bit weird to have the input cursor types be different than the type in the `Connection`, so I think I'll change it so they are both `IndexCursor`. (which is not against the spec, anyway, and just requires unwrapping a newtype before passing to the macro)

testing/jormungandr-testing-utils/resources/explorer/graphql/schema.graphql
danielSanchezQ: Pagination is a feature in `async-graphql`, but not in juniper (where @enzoc4 has to implement himself) iirc.

testing/jormungandr-testing-utils/resources/explorer/graphql/schema.graphql
mzabaluev: I'm rather bothered by potential mistakes that were prevented by using newtypes.

testing/jormungandr-testing-utils/resources/explorer/graphql/epoch.graphql
mzabaluev: Would it be difficult to add?

jormungandr/src/explorer/graphql/mod.rs
```diff
@@ -1542,47 +1593,72 @@ impl Query {
...
-        let (hash, state_ref) = context.db.get_tip().await;
+    /// get the block that the ledger currently considers as the main branch's
+    /// tip
+    async fn tip(&self, context: &Context<'_>) -> Branch {
+        let (hash, state_ref) = extract_context(&context).await.db.get_tip().await;
         Branch::from_id_and_state(hash, state_ref)
     }
 
-    async fn branch(&self, id: String, context: &Context) -> FieldResult<Branch> {
+    pub async fn branch(&self, context: &Context<'_>, id: String) -> FieldResult<Branch> {
```
mzabaluev, line 1603: Looks a bit inconsistent that the branch ID is a `String`, but this PR does not change it.

testing/jormungandr-testing-utils/resources/explorer/graphql/epoch.graphql
mzabaluev: Would that be difficult to fix?

testing/jormungandr-testing-utils/resources/explorer/graphql/schema.graphql
```diff
@@ -29,36 +32,35 @@ type Block {
...
 }
 
+"""An edge in a connection."""
 type BlockEdge {
+  """The item at the end of the edge"""
   node: Block!
 
   """A cursor for use in pagination"""
-  cursor: IndexCursor!
+  cursor: String!
```
mzabaluev, line 55: I guess you have to decustomize `IndexCursor` for _reasons_, even though it's present in the queue parameters.

testing/jormungandr-testing-utils/resources/explorer/graphql/schema.graphql
```diff
@@ -208,10 +220,13 @@ type Query {
...
 
-  """get the ref that the ledger currently considers as the main branch's"""
+  """
+  get the block that the ledger currently considers as the main branch's
+  tip
+  """
   tip: Branch!
   branch(id: String!): Branch!
-  epoch(id: EpochNumber!): Epoch!
+  epoch(id: Int!): Epoch!
```
mzabaluev, line 229: Another de-customization of `EpochNumber` here.

testing/jormungandr-testing-utils/resources/explorer/graphql/schema.graphql
enzoc4: Hmm, this is weird, actually... So it seems `async-graphql` let's you use a particular type as a cursor, but it puts `String` in the schema regardless... Ok, I guess I'll just get rid of `IndexCursor` altogether.

mzabaluev (approved):

PR #3028, opened 2021-02-15
fix compilation issue in enclave.rs

dkijania:
When doing `cargo install --path jormungandr` some compilation issues pops up. This PR intends to fix them


mzabaluev (commented):
A non-breaking way to do it is to run `cargo update` and also commit the modified `Cargo.lock`.

Using option `--locked` with `cargo install` would also remove the need to make these changes.

Issue #3017, opened 2021-02-11
gRPC methods not behaving as expected


Zeegomo:
**Describe the bug**
tests `pull_blocks_to_tip_incorrect_hash` and `pull_headers_empty_start_hash` fails because `grpc::pull_blocks_to_tip` with an incorrect starting hash and `pull_headers` with an empty start hash only returns info about the tip instead of the whole chain

**To Reproduce**
0. checkout updated grpc tests from https://github.com/input-output-hk/jormungandr/pull/3016
1. cargo test jormungandr::grpc::client_tests::pull_blocks_to_tip_incorrect_hash
2. cargo test jormungandr::grpc::client_tests::pull_headers_empty_start_hash

**Example output**:
```
pull_headers_empty_start_hash

[...]

PR #3033, opened 2021-02-16
Refactor grpc tests

Zeegomo:
Refactor gRPC tests and uniform expected behavior for range queries as discussed in https://github.com/input-output-hk/jormungandr/issues/3017

### DIFF
--- a/testing/jormungandr-integration-tests/src/jormungandr/grpc/client_tests.rs
+++ b/testing/jormungandr-integration-tests/src/jormungandr/grpc/client_tests.rs
@@ -60,7 +60,9 @@ pub async fn handshake_sanity() {
 // L1006 Tip request
 #[tokio::test]
 pub async fn tip_request() {
-    let setup = setup::client::default().await;
+    let setup =
+        setup::client::bootstrap(ConfigurationBuilder::new().with_slot_duration(9).to_owned())
+            .await;
 
     let tip_header = setup.client.tip().await;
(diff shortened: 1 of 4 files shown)


testing/jormungandr-integration-tests/src/jormungandr/grpc/client_tests.rs
```diff
@@ -68,12 +60,15 @@ pub async fn handshake_sanity() {
...
-    let fixture = Fixture::default();
-    let (server, config) = fixture.bootstrap_node();
-    let client = Config::attach_to_local_node(config.get_p2p_listen_port()).client();
+    let setup =
+        setup::client::bootstrap(ConfigurationBuilder::new().with_slot_duration(9).to_owned())
+            .await;
 
-    let tip_header = client.tip().await;
-    let block_hashes = server.logger.get_created_blocks_hashes();
+    while setup.client.tip().await.chain_length() < 1.into() {
```
mzabaluev, line 67: This could wait indefinitely if the mock does not do the right thing?

The whole loop looks like it could be put into an utility function, but then this pattern occurs in other unchanged code, so let's leave it for later refactoring.

mzabaluev (approved):

testing/jormungandr-integration-tests/src/jormungandr/grpc/client_tests.rs
Zeegomo: correct, I will refactor it in an utility function with a timeout.

Issue #3041, opened 2021-02-18
Command line log configuration options should override the YAML config


mzabaluev:
When log backends are configured both in command line and in the configuration file, the effective configuration is a sum of the configured outputs. This is surprising and often leads to annoying results like each line output twice if stdout/stderr log settings are given in both ways.

Change the behavior so that whenever an output is specified via command line options, it replaces all outputs listed in the configuration file settings. It's also worthwhile to log an `info` notice about the override in case the configuration file has non-empty logger configuration.

The method implementation to change is https://github.com/input-output-hk/jormungandr/blob/909882b3df81bcf83920ac3b00e8ee82976a904e/jormungandr/src/settings/start/mod.rs#L62-L62

mzabaluev:
Cc @dkijania

PR #3042, opened 2021-02-18
Cli log config overrides yaml

filip-dulic-bloxico:
- changes logic for configuring log settings
- instead of appending CLI log settings to those read from a YAML file,
  when log output settings are defined in the CLI, they override those coming from a YAML file.

Implements https://github.com/input-output-hk/jormungandr/issues/3041

### DIFF
--- a/jormungandr/src/settings/logging.rs
+++ b/jormungandr/src/settings/logging.rs
@@ -134,7 +134,7 @@ impl Subscriber for BoxedSubscriber {
 impl Layer<BoxedSubscriber> for BoxedSubscriber {}
 
 impl LogSettings {
-    pub fn init_log(self) -> Result<Vec<WorkerGuard>, Error> {
+    pub fn init_log(self) -> Result<(Vec<WorkerGuard>, LogInfoMsg), Error> {
         use tracing_subscriber::prelude::*;
         let mut guards = Vec::new();
         let mut layers: Vec<Layered<_, BoxedSubscriber>> = Vec::new();
@@ -161,7 +161,7 @@ impl LogSettings {
                 .map_err(Error::SetGlobalSubscriberError)?;
         }
 
-        Ok(guards)
+        Ok((guards, self.1))
     }
 }
 
--- a/jormungandr/src/settings/logging.rs
+++ b/jormungandr/src/settings/logging.rs
@@ -16,7 +16,12 @@ use tracing_subscriber::fmt::SubscriberBuilder;
 use tracing_subscriber::layer::SubscriberExt;
 use tracing_subscriber::layer::{Layer, Layered};
 
-pub struct LogSettings(pub Vec<LogSettingsEntry>);
+pub struct LogSettings(pub Vec<LogSettingsEntry>, pub LogInfoMsg);
+
+/// A wrapper to return an optional string message that we
+/// have to manually log with `info!`, we need this because
+/// some code executes before the logs are initialized.
+pub type LogInfoMsg = Option<String>;
 
 #[derive(Debug)]
 pub struct LogSettingsEntry {
--- a/jormungandr/src/settings/start/mod.rs
+++ b/jormungandr/src/settings/start/mod.rs
@@ -60,21 +60,6 @@ impl RawSettings {
     }
 
     pub fn log_settings(&self) -> LogSettings {
-        // FIXME: remove this after refactoring is done
-        // changing the flow for creating the LogSettings inner Vec to this:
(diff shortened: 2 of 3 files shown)


jormungandr/src/settings/start/mod.rs
```diff
@@ -92,6 +83,45 @@ impl RawSettings {
...
+
+        if let Some(output) = cmd_output {
+            let cmd_level = self.command_line.log_level;
+            let cmd_format = self.command_line.log_format;
+
+            // Read the log settings, and look if we already
+            // have this output configured
+            for entry in entries.iter_mut() {
+                // If the output is aleady configured:
+                if &entry.output == &output {
```
mzabaluev: Eh, I should have specified more clearly: if the log output is set in the command line, it replaces all entries from config, rather than just modifying settings for any matching output entry.

jormungandr/src/settings/start/mod.rs
```diff
@@ -92,6 +83,45 @@ impl RawSettings {
...
+                        level: cmd_level.unwrap_or(entry.level),
+                        format: cmd_format.unwrap_or(entry.format),
+                        output: output.clone(),
+                    };
+                    // log to info! that the output was overriden
+                    tracing::info!(
+                        "Log settings overriden from cli: {:?} replaced with {:?}",
+                        entry,
+                        override_entry
+                    );
```
mzabaluev, line 111: The trouble with this is that the subscribers are not initialized yet. Maybe it's best to add an `Option` member to `LogSettings` with these details, then, after the global subscriber has been set, check it and inform.

jormungandr/src/settings/start/mod.rs
```diff
@@ -92,6 +83,45 @@ impl RawSettings {
...
+                    // overwrite entry.level if cli_entry.level is Some(level)
+                    // overwrite entry.format if cli_entry.format is Some(format)
+                    let override_entry = LogSettingsEntry {
+                        level: cmd_level.unwrap_or(entry.level),
+                        format: cmd_format.unwrap_or(entry.format),
+                        output: output.clone(),
+                    };
+                    // log to info! that the output was overriden
+                    tracing::info!(
+                        "Log settings overriden from cli: {:?} replaced with {:?}",
```
mzabaluev, line 108:
```suggestion
                        "log settings overriden from command line: {:?} replaced with {:?}",
```

jormungandr/src/settings/start/mod.rs
mzabaluev: Another problem is that the default entry is created in the code above if the config list is empty. In that case, we don't want a warning.

mzabaluev (changes requested):

jormungandr/src/settings/logging.rs
```diff
@@ -16,7 +16,12 @@ use tracing_subscriber::fmt::SubscriberBuilder;
 use tracing_subscriber::layer::SubscriberExt;
 use tracing_subscriber::layer::{Layer, Layered};
 
-pub struct LogSettings(pub Vec<LogSettingsEntry>);
+pub struct LogSettings(pub Vec<LogSettingsEntry>, pub LogInfoMsg);
+
+/// A wrapper to return an optional string message that we
+/// have to manually log with `info!`, we need this because
+/// some code executes before the logs are initialized.
+pub type LogInfoMsg = Option<String>;
```
mzabaluev, line 24: I think this way is more self-descriptive and does not require a type alias (OTOH you use `LogInfoMsg` on its own in other places):

```suggestion
pub struct LogSettings {
    /// Configured log outputs.
    pub entries: Vec<LogSettingsEntry>,
    /// An optional string message that should be logged after the logs are initialized.
    pub override_warning: Option<String>,
}

```

jormungandr/src/settings/start/mod.rs
```diff
@@ -72,18 +74,33 @@ impl RawSettings {
...
-            });
+                output,
+            };
+            // log to info! that the output was overriden,
+            // we send this as a message because tracing Subscribers
+            // do not get initiated until after this code runs
+            log_info_msg = Some(format!(
+                "log settings overriden from command line: {:?} replaced with {:?}",
+                entries, command_line_entry
+            ));
```
mzabaluev, line 96: This will leave a message also when no logs are configured in `config.yaml` which is quite OK, so to be perfect, this should be under an `if !entries.is_empty() { ... }` branch.

mzabaluev (approved):

PR #3051, opened 2021-02-22
Ignore possible invalid utf-strings in logs

Zeegomo:
I think failures like https://github.com/input-output-hk/jormungandr/runs/1952567767?check_suite_focus=true#step:11:2131 might be caused by concurrent access to the log file (maybe we changed log backend and writes are not atomic anymore?). This just discards such lines, if that is the case an operation will eventually read the right value.

### DIFF
--- a/testing/jormungandr-testing-utils/src/testing/node/logger.rs
+++ b/testing/jormungandr-testing-utils/src/testing/node/logger.rs
@@ -283,7 +283,7 @@ impl JormungandrLogger {
         let file = File::open(self.log_file_path.clone())
             .unwrap_or_else(|_| panic!("cannot find log file: {:?}", &self.log_file_path));
         let reader = BufReader::new(file);
-        reader.lines().map(|line| line.unwrap())
+        reader.lines().flat_map(|line| line.ok())
     }
 
     pub fn get_log_entries(&self) -> impl Iterator<Item = LogEntry> + '_ {


testing/jormungandr-testing-utils/src/testing/node/logger.rs
```diff
@@ -283,7 +283,7 @@ impl JormungandrLogger {
         let file = File::open(self.log_file_path.clone())
             .unwrap_or_else(|_| panic!("cannot find log file: {:?}", &self.log_file_path));
         let reader = BufReader::new(file);
-        reader.lines().map(|line| line.unwrap())
+        reader.lines().flat_map(|line| line.ok())
```
eugene-babichenko, line 286: Might be better to use `filter_map` there.

eugene-babichenko:
I don't like the fact that we are ignoring the actual problem here. This problem is worth fixing.

testing/jormungandr-testing-utils/src/testing/node/logger.rs
Zeegomo: definitely swapped them by mistake

Zeegomo:
You mean making sure that the writes are all atomic? This might be indeed useful for other processes that want to inspect how the node's doing.
As per tests, it's not actually a problem if we miss one line because it's being written, since we do not have particular synchronization guarantees and it will simply appear as a delayed write.

eugene-babichenko:
> it's not actually a problem if we miss one line because it's being written

This might be the line a particular test depends on. Skipping a line might lead to random test failures.

Zeegomo:
Superseded by https://github.com/input-output-hk/jormungandr/pull/3074

PR #3053, opened 2021-02-23
make v1/fragments/logs and v1/fragments/statuses visible

dkijania:


jormungandr/src/rest/v1/mod.rs
```diff
@@ -12,9 +12,9 @@ pub fn filter(
     let root = warp::path!("v1" / ..);
 
     let fragments = {
-        let root = warp::path!("fragments");
+        let root = warp::path!("fragments" / ..);
 
-        let post = warp::path::end()
+        let post = warp::path!("post")
```
eugene-babichenko, line 17: This change is not needed. The first one already fixes the problem.

eugene-babichenko (changes requested):

jormungandr/src/rest/v1/mod.rs
```diff
@@ -12,10 +12,9 @@ pub fn filter(
     let root = warp::path!("v1" / ..);
 
     let fragments = {
-        let root = warp::path!("fragments");
+        let root = warp::path!("fragments" / ..);
 
         let post = warp::path::end()
-            .and(warp::post())
```
eugene-babichenko, line 18: Why remove the post filter? It shouldn't work at all.

dkijania: you're right ! sorry

eugene-babichenko (approved):

PR #3062, opened 2021-02-25
[Tests] take vote plan from settings not from blockchain

dkijania:


testing/jormungandr-testing-utils/src/wallet/committee.rs
```diff
@@ -208,4 +216,46 @@ impl PrivateVoteCommitteeDataManager {
...
+            })
+            .collect::<Vec<_>>();
+
+        let absolute_max_votes = encrypted_tally
+            .iter()
+            .map(|(_encrypted_tally, max_votes)| *max_votes)
+            .max()
+            .unwrap();
+        let table =
+            chain_vote::TallyOptimizationTable::generate_with_balance(absolute_max_votes, 1);
```
Zeegomo, line 238: It's better if you leave the default balance here if you are not sure there will always be only one proposal.

dkijania: Very good catch! This is a copy-paste issue. so basically instead of 1 there should be number of proposals ?

Zeegomo: the default is 2, but I suggest to use `generate(...)` that take care of this if we ever want to change that default. `generate_with_balance` should be used only if you have to decrypt either only 1 proposal or a lot of them.

PR #3066, opened 2021-02-26
Improve dynamic composition of tracing subscribers

filip-dulic-bloxico:
Attempt at making dynamic subscribers use the `tracing_subscriber::registry::Registry` as a way to compose multiple layers into one that is configured as the global default.

Added messages to be logged after the subscribers have been initialized. These messages are used to inform when log settings have been overriden, which happens when there are multiple entries with the same output.

### DIFF
--- a/jormungandr/src/settings/logging.rs
+++ b/jormungandr/src/settings/logging.rs
@@ -202,12 +202,14 @@ impl LogSettings {
                 match settings.format {
                     LogFormat::Default | LogFormat::Plain => {
                         // create a reloadable layer, and return the handle
+                        // FIXME: use this handle to reload layer
                         let (layer, _handle) = reload::Layer::new(
                             tracing_subscriber::fmt::Layer::new().with_writer(non_blocking),
                         );
                         (Some(layer), None)
                     }
                     LogFormat::Json => {
+                        // FIXME: use this handle to reload layer
                         let (layer, _handle) = reload::Layer::new(
                             tracing_subscriber::fmt::Layer::new()
                                 .json()
--- a/jormungandr/src/settings/logging.rs
+++ b/jormungandr/src/settings/logging.rs
@@ -13,7 +13,10 @@ use tracing::subscriber::SetGlobalDefaultError;
 #[allow(unused_imports)]
 use tracing_subscriber::layer::SubscriberExt;
 
-pub struct LogSettings(pub Vec<LogSettingsEntry>, pub LogInfoMsg);
+pub struct LogSettings {
+    pub entries: Vec<LogSettingsEntry>,
+    pub msgs: LogInfoMsg,
+}
 
 /// A wrapper to return an optional string message that we
 /// have to manually log with `info!`, we need this because
@@ -175,7 +178,7 @@ impl LogSettings {
         // Parse which settings are present for possible outputs
         let mut layer_settings = LogOutputLayerSettings::default();
         let mut info_msgs: Vec<String> = Vec::new();
-        for config in self.0.into_iter() {
+        for config in self.entries.into_iter() {
             layer_settings.read_setting(config, &mut info_msgs);
         }
         let (std_out_layer, std_out_layer_json) = if let Some(settings) = layer_settings.stdout {
@@ -297,7 +300,7 @@ impl LogSettings {
         // panics if something goes wrong.
         registry.init();
 
-        let log_msgs = match (self.1, info_msgs.is_empty()) {
+        let log_msgs = match (self.msgs, info_msgs.is_empty()) {
             (None, true) => None,
             (None, false) => Some(info_msgs),
             (Some(msgs), true) => Some(msgs),
--- a/jormungandr/src/settings/logging.rs
+++ b/jormungandr/src/settings/logging.rs
@@ -215,15 +215,15 @@ impl LogSettings {
             (None, None)
         };
         #[cfg(feature = "systemd")]
-        let journald_layer = if let Some(entry) = layer_settings.journald {
-            entry.format.require_default()?;
+        let journald_layer = if let Some(settings) = layer_settings.journald {
+            settings.format.require_default()?;
             let layer = tracing_journald::layer().map_err(Error::Journald)?;
             Some(layer)
         } else {
             None
         };
         #[cfg(feature = "gelf")]
-        let gelf_layer = if let Some(entry) = layer_settings.gelf {
+        let gelf_layer = if let Some(settings) = layer_settings.gelf {
             // have to use if let because it's an enum
             if let LogOutput::Gelf { backend, .. } = settings.output {
                 let (layer, task) = tracing_gelf::Logger::builder()
--- a/jormungandr/src/settings/logging.rs
+++ b/jormungandr/src/settings/logging.rs
@@ -169,6 +169,7 @@ impl LogSettings {
             for layer in layer_iter {
                 init_layer = BoxedSubscriber(Box::new(init_layer.with(layer)));
             }
+            // FIXME: this should only be run once
             tracing::subscriber::set_global_default(init_layer)
                 .map_err(Error::SetGlobalSubscriberError)?;
         }
@@ -191,8 +192,8 @@ impl LogSettingsEntry {
 
         match output {
             LogOutput::Stdout => {
-                let (subscriber, guard) = tracing_appender::non_blocking(std::io::stdout());
-                let builder = builder.with_writer(subscriber).with_max_level(*level);
+                let (non_blocking, guard) = tracing_appender::non_blocking(std::io::stdout());
+                let builder = builder.with_writer(non_blocking).with_max_level(*level);
                 let subscriber: Box<dyn Subscriber + Send + Sync> = match format {
                     LogFormat::Default | LogFormat::Plain => Box::new(builder.finish()),
                     LogFormat::Json => Box::new(builder.json().finish()),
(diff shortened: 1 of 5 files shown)


filip-dulic-bloxico:
There are reloadable layers, which can be updated or toggled at runtime. Got this example code from someone who asked in the discord channel today https://discord.com/channels/500028886025895936/627649734592561152/816016029590618165 , which is somewhat related to what I think that we could use:

```rust
// configure stdout and file `fmt` layers
let file_layer = tracing_subscriber::fmt::layer()
    .with_writer(my_file_writer);
let stdout_layer = tracing_subscriber::fmt::layer()
    .with_writer(std::io::stdout);
// create a reloadable layer whose value is currently `None`.
// this will be the layer that writes to stderr; at runtime, to
// toggle stderr on and off, we will reload this layer to be
// `Some` to enable stderr, or `None` to turn it back off.
let (reload_handle, stderr_layer) = tracing_subscriber::reload::Layer::new(None);

// set up the global subscriber, using a `Registry` to combine
// our three layers.
tracing_subscriber::registry()
   .with(file_layer)
   .with(stdout_layer)
   .with(stderr_layer)
   .init();

// ... lots of code later ...

// here's how we would enable stderr logging:
let stderr = tracing_subscriber::fmt::layer()
   .with_writer(std::io::stderr);
reload_handle
    .reload(Some(stderr))
    // you could handle this error more gracefully, of course.
    .expect("reloading should succeed");

// ... 

// or, we could turn it back off:
reload_handle.reload(None);
```

I made a test, and using the same registry code from Friday, the file, journald and gelf output layers have been made reloadable, without anything going wrong. I believe that we can change from LogOuput::File(path) to LogOutput::File(Option<Vec<(path, format, level)>>) (could be without the Option), and try to compose into a single subscriber that can be reloaded into the registered layer.

danielSanchezQ:
I'm not sure that would work also. Think of the (extreme, and just for an example) case where we have 100 file subscribers. 
The problem is with the way they handle this:

```rust
tracing_subscriber::registry()
   .with(file_layer)
   .with(stdout_layer)
   .with(stderr_layer)
   .init();
```

We would have to manually hardcode the 100 file layers in there.

filip-dulic-bloxico:
> I'm not sure that would work also. Think of the (extreme, and just for an example) case where we have 100 file subscribers.
> The problem is with the way they handle this:
> 
> ```rust
> tracing_subscriber::registry()
>    .with(file_layer)
>    .with(stdout_layer)
>    .with(stderr_layer)
>    .init();
> ```
> 
> We would have to manually hardcode the 100 file layers in there.

Those 100 file layers can be wrapped into a single subscriber loaded into the `file_layer`, because it is a `reload::Layer`. However, the problem is displaced into Boxing and storing the `reload::Handle<L, S>` types, used to update reloadable layers on the subscriber.

I think that we will have to wait for newer versions of the `tracing` crates, as they are working on implementing `Collector`s for subscribers, as well as having a better setup for dynamic subscribers.

In any case, by changing the implementation to using `Registry`, we remove the Boxed types, and a nice `init()` method that takes care of setting global defaults with all of the registered layers.

I'm working on cleaning it up, but perhaps this branch can be frozen or postponed if it's breaking tests.

danielSanchezQ:
Implementation is good. Never meant it's wrong 😁 
Just pointing out that the tracing library is not meant to be used as we thought (and we had with slog previously). I think we should go for this for now, it's clean and clear 👍

mzabaluev:
You don't necessarily need a dynamically reloadable layer since you can just supply an `Option<T>` on construction and that [implements `Layer`](https://docs.rs/tracing-subscriber/0.2.16/tracing_subscriber/layer/trait.Layer.html#impl-Layer%3CS%3E-for-Option%3CL%3E) if `T` implements it.

We can have a static composition of all possible singular backend layers. But it's an open question to me whether it's possible to have a `Layer` impl that is dynamically composed from appenders to multiple files. We can forbid this in the config file, but the restriction would feel artificially imposed, considering that our current data model for log outputs is an arbitrary list.

To avoid dealing with the whole complexity, it may be better to simplify the log configuration to a single backend, and correspondingly have the single configurable layer implemented on the enum of all supported output types, or just a boxed `Subscriber` impl.

mzabaluev:
> We can forbid this in the config file, but the restriction would feel artificially imposed, considering that our current data model for log outputs is an arbitrary list.

After some team discussion, it seems like a passable compromise that allows us to preserve the current configuration format and enable dynamic multi-destination logging later when supported better by the framework.

So the current implementation is on target, maybe with removal of dynamic configuration features unless they are necessary.
There should be a warning message for when there is more than one logger entry for the same output type, to inform that the extra entries were ignored.

filip-dulic-bloxico:
> Implementation is good. Never meant it's wrong 
> Just pointing out that the tracing library is not meant to be used as we thought (and we had with slog previously). I think we should go for this for now, it's clean and clear

Yeah, I agree, the crate isn't meant to be used like that.

It just came to mind that it should also be possible for tracing to co-exist with slog (I imagine), in a scenario where tracing is used at runtime, and traditional file-based logging is configured and executed where appropriate. I don't know if this is something that could apply to our use cases.

filip-dulic-bloxico:
> > We can forbid this in the config file, but the restriction would feel artificially imposed, considering that our current data model for log outputs is an arbitrary list.
> 
> After some team discussion, it seems like a passable compromise that allows us to preserve the current configuration format and enable dynamic multi-destination logging later when supported better by the framework.
> 
> So the current implementation is on target, maybe with removal of dynamic configuration features unless they are necessary.
> There should be a warning message for when there is more than one logger entry for the same output type, to inform that the extra entries were ignored.

Ok, I'm cleaning the code, and adding the warning (if there are multiple entries for the same output, the one listed at the bottom/last is the one considered) before submitting this PR for review.

filip-dulic-bloxico:
> You don't necessarily need a dynamically reloadable layer since you can just supply an `Option<T>` on construction and that [implements `Layer`](https://docs.rs/tracing-subscriber/0.2.16/tracing_subscriber/layer/trait.Layer.html#impl-Layer%3CS%3E-for-Option%3CL%3E) if `T` implements it.

True, `Option<T>` implements `Layer`, but this is only part of what is needed if you want to re-configure/reload it during runtime (usually because filter directives are changed during execution, for example; or, to enable/disable a subscriber given certain conditions, etc.). There is added overhead for reloading, but the key concepts here are that the `Registry` registers the layer at initialization, if the layer is reloadable, a handle needs to be kept somewhere, and it can be used to register anything that implements `Subscriber + for<'a> LookupSpan<'a>`, which is what the registry likes. So, `Option<T>` can be registered in a reloadable layer, or not, but only in a reloadable layer can it be modified during runtime execution.

filip-dulic-bloxico:
> > You don't necessarily need a dynamically reloadable layer since you can just supply an `Option<T>` on construction and that [implements `Layer`](https://docs.rs/tracing-subscriber/0.2.16/tracing_subscriber/layer/trait.Layer.html#impl-Layer%3CS%3E-for-Option%3CL%3E) if `T` implements it.
> 
> True, `Option<T>` implements `Layer`, but this is only part of what is needed if you want to re-configure/reload it during runtime (usually because filter directives are changed during execution, for example; or, to enable/disable a subscriber given certain conditions, etc.). There is added overhead for reloading, but the key concepts here are that the `Registry` registers the layer at initialization, if the layer is reloadable, a handle needs to be kept somewhere, and it can be used to register anything that implements `Subscriber + for<'a> LookupSpan<'a>`, which is what the registry likes. So, `Option<T>` can be registered in a reloadable layer, or not, but only in a reloadable layer can it be modified during runtime execution.

At least, this is what I understand. I might be wrong. In any case, I've removed the reloadable code :)

jormungandr/src/settings/logging.rs
```diff
@@ -215,30 +232,81 @@ impl LogSettingsEntry {
...
+        #[cfg(feature = "systemd")]
+        let registry = registry.with(journald_layer);
+        #[cfg(feature = "gelf")]
+        let registry = registry.with(gelf_layer);
+
+        // configure the registry subscriber as the global default,
+        // panics if something goes wrong.
+        registry.init();
+
+        let log_msgs = match (self.1, info_msgs.is_empty()) {
```
mzabaluev, line 300: This type should be a struct so that the reader does not need to check the definition to know what `self.1` means.

mzabaluev (approved):

filip-dulic-bloxico:
I finally traced the span Id error, and with simple modifications to how the parent_span is cloned, `cargo test jcli::transaction::e2e::test_account_is_created_if_transaction_out_is_account -- --nocapture` passes.

jormungandr/src/settings/logging.rs
filip-dulic-bloxico: I will now put this into a struct.

Zeegomo:
Pushed the commit that should fix most of the test failures and left some comments in the commit message.

 I think this needs further investigation because I see similar constructions to the one fixed used somewhere else in the code (e.g. [here](https://github.com/input-output-hk/jormungandr/blob/838d732889441f65bada10a53e4b1332e7ce00bd/jormungandr/src/utils/task.rs#L119)) and it's not clear to me why those do not cause problems.

In addition, setting max log level in config file does not seem to work either

filip-dulic-bloxico:
Finally got the tests working by declaring child spans in the parent's scope for jormungandr::network::service.

filip-dulic-bloxico:
> Pushed the commit that should fix some of the test failures and left some comments in the commit message.
> 
> I think this needs further investigation because I see similar constructions to the one fixed used somewhere else in the code (e.g. [here](https://github.com/input-output-hk/jormungandr/blob/838d732889441f65bada10a53e4b1332e7ce00bd/jormungandr/src/utils/task.rs#L119)) and it's not clear to me why those do not cause problems.
> 
> In addition, setting max log level in config file does not seem to work either

The max log level config needs to be re-worked with the new registry setup. Because the registry is acting as a single top-level subscriber, the _typical_ way of implementing level filtering is by adding an EnvFilter layer (which is registry compatible), and which filters the same way for all underlying layers (stdout/stderr/file/journald/gelf).

I'm taking a look at the easiest way to get the level from the config files connected, but I think that we need to revise the structs used in the config (replace the Vec<LogSettingsEntry> with a struct named LogRegistryConfig, perhaps).

mzabaluev:
> which filters the same way for all underlying layers (stdout/stderr/file/journald/gelf).

Part of the impetus for the multi-consumer logging setup was being able to set level filters separately for different log backends.
If this is not achievable with current `tracing`, maybe we should scrap the whole thing and replace it with a single backend log configuration.

mzabaluev:
Let's kill this and go with a single subscriber and, correspondingly, a single log configuration item.

PR #3074, opened 2021-03-01
Use pipes instead of file for logging in tests

Zeegomo:
Use a pipe for reading logs from child processes instead of reading from files, because that caused synchronization problems.

The main problem is that reading from stdout is blocking, the current solution is to spawn a new thread that blocks on new messages and then send them back on a channel. It's not ideal because we spawn a new thread for each new process, but it should not be too much of a problem.
I also tried with async processes and output (`tokio::process`) but I had issues with I/O drivers not receiving updates in some cases and I dropped it. 
Another option could be to use some crate that exposes non future-based non blocking reads (mio only supports unix)

### DIFF
--- a/testing/jormungandr-integration-tests/src/common/jormungandr/mod.rs
+++ b/testing/jormungandr-integration-tests/src/common/jormungandr/mod.rs
@@ -69,6 +69,6 @@ impl FragmentNode for JormungandrProcess {
         println!("Fragment '{}' in block: {} ({})", fragment_id, block, date);
     }
     fn log_content(&self) -> Vec<String> {
-        self.logger.get_log_content()
+        self.logger.get_lines_as_string()
     }
 }
--- a/testing/jormungandr-integration-tests/src/common/jormungandr/mod.rs
+++ b/testing/jormungandr-integration-tests/src/common/jormungandr/mod.rs
@@ -12,19 +12,16 @@ use jormungandr_testing_utils::testing::MemPoolCheck;
 pub use process::*;
 pub use starter::*;
 use std::collections::HashMap;
-use std::path::PathBuf;
 use thiserror::Error;
 
 use jormungandr_testing_utils::testing::{FragmentNode, FragmentNodeError};
 
 #[derive(Error, Debug)]
 pub enum JormungandrError {
-    #[error("error in logs. Error lines: {error_lines}, logs location: {log_location}, full content:{logs}")]
-    ErrorInLogs {
-        logs: String,
-        log_location: PathBuf,
-        error_lines: String,
-    },
+    #[error("error in logs. Error lines: {error_lines}, full content:{logs}")]
+    ErrorInLogs { logs: String, error_lines: String },
+    #[error("error(s) in log detected: port already in use")]
+    PortAlreadyInUse,
 }
 
 impl FragmentNode for JormungandrProcess {
@@ -72,6 +69,6 @@ impl FragmentNode for JormungandrProcess {
         println!("Fragment '{}' in block: {} ({})", fragment_id, block, date);
     }
     fn log_content(&self) -> Vec<String> {
-        self.logger.get_lines_from_log().collect()
+        self.logger.get_log_content()
     }
 }
--- a/testing/jormungandr-integration-tests/src/common/jormungandr/process.rs
+++ b/testing/jormungandr-integration-tests/src/common/jormungandr/process.rs
@@ -15,7 +15,7 @@ use jormungandr_testing_utils::testing::{
     JormungandrParams, SyncNode, TestConfig,
 };
 use jormungandr_testing_utils::testing::{RemoteJormungandr, RemoteJormungandrBuilder};
-use jortestkit::process::{self as process_utils, output_extensions::ProcessOutput, ProcessError};
+use jortestkit::prelude::ProcessOutput;
 use std::net::SocketAddr;
 use std::path::Path;
 use std::process::Child;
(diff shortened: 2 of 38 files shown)


eugene-babichenko:
Does this replace #3051 as a proper fix to the problem?

Zeegomo:
Yes, this should solve that problem

eugene-babichenko:
Please close #3051 then.

Zeegomo:
This is ready for review, it is still a draft because it points to a patched jortestkit

mzabaluev:
https://github.com/input-output-hk/jortestkit/pull/59 merged, please update and rebase.

testing/jormungandr-integration-tests/src/common/jormungandr/mod.rs
```diff
@@ -12,19 +12,16 @@ use jormungandr_testing_utils::testing::MemPoolCheck;
...
 
 #[derive(Error, Debug)]
 pub enum JormungandrError {
-    #[error("error in logs. Error lines: {error_lines}, logs location: {log_location}, full content:{logs}")]
-    ErrorInLogs {
-        logs: String,
-        log_location: PathBuf,
-        error_lines: String,
-    },
+    #[error("error in logs. Error lines: {error_lines}, full content:{logs}")]
```
mzabaluev, line 21:
```suggestion
    #[error("error in logs. Error lines: {error_lines}, full content: {logs}")]
```
Not really sure it's useful to put the full JSON content in a single error printout. Maybe the test fixture handling this error can re-parse the `logs` field and print the lines out nicely?

testing/jormungandr-integration-tests/src/common/jormungandr/process.rs
```diff
@@ -54,6 +67,60 @@ impl JormungandrProcess {
         }
     }
 
+    pub fn status(&self, strategy: &StartupVerificationMode) -> Status {
+        let port_occupied_msgs = ["error 87", "error 98", "panicked at 'Box<Any>'"];
```
mzabaluev, line 71: The match on "panicked at 'Box<Any>'" may produce false positives. It's better to fix the diagnostic in jormungandr to provide a more specific line to match on.

testing/jormungandr-integration-tests/src/common/jormungandr/process.rs
```diff
@@ -209,10 +285,13 @@ impl SyncNode for JormungandrProcess {
     }
 
     fn get_lines_with_error_and_invalid(&self) -> Vec<String> {
-        self.logger.get_lines_with_error_and_invalid().collect()
+        self.logger
+            .get_lines_with_level(LogLevel::ERROR)
+            .map(|x| x.to_string())
+            .collect()
```
mzabaluev, line 291: This will no longer catch invalid JSON, or does the test runner panic when failing to parse a line as JSON?

testing/jormungandr-integration-tests/src/common/jormungandr/process.rs
```diff
@@ -209,10 +285,13 @@ impl SyncNode for JormungandrProcess {
...
     }
 
     fn is_running(&self) -> bool {
-        todo!()
+        matches!(self.status(&StartupVerificationMode::Log), Status::Running)
     }
```
mzabaluev, line 296: This is kinda expensive to re-parse all logs in a method simply named `is_running`, but maybe it would not have much impact on tests.

testing/jormungandr-integration-tests/src/common/jormungandr/process.rs
Zeegomo: I see no reason we should have invalid json in the logs, the logger panics if the line cannot be parsed

testing/jormungandr-integration-tests/src/common/jormungandr/process.rs
Zeegomo: they are not actually parsed again, it goes through already parsed lines

testing/jormungandr-integration-tests/src/common/jormungandr/process.rs
Zeegomo: agree, I would rather fix this on another PR though, this comes from previous implementation.

testing/jormungandr-integration-tests/src/common/jormungandr/mod.rs
Zeegomo: Maybe we could reuse log formatting from `tracing_subscriber::fmt` but I'm not sure if it can be used independently, need to check.

testing/jormungandr-integration-tests/src/common/jormungandr/process.rs
mzabaluev: OK

mzabaluev:
The nightly failure should be fixed by https://github.com/input-output-hk/jormungandr/pull/2888

testing/jormungandr-scenario-tests/src/interactive/args/mod.rs
```diff
@@ -105,23 +123,42 @@ impl UserInteractionController {
...
+                &Choice::new(choice),
+                node,
+            )?,
+            (None, Some(node)) => self.controller.fragment_sender().send_vote_cast(
+                wallet,
+                &vote_plan_def.into(),
+                proposal_index as u8,
+                &Choice::new(choice),
+                node,
+            )?,
```
mzabaluev, line 143: The arm bodies are the same. Use

```suggestion
            (Some(node), None) | (None, Some(node)) => self.controller.fragment_sender().send_vote_cast(
                wallet,
                &vote_plan_def.into(),
                proposal_index as u8,
                &Choice::new(choice),
                node,
            )?,
```

testing/jormungandr-scenario-tests/src/interactive/args/mod.rs
```diff
@@ -140,10 +177,24 @@ impl UserInteractionController {
...
+
+        let check = match (node, legacy_node) {
+            (Some(node), None) => self
+                .controller
+                .fragment_sender()
+                .send_transaction(from, &to, node, value)?,
+            (None, Some(node)) => self
+                .controller
+                .fragment_sender()
+                .send_transaction(from, &to, node, value)?,
```
mzabaluev, line 191: Same as above.

testing/jormungandr-scenario-tests/src/interactive/args/mod.rs
Zeegomo: Those bindings are of different types, I don't like this solution but I do not know how that could be rewritten in a better fashion

mzabaluev: Ah, I did not realize there's static polymorphism over the parameter.

mzabaluev (approved):

PR #3087, opened 2021-03-03
reject fragments that will overflow the fragments pool

### LINKED ISSUE #3064: Change from oldest fragment eviction to rejecting new fragments when the pool is full
When the leadership task's throughput for processing incoming fragments is exceeded and the pool has been filled to the limit, some fragments need to be rejected. The current behavior of evicting the least recently added fragments is problematic:
* Some newer fragments may optimistically depend on fragments added to the pool earlier.
* A dishonest participant can attempt to flush out an unwanted transaction after observing it in the pool. It does not cost them anything to fill up the pool with invalid transactions.

Instead, the leadership task should switch to rejecting any new incoming transactions while the pool capacity is full.

eugene-babichenko:
closes #3064

### DIFF
--- a/jormungandr/Cargo.toml
+++ b/jormungandr/Cargo.toml
@@ -72,6 +72,11 @@ features = ["rustls-tls"]
 [dev-dependencies]
 rand_core = "0.5"
 tokio = { version = "^1.3", features = ["full"] }
+quickcheck = "0.9"
+quickcheck_macros = "0.9"
+chain-impl-mockchain = { git = "https://github.com/input-output-hk/chain-libs.git", branch = "master", features = [ "property-test-api" ] }
+chain-addr = { git = "https://github.com/input-output-hk/chain-libs.git", branch = "master", features = [ "property-test-api" ] }
+chain-crypto = { git = "https://github.com/input-output-hk/chain-libs.git", branch = "master", features = [ "property-test-api" ] }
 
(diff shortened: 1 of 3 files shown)


mzabaluev (commented):
Looks like `Pool::insert` still does not enforce the capacity limit. Maybe remove that method if it's not used anywhere.

jormungandr/Cargo.toml
```diff
@@ -73,6 +73,11 @@ features = ["rustls-tls"]
 [dev-dependencies]
 rand_core = "0.5"
 tokio = { version = "^1.0", features = ["full"] }
+quickcheck = "0.9"
+quickcheck_macros = "0.9"
```
mzabaluev, line 77: It's better to use quickcheck 1.0 while we are at it, possibly integrating #2927, https://github.com/input-output-hk/chain-libs/pull/506, and https://github.com/input-output-hk/chain-libs/pull/507. Unless this is too much work and requires another PR.

eugene-babichenko: I think it does require another PR, there are breaking interface changes.

mzabaluev (approved):
LGTM

With this in, does `self.entries` still need to be an LRU?

eugene-babichenko:
Can be turned into a simple queue actually.

PR #3094, opened 2021-03-04
doc: fix v1 FragmentStatuses docs

### LINKED ISSUE #3063: Wrong API description for /v1/fragments/statuses
The [FragmentStatuses](https://github.com/input-output-hk/jormungandr/blob/c18270fecb5d0a90a799ea14218cd8440156fc2c/doc/api/v1.yaml#L49-L49) method description seems wrong: the method should be `GET`, the parameter is in the query string, and it's a comma-separated list and not an array.

eugene-babichenko:
fixes #3063

### DIFF
--- a/doc/api/v1.yaml
+++ b/doc/api/v1.yaml
@@ -46,23 +46,20 @@ paths:
           description: One of messages is malformed
 
   /api/v1/fragments/statuses:
-    post:
+    get:
       description: Get statuses of fragments
       operationId: FragmentStatuses
       tags:
         - fragment
-      requestBody:
-        description: List of fragment IDs
-        required: true
-        content:
-          application/json:
-            schema:
-              type: array
-              items:
-                type: string
-                pattern: '[0-9a-f]+'
-            example: ["68dcc12fe0dfe5e7b66ca6f8c959f9aa43b273e120a77fc3e4e2f04f1ecd7968"]
-          
+      parameters:
+        - in: query
+          name: fragment_ids
+          description: Comma-separated list of fragment IDs
+          schema:
+            type: string
+            pattern: '[0-9a-f]+(,[0-9a-f]+)+'
+          required: true
+          example: '68dcc12fe0dfe5e7b66ca6f8c959f9aa43b273e120a77fc3e4e2f04f1ecd7968,b34bf76412138543d7b57ea23ff58b6b6c04a954c0badc3cadeafc564f1a771'
       responses:
         '200':
           description: Success. If a fragment is missing from in-memory fragment logs it will not show up in the result dictionary.


doc/api/v1.yaml
```diff
@@ -46,23 +46,19 @@ paths:
...
-                pattern: '[0-9a-f]+'
-            example: ["68dcc12fe0dfe5e7b66ca6f8c959f9aa43b273e120a77fc3e4e2f04f1ecd7968"]
-          
+      parameters:
+        - in: path
+          name: fragment_ids
+          description: Comma-separated list of fragment IDs
+          schema:
+            type: string
+            pattern: '[0-9a-f]+(,[0-9a-f]+)+'
```
mzabaluev, line 60: I wonder how common this format is in our API. What is an idiomatic way to pass an array of simple values in REST/OpenAPI?

doc/api/v1.yaml
```diff
@@ -46,23 +46,19 @@ paths:
...
-            schema:
-              type: array
-              items:
-                type: string
-                pattern: '[0-9a-f]+'
-            example: ["68dcc12fe0dfe5e7b66ca6f8c959f9aa43b273e120a77fc3e4e2f04f1ecd7968"]
-          
+      parameters:
+        - in: path
+          name: fragment_ids
```
mzabaluev, line 56: Seems like the linter wants this to be in the endpoint path.

doc/api/v1.yaml
```diff
@@ -46,23 +46,19 @@ paths:
...
-        description: List of fragment IDs
-        required: true
-        content:
-          application/json:
-            schema:
-              type: array
-              items:
-                type: string
-                pattern: '[0-9a-f]+'
-            example: ["68dcc12fe0dfe5e7b66ca6f8c959f9aa43b273e120a77fc3e4e2f04f1ecd7968"]
```
mzabaluev, line 64: Would be good to provide an updated example.

doc/api/v1.yaml
eugene-babichenko: I am not sure we have it in other places. The problem is that it was the only way to pass an array of these as an argument. Initially I wanted to use repeated `fragment_id`, but I iirc it just did was not possible.

doc/api/v1.yaml
eugene-babichenko: Ah, that is incorrect. It is not in path, but in query.

doc/api/v1.yaml
eugene-babichenko: Done.

doc/api/v1.yaml
mzabaluev: Query string does not forbid repeating keys AFAIK (worked with PHP array args in 2001!), but this way is more compact.

mzabaluev (approved):

doc/api/v1.yaml
eugene-babichenko: This was the problem with `warp`, not the standard restricting me from doing that.

PR #3121, opened 2021-03-08
Improve log formatting

Zeegomo:
Improve formatting of logs in testing environment by mimicking `tracing_subscriber` default behavior instead of using json.
This pr also rework the structure of `LogEntry` struct and how it's handled to make it closer to what we get from the logging output (i.e. there are no fixed fields, just multiple key-value pairs)

**Example:**
`Mar 08 09:30:04.007 DEBUG service{kind=block}: process_leadership_block{hash=7c4052f3f017488e09f9566ab74e1951e6df645203f9adae88b215dcdc2ab50e,parent=3ce3cf198285607b1c2616c26f038118b6ea10fbaa8940fdd6a23e3452027c92,date=0.14}:  jormungandr::blockchain::process: propagating block to the network 
`
instead of 
` {"level":"DEBUG","ts":"Mar 08 09:30:04.007","fields":{"message":"propagating block to the network"},"target":"jormungandr::blockchain::process","span":{"name":"process_leadership_block","hash":"7c4052f3f017488e09f9566ab74e1951e6df645203f9adae88b215dcdc2ab50e","date":"0.14","parent":"3ce3cf198285607b1c2616c26f038118b6ea10fbaa8940fdd6a23e3452027c92"},"spans":[{"kind":"block","name":"service"},{"name":"process_leadership_block","hash":"7c4052f3f017488e09f9566ab74e1951e6df645203f9adae88b215dcdc2ab50e","parent":"3ce3cf198285607b1c2616c26f038118b6ea10fbaa8940fdd6a23e3452027c92","date":"0.14"}]}
`

### DIFF
--- a/testing/jormungandr-testing-utils/src/testing/node/logger.rs
+++ b/testing/jormungandr-testing-utils/src/testing/node/logger.rs
@@ -2,15 +2,19 @@ use crate::testing::Timestamp;
 use chain_core::property::FromStr;
 use chain_impl_mockchain::{block, key::Hash};
 use jormungandr_lib::{interfaces::BlockDate, time::SystemTime};
-use serde::de::Error;
 use serde::{Deserialize, Serialize};
+
 use std::cell::{Ref, RefCell};
 use std::collections::HashMap;
+use std::convert::AsRef;
 use std::fmt;
 use std::io::BufRead;
+use std::ops::Index;
+use std::ops::IndexMut;
 use std::process::ChildStdout;
 use std::sync::mpsc::{self, Receiver};
 use std::time::Instant;
+use strum::AsRefStr;
 
 // TODO: we use a RefCell because it would be very labor intensive to change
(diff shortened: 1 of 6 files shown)


testing/jormungandr-testing-utils/src/testing/node/logger.rs
```diff
@@ -257,35 +305,42 @@ impl JormungandrLogger {
...
+                            if v.is_string() {
+                                v.to_owned()
+                            } else {
+                                serde_json::Value::String(v.to_string())
+                            },
+                        )
+                    })
+                    .collect();
+                *$container.index_mut($field) = mapped;
+            };
```
danielSanchezQ, line 330: Could this be a function instead? I don't see the advantage of this being a macro. 
What am I missing? 🤔

danielSanchezQ (approved):
LGTM

PR #3124, opened 2021-03-09
CI test: Ditch update_deps job on Windows

mzabaluev:
The cache update on Windows is slow, blocking testing jobs for more
than a minute. It's better to let the Windows jobs miss the cache
sometimes and restore an older snapshot of thecrate repository index,
but let test jobs for the other platforms run fast.


danielSanchezQ (approved):
LGTM

PR #3127, opened 2021-03-10
Remove unneeded async/await calls and update reqwest

Zeegomo:
Isolate gRPC calls that needs async in their own context so that the rest of the testing framework can use them as if they were standard blocking calls.

That also enable us to update reqwest to 0.11


mzabaluev (approved):
:+1:

PR #3132, opened 2021-03-11
Filter in-use ports when assigning a new one

Zeegomo:
There are some ports in use in the range that we use for nodes (e.g. mono on 8084 on ubuntu https://github.com/actions/virtual-environments/issues/2821)
Let's check before assigning ports that they are indeed free

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2009,6 +2009,7 @@ dependencies = [
  "jortestkit",
  "json",
  "lazy_static",
+ "netstat2",
  "os_info 3.0.1",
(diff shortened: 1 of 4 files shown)


danielSanchezQ (approved):
LGTM, seems simple enough. Hopefully, this kill this kind of error from now on/

Zeegomo:
Earlier failure seems unrelated and caused by a slow runner (db was not correctly synced even though we wait for double the default sync time), if that pops up again we should probably do something (wait even more?)

PR #3133, opened 2021-03-11
Poldercast [1/3]: Update to poldercast 1.1 and newer multiaddr

mzabaluev:
Update to poldercast 1.1 and parity-multiaddr 0.11.

Change data structures used to identify peers in new poldercast:
- Replace `Address` identifier for a peer with `SocketAddr`.
- Reintroduce `NodeId` wrapping a public key, which can be configured.


mzabaluev:
What needs to be done:

- [ ] Replace use of multiaddr `Address` with `SocketAddr` and `NodeId` (interworking with `poldercast::PublicKey`) in the p2p code.
- [ ] Replace the `policy` stuff with calling `promote_peer`/`remove_peer` in the new `Topology` API.
- [ ] Redo the custom `Layer` for preferred nodes.

Zeegomo:
**Changed**:
*  Trusted peers are not inserted in the topology right after startup anymore. Instead, we establish a connection and let them advertise themselves to the node. Basically that means that the node will need a bit more time to discover peers, and if a trusted peer is not available for gossip at startup we will forget about it and we will add it to the topology only after receiving a gossip about it.
* Use `SocketAddr` instead of `Multiaddr` in p2p code and config for listen address.
*  Re-implemented quarantine with `promote_peer` / `remove_peer`.
* Various rework of stuff that relied on old poldercast features

**What needs to be done**:
* Implement custom layer for preferred nodes (and maybe random direct connections?). It will not insert such nodes in the topology on its own but instead it will return a view based only on those nodes if they are already in the topology.
* Implement custom ring layer to be able to specify per-topic interest as it is now
* There are a bunch of things that the new poldercast version does not support:
    * Layers are trait objects and do not implement `Sync + Send` (currently using a fork, we might open a PR for that, or solve it using a task if that is possible)
    * We do not have access to the inner nodes known to the topology. There are a bunch of methods used only by rest API IIRC that rely on this information. We could "reverse engineer" this info but it would be pretty inefficient because internally it uses a lru cache and we don't know when evictions happen. At the moment it just return the current view, so this is just a subset of known nodes.
* Might need to review changes in tests
* Make `node_id` of preferred nodes (and trusted peers) compulsory?

**Possible additions**:
* Move topology to its own task (and possibly trim down the global state that needs to be passed around in networking)
* Actually check that the `node_id` for trusted_peers / preferred nodes is the one that we expect and that the ids coming from external nodes do not change over time

doc/configuration/introduction.md
```diff
@@ -32,7 +32,7 @@ bootstrap_from_trusted_peers: false
 
 p2p:
   public_address: "/ip4/X.X.X.X/tcp/Y" # This should match your public IP address (X) and port number (Y)
-  #listen_address: /ip4/0.0.0.0/tcp/Z
+  #listen: 0.0.0.0:Y
```
eugene-babichenko, line 35: Any particular reason for that change? I mean both moving away from multiaddr and the name change.

doc/configuration/network.md
```diff
@@ -86,12 +86,14 @@ Use the CA certificate with `jcli`.
...
-- `public_id`: (optional) This is a static identifier, 24 bytes encoded in hexadecimal. They are used
-  to bootstrap the connection to the node if the node introduce itself as a trusted peer.
-  **Most of the user don't need to set this value** and in fact we are working toward potentially
-  removing the need for this value.
-- `listen_address`: (optional) [multiaddr][multiaddr] specifies the address the node
-    will listen to to receive p2p connection. Can be left empty and the node will listen
+- `node_key_file`: (optional) Path to a file containing a bech32-encoded ed25519 secret key.
+  The keys are used to advertize the node in network gossip and to authenticate
+  a connection to the node if the node is used as a trusted peer.
+  **Most of the users don't need to set this value** as the key will be randomly
```
eugene-babichenko, line 92: Will it be unique for each jormungandr run, or will it be generated and saved on the first run?

jormungandr/Cargo.toml
```diff
@@ -36,10 +36,11 @@ hex = "0.4"
 http = "0.2.2"
 humantime = "2.0"
 jormungandr-lib = { path = "../jormungandr-lib" }
+keynesis = "1.1"
```
eugene-babichenko, line 39: Is a new dependency for `ed25519` keys strictly required?

doc/configuration/introduction.md
Zeegomo: For a listen address a multiaddr does not make a lot of sense. As for the name, it is just to uniform with the rest config, where the field is also called just `listen`

jormungandr/Cargo.toml
Zeegomo: `poldercast` use key types from `keynesis`, so we need to import that library if we want to interact with it (e.g. conversion between 'our' type and their and implementing custom layers).

doc/configuration/network.md
Zeegomo: That is yet to be decided, as it requires some thought on how we want to handle identification. I think both choices make sense, probably a node that wants to serve as a trusted peer would want to reuse the same key, while 'casual' nodes can use this as an ephemeral key, but again I haven't thought this through. Identification / Verification of identity is out of scope for this PR (but I still had to add handling of ids on the topology side since they are required by poldercast)

eugene-babichenko: Well, I think keeling this key ephemeral when not provided by a user is fine. Let them generate a key by hand if they want it to be persistent.

jormungandr/src/network/p2p/topology.rs
```diff
@@ -2,171 +2,207 @@
...
-        })
+    /// register a strike against the given peer
+    pub async fn report_node(&self, node_id: &NodeId) {
+        let mut inner = self.lock.write().await;
+        if let Some(node) = inner.topology.get(node_id).cloned() {
+            if inner.quarantine.quarantine_node((&node).into()) {
+                inner.topology.remove_peer(node_id);
+                // Don't know what is the purpose of trusted peers in poldercast,
+                // this is a quick hack to treat those as standard ones
+                inner.topology.remove_peer(node_id);
```
eugene-babichenko, line 204: 1. Please point me at the doc/cite a message about that.
2. What is the purpose of calling `remove_peer` twice?

jormungandr/src/network/p2p/topology.rs
```diff
@@ -2,171 +2,207 @@
...
 
 impl P2pTopology {
-    pub fn new(config: &Configuration, span: Span, rng: ChaChaRng) -> Self {
-        Builder::new(config.profile.clone(), span)
-            .set_poldercast_modules()
-            .set_custom_modules(&config, rng)
-            .set_policy(config.policy.clone())
-            .build()
+    pub fn new(config: &Configuration) -> Self {
+        // FIXME: How should we handle cases where the is not listen set? Can a node just receive?
```
eugene-babichenko, line 81: @mzabaluev we need your clarification on that.

jormungandr/src/rest/v0/logic.rs
```diff
@@ -228,11 +233,11 @@ async fn create_stats(context: &Context) -> Result<Option<NodeStats>, Error> {
...
         last_received_block_time: stats.slot_start_time().map(SystemTime::from),
-        peer_available_cnt: nodes_count.available_count,
+        peer_available_cnt,
         peer_connected_cnt: stats.peer_connected_cnt(),
-        peer_quarantined_cnt: nodes_count.quarantined_count,
-        peer_total_cnt: nodes_count.all_count,
-        peer_unreachable_cnt: nodes_count.not_reachable_count,
+        peer_quarantined_cnt,
+        peer_total_cnt,
+        peer_unreachable_cnt: 0, // FIXME
```
eugene-babichenko, line 240: Why is that?

jormungandr/src/settings/start/mod.rs
```diff
@@ -264,44 +269,40 @@ fn generate_network(
...
-        .topics_of_interest
-        .unwrap_or_else(config::default_interests)
-    {
-        let sub = poldercast::Subscription {
-            topic: topic.0,
-            interest: interest_level.0,
-        };
-        profile.add_subscription(sub);
-    }
+    // TODO: do we want to check that we end up with a valid address?
```
eugene-babichenko, line 272: I think currently the answer is "no" to both.

jormungandr-lib/Cargo.toml
```diff
@@ -22,8 +22,8 @@ rand_chacha = "0.3"
 chrono = { version = "0.4", features = ["serde"] }
 humantime = "2.0"
 thiserror = "1.0"
-poldercast = "0.14"
-multiaddr = { package = "parity-multiaddr", version = "0.9" }
+poldercast = { git = "https://github.com/Zeegomo/poldercast", branch = "master" }
```
eugene-babichenko, line 25: Do we still need it in chain-libs? IIRC multiaddr should be sufficient. It is better to keep deps of `jormungandr-lib` minimal since it is used in places like `jcli`.

jormungandr/src/network/mod.rs
```diff
@@ -373,24 +369,40 @@ async fn handle_propagation_msg(msg: PropagateMsg, state: GlobalStateR, channels
...
-                    .view(poldercast::Selection::Topic {
+                    .view(poldercast::layer::Selection::Topic {
                         topic: p2p::topic::BLOCKS,
                     })
                     .await;
-                prop_state.peers.propagate_block(view.peers, header).await
+                prop_state
+                    .peers
+                    .propagate_block(
+                        view.peers
```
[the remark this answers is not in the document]
Zeegomo: You mean `view.peers.into_iter()[...]collect()`? The view selection is different based on what we are propagating so I'm not sure how I could move it before the match. Btw, this is greatly changed by subsequent PRs so I would not focus on this too much.

jormungandr/src/network/p2p/topology.rs
Zeegomo: I'm not sure this is the intended solution, but reading the source code I understand that:
* `remove_peer` [(source)](https://github.com/primetype/poldercast/blob/c5a8603addb7e0135bbfbebce0e6ffe38c68edcf/src/topology.rs#L90) remove the peer from each of the layers and also demote it.
* peers are divided into three levels [(source)](https://github.com/primetype/poldercast/blob/c5a8603addb7e0135bbfbebce0e6ffe38c68edcf/src/profiles.rs#L6), `trusted`, `pool` and `dirty`. The only difference between these that I can see is that when receiving a gossip about a peer, if it is currently on our `dirty` pool  we do not re-add it to the topology [(source)](https://github.com/primetype/poldercast/blob/c5a8603addb7e0135bbfbebce0e6ffe38c68edcf/src/topology.rs#L116)
* You are supposed to call `promote_peer` when you can successfully connect to a peer, and that moves a peer up 1 pool (dirty -> pool -> trusted) 
* Imagine this scenario: node A connects to node B and calls `promote_peer` -> B gets promoted to the `trusted` pool. Then, for some reason, the connection drops, and we would like to quarantine B for some time. If we simply call `remove_peer` once, it gets remove from the layers, but the next gossip would re-add it right away, since it got demoted from `trusted` to `pool`. By calling `remove_peer` again, we demote it all the way to `dirty`, so that even if we receive gossips about it while it should be quarantined they will be ignored

jormungandr/src/rest/v0/logic.rs
Zeegomo: Those definitions were previously used by poldercast, I don't know if we want to keep them all or not and I did not want to break the rest api. For example, in this case, what would `peer_unreachable` mean? If it means nodes we could not connect to it seems pretty similar to the quarantined ones atm.

jormungandr-lib/Cargo.toml
Zeegomo: Well spotted!

jormungandr/src/network/p2p/mod.rs
```diff
@@ -1,23 +1,41 @@
...
 /**
 # topics definition for p2p interest subscriptions
 */
 pub mod topic {
     use poldercast::Topic;
 
-    pub const MESSAGES: Topic = Topic::new(0u32);
-    pub const BLOCKS: Topic = Topic::new(1u32);
+    pub const MESSAGES: Topic = Topic::new([0; 32]);
+    pub const BLOCKS: Topic = Topic::new([1; 32]);
```
mzabaluev, line 38: Hmm; maybe 1 and thirty one 0 instead?

jormungandr/src/network/p2p/topology.rs
mzabaluev: I guess this is where the public address thing comes to the fore.
My first approach would be to:
- Always require to listen on some socket, just to simplify all code paths;
- If the public address is not set, try to create a dummy profile with the local binding address retrieved from `TcpListener::local_addr` and never advertise it in gossip?

jormungandr/src/network/p2p/topology.rs
```diff
@@ -2,171 +2,207 @@
...
-    /// Create a new topology for the given node profile
-    fn new(node: poldercast::NodeProfile, span: Span) -> Self {
-        Builder {
-            topology: Topology::new(node),
-            span,
-        }
-    }
+#[derive(Eq, Clone, Serialize, Debug)]
+pub struct ProfileInfo {
+    #[serde(serialize_with = "serialize_display")]
```
mzabaluev, line 28: Protip: You can use [`serde_with`](https://crates.io/crates/serde_with) to avoid reinventing some common helpers.

jormungandr/Cargo.toml
```diff
@@ -36,10 +36,11 @@ hex = "0.4"
...
 lazy_static = "1.4"
 linked-hash-map = "0.5"
-poldercast = "0.14"
-multiaddr = { package = "parity-multiaddr", version = "0.9" }
+poldercast = { git = "https://github.com/Zeegomo/poldercast", branch = "master" }
```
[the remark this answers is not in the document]
Zeegomo: No they are all released, following prs points to the upstream repo. This points to the fork because since topology does not live on its own task yet we need to add `Send + Sync` bounds that are not required once make this change.

jormungandr/src/network/p2p/mod.rs
Zeegomo: we can do that, what's the rationale?

jormungandr/src/network/p2p/topology.rs
Zeegomo: What about we use some constant that can't be for sure used in a network (even a private one) like `127.0.0.1:0` if the public address is not defined and then manually filter outgoing gossips? On a side note, what would happen if a node tries to connect to itself (e.g because a malicious user create another routing for the public address of one of the nodes)?

As for the requirement on listen address, I think the only place where it is used is when starting the grpc server, so we can simply not do that if it's not set (and we already do that), I don't see how always enforcing one would make things simpler.

jormungandr/src/network/p2p/mod.rs
mzabaluev: It's just that the all-ones topic ID seems selected from the "easier to write in Rust" than any other rationale :)

jormungandr/Cargo.toml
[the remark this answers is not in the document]
mzabaluev: Does poldercast 1.2 fix this?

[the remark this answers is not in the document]
Zeegomo: Nope, that change was never proposed, we solved that by moving topology to its own task

jormungandr/src/network/p2p/mod.rs
Zeegomo: That's true :) As long as they are different I don't see why it matters but it's an easy fix

jormungandr/src/network/p2p/topology.rs
```diff
@@ -2,171 +2,204 @@
...
+            }
+        }
     }
 
+    #[instrument(skip(self), level = "debug")]
     pub async fn force_reset_layers(&self) {
-        let mut topology = self.lock.write().await;
-        topology.force_reset_layers()
+        tracing::warn!("resetting layers is not supported in this poldercast version");
     }
```
mzabaluev, line 151: Hmm, maybe just go to the now useless periodic task that calls it and cut it out?

jormungandr/src/network/p2p/topology.rs
mzabaluev: This really indicates a lack of proper method in poldercast. OK to use it this way, as long as the hack is documented.

jormungandr/src/rest/v0/logic.rs
```diff
@@ -570,9 +575,7 @@ pub async fn get_diagnostic(context: &Context) -> Result<Diagnostic, Error> {
     Ok(diagnostic_data.clone())
 }
 
-pub async fn get_network_p2p_quarantined(
-    context: &Context,
-) -> Result<Vec<poldercast::Node>, Error> {
+pub async fn get_network_p2p_quarantined(context: &Context) -> Result<Vec<ProfileInfo>, Error> {
```
mzabaluev, line 578: Does this fn and those below need to be public?

jormungandr/src/settings/start/config.rs
```diff
@@ -197,11 +187,11 @@ pub fn default_interests() -> BTreeMap<Topic, InterestLevel> {
     BTreeMap::from_iter(vec![
         (
             Topic(topic::MESSAGES),
-            InterestLevel(poldercast::InterestLevel::Low),
+            InterestLevel(poldercast::InterestLevel::new(1)),
```
mzabaluev, line 190: Maybe define our own consts with application-defined meanings of low and normal.

jormungandr/src/settings/start/config.rs
```diff
@@ -324,9 +276,9 @@ impl<'de> Deserialize<'de> for InterestLevel {
                 use serde::de::Unexpected;
 
                 match v {
-                    "low" => Ok(InterestLevel(poldercast::InterestLevel::Low)),
-                    "normal" => Ok(InterestLevel(poldercast::InterestLevel::Normal)),
-                    "high" => Ok(InterestLevel(poldercast::InterestLevel::High)),
+                    "low" => Ok(InterestLevel::LOW),
```
mzabaluev, line 279: Wait, there is an enum for this.

jormungandr/src/settings/start/mod.rs
```diff
@@ -264,44 +269,40 @@ fn generate_network(
...
+    // Is it possible for a node to specify no public address?
```
mzabaluev, line 273: Should be possible, it's a non-public node then which should also not gossip about itself.

testing/jormungandr-integration-tests/src/networking/p2p.rs
```diff
@@ -181,6 +180,9 @@ pub fn node_does_not_quarantine_whitelisted_node() {
         .spawn_custom(params(CLIENT).policy(policy))
         .unwrap();
 
+    // Give time to the client to accept incoming gossip from the server and
+    // add it to its topology
+    process_utils::sleep(5);
     server.shutdown();
 
     process_utils::sleep(10);
```
mzabaluev, line 188: The amount of sleep in a test is getting annoyingly long.

jormungandr/src/network/p2p/topology.rs
```diff
@@ -13,8 +13,13 @@ use tracing::instrument;
 use super::{topic, NodeId};
 
 use std::hash::{Hash, Hasher};
+use std::net::{IpAddr, Ipv4Addr};
 use std::sync::Arc;
 
+lazy_static! {
+    static ref LOCAL_ADDR: Address = Address::new(IpAddr::V4(Ipv4Addr::new(127, 0, 0, 1)), 0);
```
eugene-babichenko, line 20: Should not this be doable with just `const`? `Address::new` should be edited into `const fn` if it's not already.

mzabaluev (commented):
Looks good. Just remove the now unnecessary periodic task to call `force_reset_layers` and maybe clean up the `InterestLevel` magic numbers and we should merge this.

jormungandr/src/network/p2p/topology.rs
Zeegomo: Should we remove that option from the config file as well?

jormungandr/src/rest/v0/logic.rs
Zeegomo: They are called from `jormungandr/src/rest/v0/handlers.rs`

jormungandr/src/settings/start/config.rs
Zeegomo: Agree, this is indeed reworked in https://github.com/input-output-hk/jormungandr/pull/3178. Now I see that the splitting of the PRs is starting to make some confusion, not sure what the best method would be 🤔

jormungandr/src/network/p2p/topology.rs
Zeegomo: `Address` is just an alias to `SocketAddr`, we could create our own type for this but I don't know if it is worth it

mzabaluev: Then `Address::new` is already a `const fn` so this does not need a lazy static.

mzabaluev (commented):
Can't approve my own PR, but it's good to go when we merge this with 2/3 and 3/3.

jormungandr/src/network/p2p/topology.rs
Zeegomo: is it https://doc.rust-lang.org/src/std/net/addr.rs.html#128?

PR #3141, opened 2021-03-16
Correct API endpoint path for /v1/fragments

mzabaluev:
Also apply formatting fixes auto-made by my YAML extension and fix lints.


eugene-babichenko (approved):
Are there lints for that in the repo?

mzabaluev:
> Are there lints for that in the repo?

We do checks with spectral in the io, but not the same lints. These did not look much important, and I've filed an issue for the issues with another security audit that I have used.

PR #3145, opened 2021-03-17
Change tracing to single configurable backend subscriber

### LINKED ISSUE #3142: Change the tracing subscriber to a single configurable backend
Taking another path after https://github.com/input-output-hk/jormungandr/issues/3040#issuecomment-800936871

Simplify the logging architecture and configuration to only have to deal with a single subscriber, configurable to any of the supported `Subscriber`/`Layer` implementations.

filip-dulic-bloxico:
This should close #3142 , implementing a single subscriber from any of the expected LogOutput variants.

* adds a DEFAULT_LOG_SETTINGS_ENTRY, which is the default configuration
* changes the code to import log settings from yaml, reading only the last
entry (ignoring the rest)
* command-line settings override anything else
* updates `LogSettings::init_log` method

### DIFF
--- a/doc/configuration/logging.md
+++ b/doc/configuration/logging.md
@@ -10,10 +10,6 @@ The following options are available in the `log` section:
 - `output`: Log output destination (multiple destinations are supported). Possible values are:
   - `stdout`: standard output
   - `stderr`: standard error
-  - `syslog`: syslog (only available on Unix systems)
-  - `syslogudp`: remote syslog  (only available on Unix systems)
-    - `host`: address and port of a syslog server
-    - `hostname`: hostname to attach to syslog messages
   - `journald`: journald service (only available on Linux with systemd,
     (if jormungandr is built with the `systemd` feature)
   - `gelf`: Configuration fields for GELF (Graylog) network logging protocol
--- a/doc/configuration/logging.md
+++ b/doc/configuration/logging.md
@@ -24,15 +24,18 @@ The following options are available in the `log` section:
 
 ## Example
 
-Multiple logging backends are supported.
+A single configurable backend is supported.
 
 ```yaml
 log:
   - output: stdout
     level:  trace
     format: plain
+```
+
+```yaml
   - output:
-      file: example.log
+    file: example.log
     level: info
     format: json
 ```
--- a/jormungandr/src/settings/logging.rs
+++ b/jormungandr/src/settings/logging.rs
@@ -202,12 +202,14 @@ impl LogSettings {
                 match settings.format {
                     LogFormat::Default | LogFormat::Plain => {
                         // create a reloadable layer, and return the handle
+                        // FIXME: use this handle to reload layer
                         let (layer, _handle) = reload::Layer::new(
                             tracing_subscriber::fmt::Layer::new().with_writer(non_blocking),
                         );
                         (Some(layer), None)
                     }
                     LogFormat::Json => {
+                        // FIXME: use this handle to reload layer
                         let (layer, _handle) = reload::Layer::new(
                             tracing_subscriber::fmt::Layer::new()
                                 .json()
--- a/jormungandr/src/settings/logging.rs
+++ b/jormungandr/src/settings/logging.rs
@@ -1,6 +1,6 @@
 use std::fmt::{self, Display};
 use std::fs;
-use std::io::{self};
+use std::io;
 #[cfg(feature = "gelf")]
 use std::net::SocketAddr;
 use std::path::PathBuf;
--- a/jormungandr/src/settings/logging.rs
+++ b/jormungandr/src/settings/logging.rs
@@ -13,7 +13,10 @@ use tracing::subscriber::SetGlobalDefaultError;
 #[allow(unused_imports)]
 use tracing_subscriber::layer::SubscriberExt;
 
-pub struct LogSettings(pub Vec<LogSettingsEntry>, pub LogInfoMsg);
+pub struct LogSettings {
+    pub entries: Vec<LogSettingsEntry>,
+    pub msgs: LogInfoMsg,
+}
 
 /// A wrapper to return an optional string message that we
 /// have to manually log with `info!`, we need this because
@@ -175,7 +178,7 @@ impl LogSettings {
         // Parse which settings are present for possible outputs
         let mut layer_settings = LogOutputLayerSettings::default();
         let mut info_msgs: Vec<String> = Vec::new();
-        for config in self.0.into_iter() {
+        for config in self.entries.into_iter() {
             layer_settings.read_setting(config, &mut info_msgs);
         }
         let (std_out_layer, std_out_layer_json) = if let Some(settings) = layer_settings.stdout {
@@ -297,7 +300,7 @@ impl LogSettings {
         // panics if something goes wrong.
         registry.init();
 
-        let log_msgs = match (self.1, info_msgs.is_empty()) {
+        let log_msgs = match (self.msgs, info_msgs.is_empty()) {
             (None, true) => None,
             (None, false) => Some(info_msgs),
             (Some(msgs), true) => Some(msgs),
--- a/jormungandr/src/settings/logging.rs
+++ b/jormungandr/src/settings/logging.rs
@@ -215,15 +215,15 @@ impl LogSettings {
             (None, None)
         };
         #[cfg(feature = "systemd")]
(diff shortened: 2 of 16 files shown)


Zeegomo (changes requested):
I think we should either reject a config with more than one backend configured or warn the user when that happens.

In tests we use multiple backends as well and in the opposite order (previously the only accepted one was the first in the list), so that needs to be changed as well

filip-dulic-bloxico:
@Zeegomo I've made the requested changes, let me know if there is anything else.

Zeegomo (commented):
Good work! I think we should also remove multiple logs backends from test configs to avoid confusion and update the doc.

Zeegomo (approved):

jormungandr/src/settings/logging.rs
```diff
@@ -1,29 +1,29 @@
 use std::fmt::{self, Display};
 use std::fs;
-use std::io::{self, Write};
+use std::io::{self};
```
eugene-babichenko, line 3: Just `std::io`. Don't why the linters are not triggered.

eugene-babichenko (commented):
If we only allow one backend, shouldn't it make more sense to **not** have a list here? https://github.com/input-output-hk/jormungandr/blob/b51eec8f09faf985c28bc499d6a2a4857867986a/jormungandr/src/settings/start/config.rs#L64

jormungandr/src/settings/logging.rs
filip-dulic-bloxico: Done.

doc/configuration/logging.md
```diff
@@ -24,15 +20,18 @@ The following options are available in the `log` section:
...
+A single configurable backend is supported.
 
 ```yaml
 log:
   - output: stdout
     level:  trace
     format: plain
+```
+
+```yaml
```
eugene-babichenko, line 32: One last question: I understand that doing that will break the configuration format and probably many tests. But shouldn't it be more obvious to do the following since we allow only one backend (note there is no list of backends, all configs are top-level):

```yaml
log:
  output: stdout
  level: trace
  format: plain
```

filipdulic: It would, but like you said it will break many tests etc. Perhaps we should create a new task, with this change, and test/code cleanup for it?

eugene-babichenko: Yeah, I think so. Please create a separate issue for that. Doing it here will make this PR too big.

filip-dulic-bloxico:
There tests that are failing are doing so because of a confusion caused by structs with the same name but different functionality:

```
testing/jormungandr-testing-utils/src/testing/node/logger.rs
44:pub struct LogEntry {
45-    pub level: Level,
46-    #[serde(alias = "timestamp")]
47-    pub ts: String,
48-    pub fields: RawFields,
49-    pub target: String,
50-    pub span: Option<RawFields>,
51-    pub spans: Option<Vec<RawFields>>,
52-}

jormungandr-lib/src/interfaces/config/log.rs
8:pub struct LogEntry {
9-    pub format: String,
10-    pub level: String,
11-    pub output: LogOutput,
12-}
```

It's unclear to me as of now how or if they should both be affected by changing to a single subscriber. I'm investigating by debugging the tests.

Zeegomo:
At least the first one (testing/jormungandr-testing-utils/src/testing/node/logger.rs) is just a struct to represent a logged line and deserialize it from the json output, so it should not be affected by how many subscribers there are.

The problem here is that tests configs in some places were changed to only use a file, while `JormungandrLogger` reads from the stdout of the process to avoid data races in files, it should be sufficient to change the config to just log to stdout.

filip-dulic-bloxico:
> At least the first one (testing/jormungandr-testing-utils/src/testing/node/logger.rs) is just a struct to represent a logged line and deserialize it from the json output, so it should not be affected by how many subscribers there are.
> 
> The problem here is that tests configs in some places were changed to only use a file, while `JormungandrLogger` reads from the stdout of the process to avoid data races in files, it should be sufficient to change the config to just log to stdout.

Thanks, @Zeegomo I'll try reverting to stdout/stderr in those places where I changed them.

filip-dulic-bloxico:
> > At least the first one (testing/jormungandr-testing-utils/src/testing/node/logger.rs) is just a struct to represent a logged line and deserialize it from the json output, so it should not be affected by how many subscribers there are.
> > The problem here is that tests configs in some places were changed to only use a file, while `JormungandrLogger` reads from the stdout of the process to avoid data races in files, it should be sufficient to change the config to just log to stdout.
> 
> Thanks, @Zeegomo I'll try reverting to stdout ~/stderr~ in those places where I changed them.

This seems to have fixed all except one test (`jormungandr::legacy::test_legacy_node_all_fragments`), there's something going on in the ConfiguredStarter<_> struct. I'm checking it now.

testing/jormungandr-scenario-tests/src/node.rs
```diff
@@ -760,21 +760,14 @@ impl<'a, R: RngCore, N> SpawnBuilder<'a, R, N> {
...
-                output: LogOutput::Stderr,
-            },
-            LogEntry {
-                format: format.to_string(),
-                level,
-                output: LogOutput::File(log_file.to_path_buf()),
-            },
-        ]));
+        self.node_settings.config.log = Some(Log(LogEntry {
+            format: format.to_string(),
```
Zeegomo, line 767: This will break because we are only able to read from json encoded logs, while plain format is used here

Zeegomo:
> This seems to have fixed all except one test (jormungandr::legacy::test_legacy_node_all_fragments)

I guess this is because we changed config format (i.e. only one backend)

filipdulic:
> > This seems to have fixed all except one test (jormungandr::legacy::test_legacy_node_all_fragments)
> 
> I guess this is because we changed config format (i.e. only one backend) and this test use the last release on Jormungandr (which does not have those changes)
What do you suggest?

Zeegomo:
I would start from https://github.com/input-output-hk/jormungandr/blob/32d0018f5b882bed0a073ba780c9cb89cfb0e640/testing/jormungandr-testing-utils/src/testing/node/configuration/legacy/config.rs#L52 (or that file in general as it contains config options for legacy nodes) this probably serialize the config using a yaml type that is not compatible with what the node was expecting before (i.e. it was a list and this is just a struct)

filip-dulic-bloxico:
I think that the problem comes from the fact that the `jormungandr::legacy::test_legacy_node_all_fragments` test downloads the latest release of jormungandr, which at this moment fails when reading the node configuration.

To debug, I ran the test with `--nocapture`, this usually shows how the jormungandr binary is called, along with the args used. In this case, for this legacy test, the downloaded binary fails when running:

`/tmp/.tmpooQF8f/release-0.11.0-rc1/jormungandr --secret /tmp/.tmpooQF8f/node_secret.yaml --config /tmp/.tmpooQF8f/node_config.yaml --genesis-block /tmp/.tmpooQF8f/block-0.bin` <--- **FAILS**

However, if I build the binary locally on the `change-tracing-to-single-configurable-backend-subscriber`, and use it, it works:

```
cargo build -p jormundandr --release
./target/release/jormungandr  --secret /tmp/.tmpooQF8f/node_secret.yaml --config /tmp/.tmpooQF8f/node_config.yaml --genesis-block /tmp/.tmpooQF8f/block-0.bin
```

It works fine.

Zeegomo:
Those changes would defy the idea of a legacy test. The fix itself is quite easy and it is to change (or better, revert the changes of this pr for the legacy config) the `legacy::NodeConfig` to still serialize logs config to a list instead of a single struct , here is an example https://github.com/Zeegomo/jormungandr/tree/change-tracing-to-single-configurable-backend-subscriber

PS: there is a bug in the starter code that does not correctly serialize the legacy config but the branch I linked also contains a small fix for it

filip-dulic-bloxico:
> Those changes would defy the idea of a legacy test. The fix itself is quite easy and it is to change (or better, revert the changes of this pr for the legacy config) the `legacy::NodeConfig` to still serialize logs config to a list instead of a single struct , here is an example https://github.com/Zeegomo/jormungandr/tree/change-tracing-to-single-configurable-backend-subscriber
> 
> PS: there is a bug in the starter code that does not correctly serialize the legacy config but the branch I linked also contains a small fix for it

I incorporated the changes @Zeegomo made to this branch, as they solve the failing tests perfectly. This PR should be ready to go.

PR #3153, opened 2021-03-18
add option to get public key from bech32

dkijania:
For debugging purposes we need to convert public key from bytes. This pr try to solve this.

### DIFF
--- a/jcli/src/jcli_lib/key.rs
+++ b/jcli/src/jcli_lib/key.rs
@@ -65,7 +65,7 @@ pub enum Key {
     Generate(Generate),
     /// get the public key out of a given private key
     ToPublic(ToPublic),
-    /// retrive a private key from the given bytes
+    /// retrive a private/public key from the given bytes
     FromBytes(FromBytes),
     /// get the bytes out of a private key
     ToBytes(ToBytes),
@@ -91,6 +91,9 @@ pub struct FromBytes {
 
     #[structopt(flatten)]
     output_file: OutputFile,
+
+    #[structopt(long = "public")]
+    public_key: bool,
 }
 
 #[derive(StructOpt, Debug)]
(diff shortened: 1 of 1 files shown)


jcli/src/jcli_lib/key.rs
```diff
@@ -270,6 +273,21 @@ impl FromBytes {
...
+        if self.public_key {
+            let key_bech32 = match self.key_type {
+                GenPrivKeyType::Ed25519 => bytes_to_pub_key::<Ed25519>(&bytes)?,
+                GenPrivKeyType::Ed25519Bip32 => bytes_to_pub_key::<Ed25519Bip32>(&bytes)?,
+                GenPrivKeyType::Ed25519Extended => unimplemented!(),
+                GenPrivKeyType::SumEd25519_12 => bytes_to_pub_key::<SumEd25519_12>(&bytes)?,
+                GenPrivKeyType::RistrettoGroup2HashDh => {
+                    bytes_to_pub_key::<RistrettoGroup2HashDh>(&bytes)?
+                }
+            };
```
danielSanchezQ, line 285: I would make this a function itself.

danielSanchezQ: And something like?:

```rust
let bench32_key = if self.public_key {
    process_public_key(self.key_type, bytes)
} else {
    process_private_key(self.key_type, bytes)
};
let mut output = self.output_file.open()?;
writeln!(output, "{}", key_bech32)?;
return Ok(());
```

jcli/src/jcli_lib/key.rs
```diff
@@ -270,6 +273,21 @@ impl FromBytes {
...
+            let mut output = self.output_file.open()?;
+            writeln!(output, "{}", key_bech32)?;
+            return Ok(());
+        }
```
mzabaluev, line 289: Agree with @danielSanchezQ. At least, for better readability, there should be an else branch for the private key output code below.

mzabaluev (approved):

jcli/src/jcli_lib/key.rs
```diff
@@ -91,6 +91,9 @@ pub struct FromBytes {
 
     #[structopt(flatten)]
     output_file: OutputFile,
+
+    #[structopt(long = "public")]
+    public_key: bool,
```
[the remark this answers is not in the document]
dkijania: yes

danielSanchezQ (approved):

PR #3140, opened 2021-03-16
do not re-run blocks from the leadership service

eugene-babichenko:
depends on input-output-hk/chain-libs#543

PR #3156, opened 2021-03-19
handle execution of time-consuming transactions

eugene-babichenko:
The process of building a block now has two deadline:

1. The soft deadline is equal to the end of the leadership event.
2. The hard deadline is a multiple of the slot duration. The multiplier
   is configured on the node start up.

If a single transaction was not able to complete until the hard deadline
it is forever rejected by the current node.

This patch is the 3rd and the last in the series of patches (previous: #3140, #3152) that implement the handling of time-consuming transactions, required for the successful handling of private voting.

### DIFF
--- a/jormungandr/src/settings/start/mod.rs
+++ b/jormungandr/src/settings/start/mod.rs
@@ -14,6 +14,7 @@ const DEFAULT_FILTER_LEVEL: LevelFilter = LevelFilter::TRACE;
 const DEFAULT_LOG_FORMAT: LogFormat = LogFormat::Default;
 const DEFAULT_LOG_OUTPUT: LogOutput = LogOutput::Stderr;
 const DEFAULT_NO_BLOCKCHAIN_UPDATES_WARNING_INTERVAL: u64 = 1800; // 30 min
+const DEFAULT_BLOCK_HARD_DEADLINE: u32 = 50;
 
 #[derive(Debug, Error)]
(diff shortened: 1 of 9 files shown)


eugene-babichenko:
oopsie, the new config option needs documentation

jormungandr/src/settings/start/mod.rs
```diff
@@ -14,6 +14,7 @@ const DEFAULT_FILTER_LEVEL: LevelFilter = LevelFilter::TRACE;
 const DEFAULT_LOG_FORMAT: LogFormat = LogFormat::Default;
 const DEFAULT_LOG_OUTPUT: LogOutput = LogOutput::Stderr;
 const DEFAULT_NO_BLOCKCHAIN_UPDATES_WARNING_INTERVAL: u64 = 1800; // 30 min
+const DEFAULT_BLOCK_HARD_DEADLINE: u32 = 50;
```
Zeegomo, line 17: How did you choose this value?

Zeegomo (approved):

jormungandr/src/settings/start/mod.rs
eugene-babichenko: Really depends on workload and environment. For Fund-4 benchmark tallying took 180 seconds on t3a.2xlarge. This is the equivalent of 18 slots if we go with 10 seconds per slot.

PR #3157, opened 2021-03-19
Voters rewards

danielSanchezQ:


jcli/src/jcli_app/rewards/voters.rs
```diff
@@ -0,0 +1,114 @@
...
+use std::ops::{Div, Mul};
+
+#[derive(StructOpt)]
+#[structopt(rename_all = "kebab-case")]
+pub struct VotersRewards {
+    #[structopt(flatten)]
+    common: Common,
+    /// Reward (in ADA) to be distributed
+    #[structopt(long = "total-rewards")]
+    total_rewards: f64,
```
eugene-babichenko, line 19: Using floating point numbers for anything money-related is not advised. We should used fixed-point numbers (or just scale everything up to `u64`/`u128`) and do appropriate rounding. And conversions between `u64` and `f64` will result in precision loss.

eugene-babichenko (commented):
Apart from `f64` use notice, should this really be included to `jcli`? This is very specific to Catalyst.

jcli/src/jcli_app/rewards/voters.rs
danielSanchezQ: Yeah, maybe it would be better to use Lovelace instead of ADA for this. But what rounding should we use if it is complete fractional?

danielSanchezQ:
Im happy to move it somewhere else. Maybe it's time for a catalyst related project for this kind of tools.

Maybe is time that we make a lib out of jcli functionality too. I recall @dkijania also mentioned it at some point.

danielSanchezQ:
We can move to an external project after #3163 is merged.

danielSanchezQ:
Moved to external repository: https://github.com/input-output-hk/catalyst-toolbox/pull/1

PR #3158, opened 2021-03-19
Pack logs for failed tests

filip-dulic-bloxico:
Reading failed tests output from the web UI of GitHub Actions is not very user friendly. We should pack logs of failed tests in an artifact published at the end of the workflow.

Ok, I'm implementing this in the following steps:

- [X] make temp_dir persistent using std::thread:panicking in the Drop impl for JormungandrProcess
- [X] in `.github/workflows/test.yml` add `TMPDIR: ${{ runner.temp }}/logs` to the action that runs tests as root TempDir (the auto-generated directories are still made in dirs with names like `.tmp*****`) to generate the artifact of the contents of TMPDIR.
- [ ] Once that's working, hook up the test name (using the `function_name` crate) with the JormungandrProcess `temp_dir`, maybe by printing it to stdout during tests, maybe by `touch`ing a file with the test name for filename.

### DIFF
--- a/testing/jormungandr-integration-tests/src/common/jormungandr/process.rs
+++ b/testing/jormungandr-integration-tests/src/common/jormungandr/process.rs
@@ -1,6 +1,6 @@
 use super::JormungandrError;
 use crate::common::jcli::{JCli, JCliCommand};
-use assert_fs::TempDir;
+use assert_fs::{fixture::PathChild, TempDir};
 use chain_impl_mockchain::fee::LinearFee;
 use chain_time::TimeEra;
 use jormungandr_lib::{
@@ -240,32 +240,21 @@ impl JormungandrProcess {
 
 impl Drop for JormungandrProcess {
     fn drop(&mut self) {
-        let errors = self
-            .logger
-            .get_lines_with_level(LogLevel::ERROR)
-            .collect::<Vec<_>>();
-        if !errors.is_empty() {
-            println!("Error lines:");
-            for line in errors {
-                println!("{}", line);
-            }
-        }
         // There's no kill like overkill
         let _ = self.child.kill();
         // FIXME: These should be better done in a test harness
         self.child.wait().unwrap();
-
-        if let Some(temp_dir) = self.steal_temp_dir() {
-            if panicking() {
-                println!(
-                    "persisting node temp_dir after panic: {:?}",
-                    temp_dir.path()
-                );
-                temp_dir.into_persistent();
-            } else {
-                // Ensure deletion happens.
-                temp_dir.close().unwrap();
-            }
+        if panicking() {
+            let temp_dir = self.temp_dir.take().unwrap_or_else(|| {
+                println!("No directory passed to the process, creating a new one");
+                TempDir::new().unwrap()
+            });
+            println!(
+                "persisting node temp_dir after panic: {}",
+                temp_dir.path().display()
+            );
+            std::fs::write(temp_dir.child("node.log").path(), self.log_content()).unwrap();
+            temp_dir.into_persistent();
         }
     }
 }
--- a/testing/jormungandr-integration-tests/src/common/jormungandr/process.rs
+++ b/testing/jormungandr-integration-tests/src/common/jormungandr/process.rs
@@ -255,8 +255,7 @@ impl Drop for JormungandrProcess {
         // FIXME: These should be better done in a test harness
         self.child.wait().unwrap();
 
-        if self.temp_dir.is_some() {
-            let temp_dir = self.steal_temp_dir().unwrap();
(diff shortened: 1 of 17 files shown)


testing/jormungandr-integration-tests/src/common/jormungandr/process.rs
```diff
@@ -251,9 +252,22 @@ impl Drop for JormungandrProcess {
...
         self.child.wait().unwrap();
+
+        if self.temp_dir.is_some() {
+            let temp_dir = self.steal_temp_dir().unwrap();
+            if panicking() {
+                println!(
+                    "persisting node temp_dir after panic: {:?}",
+                    temp_dir.path()
+                );
+                temp_dir.into_persistent();
```
Zeegomo, line 265: We should save the log content to a file in the dir as well

testing/jormungandr-integration-tests/src/common/jormungandr/process.rs
```diff
@@ -251,9 +252,22 @@ impl Drop for JormungandrProcess {
         }
         // There's no kill like overkill
         let _ = self.child.kill();
-
         // FIXME: These should be better done in a test harness
         self.child.wait().unwrap();
+
+        if self.temp_dir.is_some() {
+            let temp_dir = self.steal_temp_dir().unwrap();
```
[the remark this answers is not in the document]
filip-dulic-bloxico: Yup, vestigial code from prototyping the solution (TempDir doesn't implement copy, and it's inside a shared reference, so `if let Some(dir) = self.temp_dir` will not work, however, `if let Some(dir) = self.steal_temp_dir()` does.

Updated it now.

testing/jormungandr-integration-tests/src/common/jormungandr/process.rs
filip-dulic-bloxico: I'm not clear on this, when the temp_dir has files inside, and the temp_dir is persisted, the files inside survive (unless there's a place where the `node.log` file is being deleted and that I'm unaware of.

Zeegomo: What I mean is that there is no such `node.log` file already in the dir, because we only collect the logs in memory. In pratice, depending on the exact configuration there might be cases in which there is a `node.log` file, but that is completely empty (due to limitations to the current subscribers), and will ultimately be removed by https://github.com/input-output-hk/jormungandr/pull/3145.

Zeegomo:
I intentionally left a panic inside one of the tests so that we can see how that behaves

testing/jormungandr-integration-tests/src/common/jormungandr/process.rs
```diff
@@ -239,21 +240,55 @@ impl JormungandrProcess {
...
-            }
-        }
+        use rand::Rng;
         // There's no kill like overkill
         let _ = self.child.kill();
-
         // FIXME: These should be better done in a test harness
         self.child.wait().unwrap();
+        if panicking() {
+            let dir_name = rand::thread_rng()
```
eugene-babichenko, line 249: Just use `tempfile`, we already use it across all of our projects. It will take care of generating a name and creating a directory. Hint: [`TempDir::into_path`](https://docs.rs/tempfile/3.2.0/tempfile/struct.TempDir.html#method.into_path) prevents a temporary directory from being removed when a `TempDir` goes out of scope.

testing/jormungandr-integration-tests/src/common/jormungandr/process.rs
```diff
@@ -239,21 +240,55 @@ impl JormungandrProcess {
...
+                "persisting node temp_dir after panic: {}",
+                logs_dir.display()
+            );
+
+            // This code will not panic, as it will pollute the backtrace
+            // by doing so when the process is already panicking
+            std::fs::create_dir_all(&logs_dir)
+                .unwrap_or_else(|e| eprintln!("Could not create logs dir: {}", e));
+
+            if let Some(dir) = self.temp_dir.take() {
```
eugene-babichenko, line 269: Should be doable with `fs_extra`. Already imported in the project as well. Although I am not sure if it is actually used and is not a leftover dependency from long ago.

eugene-babichenko: (by this I mean that `rg fs_extra` gives me only `Cargo.toml` files)

eugene-babichenko (approved):
The code is looking good, can we remove the test panic? :)

Zeegomo:
Sure, I will do that and squash/merge the pr later today

eugene-babichenko:
Good. Feel free to merge it whenever you think it is ready. I have nothing to add from my side.

PR #3163, opened 2021-03-22
Jcli lib+bin

danielSanchezQ:
We already have a lot of building blocks in `jcli` that we can reuse for other apps or for testing.
I made a refactor to rename jcli_app into jcli_lib and created the pertinent changes in cargo so it can be used as both lib and the compiled bin we currently have.


jcli/Cargo.toml
```diff
@@ -54,3 +54,11 @@ predicates = "1.0"
...
 [build-dependencies]
 versionisator = "1.0.2"
+
+[lib]
+name = "jcli_lib"
+path = "src/lib.rs"
+
+[[bin]]
+name = "jcli"
+path = "src/main.rs"
```
eugene-babichenko, line 64: missing newline

eugene-babichenko (commented):
What is an example use of this? Just to make sure the exposes API is pretty and easy to use.

A potentially problematic part of this is dependency management. `jcli` contains deps that are only required for the app itself and would be completely useless in the lib. For example, things like `structopt` that are rather heavy in the sense they bring in a huge tree of dependencies. That said, IMO it would be better to keep `jcli_lib` as a separate crate within the workspace.

danielSanchezQ:
I agree. It would be much better to keep `jcli_lib` as a separate crate. Also `jcli_lib` API should be taken a look at and probably refactor. Right now it just exposes utils and some things in `block`. 

An example is the voter reward command that was extracted to the catalyst toolbox project: https://github.com/input-output-hk/catalyst-toolbox/pull/1

eugene-babichenko:
>  Specifically StructOpt I don't think it can't be removed, since it is part of the helpers that we can use for creating other cli tools.

Can be made feature-gated at least.

danielSanchezQ:
yeah, that sound good. But anyway that may be part of another PR where we do really think about how to refactor everything. I'll open an issue.

PR #3168, opened 2021-03-24
Move topology to task

Zeegomo:
Second part of https://github.com/input-output-hk/jormungandr/issues/3169. This PR allow us to directly use the poldercast official repo (we have to wait for a release to be able to pull the crate from crates.io), while also cleaning the codebase a bit refactoring the topology in its own task.

This PR introduces a new representation of a peer (`topology/Peer`, it was implicit before by using poldercast data types, it is more visible due to being in another task), so that now we have two: the one from p2p/peer (connection layers) and the one from the topology task. These two representation, while having both a id to identify a node,  only share the address at the moment, since we do not check the identity of a node and the precise specification for doing that are not decided yet. The two IDs are therefore independent, this PR does not attempt to unify them due to requirements still not being clear as detailed above.


Zeegomo:
superseded by #3177

PR #3184, opened 2021-03-29
Add nomad entrypoint and hydra jobs

disassembler:


jormungandr-lib/Cargo.toml
```diff
@@ -28,6 +28,7 @@ hex = "0.4"
 bech32 = "0.7"
 base64 = "0.13.0"
 http = "0.2.2"
+eccoxide = { git = "https://github.com/eugene-babichenko/eccoxide.git", branch = "fast-u64-scalar-mul", features = ["fast-u64-scalar-mul"], optional = true }
```
eugene-babichenko, line 31: What is the reason for adding this dependency here?

manveru: At the moment rust.nix doesn't support recursively collecting dependencies for git dependencies that themselves depend on git dependencies. This is a dependency from chain-libs and we decided to add it here temporarily until either chain-libs switches to the non-git version, or we implement support for this kind of project in rust.nix.

PR #3187, opened 2021-03-30
Switch to circleci image

Zeegomo:
Use CircleCI supported images instead of building our own container, should also resolve current issues with glibc versions for legacy tests


danielSanchezQ (approved):
Sweet, all green!

PR #3190, opened 2021-04-01
Add buffer zone for candidate block time check

Zeegomo:
The current candidate block selection algorithm is overly precise
in the check for blocks that appear to be in the future.
A small skew between nodes clocks could make a block appear to be
in the future with respect to the current node and cause it to
start its own fork.

Add a buffer zone (1 sec) to allow small discrepancies in the
clocks of participants, and warn when we receive a block that
appears to be in the future.


Zeegomo:
Failure is unrelated and hopefully fixed by https://github.com/input-output-hk/jormungandr/pull/3185

eugene-babichenko (approved):

PR #3192, opened 2021-04-05
replace LRU in fragment pool with the custom queue implementation & preserve the order of fragments returned to pool

eugene-babichenko:
It fixes a possible situation when fragments are returned to pool in such a way they would executed before some other fragments depending on them.

### DIFF
--- a/jormungandr/src/fragment/pool.rs
+++ b/jormungandr/src/fragment/pool.rs
@@ -162,30 +162,39 @@ fn is_transaction_valid<E>(tx: &Transaction<E>) -> bool {
 }
 
 pub(super) mod internal {
-    use std::{collections::HashMap, ptr};
-
     use super::*;
 
-    struct QueueEntry<K, V> {
+    use std::{
+        collections::HashMap,
+        hash::{Hash, Hasher},
+        ptr,
+    };
+
+    /// Doubly-linked queue with the possibility to remove elements from the middle of the list by
+    /// their keys.
+    struct IndexedDeqeue<K, V> {
+        head: *mut IndexedDequeueEntry<K, V>,
+        tail: *mut IndexedDequeueEntry<K, V>,
+
+        index: HashMap<IndexedDequeueKeyRef<K>, Box<IndexedDequeueEntry<K, V>>>,
+    }
+
+    struct IndexedDequeueEntry<K, V> {
         key: K,
         value: V,
 
-        prev: *mut QueueEntry<K, V>,
-        next: *mut QueueEntry<K, V>,
+        prev: *mut IndexedDequeueEntry<K, V>,
+        next: *mut IndexedDequeueEntry<K, V>,
     }
 
-    struct QueueKeyRef<K>(*const K);
+    /// A wrapper around the pointer to the key of the queue element. This wrapper forwards the
+    /// implementations of `Eq` and `Hash` to `K`. This is required becuase by default the
+    /// implementations of `Eq` and `Hash` from the pointer itself will be used.
+    struct IndexedDequeueKeyRef<K>(*const K);
 
-    struct Queue<K, V> {
-        head: *mut QueueEntry<K, V>,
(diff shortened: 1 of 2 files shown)


Zeegomo (commented):
I'm not entirely sure that implementing our own data-structure from scratch in this case is worth it. Maybe we can consider [intrusive-collections](https://crates.io/crates/intrusive-collections/0.9.0) for the implementation of the linked list with the ability to peek into it and delete elements in the middle?

eugene-babichenko:
With that kind of libraries it sometimes takes more time and effort to figure out how to adapt them for a particular case :)

jormungandr/src/fragment/pool.rs
```diff
@@ -163,16 +163,154 @@ fn is_transaction_valid<E>(tx: &Transaction<E>) -> bool {
...
+                .insert(IndexedDequeueKeyRef(&entry.key), entry)
+                .is_some()
+            {
+                panic!("inserted an already existing key");
+            }
+        }
+
+        fn pop_back(&mut self) -> Option<(K, V)> {
+            let tail = unsafe { self.tail.as_mut() }?;
+            let entry = self.index.remove(&IndexedDequeueKeyRef(&tail.key)).unwrap();
```
Zeegomo, line 254: I'm not sure on this, but aren't `entry` (`Box<IndexedDequeueEntry>`) and `tail` (`mut *`) aliases to the same memory location? If so, that breaks Rust aliasing rules

eugene-babichenko: Well, the whole idea was to bypass some of Rust rules to write C-style code with minimal overhead.

Zeegomo: But taking care of multiple aliases is explicitly mentioned in the doc for `unsafe as_mut()`, I guess this has to do with how the compiler does optimizations.

eugene-babichenko: Ah, you got a point, I missed that piece of the documentation. From my understanding, completing all operations involving `tail` before calling `HashMap::remove` and manually dropping it afterwards should do the trick. This should be fine, `HashMap::remove` does not actually access the data.

enzoc4:
I have not read the code here too deeply, but just in case, I think we have a really similar data structure in chain-wallet-libs: https://github.com/input-output-hk/chain-wallet-libs/blob/master/wallet/src/states.rs

It has a few differences here and there, because it is a bit more tailored for that particular use-case, but maybe there is room to have a common part in `chain-libs` or something (if we end up using this, of course).

(But it may be not worth it overall, I don't know)

jormungandr/src/fragment/pool.rs
```diff
@@ -163,16 +163,154 @@ fn is_transaction_valid<E>(tx: &Transaction<E>) -> bool {
...
+        fn len(&self) -> usize {
+            self.index.len()
+        }
+
+        fn contains(&self, key: &K) -> bool {
+            self.index.contains_key(&IndexedDequeueKeyRef(key))
+        }
+    }
+
+    unsafe impl<K, V> Send for IndexedDeqeue<K, V> {}
```
enzoc4, line 288:
```suggestion
    unsafe impl<K: Send, V: Send> Send for IndexedDeqeue<K, V> {}
```

eugene-babichenko:
> I have not read the code here too deeply, but just in case, I think we have a really similar data structure in chain-wallet-libs: [https://github.com/input-output-hk/chain-wallet-libs/blob/master/wallet/src/states.rs](https://github.com/input-output-hk/chain-wallet-libs/blob/master/wallet/src/states.rs?rgh-link-date=2021-04-06T18%3A19%3A54Z)

Well, I am not sure it's worth it, but yes, the overall idea is very similar.

Zeegomo (approved):
I'm not too comfortable with this and miri complains a little about our usage of pointers, but those rules are still experimental and I don't see anything that is obviously wrong, so if you are confident enough we can merge it.

PR #3193, opened 2021-04-06
reverse order of logger config until fix is provided

dkijania:


Zeegomo:
This will likely break all tests, since they rely on being able to read logs from stdout

PR #3215, opened 2021-04-13
explorer: use thiserror instead of error-chain

enzoc4:
There is no real reason to use `error-chain` there, besides that no one took the time to replace it. `thiserror` is much nicer to work with anyway (it's used in other places, though, so it can not be removed anyway).


eugene-babichenko (approved):

eugene-babichenko:
Great work on removing legacy stuff. We should prepare another PR remove that from all other places on drop that dependency finally.

PR #3219, opened 2021-04-14
Use longer epoch in send_all_fragments

Zeegomo:
When runners are particularly slow to start up, only a few slots remain
usable to settle transactions before the change of the epoch and
of rules (voting period / tally /ecc).

This commits adjust the epoch duration to accommodate such slow
runners.


mzabaluev (approved):
Maybe add a comment as to why the parameter had to be tuned here.

PR #3221, opened 2021-04-14
Improvement of CI script for versioning bundle of Cargo workspace crates/packages

andrcmdr:
Improvement of CI script for versioning bundle of Cargo workspace crates/packages (Cargo workspace metadata parsing in v3 script and Cargo read-manifest improvements in v2 script).

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -30,7 +30,7 @@ jobs:
 
       - id: release_info
         name: Get release information
-        run: python3 ci/release-info.py "$GITHUB_EVENT_NAME"
+        run: python3 ./ci/release-info-v2.py "$GITHUB_EVENT_NAME"
 
       - if: ${{ steps.release_info.outputs.release_type == 'nightly' }}
         name: Delete existing nightly releases
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -30,7 +30,7 @@ jobs:
 
       - id: release_info
         name: Get release information
-        run: python3 ./ci/release-info-v2.py "$GITHUB_EVENT_NAME"
+        run: python3 ./ci/release-info-v3.py "$GITHUB_EVENT_NAME"
 
       - if: ${{ steps.release_info.outputs.release_type == 'nightly' }}
         name: Delete existing nightly releases
--- a/ci/release-info-v2.py
+++ b/ci/release-info-v2.py
@@ -1,0 +1,96 @@
+import json
+import os
+import re
+import sys
+from datetime import date
+from subprocess import Popen, PIPE
+
+def read_version(crate):
+    # Reads package version from workspace manifest file
+    channel = Popen(
+        ['cargo', 'read-manifest', '--manifest-path', './{}/Cargo.toml'.format(crate)],
+        stdout=PIPE
+    )
+
+    # parse json data
+    data = json.load(channel.stdout)
+    version = data['version']
+    return version
+
+def check_version(ver):
+    # Checks package version for matching with the current tag reference
+    if ref is not None and ref != 'refs/tags/v' + str(ver):
+        return 0
+    else:
+        return 1
+
+def print_error(crate, ver, v):
+    # print(crate, ver, v)
+    if not v:
+        print(
+            '::error file={path}::version {0} does not match release tag {1}'
+            .format(str(ver), ref, path = str(crate) + '/Cargo.toml')
+        )
+
+def bundle_version(crates):
+    # Checks package versions of the crates bundle for consistency with the given tag reference
+    bundle = list(map(read_version, crates))
+    consistency = list(map(check_version, bundle))
+    # print(crates, bundle, consistency)
+
+    # print errors for packages which versions didn't match tag reference
+    if not all(consistency):
+        list(map(print_error, crates, bundle, consistency))
+        sys.exit(1)
+    else:
+        version = bundle[0]
+        return version
+
+
+event_name = sys.argv[1]
+
+date = date.today().strftime('%Y%m%d')
+
+ref = None
+if event_name == 'push':
+    ref = os.getenv('GITHUB_REF')
+    if ref.startswith('refs/tags/'):
+        release_type = 'tagged'
+    elif ref == 'refs/heads/ci/test/nightly':
+        # emulate the nightly workflow
+        release_type = 'nightly'
+        ref = None
+    else:
+        raise ValueError('unexpected ref ' + ref)
+elif event_name == 'schedule':
+    release_type = 'nightly'
+else:
+    raise ValueError('unexpected event name ' + event_name)
+
+
+# Cargo workspace crates/packages for versioning bundle
+crates = [
+    'jormungandr',
+    'jormungandr-lib',
+    'jcli',
+    'testing/jormungandr-testing-utils',
+    'testing/jormungandr-integration-tests',
+    'testing/jormungandr-scenario-tests'
+]
+
+version = bundle_version(crates)
+release_flags = ''
+if release_type == 'tagged':
+    tag = 'v' + version
+elif release_type == 'nightly':
+    version = re.sub(
+        r'^(\d+\.\d+\.\d+)(-.*)?$',
+        r'\1-nightly.' + date,
+        version,
+    )
+    tag = 'nightly.' + date
+    release_flags = '--prerelease'
+
+for name in 'version', 'date', 'tag', 'release_type', 'release_flags':
+    print('::set-output name={0}::{1}'.format(name, globals()[name]))
+
--- a/ci/release-info-v3.py
+++ b/ci/release-info-v3.py
@@ -1,0 +1,100 @@
+import json
+import os
+import re
+import sys
+from datetime import date
+from subprocess import Popen, PIPE
+
+def check_version(ver):
+    # Checks package version for matching with the current tag reference
+    # print(ver[0])
+    if ref is not None and ref != 'refs/tags/v' + str(ver[0]):
+        return 0
+    else:
+        return 1
+
+def print_error(crate, ver, v):
+    # print(crate, ver[0], v)
+    if not v:
+        print(
+            '::error file={path}::version {0} does not match release tag {1}'
+            .format(str(ver[0]), ref, path = str(crate) + '/Cargo.toml')
+        )
+
+def bundle_version(crates):
+    # Reads package versions from workspace manifest file
+    channel = Popen(
+        ['cargo', 'metadata', '--format-version=1', '--no-deps'],
+        stdout=PIPE
+    )
+
+    # parse json data
+    data = json.load(channel.stdout).get('packages')
+
+    # read, map and assign workspace crates versions to bundle package versions
+    for package_id, _ in enumerate(data):
+        if data[package_id]['name'] in crates:
+            crates[data[package_id]['name']].append(data[package_id]['version'])
+    #       print(package_id, data[package_id]['name'], data[package_id]['version'], crates)
+
+    # Checks package versions of the crates bundle for consistency with the given tag reference
+    consistency = list(map(check_version, list(crates.values())))
+    # print(crates, consistency)
+
+    # print errors for packages which versions didn't match tag reference
+    if not all(consistency):
+        list(map(print_error, list(crates.keys()), list(crates.values()), consistency))
+        sys.exit(1)
+    elif all(consistency):
+        version = list(crates.values())[0][0]
+    #   print(version)
+        return version
+
+
+event_name = sys.argv[1]
+
+date = date.today().strftime('%Y%m%d')
+
+ref = None
+if event_name == 'push':
+    ref = os.getenv('GITHUB_REF')
+    if ref.startswith('refs/tags/'):
+        release_type = 'tagged'
+    elif ref == 'refs/heads/ci/test/nightly':
+        # emulate the nightly workflow
+        release_type = 'nightly'
+        ref = None
+    else:
+        raise ValueError('unexpected ref ' + ref)
+elif event_name == 'schedule':
+    release_type = 'nightly'
+else:
+    raise ValueError('unexpected event name ' + event_name)
+
+
+# Cargo workspace crates/packages for versioning bundle
+crates = {
+    'jormungandr':[],
+    'jormungandr-lib':[],
+    'jcli':[],
+    'jormungandr-testing-utils':[],
+    'jormungandr-integration-tests':[],
+    'jormungandr-scenario-tests':[]
+}
+
+version = bundle_version(crates)
+release_flags = ''
+if release_type == 'tagged':
+    tag = 'v' + version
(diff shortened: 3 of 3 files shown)


.github/workflows/release.yml
```diff
@@ -30,7 +30,7 @@ jobs:
 
       - id: release_info
         name: Get release information
-        run: python3 ci/release-info.py "$GITHUB_EVENT_NAME"
+        run: python3 ./ci/release-info-v3.py "$GITHUB_EVENT_NAME"
```
mzabaluev, line 33: Why create versioned scripts? We need just one.

ci/release-info-v3.py
```diff
@@ -0,0 +1,100 @@
...
+        return 0
+    else:
+        return 1
+
+def print_error(crate, ver, v):
+    # print(crate, ver[0], v)
+    if not v:
+        print(
+            '::error file={path}::version {0} does not match release tag {1}'
+            .format(str(ver[0]), ref, path = str(crate) + '/Cargo.toml')
```
mzabaluev, line 21: This will not produce a correct path, because some crates live under `./testing`.

ci/release-info-v3.py
```diff
@@ -0,0 +1,100 @@
...
+    )
+
+    # parse json data
+    data = json.load(channel.stdout).get('packages')
+
+    # read, map and assign workspace crates versions to bundle package versions
+    for package_id, _ in enumerate(data):
+        if data[package_id]['name'] in crates:
+            crates[data[package_id]['name']].append(data[package_id]['version'])
+    #       print(package_id, data[package_id]['name'], data[package_id]['version'], crates)
```
mzabaluev, line 38: Remove any temp lines that were commented out.

ci/release-info-v3.py
```diff
@@ -0,0 +1,100 @@
...
+
+    # Checks package versions of the crates bundle for consistency with the given tag reference
+    consistency = list(map(check_version, list(crates.values())))
+    # print(crates, consistency)
+
+    # print errors for packages which versions didn't match tag reference
+    if not all(consistency):
+        list(map(print_error, list(crates.keys()), list(crates.values()), consistency))
```
mzabaluev, line 46: Since you're modifying crates dictionary anyway, the consistency flag could go as a field in each map value. This is a small stylistic matter, though, this works.
Why is the top-level `list` constructor and a `map` needed? `print_error` is used for the side effect not an output value. A for loop would better illustrate what the code does than the functional obfuscation.

mzabaluev (changes requested):
Thanks for the rework, the `cargo metadata` command is the right way to do it.

.github/workflows/release.yml
mzabaluev: Replace `release-info.py` with this variant and drop the rest; we use version control to, well, control versions, and it's easier to track changes when the file gets modified in place.

ci/release-info-v2.py
```diff
@@ -0,0 +1,96 @@
...
+    )
+
+    # parse json data
+    data = json.load(channel.stdout)
+    version = data['version']
+    return version
+
+def check_version(ver):
+    # Checks package version for matching with the current tag reference
+    if ref is not None and ref != 'refs/tags/v' + str(ver):
```
eugene-babichenko, line 22: This line is **very** confusing since `ref` is defined somewhere else and not before this function. I would rather write `def check_version(ref, ver)`.

ci/release-info-v2.py
```diff
@@ -0,0 +1,96 @@
...
+        return 0
+    else:
+        return 1
+
+def print_error(crate, ver, v):
```
eugene-babichenko, line 27: What is `v` exactly?

ci/release-info-v2.py
danielSanchezQ, line 22: Better use formatting: `f'refs/tags/v{str(ver)}'`

ci/release-info-v2.py
```diff
@@ -0,0 +1,96 @@
...
+
+    # parse json data
+    data = json.load(channel.stdout)
+    version = data['version']
+    return version
+
+def check_version(ver):
+    # Checks package version for matching with the current tag reference
+    if ref is not None and ref != 'refs/tags/v' + str(ver):
+        return 0
```
danielSanchezQ, line 23: Why use numbers here?, you could simply return a boolean with the whole expression in the if clause.

ci/release-info-v2.py
```diff
@@ -0,0 +1,96 @@
...
+    else:
+        return 1
+
+def print_error(crate, ver, v):
+    # print(crate, ver, v)
+    if not v:
+        print(
+            '::error file={path}::version {0} does not match release tag {1}'
```
danielSanchezQ, line 31: `f'::error file={str(crate)}/Cargo.toml::version {str(ver)} does not match release tag {ref}'`

ci/release-info-v2.py
```diff
@@ -0,0 +1,96 @@
...
+    consistency = list(map(check_version, bundle))
+    # print(crates, bundle, consistency)
+
+    # print errors for packages which versions didn't match tag reference
+    if not all(consistency):
+        list(map(print_error, crates, bundle, consistency))
+        sys.exit(1)
+    else:
+        version = bundle[0]
+        return version
```
danielSanchezQ, line 47: This block is not needed. Could be outside the else.

ci/release-info-v3.py
```diff
@@ -0,0 +1,100 @@
...
+from datetime import date
+from subprocess import Popen, PIPE
+
+def check_version(ver):
+    # Checks package version for matching with the current tag reference
+    # print(ver[0])
+    if ref is not None and ref != 'refs/tags/v' + str(ver[0]):
+        return 0
+    else:
+        return 1
```
danielSanchezQ, line 14: This duplicated code could be avoided if making an `__init__.py` file in the `ci` folder and transforming it into a module iirc.

ci/release-info-v3.py
danielSanchezQ: +1, use a for loop instead

ci/release-info-v3.py
```diff
@@ -0,0 +1,100 @@
...
+    elif ref == 'refs/heads/ci/test/nightly':
+        # emulate the nightly workflow
+        release_type = 'nightly'
+        ref = None
+    else:
+        raise ValueError('unexpected ref ' + ref)
+elif event_name == 'schedule':
+    release_type = 'nightly'
+else:
+    raise ValueError('unexpected event name ' + event_name)
```
danielSanchezQ, line 72: We could avoid duplicates and this could be a function too.

danielSanchezQ:
I would encourage the use of formatting strings wherever is possible to use.

ci/release-info-v3.py
andrcmdr: OK, yeah, this was for testing and control of the correct output for this preliminary PR. Thanks for mention that! Will remove for sure!

ci/release-info-v2.py
andrcmdr: This is forming a bit mask for checking the crates version consistency in a whole bundle.

ci/release-info-v2.py
```diff
@@ -0,0 +1,96 @@
...
+import os
+import re
+import sys
+from datetime import date
+from subprocess import Popen, PIPE
+
+def read_version(crate):
+    # Reads package version from workspace manifest file
+    channel = Popen(
+        ['cargo', 'read-manifest', '--manifest-path', './{}/Cargo.toml'.format(crate)],
```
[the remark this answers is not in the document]
andrcmdr: Popen used previously and works fine - Python has a subprocess.Popen(), subprocess.call(), os.system(). I don't see an actual advantages of one or another in our case, we need just correctly transfer a JSON string through the PIPE, so I prefer to leave it with subprocess.Popen() and PIPE, as it was before.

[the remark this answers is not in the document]
andrcmdr: And this is v2 script - I rewrote it later to v3 script which usied cargo metadata to gather the data about workspace and select exact packages for bundle from workspace.

ci/release-info-v2.py
andrcmdr: ref is defined before this function and passing by reference, passing ref as parameter for this function prevents the use of this function in parallel map, 'cause map stops on most short iterable and ref is a one item container.

ci/release-info-v2.py
danielSanchezQ: This is used here: `consistency = list(map(check_version, bundle))`. Which value is used in an `all` call. `all` already check boolean(y) values. So we could actually keep the bools.

ci/release-info-v2.py
andrcmdr: v is a bit filed mapped from the consistence list to check if the version of exact package is differ from the other packages in a bundle.

ci/release-info-v2.py
andrcmdr: I understand this relative new approach in Python for forming the strings, but this involve more complication and decrease the readability, so I leave this as it was in original script, just added typecast to string for version.

ci/release-info-v2.py
andrcmdr: This return actually used below for version variable used in the script main body.

ci/release-info-v3.py
andrcmdr: I'd prefer to have a one script to rule the whole CI pipeline for making the release and then packages, not to split'em or use __init__, this is uneeded complication, so I'd prefer to leave it as it was in very first version of script.

ci/release-info-v2.py
danielSanchezQ: Yeah, I mean:
```python
if ____:
    return
return bundle[0]
```

ci/release-info-v3.py
andrcmdr: This block of the script body just works fine and remain as it is from very first script, It checks the type of event received from GH Actions and I'm not changing this part, it just works well enough and clearly readable.

ci/release-info-v2.py
andrcmdr: Yeah, thanks to mention that! I will add the check of consistency here, as I did already in futher version 3 of this script. Mainly bundle_version() return is conditioned by consistency check for versions of packages in bundle, versions must be similar, and only then the return of one consolidated version must be performed.

ci/release-info-v3.py
danielSanchezQ: Btw the wrapping list for crates./keys/values are useless here, it would work without the extra allocation.

.github/workflows/release.yml
andrcmdr: Thanks for mention that! Yeah, I will make only one script version, for sure, but I placed a three versions of scripts to show the differences between versions and use of different approaches, for preliminary review and for testing purposes, and for more simple changing scripts in [release.yml](https://github.com/input-output-hk/jormungandr/blob/master/.github/workflows/release.yml).

ci/release-info-v3.py
andrcmdr: Sure, will rework that - will retrieve paths from manifest json data, 'cause in v3 script the paths are can be taken only from metadata, that's why here I've decided to use only crate names, to not introduce more complexity into parsing and dict with packages bundle.

ci/release-info-v2.py
andrcmdr: I'd prefer to use "string".format() of printf(), 'cause this improves readability and used in Python and C for many years, rather than use of such string formatting - this is relatively recent addition in Python 3 and doesn't have any advantages over format() or printf() functions, nor improves the readability. But here all of this is unnecessary - str() here is only for type cast the version into string in side cases, all other condition check expression remain as it was before in very first version of script.

ci/release-info-v2.py
andrcmdr: Isn't Python resolve 1 and 0 into booleans in condition checks? This used in a simple typecast in condition checks - if v (element in consistency) is 0 this means that this package doesn't compliant with tag and bundle version.

ci/release-info-v3.py
andrcmdr: > print_error is used for the side effect not an output value.

Nope - print_error exactly print and output all errors in parallel, and this is how maps works

> Why is the top-level list constructor and a map needed?

Without collecting an output of map (errors) in a list data structure in the main consistency check - there would be no output at all for the program, 'cause map should output error messages into container as collector, this list doesn't assigned to some variable, 'cause this isn't used anywhere else, it's the final step of execution, if consistency isn't reached and some of the package versions didn't correspond to tag and version of the whole bundle.

> Since you're modifying crates dictionary anyway, the consistency flag could go as a field in each map value. This is a small stylistic matter, though, this works.

I thought about that - it's better to separate a bit field mask (consistency of package versions) from main data structure with crates in a bundle (with package names and versions), 'cause the crates dictionary may be used in further for other information from metadata, like paths to manifests (this is planned next change), and binding to array element numbers in that way will decrease the flexibility of future use of main data structure (dict in this case in v3 script). That's why it's separated.

> A for loop would better illustrate what the code does than the functional obfuscation.

> +1, use a for loop instead

In first implementation I was used for loops. But because there's a need to iterate on elements from several data structures I was using map(). There are at least two, or even three data structures used for crate names and versions (dict used for this in v3 script - pairs of elements at least are used in one dict, and this has the same time complexity, only container format changed and space complexity little bit decreased), and bit mask flags/fields for consistency check - there always will be nested for loop (in Python manifest - flat is better than nested!) for traversing all elements, check consistency, output versions and package name, and all in one cycle run.
Maps do this in parallel - it takes elements from every iterable/enumerable data structure queue and runs function for this bunch of elements, thus map works and apply function in parallel for all data structures, has linear time complexity, perfect scalability and readaility.
**map(), reduce(), filter(), all(), even fold/foldl/foldr all has a O(n) linear time complexity. One loop nested in another will cause O(n²) quadratic time complexity.**
Even non-nested loop is way more slower than parallel map - even for such little queue of elements it's sensitive for time of output for CI (even milliseconds counts in CI!). For v2 it even more sensitive 'cause script parse in parallel map a massive json string output of `cargo read-manifest` for every package - if it'll be a loop, even non-nested, this will be way less effective.

[https://www.skoumal.com/en/performance-of-built-in-higher-order-functions-map-filter-reduce-and-flatmap-vs-for-in-loop-in-swift/](https://www.skoumal.com/en/performance-of-built-in-higher-order-functions-map-filter-reduce-and-flatmap-vs-for-in-loop-in-swift/)

And there's no any obfuscation - with map(), reduce(), filter(), all() and so on, code is more clear, plain, flat and readable, than with loops and conditions, and readability always counts.

andrcmdr:
Thanks for review guys! Will little bit rework script and push it.

ci/release-info-v3.py
danielSanchezQ: Just trying to help here, it does not work as that in most of the cases.

PR #3224, opened 2021-04-16
Update bootstrap

Zeegomo:
Collect information about network topology when bootstrapping so that the node does not have to wait until the first gossip.
This should fix a bunch of failing scenario tests (https://cardano-rust-testrun-logs.s3.eu-central-1.amazonaws.com/private-network/last_test_report.html)

The downside of this is that nodes that skip the bootstrapping phase will have no peers to connect to, and will rely only on incoming connections to be able join the topology, unless we include the step that fetches available peers from trusted peers anyway.
I'm not really sure of what is the purpose of the `skip_bootstrap` option, if it is used only when starting the first node in a chain then this behavior makes sense.

waiting for https://github.com/input-output-hk/chain-libs/pull/552 to be merged

### DIFF
--- a/jormungandr/src/topology/mod.rs
+++ b/jormungandr/src/topology/mod.rs
@@ -86,18 +86,7 @@ impl Serialize for NodeId {
 
 /// This represents a peer and its public key used for
 /// identification in the topology.
-#[derive(Debug, Hash, Clone, PartialEq, Eq)]
-pub struct Peer(Gossip);
-
-impl Peer {
-    pub fn address(&self) -> Address {
-        self.0.address()
-    }
-
-    pub fn id(&self) -> NodeId {
-        self.0.id()
-    }
-}
+pub type Peer = Gossip;
 
 #[derive(Eq, Clone, Serialize, Debug)]
 pub struct PeerInfo {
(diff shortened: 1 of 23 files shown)


eugene-babichenko (approved):

jormungandr/src/topology/mod.rs
```diff
@@ -86,13 +86,17 @@ impl Serialize for NodeId {
...
 /// identification in the topology.
-#[derive(Debug, Clone)]
-pub struct Peer {
-    pub addr: Address,
-    // This will be missing only for trusted peers before we connect
-    // to them if no if is specified if the config file.
-    // We might enforce this field to be present in the future.
-    pub id: Option<NodeId>,
+#[derive(Debug, Hash, Clone, PartialEq, Eq)]
+pub struct Peer(Gossip);
```
mzabaluev, line 90: Might as well just be a type alias of `Gossip`.

testing/jormungandr-testing-utils/Cargo.toml
```diff
@@ -63,6 +63,8 @@ tracing-subscriber = "0.2"
 tracing = "0.1"
 netstat2 = "0.9"
 multiaddr = { package = "parity-multiaddr", version = "0.11" }
+keynesis = "1"
+bincode = "1"
```
mzabaluev, line 67: It would be good to be consistent with the version spec elsewhere:

```suggestion
bincode = "1.3"
```

testing/jormungandr-testing-utils/src/testing/node/grpc/server/mod.rs
```diff
@@ -129,7 +129,17 @@ impl Node for JormungandrServerImpl {
...
         info!(method = %MethodType::GetPeers, "Get peers request received");
-        Ok(Response::new(PeersResponse::default()))
+        use bincode::Options;
+        let data = self.data.read().unwrap();
+        let mut self_gossip = Vec::new();
+        let config = bincode::options();
+        config.with_limit(512);
+        config
+            .serialize_into(&mut self_gossip, data.profile().gossip().as_ref())
+            .unwrap();
```
mzabaluev, line 139: The code duplication is because the gossip format is controlled by the poldercast crate and we decided not to expose it in chain-libs, correct?

mzabaluev (approved):

testing/jormungandr-testing-utils/src/testing/node/grpc/server/mod.rs
Zeegomo: The gossip format is controlled by Jormungandr's `topology::gossip`, and that is not exposed outside of the node.

Zeegomo:
The vicinity layer is disable because of issue in poldercast (fix here https://github.com/primetype/poldercast/pull/36). I'm not sure if we want to wait for that or go with this anyway. The impact of the vicinity layer missing should only be a slower propagation of information in large networks, while this PR alone introduces several improvements.

eugene-babichenko:
Fine by me. Currently we do not have large enough deployments to be affected by this. Other commits are looking good as well.

PR #3225, opened 2021-04-18
prepare bootstrap.py as a replacement for shell and PowerShell scripts

eugene-babichenko:
Nits:

* This script is using `colorama` (when available in the system; it is a very widely used library, so should not be a problem) for cross-platform output coloring. This solution is a bit debatable and suggestions on how to do better are welcome.
* We should agree on a formatter/linter for Python scripts. Currently I am using `autopep8` and its defaults are terrible.

TODO:

- [] Re-implement `*.shtempl` stuff.
- [] Remove old scripts.


scripts/bootstrap.py
```diff
@@ -0,0 +1,271 @@
...
+
+color_red = ''
+color_white = ''
+color_green = ''
+color_blue = ''
+
+if not args.no_colors:
+    try:
+        # colorama is often present in a system and provides cross-platform coloring
+        from colorama import init, Fore
```
danielSanchezQ, line 87: If we are about to add extra deps it may be work using some higher abstraction libs, we may or may not but at least you may be interested in:

[rich](https://github.com/willmcgugan/rich)
[typer](https://typer.tiangolo.com/)

scripts/bootstrap.py
```diff
@@ -0,0 +1,271 @@
...
+    [jcli, 'certificate', 'new', 'stake-delegation', faucet_public_key, stake_pool_id], capture_output=True, check=True).stdout
+stake_delegation1 = sign_certificate(
+    stake_delegation1_unsigned, faucet_secret_key)
+
+stake_delegation2_unsigned = subprocess.run(
+    [jcli, 'certificate', 'new', 'stake-delegation', fixed_public_key, stake_pool_id], capture_output=True, check=True).stdout
+stake_delegation2 = sign_certificate(
+    stake_delegation2_unsigned, fixed_secret_key)
+
+genesis_yaml = '''\
```
danielSanchezQ, line 150: We could use [pydantic](https://pydantic-docs.helpmanual.io/) for this

danielSanchezQ:
I see some of the lines and the calls are quite large. In my experience, autopep8 and tools like that are not quite good (all hail `rustfmt`). Usually, I just format everything by hand.

danielSanchezQ:
I know it is a script. But usually, I think is better to include things in functions as possible and then make a `__main__` block in where those methods are called so that they are read as a story.

scripts/bootstrap.py
eugene-babichenko: Not at all. I want to keep them optional and use the most common thing available. Many PyPi packages include `colorama` as a dependency, so I decided to go with it.

Now as I think of it, Windows 10 is said to [support ANSI codes](https://docs.microsoft.com/en-us/windows/console/console-virtual-terminal-sequences), so we may try to go without introducing extra dependencies and see how it goes.

scripts/bootstrap.py
eugene-babichenko: We can achieve pretty much the same thing by just passing a `kwargs` dictionary. Or use [template strings](https://docs.python.org/3.4/library/string.html#template-strings).

eugene-babichenko:
> I see some of the lines and the calls are quite large. In my experience, autopep8 and tools like that are not quite good (all hail `rustfmt`). Usually, I just format everything by hand.

[`black`]() does quite a good job. But it would require some extra configuration to not break everything (for instance, it enforces the use of `"` by default). Since the amount of Python scripts is getting quite big, I would prefer to have some conventional & automated formatting at this point.

scripts/bootstrap.py
danielSanchezQ: Ok, no lib no (3rd) party :P

danielSanchezQ (approved):
LGTM

PR #3226, opened 2021-04-18
CI checks for Python scripts & better formatting

eugene-babichenko:


danielSanchezQ (approved):
Sweet, gonna use it in catalyst toolbox too. Thanks!

.github/workflows/python-scripts.yml
```diff
@@ -0,0 +1,16 @@
...
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Install linters
+        run: pip3 install black
+
+      - uses: actions/checkout@v2
+
+      - name: Check formatting (black)
+        run: black . --check
```
danielSanchezQ, line 16: Could we check the formatting but filter just the scripts folder?

eugene-babichenko: It would actually be `scripts` and `ci`, but yes, definitely need to do that.

PR #3230, opened 2021-04-20
Jcli make transaction

danielSanchezQ:
I'm starting this as a draft. It is a close example of what jcli path to being a lib should lead to. 

The simplified transaction command uses several of the other transaction commands.
I started extracting the common code into external functions. Also stripping some functionality like reading and writing to files that shouldn't be within the reusable computations.

### DIFF
--- a/jcli/src/jcli_lib/address.rs
+++ b/jcli/src/jcli_lib/address.rs
@@ -131,7 +131,7 @@ fn mk_delegation(prefix: &str, s: PublicKey<Ed25519>, testing: bool, d: PublicKe
     mk_address_2(prefix, s, d, testing, Kind::Group)
 }
 
-fn mk_account(prefix: &str, s: PublicKey<Ed25519>, testing: bool) {
+pub fn mk_account(prefix: &str, s: PublicKey<Ed25519>, testing: bool) {
     mk_address_1(prefix, s, testing, Kind::Account)
 }
 
--- a/jcli/src/jcli_lib/address.rs
+++ b/jcli/src/jcli_lib/address.rs
@@ -131,7 +131,7 @@ fn mk_delegation(prefix: &str, s: PublicKey<Ed25519>, testing: bool, d: PublicKe
     mk_address_2(prefix, s, d, testing, Kind::Group)
 }
 
-pub fn mk_account(prefix: &str, s: PublicKey<Ed25519>, testing: bool) {
+fn mk_account(prefix: &str, s: PublicKey<Ed25519>, testing: bool) {
     mk_address_1(prefix, s, testing, Kind::Account)
 }
 
--- a/jcli/src/jcli_lib/rest/config.rs
+++ b/jcli/src/jcli_lib/rest/config.rs
@@ -7,7 +7,7 @@ use std::path::PathBuf;
 use structopt::StructOpt;
 use thiserror::Error;
 
-#[derive(StructOpt)]
+#[derive(StructOpt, Clone)]
 pub struct RestArgs {
     /// node API address. Must always have `http://` or `https://` prefix.
     /// E.g. `-h http://127.0.0.1`, `--host https://node.com:8443/cardano/api`
--- a/jcli/src/jcli_lib/rest/config.rs
+++ b/jcli/src/jcli_lib/rest/config.rs
@@ -12,7 +12,7 @@ pub struct RestArgs {
     /// node API address. Must always have `http://` or `https://` prefix.
     /// E.g. `-h http://127.0.0.1`, `--host https://node.com:8443/cardano/api`
     #[structopt(short, long, env = "JORMUNGANDR_RESTAPI_URL")]
-    host: Url,
+    pub host: Url,
     /// print additional debug information to stderr.
     /// The output format is intentionally undocumented and unstable
     #[structopt(long)]
--- a/jcli/src/jcli_lib/rest/v0/account/mod.rs
+++ b/jcli/src/jcli_lib/rest/v0/account/mod.rs
@@ -1,5 +1,6 @@
 use crate::jcli_lib::rest::{Error, RestArgs};
 use crate::jcli_lib::utils::{AccountId, OutputFormat};
+use jormungandr_lib::interfaces::AccountState;
 use structopt::StructOpt;
 
 #[derive(StructOpt)]
@@ -24,13 +25,20 @@ impl Account {
             output_format,
             account_id,
         } = self;
-        let state = args
-            .client()?
-            .get(&["v0", "account", &account_id.to_url_arg()])
-            .execute()?
-            .json()?;
-        let formatted = output_format.format_json(state)?;
+        let state = request_account_information(args, account_id)?;
+        let formatted = output_format.format_json(serde_json::to_value(state)?)?;
         println!("{}", formatted);
         Ok(())
     }
+}
+
+pub fn request_account_information(
+    args: RestArgs,
+    account_id: AccountId,
+) -> Result<AccountState, Error> {
+    Ok(args
+        .client()?
+        .get(&["v0", "account", &account_id.to_url_arg()])
+        .execute()?
+        .json()?)
 }
--- a/jcli/src/jcli_lib/rest/v0/account/mod.rs
+++ b/jcli/src/jcli_lib/rest/v0/account/mod.rs
@@ -36,9 +36,9 @@ pub fn request_account_information(
     args: RestArgs,
     account_id: AccountId,
 ) -> Result<AccountState, Error> {
-    Ok(args
-        .client()?
+    args.client()?
         .get(&["v0", "account", &account_id.to_url_arg()])
         .execute()?
-        .json()?)
+        .json()
+        .map_err(Into::into)
 }
--- a/jcli/src/jcli_lib/rest/v0/message/mod.rs
+++ b/jcli/src/jcli_lib/rest/v0/message/mod.rs
@@ -2,7 +2,7 @@ use crate::jcli_lib::{
     rest::{Error, RestArgs},
     utils::{io, OutputFormat},
 };
-use chain_core::property::Deserialize;
+use chain_core::property::{Deserialize, Serialize};
 use chain_impl_mockchain::fragment::Fragment;
 use std::path::PathBuf;
 use structopt::StructOpt;
@@ -57,14 +57,19 @@ fn get_logs(args: RestArgs, output_format: OutputFormat) -> Result<(), Error> {
 fn post_message(args: RestArgs, file: Option<PathBuf>) -> Result<(), Error> {
     let msg_hex = io::read_line(&file)?;
     let msg_bin = hex::decode(&msg_hex)?;
-    let _fragment =
+    let fragment =
         Fragment::deserialize(msg_bin.as_slice()).map_err(Error::InputFragmentMalformed)?;
+    let fragment_id = post_fragment(args, fragment)?;
+    println!("{}", fragment_id);
+    Ok(())
+}
+
+pub fn post_fragment(args: RestArgs, fragment: Fragment) -> Result<String, Error> {
     let fragment_id = args
         .client()?
         .post(&["v0", "message"])
-        .body(msg_bin)
+        .body(fragment.serialize_as_vec()?)
         .execute()?
         .text()?;
-    println!("{}", fragment_id);
-    Ok(())
+    Ok(fragment_id)
 }
--- a/jcli/src/jcli_lib/transaction/mod.rs
+++ b/jcli/src/jcli_lib/transaction/mod.rs
@@ -64,8 +64,8 @@ pub enum Transaction {
     Auth(auth::Auth),
     /// get the message format out of a sealed transaction
     ToMessage(common::CommonTransaction),
-    /// send a transaction from faucet (simplified method)
-    Simplified(simplified::SimplifiedTransaction),
+    /// send a transaction from one account to another (simplified method)
+    MakeTransaction(simplified::MakeTransaction),
 }
 
 type StaticStr = &'static str;
@@ -203,6 +203,9 @@ pub enum Error {
 
     #[error("could not generate random key")]
     RandError(#[from] rand::Error),
+
+    #[error("invalid block0 header hash")]
+    InvalidBlock0HeaderHash,
 }
 
 /*
@@ -231,7 +234,7 @@ impl Transaction {
             Transaction::MakeWitness(mk_witness) => mk_witness.exec(),
             Transaction::Auth(auth) => auth.exec(),
             Transaction::ToMessage(common) => display_message(common),
-            Transaction::Simplified(send) => send.exec(),
+            Transaction::MakeTransaction(send) => send.exec(),
         }
     }
 }
--- a/jcli/src/jcli_lib/transaction/mod.rs
+++ b/jcli/src/jcli_lib/transaction/mod.rs
@@ -64,8 +64,8 @@ pub enum Transaction {
     Auth(auth::Auth),
     /// get the message format out of a sealed transaction
     ToMessage(common::CommonTransaction),
-    /// send  a transaction from faucet (simplified method)
-    Send(simplified::SimplifiedTransaction),
+    /// send a transaction from faucet (simplified method)
+    Simplified(simplified::SimplifiedTransaction),
 }
 
 type StaticStr = &'static str;
@@ -231,7 +231,7 @@ impl Transaction {
             Transaction::MakeWitness(mk_witness) => mk_witness.exec(),
             Transaction::Auth(auth) => auth.exec(),
             Transaction::ToMessage(common) => display_message(common),
-            Transaction::Send(send) => send.exec(),
+            Transaction::Simplified(send) => send.exec(),
         }
     }
 }
--- a/jcli/src/jcli_lib/transaction/mod.rs
+++ b/jcli/src/jcli_lib/transaction/mod.rs
@@ -1,15 +1,16 @@
-mod add_account;
+pub mod add_account;
 mod add_certificate;
 mod add_input;
-mod add_output;
+pub mod add_output;
 mod add_witness;
 mod auth;
 mod common;
-mod finalize;
+pub mod finalize;
 mod info;
 mod mk_witness;
-mod new;
+pub mod new;
 mod seal;
+mod simplified;
 mod staging;
 
 use self::staging::StagingKind;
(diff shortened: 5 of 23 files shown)


danielSanchezQ:
This is related to #3216

jcli/src/jcli_lib/transaction/simplified.rs
```diff
@@ -0,0 +1,68 @@
...
+use chain_impl_mockchain::transaction::Output;
+use jormungandr_lib::interfaces;
+use structopt::StructOpt;
+
+#[derive(StructOpt)]
+#[structopt(rename_all = "kebab-case")]
+pub struct SimplifiedTransaction {
+    /// the account to debit the funds from
+    #[structopt(name = "ACCOUNT")]
+    pub faucet_address: interfaces::Address,
```
enzoc4, line 13: I don't really see anything specific to faucets here. Why not just `sender_account` or something?

danielSanchezQ: It was called `faucet_address` in the original script, so I kept it like that. But you are right, gonna rename it :)

jcli/src/jcli_lib/transaction/mod.rs
```diff
@@ -62,6 +63,8 @@ pub enum Transaction {
     Auth(auth::Auth),
     /// get the message format out of a sealed transaction
     ToMessage(common::CommonTransaction),
+    /// send  a transaction from faucet (simplified method)
+    Send(simplified::SimplifiedTransaction),
```
enzoc4, line 67: also, a command named `Send` may be a bit confusing, at least to me it suggests network interaction

danielSanchezQ: Makes sense

jcli/src/jcli_lib/address.rs
```diff
@@ -131,7 +131,7 @@ fn mk_delegation(prefix: &str, s: PublicKey<Ed25519>, testing: bool, d: PublicKe
     mk_address_2(prefix, s, d, testing, Kind::Group)
 }
 
-fn mk_account(prefix: &str, s: PublicKey<Ed25519>, testing: bool) {
+pub fn mk_account(prefix: &str, s: PublicKey<Ed25519>, testing: bool) {
```
danielSanchezQ, line 134: I have to reverse this, since finally I am not really using it outside.

jcli/src/jcli_lib/transaction/mod.rs
```diff
@@ -62,6 +64,8 @@ pub enum Transaction {
     Auth(auth::Auth),
     /// get the message format out of a sealed transaction
     ToMessage(common::CommonTransaction),
+    /// send a transaction from faucet (simplified method)
+    Simplified(simplified::SimplifiedTransaction),
```
danielSanchezQ, line 68: I didn't really know how to name this (shortly). Any suggestion is welcome.

jcli/src/jcli_lib/transaction/simplified.rs
```diff
@@ -0,0 +1,161 @@
...
+        value: value.into(),
+    })?;
+
+    //finalize
+    transaction::finalize::finalize(fee, change, &mut transaction)?;
+
+    // get transaction and block0 ids
+    let transaction_sign_data_hash = transaction.transaction_sign_data_hash();
+    let block0 = load_block(open_block_file(&Some(block0_file))?)?;
+    let block0_hash = block0.header.id();
```
eugene-babichenko, line 133: If the only point of using `block0` is to take its hash, then just use its hash as a command argument.

jcli/src/jcli_lib/transaction/simplified.rs
eugene-babichenko: Don't pay too much attention to the original script naming. `jcli` is universal and does not care about the details of specific deployments.

jcli/src/jcli_lib/address.rs
danielSanchezQ: yeah, sorry, Sometimes I just make annotations so I do not forget to do so 😅

jcli/src/jcli_lib/rest/config.rs
```diff
@@ -7,12 +7,12 @@ use std::path::PathBuf;
...
 use thiserror::Error;
 
-#[derive(StructOpt)]
+#[derive(StructOpt, Clone)]
 pub struct RestArgs {
     /// node API address. Must always have `http://` or `https://` prefix.
     /// E.g. `-h http://127.0.0.1`, `--host https://node.com:8443/cardano/api`
     #[structopt(short, long, env = "JORMUNGANDR_RESTAPI_URL")]
-    host: Url,
+    pub host: Url,
```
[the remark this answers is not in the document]
danielSanchezQ: I needed to access that field that otherwise is just for internal use. If embedded in some other command (as happens here) it is handy to access it. Do you think a method would be better?

jcli/src/jcli_lib/rest/v0/account/mod.rs
```diff
@@ -24,13 +25,20 @@ impl Account {
...
         println!("{}", formatted);
         Ok(())
     }
 }
+
+pub fn request_account_information(
+    args: RestArgs,
+    account_id: AccountId,
+) -> Result<AccountState, Error> {
+    Ok(args
```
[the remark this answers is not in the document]
danielSanchezQ: Actually, yes. Thanks.

jcli/src/jcli_lib/transaction/mod.rs
```diff
@@ -62,6 +64,8 @@ pub enum Transaction {
     Auth(auth::Auth),
     /// get the message format out of a sealed transaction
     ToMessage(common::CommonTransaction),
+    /// send a transaction from faucet (simplified method)
```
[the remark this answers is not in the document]
danielSanchezQ: I guess you wanted to say it shouldn't?

jcli/src/jcli_lib/transaction/simplified.rs
```diff
@@ -0,0 +1,161 @@
...
+}
+
+fn create_receiver_secret_key_and_address(
+) -> Result<(SecretKey<Ed25519Extended>, interfaces::Address), Error> {
+    let sk = create_new_private_key()?;
+    let address = create_receiver_address(&sk);
+    Ok((sk, address))
+}
+
+#[allow(clippy::too_many_arguments)]
```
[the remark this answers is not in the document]
danielSanchezQ: It has indeed but is what is needed. Even if splitting it would have to receive such data. I'll try to do something to reduce that anyway.

jcli/src/jcli_lib/transaction/simplified.rs
danielSanchezQ: Cool. Just tried to stick to the original script here (and in other places).

jcli/src/jcli_lib/transaction/simplified.rs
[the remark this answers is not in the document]
danielSanchezQ: I could refactor things inside in some smaller sections maybe. But the grouping function would still need to receive that. At the end we need all we have in the cli input to compute :/

jcli/src/jcli_lib/address.rs
eugene-babichenko: We can make GitHub Actions fail upon meeting TODO/FIXME comments. ;)

jcli/src/jcli_lib/rest/config.rs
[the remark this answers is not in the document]
eugene-babichenko: > If embedded in some other command (as happens here) it is handy to access it.

Fine if that's the case :+1:

jcli/src/jcli_lib/transaction/mod.rs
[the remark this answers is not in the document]
eugene-babichenko: Yes :sweat_smile:

eugene-babichenko (changes requested):
`grep`-ing (or `rg`-ing for fellow Rustaceans) still shows notions of faucet and "simplified transactions".

jcli/src/jcli_lib/address.rs
danielSanchezQ: It would not work. because it was not even noted as a fixme or todo. The thing is that I usually make a self review, if I spot something that shouldn't be there for any reason I make a note in the review so I do not forget to change it 😄

eugene-babichenko:
Definitely not how errors should be reported.

```
thread 'main' panicked at 'Found positional argument which is not required with a lower index than a required positional argument: "change" index 4', /home/babichenko/.cargo/registry/src/github.com-1ecc6299db9ec823/clap-2.33.3/src/app/parser.rs:643:21
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

eugene-babichenko:
The help itself is fine and accessible via `jcli help transaction make-transaction`. This must be an issue with the library.

danielSanchezQ:
The failing test looks no related.

eugene-babichenko:
It works :+1:. Here are my concerns:

- It outputs `Private key of receiver (to revert transaction for testing purposes)` and I would rather remove it for security reasons.
- We do actual transaction sending here, but (see the word "offline")
```
	❯ jcli transaction --help
	jcli-transaction 0.11.1
	Build and view offline transaction
```
- I think it might be better to set the default values of fees to the Jormungandr defaults.
- The help is still sort of broken (some arguments not showing) and I think this is due to structure fields ordering again.
```
	❯ jcli transaction make-transaction -h
	error: The following required arguments were not provided:
    	<ACCOUNT>
    	<VALUE>
    	<block0-hash>
	
	USAGE:
```

danielSanchezQ:
@eugene-babichenko how about?:

> It outputs Private key of receiver (to revert transaction for testing purposes) and I would rather remove it for security reasons.

Maybe add flag to enable disable it?

> We do actual transaction sending here, but (see the word "offline")

Didn't understand this. Could you elaborate?

> I think it might be better to set the default values of fees to the Jormungandr defaults.

Definetly. Can you point me to the values please?

> The help is still sort of broken (some arguments not showing) and I think this is due to structure fields ordering again.

wtf...it actually works with `--help` 😞

eugene-babichenko:
> Maybe add flag to enable disable it?

Just remove IMO.

> Didn't understand this. Could you elaborate?

Everything under `jcli transactions` is dedicated to building transactions and do not contact HTTP endpoints.

> Definetly. Can you point me to the values please?

Uhm, well, I guess we don't have defaults for them, or I still did not wake up. Will need to ask.

jcli/src/jcli_lib/transaction/simplified.rs
```diff
@@ -0,0 +1,145 @@
...
+    pub change: Option<interfaces::Address>,
+
+    #[structopt(flatten)]
+    pub common: common::CommonTransaction,
+
+    #[structopt(flatten)]
+    pub fee: common::CommonFees,
+
+    #[structopt(flatten)]
+    rest_args: RestArgs,
```
eugene-babichenko, line 49: They need to be removed since we don't do transaction sending here anymore.

danielSanchezQ: It is still needed to query some account data.

eugene-babichenko: Oopsie, sorry.

danielSanchezQ:
Added request fees from rest api

testing/jormungandr-integration-tests/src/jcli/transaction/simplified.rs
```diff
@@ -0,0 +1,50 @@
...
+    let block0_hash = Hash::from_hex(config.genesis_block_hash()).unwrap();
+
+    jcli.transaction().make_transaction(
+        jormungandr.rest_uri(),
+        sender.address(),
+        100.into(),
+        block0_hash.to_string(),
+        sk_file_path,
+        staging_file,
+    );
```
eugene-babichenko, line 49: Shall we also make sure the transaction was successfully casted?

danielSanchezQ: It is internally asserted in the `make_transaction` method.

testing/jormungandr-integration-tests/src/common/jcli/command/transaction.rs
```diff
@@ -186,6 +186,30 @@ impl TransactionCommand {
...
     }
 
+    pub fn make_transaction(
+        mut self,
+        host: String,
+        sender: jormungandr_lib::interfaces::Address,
+        value: jormungandr_lib::interfaces::Value,
+        block0_hash: String,
+        secret: impl AsRef<Path>,
+        staging_file: impl AsRef<Path>,
```
eugene-babichenko, line 196: Uhm, but who is the receiver?

danielSanchezQ: The receiver is a testing account. Actually yesterday I was thinking that I may have missed making it optional. So if no address is provided a testing account is used.

testing/jormungandr-integration-tests/src/common/jcli/command/transaction.rs
```diff
@@ -204,9 +205,14 @@ impl TransactionCommand {
...
             .arg(host)
             .arg("--block0-hash")
-            .arg(block0_hash)
-            .arg(sender.to_string())
-            .arg(value.to_string());
+            .arg(block0_hash);
+
+        if let Some(receiver) = receiver {
+            self.command
+                .arg("--receiver-account")
```
eugene-babichenko, line 213: Maybe just `--receiver`?

eugene-babichenko:
1. This question `Continue? Yes[y] or No[n]?` is always returning `canceled by user` regardless of my answer.
2. I cannot post the resulting message with `jcli rest v0 message post`. Looks like we use inconsistent encodings.

danielSanchezQ:
> 1. This question `Continue? Yes[y] or No[n]?` is always returning `canceled by user` regardless of my answer.
> 2. I cannot post the resulting message with `jcli rest v0 message post`. Looks like we use inconsistent encodings.

Yep, sorry, I realize late. Was working on a fix atm.

eugene-babichenko (changes requested):
`jcli rest v0 message post` expects a hex-encoded fragment. Please add a full e2e test that would include sending so that we are sure everything actually works.

danielSanchezQ:
@eugene-babichenko I added some flag to directly post the fragment (since we already have the rest data). Also I added the test. If needed the hex encoded one it can always be done through the `to-message` command afterwards.

eugene-babichenko:
It worked perfectly, but needs some work on basic usability :)
I will walk through the code and leave my comments later.

jcli/src/jcli_lib/utils/io.rs
```diff
@@ -65,3 +65,12 @@ pub fn read_yaml<D: DeserializeOwned>(path: &Option<impl AsRef<Path>>) -> Result
     let yaml = serde_yaml::from_reader(reader)?;
     Ok(yaml)
 }
+
+pub fn ask_yes_or_no(with_output: bool) -> std::io::Result<bool> {
+    if with_output {
+        println!("Continue? Yes[y] or No[n]?");
```
eugene-babichenko, line 71:
```suggestion
        print!("Continue? Yes[y] or No[n]? ");
```

jcli/src/jcli_lib/transaction/simplified.rs
```diff
@@ -0,0 +1,218 @@
...
+
+    let settings = rest::v0::settings::request_settings(rest_args.clone())?;
+    let fee = common_fee_from_settings(&settings);
+
+    let transfer_value = value.saturating_add(transaction.fees(&settings.fees).into());
+
+    // ask for user confirmation after adding fees
+    if !force {
+        println!(
+            "Total value to transfer (including fees): {}\n",
```
eugene-babichenko, line 171:
```suggestion
            "Total value to transfer (including fees): {}",
```

eugene-babichenko, line 171:
```suggestion
            "Total value to transfer (including fees): {}",
```

jcli/src/jcli_lib/rest/v0/message/mod.rs
```diff
@@ -57,14 +57,19 @@ fn get_logs(args: RestArgs, output_format: OutputFormat) -> Result<(), Error> {
 fn post_message(args: RestArgs, file: Option<PathBuf>) -> Result<(), Error> {
     let msg_hex = io::read_line(&file)?;
     let msg_bin = hex::decode(&msg_hex)?;
-    let _fragment =
+    let fragment =
         Fragment::deserialize(msg_bin.as_slice()).map_err(Error::InputFragmentMalformed)?;
+    let fragment_id = post_fragment(args, fragment)?;
+    println!("{}", fragment_id);
```
eugene-babichenko, line 63: This needs to be replaced with a more informative message.

jcli/src/jcli_lib/transaction/simplified.rs
```diff
@@ -0,0 +1,218 @@
...
+            self.change,
+            self.force,
+        )?;
+
+        transaction.store(&self.common.staging_file)?;
+
+        if self.post {
+            let fragment = transaction.fragment()?;
+            let fragment_id = post_fragment(self.rest_args, fragment)?;
+            println!("{}", fragment_id);
```
eugene-babichenko, line 92: Same about adding more information here: right now we are just printing numbers and letters.

eugene-babichenko (commented):
We are outputting the raw fragment body somewhere regardless of settings provided by the user. This is not something desirable when, for example, we are using the `--post` flag.

On top of that, since this is a raw body there is no way it could be properly copied to a file. A good approach would be to either output this only to a file or output everything besides the fragment body to `stderr` to allow proper redirects. IMO the first option is more viable in this case (since this is an interactive utility) but I am open to a discussion on that.

jcli/src/jcli_lib/transaction/simplified.rs
danielSanchezQ: Yep, leftover from a previous message.

jcli/src/jcli_lib/rest/v0/message/mod.rs
danielSanchezQ: I kept it as it was. I understand that this may be used in order to pipe the output to something else. But I'm down to change it.

jcli/src/jcli_lib/transaction/simplified.rs
danielSanchezQ: Likewise here, but it is clearer here that we could output something better, since this is more a one-shot interactive thingy :)

eugene-babichenko (approved):

PR #3232, opened 2021-04-21
REST API: return elaborate reports about rejected fragments

eugene-babichenko:


jormungandr/src/fragment/pool.rs
```diff
@@ -26,6 +29,31 @@ pub enum Error {
     CannotPropagate(#[source] SendError),
 }
 
+#[derive(Debug, Serialize)]
+#[serde(rename_all = "snake_case")]
+pub enum FragmentRejectionReason {
+    FragmentInLog,
```
dkijania, line 35: maybe it should be called FragmentAlreadyInLog.

Zeegomo (commented):
Should we try to reuse/merge the fragment representation with the already existing `interfaces::FragmentStatus`?.
If that is cumbersome, or we don't want to do that for other reasons, I suggest to document what is the difference between these two, because I found it a little confusing.

jormungandr/src/fragment/pool.rs
```diff
@@ -47,56 +75,99 @@ impl Pools {
...
+                    id,
+                    reason: FragmentRejectionReason::FragmentInLog,
+                });
+                continue;
+            }
+
+            if !is_fragment_valid(&fragment) {
+                rejected.push(RejectedFragmentInfo {
+                    id,
+                    reason: FragmentRejectionReason::PreviousFragmentInvalid,
```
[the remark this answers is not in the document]
eugene-babichenko: It should be, thanks.

eugene-babichenko:
> Should we try to reuse/merge the fragment representation with the already existing `interfaces::FragmentStatus`?.

Well, the point was to make a clear separation of two types of errors:

- Errors during adding fragments to the pool.
- Errors during processing fragments, that were in the pool.

These two appear in very different contexts and I did not want to mix them together.

Zeegomo:
> These two appear in very different contexts and I did not want to mix them together.

That's totally fine, I would just suggest to put this in a comment so that future me won't spend 30 minutes wondering why we have two separate representations.

doc/api/v1.yaml
```diff
@@ -36,15 +36,55 @@ paths:
...
-                description: A list of fragment IDs (same order as in input)
-                type: array
-                items:
-                  type: string
-                  pattern: '[0-9a-f]+'
-              example:
-                [
-                  '68dcc12fe0dfe5e7b66ca6f8c959f9aa43b273e120a77fc3e4e2f04f1ecd7968',
-                ]
+                description: The information about whether a message was accepted or rejected
```
eugene-babichenko, line 39: These are big and should probably be reused...

eugene-babichenko:
There is the legacy test failing, which is fine, we made a breaking change to the REST API.

doc/api/v0.yaml
```diff
@@ -615,10 +615,55 @@ paths:
...
+                        id:
+                          description: The ID of a rejected fragment
+                          type: string
+                          format: '[0-9a-f]+'
+                        reason:
+                          oneOf:
+                            - description: This fragment was already processed by this node
+                              type: string
+                              enum:
+                                - FragmentAlreadyInLog
```
mzabaluev, line 645: I think a cleaner approach, that is less confusing to OpenAPI tools, would be to make `reason` a discriminator field, but then you can't probably make one of its variant values an object unto itself (it's more typical in JSON to grow extension fields alongside a particular value of the discriminator field; but it's probably not how serde-json would map a Rust enum).

Any examples of using `enum` have a unified value list under the field type. Also see the [JSON schema spec draft](https://tools.ietf.org/html/draft-bhutton-json-schema-validation-00#section-6.1.2).

doc/api/v1.yaml
mzabaluev: Dunno, we should probably migrate from v0 to v1-only API at some point. But maybe type definitions can be split into a separate file.

jormungandr/src/fragment/pool.rs
```diff
@@ -47,56 +51,99 @@ impl Pools {
...
+        }
+
+        if fail_fast {
+            for fragment in fragments {
+                rejected.push(RejectedFragmentInfo {
+                    id: fragment.id(),
+                    reason: FragmentRejectionReason::PreviousFragmentInvalid,
+                })
+            }
+        }
```
mzabaluev, line 105: Are you sure we want to do this as the general case?

Maybe a better approach for client-submitted fragments would be to only submit and log them if all of them validate (the function seems stateless between fragments), otherwise a HTTP failure code will signal atomic rejection of all of them?

mzabaluev: Thinking aloud: `/v1/fragments` might be used by a client to submit some fragments that make sense together (even if they should not be dependent on one another), so the possibility that only part of them may be accepted would bring complication into client-side error handling.

mzabaluev: OTOH, we do want the ability to spill fragments that don't fit into the mempool, so would the mempool fitting logic need to apply to the whole batch to make things consistent?

jormungandr/src/intercom.rs
```diff
@@ -500,7 +502,12 @@ pub fn stream_request<T, R>(
 #[allow(clippy::large_enum_variant)]
 #[derive(Debug)]
 pub enum TransactionMsg {
-    SendTransaction(FragmentOrigin, Vec<Fragment>),
+    SendTransactions {
+        origin: FragmentOrigin,
+        fragments: Vec<Fragment>,
+        fail_fast: bool,
```
mzabaluev, line 508: This currently seems redundant with `FragmentOrigin`, but it makes the code clearer.

mzabaluev (commented):
The code changes look good, but the OpenAPI for the enums is dodgy, and I have doubts about error handling. The rejection reason of "failed because the previous fragment in the batch was invalid" sounds a bit like fence-sitting on how do we really want to handle bad tx batches.

jormungandr/src/fragment/pool.rs
mzabaluev: Anyway, any of the fragments may end up not included into a confirmed block, so partial rejection is OK in principle. It's just that if a client sent something so wrong in one of the fragments, we should not assume any of their fragments make sense, so no need for such elaborate partial acceptance logic and weird error reasons like "it was the fault in another fragment that made this one fail".

doc/api/v0.yaml
eugene-babichenko: I do not like that hack as well and I think it is ugly. There is not much we can do about that, since we have multiple mempools for different leaders and we need to report the exact mempool for which the problem appeared.

Another thing we can do to (it would result into basically the same API): unify all errors except `PoolOverflow` under one enum and leave `PoolOverflow` as it is now.

eugene-babichenko: Also, JSON schema is pretty neat, but would require a lot of work from us :)

mzabaluev: > Also, JSON schema is pretty neat, but would require a lot of work from us :)

Well, if we want to have a neat OpenAPI spec, JSON schema is what it is based on.

mzabaluev: @eugene-babichenko So in the OpenAPI spec we stick all scalar variants into a single `enum`, define non-scalar variants as object types, and merge all this under some combination of `allOf`/`anyOf` for the containing object?

eugene-babichenko: Yes, pretty much that.

doc/api/v1.yaml
```diff
@@ -67,17 +67,14 @@ paths:
                           format: '[0-9a-f]+'
                         reason:
                           oneOf:
-                            - description: This fragment was already processed by this node
+                            - description: |
+                                * `FragmentAlreadyInLog` -- this fragment was already processed by this node.
+                                * `FragmentInvalid` -- this fragment failed validation.
+                                * `PreviousFragmentInvalid` -- one of the previous fragments was rejected and `fail_fast` is enabled.
```
mzabaluev, line 73: Much more concise, just one case of split personality now.
This is perhaps the best we can get from how serde and serde-yaml generate serialization of enums.

The JSON-friendly way would be:

```json
[
    {
        "id": "fafadeadbeef012345678",
        "reason": "FragmentAlreadyInLog"
    },
    {
        "id": "badc0ffee987654321",
        "reason": "PoolOverflow",
        "pool_number": 2
    }
]
```

BTW, don't we have blockchain-wise pool IDs?

mzabaluev:
I think we should remove `PreviousFragmentInvalid` and just fail the whole request with 400 if any of the fragments fails well-formedness and sanity checks.

eugene-babichenko:
> I think we should remove PreviousFragmentInvalid and just fail the whole request with 400 if any of the fragments fails well-formedness and sanity checks.

Does that imply that upon such failure none of submitted fragments will end up in the fragment pool?

doc/api/v1.yaml
eugene-babichenko: What do you mean exactly? This refers to mempool (which is basically the same as the leader ID). This is something specific to a particular node deployment.

eugene-babichenko:
Rebased on the current `master`. Now includes changes from #3248.

jormungandr-lib/src/interfaces/fragments_processing_summary.rs
```diff
@@ -0,0 +1,55 @@
+use chain_impl_mockchain::fragment::FragmentId;
+
+use serde::{Deserialize, Serialize};
+use serde_with::{serde_as, DisplayFromStr};
+
+/// This error is reserved for fragments that were rejected by the mempool at the time of sending
+/// them to mempool. If a fragment ended up being included to mempool, it will be listed in
+/// fragment logs and all further errors would be listed in fragment logs as well.
+#[derive(Debug, Serialize, Deserialize)]
+#[serde(tag = "type")]
```
mzabaluev, line 10: `tag` and `flatten`, should work with `Serialize`. We should have deserialization tests to verify it works on parsing.

The tag should be `reason` to match the API description, though?

mzabaluev (approved):
Good to go.

Issue #3229, opened 2021-04-19
Network outage and quarantine config


Zeegomo:
Suppose there are 2 nodes, `A`, `B` . For some reason, the link that connects `A` to the other nodes goes offline, so that `A` is isolated.
What happens is that `A` quarantines `B` , because from her perspective it is not responding, and `B` will do the same with `A`.

Now, since they will not contact each other due to quarantine, there's no way they will get to talk again on their own, and the only fix is to restart either `A` or `B`.

Some ways we deal with this is the `preferred_layer`, that prevents peers from being quarantined, or the quarantine whitelist itself.

Before the recent poldercast update, the node tried to connect to quarantined nodes as well (I'm not sure if that was a bug or not, the jormungandr doc says it should not, but there are some outdated info there), so it was more resilient to such outages even without custom layers or quarantine whitelist.

That changed with the new quarantine: a single failure will put the `A` in quarantine and prevent the `B` from connecting to it and viceversa, and if that happens both ways the nodes are isolated.

What we could do:
* Nothing, nodes should use `preferred_list` or `quarantine_whitelist` if they want such functionality
* Forgive a report for nodes we connected to in the past (this is actually the default behavior for poldercast, which I override)
* Introduce reasons for a report (that is just connection failure now, but may change in the future), and still send gossips (but not fragments or nodes) to nodes that are quarantined because of a connection failure. Upon reception of a gossip from such nodes, remove them from quarantine.

eugene-babichenko:
>  What happens is that A quarantines B , because from her perspective it is not responding, and B will do the same with A.

Shouldn't the quarantine be lifted after a timeout?

eugene-babichenko:
I think this might be resolved with a sort of circuit breaker pattern:

1. After a few failures to communicate with a node we put it into quarantine.
2. After the quarantine is lifted we return to step 1.

This should add some "grace period" during which nodes should be able to restore the connection.

Zeegomo:
> Shouldn't the quarantine be lifted after a timeout?

Yes, but the way it works is that the node will not be re-added to the view until we receive a new gossip about it, so that we won't re-add nodes that went offline "forever".

eugene-babichenko:
Then what I decscribed should work. At least because it implies attempts to reconnect. :)

Zeegomo:
[...]

PR #3233, opened 2021-04-23
Adjust quarantine tests

Zeegomo:
Quarantine / disruption tests were not working as expected (by that I mean that they were passing, but not testing what they were meant to test), because quarantine is now based on the node id and if a node is not provided with a node_key it will generate a random one each time, thus making a restart appear as another node joining the network.

There are now 2 failing scenario tests, `point_to_point_disruption_overlap` and `leader_leader_disruption_overlap`, where a root node without trusted peers is restarted,  for which a fix is being discussed here https://github.com/input-output-hk/jormungandr/issues/3229 .

### DIFF
--- a/testing/jormungandr-integration-tests/src/common/network/controller.rs
+++ b/testing/jormungandr-integration-tests/src/common/network/controller.rs
@@ -14,7 +14,9 @@ use jormungandr_testing_utils::{
 
 use assert_fs::fixture::FixtureError;
 use assert_fs::prelude::*;
+use assert_fs::NamedTempFile;
 use assert_fs::TempDir;
+use std::collections::HashMap;
 use std::path::PathBuf;
 use thiserror::Error;
 
@@ -39,6 +41,7 @@ pub struct Controller {
     working_directory: TempDir,
     block0_file: PathBuf,
     block0_hash: HeaderId,
+    node_key_files: HashMap<String, NamedTempFile>,
 }
 
 impl Controller {
@@ -52,11 +55,26 @@ impl Controller {
         let file = std::fs::File::create(&block0_file)?;
         block0.serialize(file)?;
 
+        let node_key_files = settings
+            .nodes
+            .keys()
+            .map(|alias| {
+                let key =
+                    jormungandr_lib::crypto::key::SigningKey::<chain_crypto::Ed25519>::generate(
+                        rand::thread_rng(),
+                    );
+                let file = NamedTempFile::new("node_key").unwrap();
+                std::fs::write(file.path(), key.to_bech32_str().as_bytes())?;
+                Ok((alias.to_string(), file))
+            })
+            .collect::<Result<HashMap<_, _>, ControllerError>>()?;
+
         Ok(Controller {
             settings,
             block0_file,
             block0_hash,
             working_directory,
+            node_key_files,
         })
     }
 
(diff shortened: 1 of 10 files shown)


eugene-babichenko (approved):

testing/jormungandr-integration-tests/src/common/network/controller.rs
```diff
@@ -39,6 +41,7 @@ pub struct Controller {
     working_directory: TempDir,
     block0_file: PathBuf,
     block0_hash: HeaderId,
+    node_key_files: HashMap<String, NamedTempFile>,
```
mzabaluev, line 44: Is there a special need for these files to exist outside the temp directory controlled by the `working_directory` member? If not, I think you could create them under the `working_directory` hierarchy and map the names as `PathBuf` values.

It will be simpler to keep all the files in stock if we decide we want to preserve the working directory on test failure. Now you'd also need to un-temp all the key files.

testing/jormungandr-integration-tests/src/networking/p2p.rs
```diff
@@ -278,7 +302,10 @@ pub fn node_trust_itself() {
 
     assert!(network_controller
         .expect_spawn_failed(
-            params(CLIENT).trusted_peers(vec![peer]),
+            network_controller
+                .spawn_params(CLIENT)
+                .unwrap()
+                .trusted_peers(vec![peer]),
             "unable to reach peer for initial bootstrap"
         )
```
mzabaluev, line 310: These nested calls to controller methods make me suspect they could be redesigned as a proper chain, like:

```rust
network_controller
    .spawn_params(CLIENT)
    .unwrap()
    .trusted_peers()
    .expect_failed("unable to reach peer for initial bootstrap")
```

testing/jormungandr-scenario-tests/src/scenario/controller.rs
```diff
@@ -166,6 +186,7 @@ impl Controller {
         working_directory: ChildPath,
         blockchain: Blockchain,
         topology: Topology,
+        node_key_files: HashMap<String, NamedTempFile>,
```
mzabaluev, line 189: Same issue as with the other controller. Maybe there should be better code reuse as well.

testing/jormungandr-scenario-tests/src/test/non_functional/disruption.rs
```diff
@@ -137,6 +137,9 @@ pub fn passive_leader_disruption_overlap(
     // 2. Only leader is up
     passive.shutdown()?;
 
+    // Wait a bit so that the leader can indeed notice that passive is offline
+    utils::wait(15);
```
mzabaluev, line 141: This is quite a bit of wait...

Zeegomo: I would rather wait a bit more and avoid spurious failures than do as it is now where the small wait is useless and it's not testing what it is supposed to test

testing/jormungandr-integration-tests/src/networking/p2p.rs
Zeegomo: This requires non trivial amount of work, might be worth investigating but I don't think it fits this PR

dkijania (approved):

testing/jormungandr-scenario-tests/src/test/non_functional/disruption.rs
eugene-babichenko: Agree with @Zeegomo here. We are not loosing to much time and `cargo test` runs everything in parallel anyway.

PR #3234, opened 2021-04-23
Fix docs workflow

Zeegomo:


eugene-babichenko (approved):
Why would they break it :(

PR #3236, opened 2021-04-26
Network stuck handler

Zeegomo:


jormungandr/src/topology/process.rs
```diff
@@ -3,9 +3,12 @@ use crate::intercom::{NetworkMsg, PropagateMsg, TopologyMsg};
 use crate::settings::start::network::Configuration;
 use crate::utils::async_msg::{MessageBox, MessageQueue};
 use futures::SinkExt;
-use tokio::time::Interval;
+use std::time::Duration;
+use tokio::time::{Instant, Interval};
 use tokio_stream::StreamExt;
 
+const STUCK_NETWORK_WARNING: Duration = Duration::from_secs(60 * 5); // 5 min
```
eugene-babichenko, line 10: Wondering if this value should be configurable.

eugene-babichenko (commented):
What is the default quarantine duration right now? This code would not make sense if `quaranatine_duration <= 5 mins`.

jormungandr/src/topology/process.rs
Zeegomo: Probably yes, It was just easier to set it up like this.

Zeegomo:
> What is the default quarantine duration right now? This code would not make sense if quaranatine_duration <= 5 mins

Default is 10min now, but I don't get why it would not make sense.
Removing a peer from quarantine is only triggered on receiving a gossip, which has not happened in a while if this special procedure fires, so if the node was ever in contact with some other peer that call to `lift_from_quarantine` will always return something.

eugene-babichenko:
> Removing a peer from quarantine is only triggered on receiving a gossip

Uhm. got a bit lost here, so we are trying to reconnect by ourselves, right?

Zeegomo:
> Uhm. got a bit lost here, so we are trying to reconnect by ourselves, right?

Yes

> Removing a peer from quarantine is only triggered on receiving a gossip

This is done because, given how poldercast works, we need to receive a new gossip about a peer before it being available for gossip / view again, so it makes no sense to lift the quarantine if we are not receiving a gossip anyway.
This attempt to recover goes a little out of the way of poldercast, in the sense that even if the peers manually lifted from quarantine do not appear in poldercast view, we feed them to the network to try to establish a connection. If that succeed, then we will receive gossips and they will be readded to the view, if that fails, we will quarantine them back.

PR #3244, opened 2021-04-29
fragments log: fix setting status of an unknown fragment

### LINKED ISSUE #3242: Forever pending message after Mempool overflow
**Describe the bug**
After we sends fragments count which exceeds Mempool settings all further transactions are pending forever.
Having Mempool setting like this:
```
Mempool {
            pool_max_entries: 2,
            log_max_entries: 1,
        },
```
when sending two transactions:

first tx id: `c437c7c8426e431de25a798adcaedfeb51a47cf30e509144724d96e6767d9e38`
second tx id: `484cd15f90e3538cb51c7e951574d69164712d733933785b674eca2ec2e4617b`

Then first tx is put in the block, while second it's pending forever. 

**Mandatory Information**
1. `jcli --full-version`  - newest master: `b38e067ea29267e6af65938c5603dbe05418e6f7`
2. `jormungandr --full-version` - newest master: `b38e067ea29267e6af65938c5603dbe05418e6f7`

**To Reproduce**
1. Go to pr: `https://github.com/input-output-hk/jormungandr/pull/3240`
2. Run tests:
```
    jormungandr::bft::mempool::v0::test_mempool_log_max_entries_equals_0
    jormungandr::bft::mempool::v0::test_mempool_log_max_entries_only_one_fragment
    jormungandr::bft::mempool::v1::test_mempool_log_max_entries_equals_0
    jormungandr::bft::mempool::v1::test_mempool_log_max_entries_only_one_fragment
```

**Expected behavior**

eugene-babichenko:
Fixes #3242

### DIFF
--- a/jormungandr/src/fragment/logs.rs
+++ b/jormungandr/src/fragment/logs.rs
@@ -57,15 +57,21 @@ impl Logs {
                 }
             }
             None => {
-                // while a log modification, if the log was not already present in the
-                // logs it means we received it from the a new block from the network.
-                // we can mark the status of the transaction so newly received transaction
-                // be stored.
-
-                self.entries.put(
-                    fragment_id,
-                    FragmentLog::new(fragment_id.clone().into_hash(), FragmentOrigin::Network),
-                );
+                // Possible reasons for entering this branch are:
+                //
+                // - Receiving a fragment with a network block.
+                // - Having a fragment evicted from the log due to overflow.
+                //
+                // For both scenarios the code defaults to FragmentOrigin::Network, since there are
+                // no means of knowing where the fragment came from.
+                //
+                // Also, in this scenario we accept any provided FragmentStatus, since we do not
+                // actually know what the previous status was, and thus cannot execute the correct
+                // state transition.
+                let mut entry =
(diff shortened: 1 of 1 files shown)


jormungandr/src/fragment/logs.rs
```diff
@@ -57,15 +57,21 @@ impl Logs {
...
-                // we can mark the status of the transaction so newly received transaction
-                // be stored.
-
-                self.entries.put(
-                    fragment_id,
-                    FragmentLog::new(fragment_id.clone().into_hash(), FragmentOrigin::Network),
-                );
+                // Possible reasons for entering this branch are:
+                //
+                // - Receiving a fragment from the network.
```
Zeegomo, line 62: I think a fragment received from the network would not be fed into this function at its first step, but rather inserted as pending before normally.
What could happen instead is that we try to remove all fragments in a block received from the network, but we never received some of the fragments inside it.

Zeegomo:
Actually, I'm not even sure why we have to add such fragments to the logs. What is the rationale here?

Zeegomo (approved):

mzabaluev (approved):

PR #3245, opened 2021-04-29
Network stuck handler

Zeegomo:
Replaces #3236, see it for older discussions.

### DIFF
--- a/jormungandr/src/settings/start/config.rs
+++ b/jormungandr/src/settings/start/config.rs
@@ -124,8 +124,7 @@ pub struct P2pConfig {
     pub gossip_interval: Option<Duration>,
 
     /// if no gossip has been received in the last interval, try to connect
-    /// to nodes that were put in quarantine and for which quarantine has
-    /// elapsed.
+    /// to nodes that were previously known to this node.
     ///
     /// The default value is 5 min.
     #[serde(default)]
--- a/jormungandr/src/settings/start/config.rs
+++ b/jormungandr/src/settings/start/config.rs
@@ -123,6 +123,14 @@ pub struct P2pConfig {
     #[serde(default)]
     pub gossip_interval: Option<Duration>,
 
+    /// if no gossip has been received in the last interval, try to connect
+    /// to nodes that were put in quarantine and for which quarantine has
+    /// elapsed.
+    ///
+    /// The default value is 5 min.
+    #[serde(default)]
+    pub network_stuck_check: Option<Duration>,
+
     /// The number of times to retry bootstrapping from trusted peers. The default
     /// value of None will result in the bootstrap process retrying indefinitely. A
     /// value of zero will skip bootstrap all together -- even if trusted peers are
@@ -165,6 +173,7 @@ impl Default for P2pConfig {
             layers: LayersConfig::default(),
             max_unreachable_nodes_to_connect_per_event: None,
             gossip_interval: None,
+            network_stuck_check: None,
             max_bootstrap_attempts: None,
         }
     }
--- a/jormungandr/src/topology/process.rs
+++ b/jormungandr/src/topology/process.rs
@@ -7,6 +7,8 @@ use tokio::time::{Instant, Interval};
 use tokio_stream::StreamExt;
 
 pub const DEFAULT_NETWORK_STUCK_INTERVAL: Duration = Duration::from_secs(60 * 5); // 5 min
+const QUARANTINE_CHECK: Duration = Duration::from_secs(60);
+const MAX_GOSSIP_SIZE: usize = 10;
 
 struct Process {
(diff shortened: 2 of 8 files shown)


jormungandr/src/topology/topology.rs
```diff
@@ -223,4 +210,22 @@ impl P2pTopology {
     pub fn update_gossip(&mut self) {
         self.topology.update_profile_subscriptions(&self.key);
     }
+    pub fn lift_nodes_from_quarantine(&mut self) -> Vec<Peer> {
```
eugene-babichenko, line 213: This is definitely missing a newline between functions.

eugene-babichenko (commented):
The idea is good. I am not sure about that, but it might be good to just make this a default behavior for quarantined nodes. E.g. if we do not receive anything from a quarantined node after the quarantine had expired, we should poke it and remove it from the list/put back to quarantine. This will improve the connectivity which is definitely good for the network.

Zeegomo:
> if we do not receive anything from a quarantined node after the quarantine had expired, we should poke it and remove it from the list/put back to quarantine

As per previous discussion, we can't actually remove a node from poldercast internal pools directly. If we contact a node right away after quarantine we would need some logic to avoid continuously recycling it back and forth from `pool` to `dirty` which would mean the node will probably not get evicted since it's periodically polled.

This PR is an easy workaround for that, we do have a little of recycling, but not very often and only when the node would not be able to join the topology anyway, without requiring extra logic for quarantine

Zeegomo:
Also, if a node is still connected to other nodes there's no need to actively ping ex-quarantined ones, because it will pick up their updates from incoming gossip from other nodes. That is only needed if there are no incoming connections , i.e. if we did not receive a gossip for a while.

eugene-babichenko:
> Also, if a node is still connected to other nodes there's no need to actively ping ex-quarantined ones, because it will pick up their updates from incoming gossip from other nodes

Makes sense to me.

jormungandr/src/settings/start/config.rs
```diff
@@ -123,6 +123,14 @@ pub struct P2pConfig {
     #[serde(default)]
     pub gossip_interval: Option<Duration>,
 
+    /// if no gossip has been received in the last interval, try to connect
+    /// to nodes that were put in quarantine and for which quarantine has
+    /// elapsed.
+    ///
+    /// The default value is 5 min.
+    #[serde(default)]
+    pub network_stuck_check: Option<Duration>,
```
eugene-babichenko, line 132: Document this setting here, please. https://github.com/input-output-hk/jormungandr/blob/master/doc/configuration/network.md

Actually, this may need a larder update since it is already outdated and I am also thinking is it is possible to generate a piece of that file based on the structure documentation. Probably, using something like [`cargo-sync-readme`](https://crates.io/crates/cargo-sync-readme).

Zeegomo: Good catch!
Part of the outdated published doc is because our doc actions has been failing for at least 2 months

jormungandr/src/topology/quarantine.rs
```diff
@@ -26,23 +26,30 @@ struct QuarantineRecord {
...
     quarantine_duration: StdDuration,
     quarantine_whitelist: HashSet<Address>,
-    quarantined_records: LruCache<PeerInfo, Instant>,
+    /// To avoid cycling down nodes back and and forth from quarantine (and as such prevent them
+    /// from being evicted from the lru cache), do not quarantine again nodes that were recently
+    /// lifted from quarantine.
+    ///
+    /// A peer is inserted in the grace list after being lifted from quarantine and is removed
+    /// from that list after we receive a new gossip about it.
+    quarantine_grace: LruCache<NodeId, ()>,
```
eugene-babichenko, line 35: We do not put any timeouts here and just allow these peers to be evicted from the cache when it is full, right?

jormungandr/src/topology/process.rs
```diff
@@ -87,17 +89,23 @@ impl Process {
...
                         last_update = Instant::now();
+                        tracing::warn!("p2p network have been too quiet for some time, will try to contanct known nodes");
+                        nodes_to_contact.extend(
+                            self.topology
+                                .list_available()
+                                .take(MAX_GOSSIP_SIZE.saturating_sub(nodes_to_contact.len()))
+                        );
                     }
+
+                    self.send_gossip_messages(nodes_to_contact);
```
eugene-babichenko, line 108: Now this makes more sense to me :)

jormungandr/src/topology/quarantine.rs
Zeegomo: yep, either when it's full or we receive a new gossip about them, which means they are online again and we should not make special rules for them.

eugene-babichenko (approved):

PR #3248, opened 2021-04-30
 implement persistent fragment log 

eugene-babichenko:
Implements the permanent log of incoming fragments as a JSON line-delimited stream.

`FragmentDef` is taken from #3232.


eugene-babichenko:
It's legacy tests again. Failing on master as well.

jormungandr/src/main.rs
```diff
@@ -592,6 +603,20 @@ fn initialize_node() -> Result<InitializedNode, start_up::Error> {
 
     let settings = raw_settings.try_into_settings()?;
 
+    let persistent_fragment_log = settings
```
Zeegomo, line 606: Is there a reason this is not done in the `start_services` function to avoid passing it through a bunch of other functions and sturcts?

jormungandr/src/fragment/pool.rs
```diff
@@ -68,6 +76,21 @@ impl Pools {
             .filter(|(_, exists_in_logs)| !exists_in_logs)
             .map(|(fragment, _)| fragment);
 
+        if let Some(mut persistent_log) = self.persistent_log.as_mut() {
```
Zeegomo, line 79: I actually thought we were going to put this at the API level to bypass any check that is done on the node

eugene-babichenko: We discussed that with @mzabaluev and decided to run it only for fragments that made it to the log. With #3232 we will have proper errors for what did not made its way to logs. And apart from pool overflow these errors are unlikely to happen. The overflow is also pretty unlikely in the production setup from what I have seen.

Zeegomo: This is also filtering for already existing fragments, but it may be intended and beneficial to reduce the amount of stuff we store

mzabaluev: For ballot fragments in particular, we can't tell if a repeated fragment is a replay attempt, or is it the combination of a user changing their mind on a ballot at least 2 times, and the client wrongly reusing the spending counter from the first fragment. I consider the likelihood of the latter low enough to matter, so we can ignore repeated fragments.

mzabaluev:
Good start. To make this work in the Catalyst deployment, we need to add rotation.
I'm not sure about putting out json logs, it's wasteful and the fragments are binary anyway, so we hardly gain anything from the text format. Will change it to bincode in a later PR.

dkijania:
Just a note, that there is a problem when `persistent_log_path` points to none existing folder. Node doesn't want to bootstrap

jormungandr/src/fragment/pool.rs
```diff
@@ -68,6 +87,24 @@ impl Pools {
...
+                    time: SecondsSinceUnixEpoch::now(),
+                    fragment,
+                };
+                // this must be sufficient: the PersistentFragmentLog format is using byte array
+                // for serialization so we do not expect any problems during deserialization
+                let codec = bincode::DefaultOptions::new().with_fixint_encoding();
+                if let Err(err) = codec.serialize_into(&mut persistent_log, &entry) {
+                    tracing::error!(err = %err, "failed to write persistent fragment log entry");
+                }
+                if let Err(err) = persistent_log.write_all("\n".as_bytes()) {
```
mzabaluev, line 102: This is now a Frankenstein format with binary payloads and newlines, how is it supposed to be parsed?
Change the entry encoding to use a length prefix.

jormungandr/src/fragment/process.rs
```diff
@@ -47,26 +48,46 @@ impl Process {
...
         persistent_log_dir: Option<P>,
     ) -> Result<(), Error> {
-        let persistent_log = match persistent_log_dir {
+        async fn hourly_wakeup(enabled: bool) {
+            if enabled {
+                let now = Utc::now();
+                let current_hour = now.duration_trunc(Duration::hours(1)).unwrap();
+                let next_hour = current_hour + Duration::hours(1);
+                let sleep_duration = (next_hour - now).to_std().unwrap();
+                tokio::time::sleep(sleep_duration).await
```
Zeegomo, line 65: Could this be rewritten as `tokio::time::interval_at(next_hour_rounded, Duration::from_secs(60 * 60))`

jormungandr/src/fragment/process.rs
```diff
@@ -41,82 +48,136 @@ impl Process {
...
-                                selection_alg,
-                                soft_deadline_future,
-                                hard_deadline_future,
-                            )
-                            .await;
-                        reply_handle.reply_ok(contents);
+                    _ = &mut wakeup => {
+                        let dir = persistent_log_dir.as_ref().unwrap();
+                        let file = open_log_file(dir.as_ref())?;
+                        pool.set_persistent_log(file);
```
mzabaluev, line 180: I forgot: `wakeup` needs to be re-assigned with a fresh future, to the effect of:
```suggestion
                        pool.set_persistent_log(file);
                        wakeup = hourly_wakeup(persistent_log_dir.is_some());
```

Might not work exactly like that due to the on-stack pin, so just convert it to a boxed future if necessary.

jormungandr/src/start_up/error.rs
```diff
@@ -18,6 +18,8 @@ pub enum ErrorKind {
 
 #[derive(Debug, Error)]
 pub enum Error {
+    #[error("Cannot open the persistent fragments log for writing")]
+    PersistentFragmentLog(#[source] io::Error),
```
mzabaluev, line 22: This is probably no longer needed with the recent changes.

jormungandr/src/fragment/process.rs
```diff
@@ -41,82 +48,136 @@ impl Process {
...
+                tokio::time::sleep(sleep_duration).await
+            } else {
+                future::pending().await
+            }
+        }
+
+        fn open_log_file(dir: &Path) -> Result<File, Error> {
+            let mut path: PathBuf = dir.into();
+            if !path.exists() {
+                std::fs::create_dir(dir).map_err(Error::PersistentLog)?;
```
mzabaluev, line 74: May want to use `create_dir_all`

mzabaluev (changes requested):

jormungandr/src/fragment/process.rs
mzabaluev: I was a bit concerned about clock or wakeup drift, as well as potential issues with machine suspend and similar when I wrote this. It's better to calculate the precise duration towards the next hour every time.

jormungandr/src/fragment/process.rs
mzabaluev: And maybe the function should return the current hour as a byproduct, to avoid any discrepancy in the file names, but this is finessing.

jormungandr/src/fragment/pool.rs
eugene-babichenko: > Change the entry encoding to use a length prefix.

Is there a particular reason to do that? We should be fine without that actually.

eugene-babichenko: > This is now a Frankenstein format with binary payloads and newlines

Removed newlines, thanks for noting.

mzabaluev: If you can decode a stream of bincode payloads, fine then.

mzabaluev: I'm not sure if this is achieved by `.allow_trailing_bytes` and deserializing entries one after another from the same stream.

mzabaluev (approved):

jormungandr/src/fragment/pool.rs
eugene-babichenko: You can use [`allow_trailing_bytes`](https://docs.rs/bincode/1.3.3/bincode/config/trait.Options.html#method.allow_trailing_bytes) and read the stream in a loop until reaching its end or meeting an invalid payload.

PR #3240, opened 2021-04-28
Mempool tests

dkijania:

PR #3250, opened 2021-05-05
make log_max_entries at least pool_max_entries * n_pools

Zeegomo:
Tests in https://github.com/input-output-hk/jormungandr/pull/3240 are fixed by this, but we need to remove some checks that are based on the previous limit on logs

### DIFF
--- a/jormungandr/src/fragment/process.rs
+++ b/jormungandr/src/fragment/process.rs
@@ -22,7 +22,7 @@ use tracing_futures::Instrument;
 
 pub struct Process {
     pool_max_entries: usize,
-    logs: Logs,
+    logs_max_entries: usize,
     network_msg_box: MessageBox<NetworkMsg>,
 }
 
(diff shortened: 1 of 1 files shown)


jormungandr/src/fragment/process.rs
```diff
@@ -84,6 +83,15 @@ impl Process {
                 .map_err(Error::PersistentLog)
         }
 
+        let min_logs_size = n_pools * self.pool_max_entries;
+        if self.logs_max_entries < min_logs_size {
+            tracing::warn!(
+                "Having 'log_max_entries' < 'pool_max_entries' * n_pools is not recommendend"
+            );
+            tracing::warn!("Overriding 'log_max_entries' to {}", min_logs_size);
+        }
```
eugene-babichenko, line 92: Should be a single log entry IMO.

Zeegomo: I was unsure about the best way to do this. I split that into two because otherwise it would be quite a long line, but I understand also why it should go together

mzabaluev (approved):

eugene-babichenko (approved):

PR #3251, opened 2021-05-05
cover some general things with editorconfig

eugene-babichenko:
We have a lot of things uncovered by linters like trailing newline that
are not covered by the set of linters we have for Rust and other file
formats. This is annoying and needs to be fixed.

On top of that, this is an automatic fix in the sense that
`.editorconfig` is automatically recognized by many IDEs.


.editorconfig
```diff
@@ -0,0 +1,7 @@
+root = true
+
+[*]
+end_of_line = lf
```
Zeegomo, line 4: Just wandering if this should not already be covered by git

eugene-babichenko: Still could be used as a guide by many editors/IDEs without additional setup required.

mzabaluev (approved):

PR #3253, opened 2021-05-05
install mdbook-linkcheck with cargo toolchain

filip-dulic-bloxico:
There is a permission error happening when installing this
crate from the release binary. Installing it via cargo seems to
solve the issue.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -14,21 +14,20 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
-      - name: Install linkcheck
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-        run: |
-          mkdir $GITHUB_WORKSPACE/mdbook-linkcheck
-          cd $GITHUB_WORKSPACE/mdbook-linkcheck
-          gh release --repo Michael-F-Bryan/mdbook-linkcheck download --pattern '*x86_64-unknown-linux-gnu.zip'
-          unzip *x86_64-unknown-linux-gnu.zip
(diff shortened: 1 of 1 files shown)


eugene-babichenko:
This is definitely a wrong solution, because the releases of `linkcheck` contain perfectly fine binaries and building binaries every single time or deal with caching is a pain.

Here is the log (from https://github.com/input-output-hk/jormungandr/runs/2506718320):

```
2021-05-05 06:06:26 [INFO] (mdbook::book): Running the linkcheck backend
2021-05-05 06:06:26 [INFO] (mdbook::renderer): Invoking the "linkcheck" renderer
Error: -05 06:06:26 [ERROR] (mdbook::utils): Error: Rendering failed
Error: -05 06:06:26 [ERROR] (mdbook::utils): 	Caused By: Unable to start the backend
Error: -05 06:06:26 [ERROR] (mdbook::utils): 	Caused By: Permission denied (os error 13)
Error: Process completed with exit code 101.
```

That implies something is wrong with the permissions.

I am not sure about what is going on in GH Actions, but I re-ran the steps locally and checked the result of the first step (`Install linkcheck`). Seems like the executable is simply not marked as such:

```
$ ls -l
total 12212
-rw-r--r--. 1 babichenko babichenko     1070 Feb 26 17:35 LICENSE
-rw-r--r--. 1 babichenko babichenko 12489480 Feb 26 17:35 mdbook-linkcheck
-rw-r--r--. 1 babichenko babichenko     5238 Feb 26 17:35 README.md
```

This case should be resolved by adding `chmod +x mdbook-linkcheck`.

eugene-babichenko:
Superseded by #3254

PR #3266, opened 2021-05-06
fix REST API fragment sending errors: err and ok were swapped

eugene-babichenko:
I'm stupid.


Zeegomo (approved):

mzabaluev (approved):
Nobody caught you (except tests that we need to fix), so it's on the team :)

PR #3267, opened 2021-05-06
Bft desync

Zeegomo:
This attempts to replicate possible issues in case of a fork due to unsynchronized clocks, where even transactions that are submitted to a leader that is part of the major component can be settled in fork.

 It requires the installation of faketime, if we want to include this in our tests we should take care of this dependency as well.

### DIFF
--- a/jormungandr/src/blockchain/process.rs
+++ b/jormungandr/src/blockchain/process.rs
@@ -380,6 +380,7 @@ pub async fn process_new_ref(
 
                 for block in stream.into_iter() {
                     let block = block.unwrap();
+                    tracing::debug!("merging {:?}", block);
                     let fragment_ids = block.fragments().map(|f| f.id()).collect();
                     if let Some(ref mut tx_msg_box) = tx_msg_box {
                         try_request_fragment_removal(tx_msg_box, fragment_ids, &block.header())
@@ -546,7 +547,7 @@ async fn process_block_announcement(
 async fn process_network_blocks(
     mut blockchain: Blockchain,
     blockchain_tip: Tip,
-    mut tx_msg_box: MessageBox<TransactionMsg>,
+    tx_msg_box: MessageBox<TransactionMsg>,
     network_msg_box: MessageBox<NetworkMsg>,
     mut explorer_msg_box: Option<MessageBox<ExplorerMsg>>,
     mut get_next_block_scheduler: GetNextBlockScheduler,
@@ -565,7 +566,6 @@ async fn process_network_blocks(
                 let res = process_network_block(
                     &blockchain,
                     block.clone(),
-                    &mut tx_msg_box,
                     explorer_msg_box.as_mut(),
                     &mut get_next_block_scheduler,
                 )
@@ -622,7 +622,6 @@ async fn process_network_blocks(
 async fn process_network_block(
     blockchain: &Blockchain,
     block: Block,
-    tx_msg_box: &mut MessageBox<TransactionMsg>,
     explorer_msg_box: Option<&mut MessageBox<ExplorerMsg>>,
     get_next_block_scheduler: &mut GetNextBlockScheduler,
 ) -> Result<Option<Arc<Ref>>, chain::Error> {
@@ -654,9 +653,7 @@ async fn process_network_block(
             Err(ErrorKind::MissingParentBlock(parent_hash).into())
         }
         PreCheckedHeader::HeaderWithCache { parent_ref, .. } => {
-            let r =
-                check_and_apply_block(blockchain, parent_ref, block, tx_msg_box, explorer_msg_box)
-                    .await;
+            let r = check_and_apply_block(blockchain, parent_ref, block, explorer_msg_box).await;
             r
         }
     }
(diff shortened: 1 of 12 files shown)


Zeegomo:
This is a first PoC for handling forks, that is, avoid losing transactions that settle in a fork (that could occur now even if they were originally submitted to a node in the main component) in the case where there are no network splits. 

The idea is to remove from the mempool only transactions that were included in a block appended from our current main chain. This way each fork will maintain a complete set of the submitted transaction (and a different view of the mempool).

In this PoC this is implemented by maintaining only one mempool, and removing from that fragments that were committed to a block between `common_ancestor(cur_tip, new_branch` and `new_branch`. A complete solution would require also to re-add  settled in `common_ancestor(cur_tip, new_branch` and `cur_tip` to the mempool.

eugene-babichenko (commented):
The overall idea is good in a sense that we ensure that we commit all fragment. This should solve some problems in case of BFT setups. So, a very good start!

We should also cover branch switching, but I think this should be outside of the scope of this PR.

Zeegomo:
If we take into account adversarial nodes as well, I think the protocol deriving from the current implementation does not satisfy the liveness property as defined in the paper

> *Liveness*. If a transaction tx is provided by the environment to all honest servers at a point of the execution when the latest slot among the honest servers is *sl*, then any server whose clock advances u slots after *sl* to a slot *sl′*, will have a ledger at a state q for which it holds q0 (\*)→q1 (tx)→q2 (\*)→q for some states q1,q2; note that (*)→includes only transactions produced by the environment Z up to slot sl′. Ouroboros-BFT p5

because a rogue server could include transaction `tx` in a block whose parent is the genesis block. Since that chain is of length 1 nobody is going to use that as its tip (assuming the protocol has been running for some time), so that transaction is not counted in the ledger state of honest servers.
However, when the rogue server broadcasts the new block, honest servers will happily remove the transaction `tx` from their mempool and set its state as `InABlock{ ... }`.

To make things a little bit worse, until the fragment log is evicted from the logs cache, the client cannot even manually add the transaction back to the mempool. If the client is using an account input, I think it will be forced to submit another transaction first (and pay its fees) just to advance the spending counter so that the fragment id will be different. 
This last part actually applies to other cases as well, e.g. when a user submits a voting ballot a little too soon.

Zeegomo:
Little update from the cardano haskell node.

In practice, their logic is pretty similar to this PoC, that is just to remove transactions that are not valid anymore with respect to the current ledger, no re-addition of transactions when we switch branches. This is what is implemented in this PR.

Not re-adding transactions that settled in a branch back to the mempool when we change branches supposes that the client is responsible for making sure that the transaction is actually confirmed in the node.
It is anyway unlikely that a fork (with the updated mempool removal logic) will make the chain lose transactions for the following reasons:
* All forks should process transactions in nearly the same order, so we can reasonably expect all transactions settled in a block in a smaller fork to be in a block in the bigger one as well.
* Even if they are processed in different orders, if at least one leader that received a transaction remains in that fork until that transaction is settled, then that fork will contain that transaction.

In addition, a client should check the outcome of a transaction anyway because it could be rejected for other reasons, so this should not add much more load on it.

PS: https://github.com/input-output-hk/jormungandr/pull/3267#issuecomment-839782997 this still contains a better explanation of this PoC

jormungandr/src/blockchain/process.rs
```diff
@@ -344,8 +355,39 @@ pub async fn process_new_ref(
...
+                    .storage()
+                    .find_common_ancestor(candidate_hash, tip_ref.hash())?;
+
+                let stream = blockchain
+                    .storage()
+                    .stream_from_to(common_ancestor.header.hash(), candidate_hash)?
+                    .collect::<Vec<_>>()
+                    .await;
+
+                for block in stream.into_iter() {
```
mzabaluev, line 381: This is async code, could have read directly from the stream?

mzabaluev (commented):
> This is sufficient under the assumption that at least one leader that received a transaction remains in that fork until that transaction is settled.

I favor this lightweight approach after reading up on details.

> * InABlock(with depth < CONFIRMED_DEPTH) -> Pending

If a client used the transition to `InABlock` to submit another transaction dependent on the one included, this could mean the new transaction no longer has a settled basis. However, this is still an improvement on the current behaviour, where the new transaction would get rejected. In this changed implementation, the first transaction may still be applied by the nodes that have received it, and receiving the second transaction out of order is very unlikely.

As I understand it, this does not resolve the problems with net splits, where fragment propagation is disrupted the same way as propagation of blocks.

As this fixes a serious problem with the protocol, please rebase the branch and undraft it, let's go ahead with the merging.

Zeegomo:
> As I understand it, this does not resolve the problems with net splits, where fragment propagation is disrupted the same way as propagation of blocks.

Yes, I don't think we can do anything there, if the p2p network is partitioned we cannot guarantee a proper execution of the protocol.

> rebase the branch and undraft it, let's go ahead with the merging.

Should parts of this PR (like `find_common_ancestor` as Eugene suggested) go in chain-libs?

mzabaluev:
Superseded by #3370 except the faketime test changes. These can be lifted in some other PR if needed.

PR #3268, opened 2021-05-06
Ignore test_legacy_node_all_fragments

mzabaluev:
The failing test creates a "broken window" in our CI,
and we don't care about maintaining backward compatibility that much at this point.


eugene-babichenko:
I do not think we should re-enable it at all. We only care about it when we do not want to make breaking changes, which is not always the case.

PR #3269, opened 2021-05-06
PersistentFragmentLogs Iterator

danielSanchezQ:
Added iterators and utils for loading PersistentFragmentLogs directly from folder and files.

### DIFF
--- a/jormungandr-lib/src/interfaces/fragment_log_persistent.rs
+++ b/jormungandr-lib/src/interfaces/fragment_log_persistent.rs
@@ -1,6 +1,6 @@
 use std::fs;
 use std::io;
-use std::io::{BufRead, BufReader};
+use std::io::BufReader;
 use std::path::{Path, PathBuf};
 
 use crate::interfaces::FragmentDef;
@@ -41,10 +41,16 @@ pub struct FileFragmentsIterator {
 }
 
 impl FileFragments {
-    pub fn from_path(file_path: PathBuf) -> std::io::Result<Self> {
-        let mut reader = BufReader::new(fs::File::open(file_path.clone())?);
-        reader.fill_buf()?;
-        Ok(Self { reader, file_path })
+    pub fn from_path(
+        file_path: PathBuf,
+    ) -> std::io::Result<Box<dyn Iterator<Item = Result<PersistentFragmentLog, DeserializeError>>>>
+    {
+        let metadata = fs::metadata(file_path.clone())?;
+        if metadata.len() == 0 {
+            return Ok(Box::new(vec![].into_iter()));
+        }
+        let reader = BufReader::new(fs::File::open(file_path.clone())?);
+        Ok(Box::new(Self { reader, file_path }.into_iter()))
     }
 }
 
@@ -66,8 +72,9 @@ impl Iterator for FileFragmentsIterator {
 
     fn next(&mut self) -> Option<Self::Item> {
         // EOF reached when buffer is empty.
-        // Then we stop the iterator. Buffer if prefilled during initialization of FileFragments
-        if self.reader.buffer().is_empty() {
+        // Then we stop the iterator. File is guaranteed to be non-empty by construction, the check
+        // cannot be done till buffer is filled, hence first time should be able to read at least something.
+        if self.reader.buffer().is_empty() && self.counter != 0 {
             return None;
         }
         let codec = bincode::DefaultOptions::new()
--- a/jormungandr-lib/src/interfaces/fragment_log_persistent.rs
+++ b/jormungandr-lib/src/interfaces/fragment_log_persistent.rs
@@ -12,13 +12,11 @@ use bincode::Options;
 use serde::{Deserialize, Serialize};
 
 #[derive(thiserror::Error, Debug)]
-pub enum Error {
-    #[error("Couldn't deserialize entry {entry} in {file} due to: {cause}")]
-    DeserializeError {
-        file: String,
-        entry: usize,
-        cause: bincode::Error,
-    },
+#[error("Couldn't deserialize entry {entry} in {file} due to: {cause}")]
+pub struct DeserializeError {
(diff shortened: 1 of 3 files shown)


mzabaluev:
Is it a duplicate with #3270?

danielSanchezQ:
> Is it a duplicate with #3270?

It is @mzabaluev .  If the dry-run branch is gonna be merged to master later we can close this one.

dkijania (approved):

danielSanchezQ:
BTW, should I suppress the unused warnings?

mzabaluev:
I see, thanks for submitting the backport as well.

jormungandr-lib/src/interfaces/fragment_log_persistent.rs
```diff
@@ -14,3 +30,89 @@ pub struct PersistentFragmentLog {
...
+        // Then we stop the iterator.
+        if self.reader.buffer().is_empty() && self.counter != 0 {
+            return None;
+        }
+        let codec = bincode::DefaultOptions::new()
+            .with_fixint_encoding()
+            .allow_trailing_bytes();
+        let current = self.counter;
+        self.counter += 1;
+        Some(
```
eugene-babichenko, line 81: nit:

```
let result = ...;
Some(result)
```

would just look better

jormungandr-lib/src/interfaces/fragment_log_persistent.rs
```diff
@@ -1,10 +1,26 @@
...
 
 use chain_impl_mockchain::fragment::Fragment;
 
+use bincode::Options;
 use serde::{Deserialize, Serialize};
 
+#[derive(thiserror::Error, Debug)]
+pub enum Error {
+    #[error("Couldn't deserialize entry {entry} in {file} due to: {cause}")]
+    DeserializeError {
```
eugene-babichenko, line 17: Could be just `struct DeserializeError` if we do not plan to expand that.

jormungandr-lib/src/interfaces/fragment_log_persistent.rs
```diff
@@ -14,3 +30,89 @@ pub struct PersistentFragmentLog {
...
+    }
+}
+
+impl Iterator for FileFragmentsIterator {
+    type Item = Result<PersistentFragmentLog, Error>;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        // EOF reached when buffer is empty after reading successfully at least one.
+        // Then we stop the iterator.
+        if self.reader.buffer().is_empty() && self.counter != 0 {
```
eugene-babichenko, line 73: Wouldn't this break for an empty file?

eugene-babichenko (changes requested):
You need to either make `interfaces::fragment_log_persistent` a public module or re-export all of its contents on `interfaces` level. The latter is what we do now for all structs under `interfaces`.

jormungandr-lib/src/interfaces/fragment_log_persistent.rs
danielSanchezQ: Not really break per se. But It would return an item with an Some(Err) because it couldn't deserialize.

eugene-babichenko: We can have empty files, so better to handle that case.

danielSanchezQ: I prefilled the buffer, so it is guaranteed to have something if files wasn't empty. If it is emtpy it will imideatly return a `None`, so it will be an empty iterator 😄

jormungandr-lib/src/interfaces/fragment_log_persistent.rs
```diff
@@ -14,3 +28,88 @@ pub struct PersistentFragmentLog {
...
+pub struct FileFragmentsIterator {
+    reader: BufReader<fs::File>,
+    file_path: PathBuf,
+    counter: usize,
+}
+
+impl FileFragments {
+    pub fn from_path(file_path: PathBuf) -> std::io::Result<Self> {
+        let mut reader = BufReader::new(fs::File::open(file_path.clone())?);
+        reader.fill_buf()?;
```
eugene-babichenko, line 46: I would rather obtain the file metadata, check its length and return an empty iterator. But that works as a workaround too.

jormungandr-lib/src/interfaces/fragment_log_persistent.rs
danielSanchezQ: I had it like that in a previous version though.

jormungandr-lib/src/interfaces/mod.rs
```diff
@@ -44,7 +44,10 @@ pub use self::committee::CommitteeIdDef;
 pub use self::config::*;
 pub use self::fragment::FragmentDef;
 pub use self::fragment_log::{FragmentLog, FragmentOrigin, FragmentStatus};
-pub use self::fragment_log_persistent::PersistentFragmentLog;
+pub use self::fragment_log_persistent::{
+    get_fragments_log_files_path, load_fragments_from_folder_path, read_entries_from_files_path,
```
eugene-babichenko, line 48: I think we need more descriptive names here, they don't say anything about fragments logs actually.

danielSanchezQ: Makes sense, also there is one unneeded export too.

jormungandr-lib/src/interfaces/fragment_log_persistent.rs
danielSanchezQ: Actually, it may be worth using the metadata. We would avoid filling the buffer till it is really needed. Will do.

eugene-babichenko (approved):

eugene-babichenko:
Yes, it is legacy tests error we plan to disable it.

PR #3277, opened 2021-05-07
check update for audit-check

eugene-babichenko:


mzabaluev:
#3284 should fix the cause of audit failures.

I'm not sure we should continue using this action, though: last release/commit was a year ago, and https://github.com/actions-rs/audit-check/issues/163 has stayed unfixed for months despite obvious security implications.

mzabaluev:
Filed https://github.com/actions-rs/meta/issues/35

mzabaluev (approved):
Let's merge this, it's unlikely the upstream will get any important updates soon.

PR #3297, opened 2021-05-18
Reorder fragments submitted with spending counters that are too high

eugene-babichenko:
Depends on https://github.com/input-output-hk/chain-libs/pull/565.

Given the explicit spending counter we can reorder fragments on the per-account basis using priority queues. This introduces breaking changes to `jcli` as the spending counter is no longer needed to build a witness, but is required for account inputs.

**TBD**:

- [ ] `insert_all`:
  - [ ] Fragments from priority queues should be included in the overall count.
  - [ ] The pool should reject fragments that are already in one of priority queues.
- [ ] Implement removal from priority queues upon calling `remove_all`.
- [ ] Evict fragments that sit in their priority queues for too long.
- [ ] Restore legacy UTXO testing API.

**Side notes**

Since witnesses no longer depend on the spending counter (e.g. ledger state) we may validate them at the time transactions come to the pool. Early validation (or removing it from here entirely) in `chain-libs` is also an option.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -435,7 +435,7 @@ dependencies = [
 [[package]]
 name = "cardano-legacy-address"
 version = "0.1.1"
-source = "git+https://github.com/input-output-hk/chain-libs.git?branch=master#2650f1f27ec8c72657d87a978b22e8e5e514899c"
+source = "git+https://github.com/input-output-hk/chain-libs.git?branch=explicit-spending-counter#8637bd0538f3f5a08e96c0fa0d9a382da288d6aa"
 dependencies = [
  "cbor_event",
  "chain-ser",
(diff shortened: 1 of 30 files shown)


Zeegomo:
> Since witnesses no longer depend on the spending counter

I think they should, but you don't have to know the current ledger state to validate them since the spending counter is clearly stated in the transaction.

eugene-babichenko:
> I think they should

Not really. Spending counter is used for producing witness data anyway since it is included into the input body.

Zeegomo:
> Spending counter is used for producing witness data

So then the witness depends on the spending counter used in the transaction?

eugene-babichenko:
> > Spending counter is used for producing witness data
> 
> So then the witness depends on the spending counter used in the transaction?

yes

Zeegomo:
Should we close this since https://github.com/input-output-hk/chain-libs/pull/565 was discarded?

eugene-babichenko:
See https://github.com/input-output-hk/chain-libs/pull/565#issuecomment-857515061

PR #3301, opened 2021-05-19
Backport fragment pool error to fund4

eugene-babichenko:


doc/api/v1.yaml
```diff
@@ -46,8 +46,19 @@ paths:
...
-
+          description: An error occurred when processing this request
+          content:
+            application/json:
+              schema:
+                description: A list of fragment IDs (same order as in input)
+                type: array
+                items:
+                  type: string
+                  pattern: '[0-9a-f]+'
```
mzabaluev, line 57: Did not discuss this in the master PR: If we fail to parse any fragments, there won't be anything to put in this array, right?

jormungandr-lib/src/interfaces/fragments_processing_summary.rs
```diff
@@ -0,0 +1,53 @@
...
+/// This error is reserved for fragments that were rejected by the mempool at the time of sending
+/// them to mempool. If a fragment ended up being included to mempool, it will be listed in
+/// fragment logs and all further errors would be listed in fragment logs as well.
+#[derive(Debug, Serialize, Deserialize)]
+pub enum FragmentRejectionReason {
+    FragmentAlreadyInLog,
+    FragmentInvalid,
+    PreviousFragmentInvalid,
+    PoolOverflow { pool_number: usize },
+}
```
mzabaluev, line 15: This looks like backporting too much, the new types should not be used in the API responses...

jormungandr/src/rest/v1/logic.rs
```diff
@@ -91,11 +93,20 @@ pub async fn post_fragments(
...
+        fragments,
+        fail_fast: true,
+        reply_handle,
+    };
+    msgbox.try_send(msg)?;
+    let reply = reply_future.await?;
+    if reply.is_error() {
+        Err(Error::Fragments(fragment_ids))
+    } else {
+        Ok(fragment_ids)
```
mzabaluev, line 108: It looks like it does not change the response in `Ok` case...

jormungandr/src/rest/v1/logic.rs
```diff
@@ -91,11 +93,20 @@ pub async fn post_fragments(
...
+    let msg = TransactionMsg::SendTransactions {
+        origin: FragmentOrigin::Rest,
+        fragments,
+        fail_fast: true,
+        reply_handle,
+    };
+    msgbox.try_send(msg)?;
+    let reply = reply_future.await?;
+    if reply.is_error() {
+        Err(Error::Fragments(fragment_ids))
```
mzabaluev, line 106: ... all this added summary serialization gets dropped at this point, right?

I think it's OK for a quick backport, though.

mzabaluev (approved):

mzabaluev:
Ignore the audit failure, this has been fixed in master in #3284.

The clippy lint looks like an easy thing to fix, though.

jormungandr/src/rest/v1/logic.rs
eugene-babichenko: This is not going to be used during fund4 anyways I think.

jormungandr/src/rest/v1/logic.rs
eugene-babichenko: See https://github.com/input-output-hk/jormungandr/pull/3301#discussion_r635829394

doc/api/v1.yaml
eugene-babichenko: Yes. The master PR has a separate array for rejects though.

jormungandr/src/rest/v1/logic.rs
eugene-babichenko: Hm? It should not.

jormungandr/src/fragment/pool.rs
```diff
@@ -98,11 +98,15 @@ impl Pools {
         for fragment in fragments.by_ref() {
             let id = fragment.id();
 
+            let _enter =
+                tracing::span!(tracing::Level::TRACE, "pool_incoming_fragment", fragment_id=?id);
```
Zeegomo, line 102: I think the `enter` guard is returned only if you call `span!(...).enter()`, otherwise the span is not active

mzabaluev: Could use `trace_span!` macro for brevity.

doc/api/v1.yaml
mzabaluev: Yeah, but in case of parsing error there won't be an ID even for rejects.

jormungandr/src/fragment/pool.rs
danielSanchezQ: uhm, take care because iirc enter do not work properly with async. I'll come back with some info.

danielSanchezQ: ﻿﻿>  ⚠ ️Warning
> Warning: In asynchronous code that uses async/await syntax, Span::enter may produce incorrect traces if  the returned drop guard is held across an await point. See the method documentation for details.

danielSanchezQ: Although it seems Ok to use here. But just to be sure.

PR #3302, opened 2021-05-19
fix ordering of persistent fragments

Zeegomo:


testing/jormungandr-integration-tests/src/jormungandr/mempool/fragments_dump.rs
```diff
@@ -157,6 +157,7 @@ pub fn non_existing_folder() {
 }
 
 #[test]
+#[ignore]
```
eugene-babichenko, line 160: Hm, this one works good for me.

testing/jormungandr-testing-utils/src/testing/fragments/export.rs
```diff
@@ -35,22 +35,26 @@ impl FragmentExporter {
     pub fn read(&self) -> Result<Vec<Fragment>, FragmentExporterError> {
         self.read_as_bytes()?
             .iter()
+            .rev()
             .map(|bytes| Ok(Fragment::deserialize(bytes.as_ref()).unwrap()))
             .collect()
```
eugene-babichenko, line 40: Is it possible to just return an iterator here? Or would it be too troublesome to change the calling code?

testing/jormungandr-integration-tests/src/jormungandr/mempool/fragments_dump.rs
eugene-babichenko: Nevermind, it hands forever.

testing/jormungandr-testing-utils/src/testing/fragments/export.rs
Zeegomo: Oops, that `rev` is  a leftover actually

eugene-babichenko:
I am wondering if the one test that is ignored is universally valid? While it might a valid test on Windows, it is not for Linux. Shall we conditionally enable it depending on the OS?

Zeegomo:
I think if we try hard enough we can break other OSs as well, like what happens if we use `/dev/null/foo`? I'm pretty sure you cannot create a directory there on *nix, and on window the `/` should be invalid I guess

eugene-babichenko:
Such paths would not be invalid per se. There would be just a permission error or something like that. The error would noy be related to the path string itself, but to the filesystem structure.

Zeegomo:
I think the test is supposed to check if the node stops with an error if logs cannot be persisted. Whether it is because there are invalid characters in the path or it is not a directory (like `/dev/null` on *nix) it should not matter much

eugene-babichenko:
Hm, seems you were right. Now it is not passing on Windows though. I think we really need to come up with separate paths for these platforms.

Zeegomo:
I think this is failing on Windows for unrelated reasons. The release build on Windows passed just fine, and in the debug build the `invalid_folder` test completed successfully

eugene-babichenko:
> The release build on Windows passed just fine

Ah yes, I overlooked that. I will re-run that job just to be completely sure. CI failure is also unrelated and I think it was already fixed on `master`.

eugene-babichenko (approved):
LGTM

eugene-babichenko:
NB: need to port that to `master`.

Zeegomo:
> NB: need to port that to master.

Yep, but I think the dump tests need to be backported as well

eugene-babichenko:
Yep, please create a PR with all required backports.

PR #3310, opened 2021-05-26
Stop using `error-chain` entirely, drop `slog` dependency

eugene-babichenko:


jormungandr/src/blockchain/process.rs
```diff
@@ -391,7 +389,7 @@ async fn process_and_propagate_new_ref(
     network_msg_box
         .send(NetworkMsg::Propagate(PropagateMsg::Block(header)))
         .await
-        .map_err(|_| "Cannot propagate block to network".into())
+        .map_err(Into::into)
```
Zeegomo, line 392: Nit: I usually see `Ok(...?)` instead of using `Into` for errors, but that's a stylistic choice

Zeegomo (approved):
Yes!

jormungandr/src/blockchain/chain.rs
```diff
@@ -63,61 +63,55 @@ use chain_time::TimeFrame;
...
-            display("Tag '{}' not found in the storage", tag),
-        }
+    #[error("failed to send message to the fragments process")]
+    TransactionSendError(#[from] crate::async_msg::TrySendError<crate::intercom::TransactionMsg>),
 
-        BlockHeaderVerificationFailed (reason: String) {
-            description("Block header verification failed"),
-            display("The block header verification failed: {}", reason),
-        }
+    #[error("Error while creating the initial ledger out of the block0")]
```
mzabaluev, line 86: Nit: the prevailing convention (not everywhere in our code, though) is to start error messages with lower case.

mzabaluev (commented):
Looks good, except that we should take care of consistent error message formatting.

mzabaluev:
The failing tests seem to search error messages that no longer match.

Issue #3227, opened 2021-04-19
Move explorer to it's own binary crate


enzoc4:
partially replaces #2759 (bundling two things in one issue was a bad idea, existing discussion was on the other topic anyway).

# About

Make the explorer a separate service instead of a mode of operation for the node (as it works right now per #694).

The idea is for this service to not be a participant on the p2p layer, but to hook to the node and receive messages in the same way that we do right now internally through the `intercom`. 
New blocks and tip changes are enough for the most part. Bootstrapping shouldn't differ much from the way the node does it, I think, except (probably) that we only have one node to pull blocks from.

## Requirements

- [ ] Create a new `explorer` binary crate in this workspace (as in #2467, but it's out of date and built on different assumptions, and just a PoC anyway). 
We may want to move some things to `jormungandr_lib` to keep some things consistent (logging setup, some settings, maybe some networking stuff or service stuff), but it's not needed from the beginning.
- [ ] Connect this new service to the explorer through grpc. This may deserve a separate issue to discuss it. I don't have much knowledge about the networking layer myself. 
    As far as I can tell, the apis from the `BlockService` should be enough, but the explorer wouldn't be a full client, and I'm not sure how it will interact with the other stuff (will it get quarantined, for example?). May require a particular `ExplorerService` for the node?
- [ ] A lot of tests use the explorer, so that would need a rework somehow.

## Other considerations

Per design, most of the explorer works independently from the `Blockchain` (basically the ledger), so migration shouldn't be a problem with just sending blocks (and tips). 

There are currently two things that need access to the blockchain (and I didn't add them myself, so I'm not sure what to do with them yet). 
One of those is the `Block::treasury` query, which I think is used in tests at least.
The other one is in the indexing part of the `VoteTally` certificate. I think this can be ignored for now, but we'll need to think if it needs something that should be provided through an api (I think it needs the tallies, basically), of if it is something that the explorer can keep up-to-date by itself.

eugene-babichenko:
> We may want to move some things to jormungandr_lib to keep some things consistent (logging setup, some settings, maybe some networking stuff or service stuff), but it's not needed from the beginning.

I am not sure about that part. `jormungandr_lib` has many clients and not all of them would need logging and networking things. Although we can just make these things feature-gated.

Just to make sure: you intend to use gRPC for explorer-to-node communications, right?

enzoc4:
[...]

PR #3322, opened 2021-05-27
tests: use rest api for wait for date

enzoc4:
some tests need to wait for a particular epoch or date, and currently they wait on a loop querying the explorer each time until they reach that point.

there is nothing wrong with using the explorer for this, but it's a bit overkill to enable it just for that, so this makes the functions more applicable in general, and last date is easily available in rest.

using rest also makes it simpler to extract the explorer (see #3227). Although tests could launch an explorer client if they really need it, I think it's better to minimize the impact.

(actually, using the explorer for this could make sense if we use the tip subscription to remove polling, but I'm not sure it's worth it)

~TODO: I haven't thoroughly check if any of the tests don't even need to enable the explorer with this change, but it requires some case by case analysis since you can't tell easily by an `unused variable` or something like that.~


dkijania:
Right! I must have missed that this information is in stats endpoint. Makes perfect sense to do this change

dkijania (approved):

eugene-babichenko (commented):
> actually, using the explorer for this could make sense if we use the tip subscription to remove polling, but I'm not sure it's worth it

I think we should just stop using explorer in tests entirely for the sake of achieving #3227.

With that in mind, what if we try to build `jormungandr` without the explorer and see what happens to tests? E.g. remove the `mod explorer` line from `main.rs` and the endpoint from the API. Or just disable it everywhere, IIRC it is needed to be explicitly enabled through settings.

PR #3323, opened 2021-05-27
Lint tests on ci

enzoc4:
I don't see any particular reason to not lint the tests (and I don't think we are doing it), so this adds that.

I also ran `clippy fix --tests`. There is only one thing I had to change manually and that is in a separate commit.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -243,4 +243,4 @@ jobs:
         uses: actions-rs/cargo@v1
         with:
           command: clippy
-          args: -- -D warnings
+          args: --tests -- -D warnings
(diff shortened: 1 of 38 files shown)


.github/workflows/test.yml
```diff
@@ -243,4 +243,4 @@ jobs:
         uses: actions-rs/cargo@v1
         with:
           command: clippy
-          args: -- -D warnings
+          args: --tests -- -D warnings
```
eugene-babichenko, line 246: Shall we just do `--all-targets --all-features`?

enzoc4: Hmmm, (to my surprise) it is actually different, as we have some tests that are behind features and those don't get processed with `--tests` only. There is one test that doesn't even compile if you use `--all-features`.

eugene-babichenko (approved):

PR #3329, opened 2021-05-28
Improve rest utils

Zeegomo:
Reuse the same client for successive requests instead of building a new one each time, which is inefficient (relevant because this is used in load tools).
One thing that I changed is that there was a config option to use https on post requests, but I changed that to apply for all requests types (I'm not sure why that was the case, but it seems better this way imo)

### DIFF
--- a/testing/jormungandr-testing-utils/src/testing/node/rest/raw.rs
+++ b/testing/jormungandr-testing-utils/src/testing/node/rest/raw.rs
@@ -8,7 +8,7 @@ use chain_impl_mockchain::fragment::Fragment;
 use jormungandr_lib::interfaces::FragmentsBatch;
 use jortestkit::process::Wait;
 use reqwest::{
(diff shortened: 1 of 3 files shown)


testing/jormungandr-testing-utils/src/testing/node/rest/raw.rs
```diff
@@ -56,42 +84,18 @@ impl RawRest {
...
-                    .add_root_certificate(cert.clone())
-                    .build()
-                    .unwrap();
-                client.get(&request).send()
-            }
-        }
+        self.client.get(request).send()
     }
 
-    fn path(&self, path: &str) -> String {
```
mzabaluev, line 80: While you are at it, the order of `api_version` and `path` is more mnemonic.

mzabaluev (approved):

PR #3349, opened 2021-06-02
rest: log all requests

eugene-babichenko:
This produces a log that looks like:

```
Jun 02 10:53:48.423  INFO request{method=GET path=/api/v0/fragment/logs version=HTTP/1.1 remote.addr=127.0.0.1:60560}: warp::filters::trace: processing request
...
Jun 02 10:53:48.424  INFO request{method=GET path=/api/v0/fragment/logs version=HTTP/1.1 remote.addr=127.0.0.1:60560}: warp::filters::trace: finished processing with success status=200
```

Note that is automatically producing a tracing span.

### DIFF
--- a/jormungandr/src/rest/mod.rs
+++ b/jormungandr/src/rest/mod.rs
@@ -32,7 +32,16 @@ pub async fn start_rest_server(config: Rest, explorer_enabled: bool, context: Co
 
     let api = warp::path!("api" / ..)
         .and(v0::filter(context.clone()).or(v1::filter(context.clone())))
-        .with(warp::filters::trace::request());
+        .with(warp::filters::trace::trace(|info| {
+            tracing::span!(
+                tracing::Level::DEBUG,
+                "rest_api_request",
+                method = %info.method(),
+                path = %info.path(),
+                version = ?info.version(),
+                remote_addr = ?info.remote_addr(),
+            )
+        }));
     if explorer_enabled {
         let explorer = explorer::filter(context);
         setup_cors(api.or(explorer), config, stopper_rx).await;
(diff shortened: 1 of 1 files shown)


jormungandr/src/rest/mod.rs
```diff
@@ -30,8 +30,9 @@ pub async fn start_rest_server(config: Rest, explorer_enabled: bool, context: Co
         .await
         .set_server_stopper(ServerStopper(stopper_tx));
 
-    let api =
-        warp::path!("api" / ..).and(v0::filter(context.clone()).or(v1::filter(context.clone())));
+    let api = warp::path!("api" / ..)
+        .and(v0::filter(context.clone()).or(v1::filter(context.clone())))
+        .with(warp::filters::trace::request());
```
mzabaluev, line 35: Looks like a closure passed to `trace` function would be an improvement because:

* It will allow us to choose the log level.
* Passing B3 information from the headers needs `tracing::Info`.

mzabaluev (approved):
Good for the initial implementation (and `info` level is correct for a basic request log), but see below.

jormungandr/src/rest/mod.rs
eugene-babichenko: > It will allow us to choose the log level

No. Levels are hardcoded in `warp`. We can only control the verbosity level of the whole span.

If you want to move `warp` traces to another level then it is the time to implement logging by ourselves. Which is not really hard, but is still an effort.

danielSanchezQ: We could write our own log filter (later?), shouldn't be difficult.

Zeegomo: >  info level is correct for a basic request log

We log fragments received from the rest endpoint at the `debug` level, I think it would be a bit strange to have tracing about those requests at a higher level than the transactions themselves

mzabaluev: > We can only control the verbosity level of the whole span.

I meant exactly that, you can construct a debug span in the closure passed to `trace::trace`.

mzabaluev (approved):

mzabaluev:
Ignoring the CircleCI test failures as a fluke for now.

PR #3358, opened 2021-06-07
logging: basic opentelemetry integration

eugene-babichenko:
The current state of it is that it just does debug prints of traces.
Defining a proper exporter is in progress.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1849,6 +1849,7 @@ dependencies = [
  "linked-hash-map",
  "lru",
  "nix",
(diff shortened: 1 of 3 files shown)


mzabaluev (approved):
Looks OK as a start, though it's unclear to me whether the stdout exporter would be good to consume. I expected there to be some common network protocol.

mzabaluev:
Superseded by other development (#3398 )

mzabaluev:
Superseded by other development, namely #3398.

PR #3366, opened 2021-06-08
Async traits update

Zeegomo:
Extract the `try_join` macro call to an external function as a workaround for https://github.com/dtolnay/async-trait/issues/144 and update async-trait to 0.1.50


mzabaluev (approved):
Thanks for figuring out the minimal workaround!

PR #3367, opened 2021-06-08
wait for fragment to enter mempool

Zeegomo:


testing/jormungandr-testing-utils/src/testing/fragments/verifier.rs
```diff
@@ -181,6 +181,7 @@ impl FragmentVerifier {
         &self,
         duration: Duration,
         check: MemPoolCheck,
+        exit_on_pending: bool,
```
mzabaluev, line 184: There's an opaque boolean parameter that's one more than a good method signature should take.
Couple style suggestions:
- Turn it into an enum with descriptive variants.
- Maybe convert some of these parameters into setter methods on `FragmentVerifier`.

Zeegomo:
We were so lucky that a test failed because of the persistent log rotation happening at that exact time.

testing/jormungandr-testing-utils/src/testing/fragments/verifier.rs
```diff
@@ -248,3 +248,17 @@ impl FragmentVerifier {
         })
     }
 }
+
+#[derive(Clone, Copy)]
+pub enum ExitStrategy {
```
mzabaluev, line 253: The name is a bit too abstract for strategy of waiting on pending fragments.
Maybe explicitly namespacing it under `fragments::ExitStrategy`, rather than exporting at top level, would be a better style.

mzabaluev (approved):

PR #3369, opened 2021-06-11
rest: log B3 headers information

eugene-babichenko:


jormungandr/src/rest/mod.rs
```diff
@@ -33,14 +33,30 @@ pub async fn start_rest_server(config: Rest, explorer_enabled: bool, context: Co
...
                 version = ?info.version(),
-                remote_addr = ?info.remote_addr(),
-            )
+                remote_addr = Empty,
+                trace_id = Empty,
+                span_id = Empty,
+                parent_span_id = Empty,
+            );
+            if let Some(remote_addr) = info.remote_addr() {
+                span.record("remote_addr", &remote_addr.to_string().as_str());
```
mzabaluev, line 50: Heck, this API is bad. `str` implements `Value`, but you can't pass the result of `.as_str()` here because the `record` method requires reference to an implicitly `Sized` value.

mzabaluev (approved):
Looks good and minimal. I have checked that `http_zipkin::get_trace_context` is non-scary to use in async code.

jormungandr/src/rest/mod.rs
eugene-babichenko: Yes, annoying as hell.

PR #3379, opened 2021-06-17
run actions on prs

Zeegomo:
A bunch of actions were only triggered by pushes to the repository, which do not play well with prs coming from a fork

### DIFF
--- a/.github/workflows/api.yml
+++ b/.github/workflows/api.yml
@@ -3,6 +3,9 @@ on:
   push:
     paths:
       - 'doc/api/*.yaml'
+  pull_request:
+    paths:
+      - 'doc/api/*.yaml'
 jobs:
   build:
     runs-on: ubuntu-latest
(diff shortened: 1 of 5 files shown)


eugene-babichenko:
Hm, that is strange. The changes actions run should be triggered by **any** push. So there should be no need to add a `pull_request` trigger, as the action was triggered by a push already.

mzabaluev (commented):
Cc @eugene-babichenko
IIRC we removed the `pull_request` triggers from test jobs so that every push does not produce a double number of test runs. Though in principle, they are testing different targets: the `pull_request` jobs test the result of a merge with master.

mzabaluev:
> should be triggered by any push.

Apparently not by pushes to a forked repository, like what this PR is based on.
If we still don't want double CI runs, maybe we should switch to `pull_request` as the only trigger?

eugene-babichenko (approved):

PR #3380, opened 2021-06-18
Bft desync tests

Zeegomo:
Add tests for de-sync handling in bft deployments.
Currently feature gated because it relies on the installation of `libfaketime` on the host machine

### DIFF
--- a/testing/jormungandr-scenario-tests/src/scenario/repository/mod.rs
+++ b/testing/jormungandr-scenario-tests/src/scenario/repository/mod.rs
@@ -176,11 +176,10 @@ fn scenarios_repository() -> Vec<Scenario> {
             transaction_to_passive,
             vec![Tag::Short],
         ),
-        #[cfg(feature = "desync-tests")]
         Scenario::new(
             "bft_forks",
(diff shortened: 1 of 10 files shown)


testing/jormungandr-scenario-tests/src/scenario/repository/mod.rs
```diff
@@ -176,6 +176,12 @@ fn scenarios_repository() -> Vec<Scenario> {
             transaction_to_passive,
             vec![Tag::Short],
         ),
+        #[cfg(feature = "desync-tests")]
```
dkijania, line 179: I think adding another tag: Tag::FakeTime should solve the issue.  So no need to add another cfg, since it looks like this is runtime dependency

dkijania (changes requested):

testing/jormungandr-scenario-tests/src/scenario/repository/mod.rs
dkijania: I know that i was proposing that, but after looking at the code it should be able to solve with separate Tag

dkijania (approved):

PR #3392, opened 2021-06-22
nix: bump nixpkgs

nrdxp:
Needed to build jormungandr with the latest cargo/rustc from the nix shell and the nix package.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -2,11 +2,11 @@
   "nodes": {
     "nixpkgs": {
       "locked": {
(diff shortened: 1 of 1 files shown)


mzabaluev:
This has conflicts with `flake.lock` now. @dkijania @const-iohk Do we still need this?

dkijania:
it need to be rebased and flake.lock updated.

nrdxp:
I'm not sure we even still need this so I will close for now. Feel free to let me know otherwise later.

PR #3395, opened 2021-06-25
Add log lib

enzoc4:
This moves the logging config/setup to a new lib/crate.

I'm not fully convinced this is worth it, but it didn't take much work anyway, so I'll leave it here to see what you think. This doesn't really block anything (see below), so there is no need to merge it any time soon.

The motivation would be to use it on #3350 too, but I currently went for the copy-paste option in that branch because otherwise it would be hard to merge, that's why we could leave this for a later date. Maybe even until the next time we want to change anything about the logging.

Also, this probably conflicts with #3358 I guess.


minikin:
@enzoc4  Closing it. Feel free to reopen if we really need it.

PR #3398, opened 2021-06-29
implement prometheus support; drop unused metrics

eugene-babichenko:


jormungandr/src/metrics/backends/prometheus_exporter.rs
```diff
@@ -0,0 +1,248 @@
...
+                .as_secs(),
+        );
+
+        let block_hash = block.header.hash();
+        let block_hash_bytes = block_hash.as_bytes();
+        for i in 0..4 {
+            let mut value_bytes = [0u8; 8];
+            value_bytes.copy_from_slice(&block_hash_bytes[(8 * i)..(8 * (i + 1))]);
+            let value = u64::from_le_bytes(value_bytes);
+            self.block_hash[i].set(value);
```
Zeegomo, line 245: Just wondering, splitting a block hash in 4 atomics u64 prevents us from blocking the thread, but could potentially return invalid values for a brief moment. Maybe we could put a (acq / rel) fence ? Not sure if the library allows such flexibility though

jormungandr/src/topology/quarantine.rs
```diff
@@ -62,24 +68,30 @@ impl ReportRecords {
...
-            false
+            ReportNodeResult::Ok
         } else {
             let mut peer_info = PeerInfo::from(node);
             tracing::debug!(node = %peer_info.address, id=?peer_info.id, ?self.report_duration, "reporting node");
             // If we'll handle report reasons other that a connectivity issue in the future, we may want to
             // demote a peer all the way down to dirty in case of a serious violation.
             topology.remove_peer(peer_info.id.as_ref());
 
+            let mut result = ReportNodeResult::Trusted;
```
Zeegomo, line 93: Not too clear what `ReportNodeResult::Trusted` should mean. At this stage a node is either in the `pool` pool, which is standard trust, or quarantined (where is correctly flagged as `ReportNodeResult::Quarantine`), but it cannot be in the `trusted` state.

jormungandr/src/topology/topology.rs
```diff
@@ -226,6 +238,7 @@ impl P2pTopology {
                 if let Some(node) = &node {
                     tracing::debug!(node = %node.address(), id=?node.id(), "lifting node from quarantine");
                     self.topology.promote_peer(&node.id());
+                    self.stats_counter.sub_peer_quarantined_cnt(1);
```
Zeegomo, line 241: Just noting that this counter might be lagging behind the real count of quarantined nodes because evictions may happen from the lru cache and we are not able to intercept those

jormungandr/src/metrics/backends/prometheus_exporter.rs
eugene-babichenko: That's a good idea, but yeah, there is no such functionality and getting invalid values is possible. I am not sure how to deal with that without introducing a lock. As you mentioned, this library is not very flexible.

jormungandr/src/topology/quarantine.rs
eugene-babichenko: The thing is that we do not really put trusted nodes in quarantine (see the code below), it is sort of a special status. And these nodes should not be counted in the quarantine counter.

Zeegomo: But once they're reported they're no longer trusted. The logic seems good, is just that I find the names a little confusing, like `ReportNodeResult::Quarantine` is the status *after* the report, but `ReportNodeResult::Trusted` is *before*.

jormungandr/src/metrics/backends/prometheus_exporter.rs
Zeegomo: Another idea could be to store the hash inside the `Prometheus` struct with some other atomic thing like arcswap and then update the registry only inside `http_response` before gathering metrics *if* the block hash has been updated.

eugene-babichenko: Yeah, that's what I thought.

jormungandr/src/topology/quarantine.rs
eugene-babichenko: Maybe, `Preferred` would be the right term?

jormungandr/src/topology/topology.rs
eugene-babichenko: We do not even have tests that account for such case in the first place, but that's a good thing to fix while we are on it.

jormungandr/src/topology/quarantine.rs
Zeegomo: mmm, something like `LightReport`? The peer is indeed reported and removed from the topology, but it's not quarantined. We do not have a word for that, but  `LightReport` should convey the idea I hope

jormungandr/src/metrics/backends/prometheus_exporter.rs
```diff
@@ -0,0 +1,264 @@
...
+    block_hash_value: ArcSwapOption<BlockContentHash>,
+}
+
+impl Prometheus {
+    pub fn new() -> Self {
+        Default::default()
+    }
+
+    pub fn http_response(&self) -> Result<impl warp::Reply, warp::Rejection> {
+        if let Some(block_hash) = self.block_hash_value.load_full() {
```
Zeegomo, line 47: If I'm not wrong this resets the counters each time the function is called even if the values did not change. Maybe we could put a `None` back in the pointer so that successive calls do not do unnecessary work? (I would avoid using a swap as it would require a write each time)

jormungandr/src/metrics/backends/prometheus_exporter.rs
```diff
@@ -31,6 +34,8 @@ pub struct Prometheus {
     block_chain_length: UIntGauge,
     block_time: UIntGauge,
     block_hash: Vec<UIntGauge>,
+
```
Zeegomo, line 37: Is this new line intended?

eugene-babichenko: Yes

jormungandr/src/topology/quarantine.rs
```diff
@@ -62,24 +68,30 @@ impl ReportRecords {
     }
 
     /// Returns whether the node has been quarantined or not.
-    pub fn report_node(&mut self, topology: &mut poldercast::Topology, node: Peer) -> bool {
+    pub fn report_node(
+        &mut self,
+        topology: &mut poldercast::Topology,
+        node: Peer,
+    ) -> ReportNodeResult {
```
mzabaluev, line 75: Bikeshed: don't like "result" naming, should not commingle with `Result`. Maybe `ReportNodeStatus`?

jormungandr/src/metrics/mod.rs
```diff
@@ -0,0 +1,66 @@
+use crate::blockchain::Ref;
+
+use chain_impl_mockchain::block::Block;
+use jormungandr_lib::time::SecondsSinceUnixEpoch;
+
+use std::sync::Arc;
+
+pub mod backends;
+
+pub trait MetricsBackend {
```
mzabaluev, line 10: Good work on the backends. Though now I'm thinking if maintaining the non-prometheus backend would be too much of a burden.

jormungandr/Cargo.toml
```diff
@@ -61,6 +61,7 @@ lru = "^0.6.1"
 warp = { version = "0.3.0", features = ["tls"] }
 serde_with = { version = "1.8", features = ["macros"] }
 http-zipkin = "0.3.0"
+prometheus = "0.12"
```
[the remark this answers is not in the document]
eugene-babichenko: Going to implement that soon.

jormungandr/src/metrics/mod.rs
eugene-babichenko: I don't think so, it is rather simple. And I am going to extract the common piece that is responsible for new block processing to make it even easier.

Anyway, right now we have a tendency for removing metrics, not adding them. :)

jormungandr/src/rest/mod.rs
```diff
@@ -58,19 +68,62 @@ pub async fn start_rest_server(config: Rest, explorer_enabled: bool, context: Co
...
+    App: Filter<Error = warp::Rejection> + Clone + Send + Sync + 'static,
+    App::Extract: warp::Reply,
+{
+    if config.enable_explorer {
         let explorer = explorer::filter(context);
-        setup_cors(api.or(explorer), config, stopper_rx).await;
+        setup_cors(app.or(explorer), config, shutdown_signal).await;
     } else {
-        setup_cors(api, config, stopper_rx).await;
+        setup_cors(app, config, shutdown_signal).await;
```
mzabaluev, line 120: I don't like how `setup_cors` is now called inside of one of the app setup functions.
Can we make each filter setup function return an `impl Filter + ...` to be able to chain them like warp intended?

jormungandr/src/topology/quarantine.rs
```diff
@@ -29,7 +29,7 @@ struct ReportRecord {
     report_time: Instant,
 }
 
-pub enum ReportNodeResult {
+pub enum ReportNodeStatus {
```
mzabaluev, line 32: Good.

jormungandr/src/rest/mod.rs
mzabaluev: Tried it, and it's not actually easy without lots of boxing.

mzabaluev (approved):
Looks OK. CircleCI fell off in a weird way.

PR #3401, opened 2021-07-01
Fix config generated by bootstrap script on Windows

eugene-babichenko:


enzoc4 (approved):
shouldn't jormungandr be able to handle Windows paths on Windows though?

eugene-babichenko:
The problem is that backslashes in YAML need to be escaped. I just found using slashes more aesthetic than `\\` :)

PR #3403, opened 2021-07-01
Async graphql warp 2.9.5

enzoc4:
closes #3399 

I'm actually not sure why it didn't work, but it's just removing that line.


eugene-babichenko (approved):
Test failure is unrelated.

PR #3404, opened 2021-07-01
Update chain-libs deps

Zeegomo:
Update jormungandr and jcli after recent refactor of chain-vote

### DIFF
--- a/jcli/Cargo.toml
+++ b/jcli/Cargo.toml
@@ -29,7 +29,7 @@ chain-impl-mockchain = { git = "https://github.com/input-output-hk/chain-libs.gi
 chain-addr      = { git = "https://github.com/input-output-hk/chain-libs.git", branch = "master" }
 chain-crypto    = { git = "https://github.com/input-output-hk/chain-libs.git", branch = "master" }
 chain-time    = { git = "https://github.com/input-output-hk/chain-libs.git", branch = "master" }
-chain-vote = { git = "https://github.com/input-output-hk/chain-libs.git", branch = "master", features = ["p256k1"] }
+chain-vote = { git = "https://github.com/input-output-hk/chain-libs.git", branch = "master" }
 jormungandr-lib = { path = "../jormungandr-lib" }
 gtmpl = "0.6.0"
 ed25519-bip32 = "0.3"
--- a/jcli/Cargo.toml
+++ b/jcli/Cargo.toml
@@ -32,7 +32,7 @@ chain-time    = { git = "https://github.com/input-output-hk/chain-libs.git", bra
 chain-vote = { git = "https://github.com/input-output-hk/chain-libs.git", branch = "master" }
 jormungandr-lib = { path = "../jormungandr-lib" }
 gtmpl = "0.6.0"
-ed25519-bip32 = "0.3"
+ed25519-bip32 = "0.4"
 thiserror = "1.0"
 bytes = "1.0"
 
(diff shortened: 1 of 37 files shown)


jcli/src/jcli_lib/utils/vote.rs
```diff
@@ -70,6 +70,8 @@ pub enum SharesError {
     InsufficientShares,
     #[error("invalid binary share data")]
     InvalidBinaryShare,
+    #[error("decryption share cannot be validated")]
```
eugene-babichenko, line 73: "is not valid"?

eugene-babichenko (approved):

Zeegomo:
Failure is unrelated

jcli/src/jcli_lib/vote/mod.rs
```diff
@@ -61,8 +61,8 @@ pub enum Error {
 pub enum Vote {
     /// Create committee member keys
     Committee(committee::Committee),
-    /// Build an encryption key from committee member keys
-    EncryptingKey(encrypting_vote_key::EncryptingVoteKey),
```
mzabaluev, line 65: Please change the example in docs that uses `--encrypting-key`.
Some other tooling may need to be updated as well; Cc @dkijania @danielSanchezQ

jcli/Cargo.toml
```diff
@@ -29,10 +29,10 @@ chain-impl-mockchain = { git = "https://github.com/input-output-hk/chain-libs.gi
 chain-addr      = { git = "https://github.com/input-output-hk/chain-libs.git", branch = "master" }
 chain-crypto    = { git = "https://github.com/input-output-hk/chain-libs.git", branch = "master" }
```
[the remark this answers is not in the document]
mzabaluev: OK, saw the other PR that changes this.

jcli/src/jcli_lib/key.rs
```diff
@@ -2,9 +2,9 @@ use crate::jcli_lib::utils::io;
 use crate::jcli_lib::utils::output_file::{self, OutputFile};
 use bech32::{self, u5, FromBase32, ToBase32};
 use chain_crypto::{
-    bech32::Bech32 as _, AsymmetricKey, AsymmetricPublicKey, Curve25519_2HashDh, Ed25519,
-    Ed25519Bip32, Ed25519Extended, SecretKey, SigningAlgorithm, SumEd25519_12, Verification,
-    VerificationAlgorithm,
+    bech32::Bech32 as _, AsymmetricKey, AsymmetricPublicKey, Ed25519, Ed25519Bip32,
+    Ed25519Extended, EllipticCurve2hashDh, SecretKey, SigningAlgorithm, SumEd25519_12,
```
mzabaluev, line 6: So now we have explicitly selected algorithms for most cryptographic things in jormungandr, but a feature-selected alias for the ECC which affects both the vote protocol and Praos.
This diverges from the general approach of providing generic crypto APIs in the libraries and selecting specific implementations in jormungandr (`blockcfg` is where it should be concentrated, actually).

Can we export both ECC backends in chain-libs unaliased, so that jormungandr can select one of them explicitly? This will remove the issue with being unable to build the library with `--all-features` because of the conflicting backends.

mzabaluev (approved):
Approving with actions since chain-libs already has the changes committed to master, will file a follow-up issue about that.

mzabaluev (approved):

PR #3420, opened 2021-07-08
Clean up the list of dependabot ignores

mzabaluev:
Most of these ignores are obsolete and were created by commands to
old dependabot.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,81 +1,17 @@
 version: 2
 updates:
-- package-ecosystem: cargo
-  directory: "/"
-  schedule:
-    interval: daily
-  open-pull-requests-limit: 10
-  ignore:
-  - dependency-name: bytes
-    versions:
-    - ">= 0.6.a, < 0.7"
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -1,81 +1,17 @@
...
-  - automated
-  - dependencies
-  - submodules
+  - package-ecosystem: cargo
+    directory: '/'
+    schedule:
+      interval: daily
+    open-pull-requests-limit: 10
+    ignore: []
+  - package-ecosystem: gitsubmodule
```
Zeegomo, line 9: Do we still need the submodule part?

Zeegomo (approved):

.github/dependabot.yml
mzabaluev: Actually not. We can remove this as well, but it does not cost us anything and it may need to be re-added if we ever revert to using submodules.

eugene-babichenko (approved):

PR #3422, opened 2021-07-09
rest/v1: get all proposals in a vote plan a user have voted for

eugene-babichenko:


jormungandr/src/rest/v1/logic.rs
```diff
@@ -38,6 +39,8 @@ pub enum Error {
     Hex(#[from] hex::FromHexError),
     #[error("Could not process all fragments")]
     Fragments(FragmentsProcessingSummary),
+    #[error("Unexpected address type")]
+    UnexpectAddressType,
```
mzabaluev, line 43: `Unexpected`?

jormungandr/src/rest/v1/logic.rs
```diff
@@ -111,3 +114,49 @@ pub async fn get_fragment_logs(context: &Context) -> Result<Vec<FragmentLog>, Er
...
+            .blockchain_tip()?
+            .get_ref()
+            .await
+            .ledger()
+            .active_vote_plans()
+            .into_iter()
+            .find(|x| x.id == vote_plan_id);
+        let vote_plan = match maybe_vote_plan {
+            Some(vote_plan) => vote_plan,
+            None => return Ok(None),
```
mzabaluev, line 149: Should we respond with 404 when the vote plan is not found?

eugene-babichenko: It is converted into 404 in `handlers.rs` the same way we do in `v0` since statuses handling is specific to a particular framework.

mzabaluev: OK. It could be another internal error case instead.

mzabaluev (commented):
Looks OK apart from the schema lint and error case naming, which is internal anyway.

mzabaluev (approved):

PR #3423, opened 2021-07-09
ci: use cargo-audit without a wrapper Action

eugene-babichenko:


.github/workflows/audit.yml
```diff
@@ -33,7 +33,8 @@ jobs:
           restore-keys: |
             cargo-index-
 
-      # FIXME switch back to actions-rs/audit-check when the fix is merged
-      - uses: eugene-babichenko/audit-check@dont-override-cargo-lock
-        with:
-          token: ${{ secrets.GITHUB_TOKEN }}
+      - name: Install cargo audit
+        run: cargo install cargo-audit
```
mzabaluev, line 37: This takes a long time to build.

Can we supply a docker container, for example, reusing our old [CI Dockerfile](https://github.com/input-output-hk/rust-ci-docker/blob/master/stable/Dockerfile) which used to also have cargo audit?

eugene-babichenko: Yeah I think we can. What is the difference between `circleci/rust` and `rust` images?

Zeegomo: `circleci/rust` is a legacy image  (`cimg/rust` is the replacement), but I don't know the difference between that and `rust`.

eugene-babichenko:
Superseded by #3434

PR #3434, opened 2021-07-13
ci: use custom Docker image for audit

eugene-babichenko:
Example of workflow runs:

- Docker build: https://github.com/input-output-hk/jormungandr/actions/runs/1026310518
- Audit: https://github.com/input-output-hk/jormungandr/runs/3056159930?check_suite_focus=true


docker/ci/Dockerfile
```diff
@@ -0,0 +1,2 @@
+FROM rust:1.53-slim
```
mzabaluev, line 1: So we have two directories where it could make sense to host it: `ci` and `docker/ci`.
I think it makes more sense to do under `ci`, like the other stuff we use with the workflows.
Especially if this is meant to be an image just for the audit job.

`./docker` is more intended for the community users who want to bake their own jormungandr images.

docker/ci/Dockerfile
```diff
@@ -0,0 +1,2 @@
...
+RUN apt-get update && apt-get install -y pkg-config libssl-dev && cargo install cargo-audit
```
mzabaluev, line 2: There ought to be an `apt-get clean` in the same layer step, like [here](https://github.com/input-output-hk/rust-ci-docker/blob/0e678de1fe98f23455262bc44e737b4299f0001e/stable/Dockerfile#L8).

Conversely, `cargo install` might be done in a different later `RUN` step from the apt-get installs; this way layer caching is more affine to the respective package source's changes.

mzabaluev:
```suggestion
RUN apt-get update && apt-get install -y pkg-config libssl-dev && apt-get clean
RUN cargo install cargo-audit
```

mzabaluev (approved):

PR #3447, opened 2021-07-16
Fix schedule missing epoch

Zeegomo:
Re-use the schedule epoch when waiting for the next one to avoid
missing a whole new epoch when doing the check at epoch boundaries

I'm not sure how to go about testing, adding a unit test is a bit of a mess because I would basically need to bootstrap a whole node and adding an integration test has its challenges in dealing with tricky time logic.

One way to approach this could be to use libfaketime and add a test in scenario tests, but it won't be tested by the CI.

### DIFF
--- a/jormungandr/src/leadership/process.rs
+++ b/jormungandr/src/leadership/process.rs
@@ -146,24 +146,14 @@ impl Module {
         }
     }
 
-    /// this function compute when the next epoch will start, next epoch
-    /// from the local system time point of view. Meaning this is not the
-    /// epoch of the current tip
-    fn next_epoch_time(&self) -> Result<SystemTime, LeadershipError> {
-        let current_position = self.current_slot_position()?;
-        let epoch = Epoch(current_position.epoch.0 + 1);
-        let slot = EpochSlotOffset(0);
-
-        Ok(self.slot_time(epoch, slot))
+    fn epoch_time(&self, epoch: Epoch) -> Result<SystemTime, LeadershipError> {
+        Ok(self.slot_time(epoch, EpochSlotOffset(0)))
     }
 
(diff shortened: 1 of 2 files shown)


mzabaluev (approved):

jormungandr/src/leadership/process.rs
```diff
@@ -262,7 +252,7 @@ impl Module {
                 // wait for the next epoch
 
                 tracing::debug!("no item scheduled, waiting for next epoch");
-                self.next_epoch_instant()
+                self.epoch_instant(Epoch(self.schedule.as_ref().unwrap().epoch().0 + 1))
```
mzabaluev, line 255: Is this always in the future?

jormungandr/src/leadership/process.rs
```diff
@@ -146,24 +146,14 @@ impl Module {
...
+    fn epoch_instant(&self, epoch: Epoch) -> Result<Instant, LeadershipError> {
+        let epoch_time = self.epoch_time(epoch)?;
 
-        match next_epoch_time
-            .as_ref()
-            .duration_since(SystemTime::now().into())
-        {
+        match epoch_time.as_ref().duration_since(SystemTime::now().into()) {
             Err(err) => {
                 // only possible if `next_epoch_time` is earlier than now. I.e. if the next
```
eugene-babichenko, line 158:
```suggestion
                // only possible if `epoch_time` is earlier than now. I.e. if the next
```

jormungandr/src/leadership/process.rs
Zeegomo: Not necessarily. For example, in the case that raised the bug this would result in an `Instant` that is before `Instant::now`

PR #3458, opened 2021-07-22
Clean up serialization of vote plans

mzabaluev:
Deduplicate code in serde impls. VotePlan DTO is defined as a struct
with its own serde trait impls and From conversions from/to
the chain-libs VotePlan struct.

Use the BlockDate DTO and serialization consistently everywhere,
meaning that the string format will be used in all human-readable
serializations.
To support deserializing existing vote plan JSON data with object format
for blockdates, define adaptive serialization with a Visitor.


danielSanchezQ (approved):
LGTM

PR #3470, opened 2021-07-27
[Tests] Non duplicated votes update tests

dkijania:
Updated chain-libs with newest changed regarding non-duplicated votes feature. Added new test cases for positive and negative scenario. Introduced FragmentChainSender which should improve test readibility and VotesCastCounter which should help adjusting load test for new requirements

### DIFF
--- a/testing/jormungandr-testing-utils/src/testing/fragments/generator.rs
+++ b/testing/jormungandr-testing-utils/src/testing/fragments/generator.rs
@@ -1,4 +1,5 @@
 use super::{FragmentSender, FragmentSenderError, MemPoolCheck};
+use crate::testing::vit::VoteCastCounter;
 use crate::testing::SyncNode;
 use crate::{
     stake_pool::StakePool,
@@ -22,14 +23,16 @@ pub struct FragmentGenerator<'a, S: SyncNode + Send> {
     sender: Wallet,
     receiver: Wallet,
     active_stake_pools: Vec<StakePool>,
-    vote_plan_for_casting: Option<VotePlan>,
+    vote_plans_for_casting: Vec<VotePlan>,
     vote_plans_for_tally: Vec<VotePlan>,
     node: RemoteJormungandr,
     rand: OsRng,
+    vote_cast_register: Option<VoteCastCounter>,
     slots_per_epoch: u32,
     fragment_sender: FragmentSender<'a, S>,
     stake_pools_count: usize,
     vote_plans_for_tally_count: usize,
+    vote_plans_for_casting_count: usize,
 }
 
 impl<'a, S: SyncNode + Send> FragmentGenerator<'a, S> {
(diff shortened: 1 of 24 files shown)


testing/jormungandr-testing-utils/src/testing/vit/vote_cast_register.rs
```diff
@@ -0,0 +1,527 @@
...
+    pub fn new(vote_plans: Vec<(VotePlanId, u8)>) -> Self {
+        Self {
+            register: vote_plans
+                .iter()
+                .map(|(id, limit)| VotePlanVoteCastPosition::new(id.clone(), *limit))
+                .collect(),
+        }
+    }
+
+    pub fn has_next(&self) -> bool {
```
Zeegomo, line 72: nit: this sounds like I'm reading a java iterator implementation

testing/jormungandr-testing-utils/src/testing/vit/vote_cast_register.rs
```diff
@@ -0,0 +1,527 @@
...
+    }
+
+    pub fn available_to_send(&self) -> usize {
+        (self.limit - self.current).into()
+    }
+
+    pub fn advance_single_unsafe(&mut self) -> Result<VotesToCast, Error> {
+        self.advance_batch_unsafe(1)
+    }
+    pub fn advance_batch_unsafe(
```
Zeegomo, line 174: Nit: I think unsafe usually conveys a particular concept in rust, which is not what it means here

testing/jormungandr-testing-utils/src/testing/fragments/generator.rs
```diff
@@ -206,16 +233,31 @@ impl<'a, S: SyncNode + Send> FragmentGenerator<'a, S> {
...
-                0,
-                &Choice::new(1),
-                &self.node,
-            ),
+            8 => {
+                let vote_cast_register = self
+                    .vote_cast_register
+                    .as_mut()
+                    .expect("please run 'prepare' method before running load");
+                let wallet_idx = 0;
```
Zeegomo, line 241: Why always with wallet 0?

Zeegomo (commented):
Maybe I've missed it, but did you plan to update the `VoteCastsGenerator` in `testing/jormungandr-testing-utils/src/testing/fragments/load/vote_casts_generator.rs` as well?

testing/jormungandr-testing-utils/src/testing/vit/vote_cast_register.rs
dkijania: I guess it is not good :) removing then

testing/jormungandr-testing-utils/src/testing/vit/vote_cast_register.rs
dkijania: renaming

testing/jormungandr-testing-utils/src/testing/fragments/generator.rs
dkijania: added comment

mzabaluev:
Needs to be rebased, also the lints have been fixed in master.

PR #3495, opened 2021-08-08
Add network tests

mrzenioszeniou:
Added tests for the following P2P network parameters:
* `gossip_interval`
* `network-stuck_check`
* `max_bootstrap_attempt`

I also did some housekeeping in our test code:
* We had a few module directories with a single `mod.rs` file in them. I got rid of the unnecessary nesting:
    * `testing/jormungandr-integration-tests/src/common/data`
    * `testing/jormungandr-integration-tests/src/common/startup`
    * `testing/jormungandr-integration-tests/src/jormungandr/explorer`
    * `testing/jormungandr-integration-tests/src/jormungandr/legacy`
* Some functions where unnecessarily declared public so I made them private.
* `jormungandr_integration_tests::common::network::builder::NetworkBuilder.configs` was not meaningfully used, so I got rid of it. Consequentially I removed `testing/jormungandr-integration-tests/src/networking/p2p/topic_of_interest_influences_node_sync_ability` as well.
* There were a couple of loose constructors, so I associated them with the appropriate structs.

__Note:__ I ended up using the logs to figure out if -and when- certain events took place. For example, to figure out the interval between to consecutive gossip attempts. I don't like this, mainly because our tests' validity hangs by a string literal or a time format somewhere. If said literal or time format changes, the compiler can't help us fix the now-failing tests. If there's a cleaner way, feel free point it out.


testing/jormungandr-integration-tests/src/networking/p2p.rs
```diff
@@ -334,63 +343,156 @@ pub fn node_put_itself_in_preffered_layers() {
...
+        }
+    }
+}
 
-    assert!(
-        fast_client_block_recv_cnt > slow_client_block_recv_cnt,
-        "node with high block topic of interest should have more recieved blocks"
+#[ignore = "long runner"]
+#[test]
+fn network_stuck_check() {
```
dkijania, line 403: for long running tests there is a feature: `soak-non-functional`.  
```
#[cfg(feature = "soak-non-functional")]
```

mrzenioszeniou: Nice, thanks!

dkijania:
IMHO we can try to pursue with assertion in logs for now. We ran tests on PR, so there would be a feedback if format would change (maybe not so convenient as during compilation).

dkijania (approved):

mrzenioszeniou:
@dkijania the `gossip_interval` test fails occasionally which made me rethink the logic thereof. Even if the sender times their gossiping attempts correctly, there's no guarantee that the receiver will get them with at least `gossip_interval` seconds in between. So instead of checking when the gossip was received, I added a trace log that we can use to check when the node is attempting to gossip. On top of that I noticed that we're using [tokio::time::Interval](https://docs.rs/tokio/1.9.0/tokio/time/struct.Interval.html) to time this, which has a `tokio::time::MissedTickBehavior` field that defaults to [Burst](https://docs.rs/tokio/1.9.0/tokio/time/enum.MissedTickBehavior.html#variant.Burst). This means that if the node misses a tick, it will try to make up for it as fast as possible, without respecting the `gossip_interval`. That doesn't seem beneficial for the network's health anyway, so I thought we should switch to [Delay](https://docs.rs/tokio/1.9.0/tokio/time/enum.MissedTickBehavior.html#variant.Delay) instead. It doesn't try to make up for lost ticks and ensures that the duration between ticks is always respected. Check out the last commit I just pushed.

dkijania (approved):

PR #3497, opened 2021-08-09
add buffering for persistent logs

Zeegomo:
Add buffering to persistent logs to improve performance and switch to use `tokio::fs::File` instead of `std::fs::File` not to block the runtime for the duration of the write.
Not really sure of what size of the buffer to pick, but I think it does not make too much of a difference provided it's more than the underlying block size and we flush it every batch. The one that is currently used is quite big to avoid lots of flushes while processing a single batch.

We should benchmark this and see if it works for our deployment, or if we should not forcibly flush the buffer.


jormungandr/src/fragment/pool.rs
```diff
@@ -123,15 +134,17 @@ impl Pools {
...
-            if let Some(mut persistent_log) = self.persistent_log.as_mut() {
+            if let Some(persistent_log) = self.persistent_log.as_mut() {
                 let entry = PersistentFragmentLog {
                     time: SecondsSinceUnixEpoch::now(),
                     fragment: fragment.clone(),
                 };
                 // this must be sufficient: the PersistentFragmentLog format is using byte array
                 // for serialization so we do not expect any problems during deserialization
                 let codec = bincode::DefaultOptions::new().with_fixint_encoding();
-                if let Err(err) = codec.serialize_into(&mut persistent_log, &entry) {
```
mzabaluev, line 134: That's why it was slow: we gave the `File` writer to bincode to do lots of small writes.

mzabaluev (approved):
Looks good.

mzabaluev:
Needs rebasing.

PR #3500, opened 2021-08-10
do not propagate overflowing fragments

eugene-babichenko:


jormungandr/src/fragment/pool.rs
```diff
@@ -188,6 +188,9 @@ impl Pools {
             }
         }
 
+        // we are not concerned about the fragments that were rejected by all pools
+        filtered_fragments.truncate(accepted.len());
```
Zeegomo, line 192: I'm not super confident with truncating it like this. It's not necessarily a problem for overflowing transactions but this relies on the assumption that the accepted fragments are always the first X ones, while this is not true.

For example, imagine the the logs and mempool of a passive node both have the same size T, and they hold the last T transactions submitted to the node (i.e. they are both full). Now, the node receives a block announcement with U transactions, completely different from the ones currently in the mempool. This will make the logs evict the last U transaction to make space for the new `InABlock` records.
After it, I try to submit two transactions in a batch: the first was the first transaction ever received by the node, while the second is a completely new one. What will happen is that even though the first transaction will make it through the initial filtering (it was evicted from the logs), it will not be included in the pool, because it is still there (remember this is a passive node) and will not be inserted in the `accepted`  set, while the second one will.
Unfortunately, truncating `filtered_fragments` to length 1 will keep the wrong transaction and never propagate the correct one

eugene-babichenko: > It's not necessarily a problem for overflowing transactions but this relies on the assumption that the accepted fragments are always the first X ones, while this is not true.

Well, we filter out invalid and logged fragments here https://github.com/input-output-hk/jormungandr/blob/dont-prop-overflowing-fragments/jormungandr/src/fragment/pool.rs#L101-L108
so I think the assumption you are talking about is safe. Especially given that we do not have parallel operations on the mempool module.

Zeegomo: Logs and mempool are not always guaranteed to be perfectly synced. I think the example above explains a situation in which that happens.

eugene-babichenko: Yeah, I get it now. A bit problematic, but I think I have a workaround for that.

Zeegomo: Maybe we could turn the `accepted` `HashSet` into a `HashMap` and return its values for this new purpose

Zeegomo (approved):

PR #3501, opened 2021-08-10
derive Eq for VotePlanStatus

Zeegomo:


danielSanchezQ (approved):
Check the lints, otherwise lgtm.

PR #3505, opened 2021-08-11
[Test] Adjust test tools in preparation for vote check tool

Zeegomo:
This PR adjust some of the tools the are currently used in testing to be able to reuse them for the vote check tool.
In particular, it contains: 
* a quick change to make `Starter` more binary friendly by:
  * adding a verbose flag to suppress unwanted output
  * lazily get the local jormungandr binary
  * remove unnecessary panics

  If we plan to use such struct more broadly I think it would be better to refactor it (and some of its dependencies, like `JormungandrProcess`) in some shared tool space (e.g. jormungandr-(testing)-utils).

  Due to the challenges of such refactor and the close deadline for code freeze I chose to go with this simpler solution for now. 
* an addition to be able to run arbitrary queries with the explorer interface tool


dkijania (approved):

dkijania:
very good stuff

PR #3529, opened 2021-08-19
tests: remove hidden TTL manipulations from builders

eugene-babichenko:


testing/jormungandr-testing-utils/src/testing/fragments/mod.rs
```diff
@@ -109,51 +105,50 @@ impl FragmentBuilder {
...
     pub fn full_delegation_cert_for_block0(
-        date: BlockDate,
+        valid_until: BlockDate,
         wallet: &Wallet,
         pool_id: PoolId,
     ) -> Initial {
         let expiry_date = BlockDate {
-            epoch: date.epoch + 1,
-            slot_id: date.slot_id,
+            epoch: valid_until.epoch + 1,
```
mrzenioszeniou, line 124: Why are we keeping this manipulation here? If the intend is indeed for the expiration to fall an epoch after `valid_until` I'd like to note that:
* the parameter here is named `valid_until`
* the variable is then named `expiry_date`
* the instance then finds its way to `chain_impl_mockchain::transaction::builder::TxBuilderState::set_expiry_date` whose parameter is called `valid_until` once more.

At this point I'm struggling to grasp what is semantically correct.

I'm also confused by the name of this function – which I know is unrelated to this PR. If it's a delegation certificate for block0, I'd expect the `BlockDate` to be statically instantiated to the first epoch or the one after that.

eugene-babichenko: I overlooked this one, thanks.

eugene-babichenko: > I'm also confused by the name of this function – which I know is unrelated to this PR. If it's a delegation certificate for block0, I'd expect the BlockDate to be statically instantiated to the first epoch or the one after that.

We may utilize it for some testing later, but I think having this now is the consequence of copy-pasting in the first place.

mrzenioszeniou (approved):

mzabaluev:
Needs rebasing to resolve the conflicts.

PR #3530, opened 2021-08-20
consistent transaction fields order

Zeegomo:
It was bothering me a little that different `Wallet` transactions had the block date field in different positions.


eugene-babichenko (approved):

mzabaluev (approved):
We should just have builder APIs everywhere to not have to deal with this.

PR #3531, opened 2021-08-20
add pending transactions to node stats

Zeegomo:


jormungandr/src/fragment/pool.rs
```diff
@@ -562,13 +574,23 @@ pub(super) mod internal {
...
         }
 
-        pub fn remove_all<'a>(&mut self, fragment_ids: impl IntoIterator<Item = &'a FragmentId>) {
-            for fragment_id in fragment_ids {
-                let maybe_fragment = self.entries.remove(fragment_id);
-                if let Some(fragment) = maybe_fragment {
-                    self.timeout_queue_remove(&fragment);
-                }
-            }
+        /// returns the ids of updated fragments
```
mzabaluev, line 577: Perhaps adding a doc on what the method actually does before returning would be good too.

mzabaluev (commented):
Looks good except for some type mismatches, it's still unfinished I guess.

Zeegomo:
Yep, I pushed it because I didn't want to forget about it but it still needs some polishing touches, hence the draft status

PR #3537, opened 2021-08-24
TTL Functional Tests

mrzenioszeniou:
Introduced some TTL tests to complement #3504.

* Added a simple unit test in to ensure transactions are removed as soon as they're expired.
* Added JCLI test to ensure no transactions can be created without a time validity.
* Added JCLI test to ensure different time validity yields different transaction.
* Added single-node test to ensure expired transactions are rejected.
* Added single-node test to ensure transactions with long TTLs are rejected.


testing/jormungandr-integration-tests/src/jormungandr/transactions.rs
```diff
@@ -89,3 +93,124 @@ pub fn accounts_funds_are_updated_after_transaction() {
...
+    )
+    .unwrap();
+
+    let jcli = JCli::default();
+
+    let valid_transaction = sender
+        .transaction_to(
+            &jormungandr.genesis_block_hash(),
+            &jormungandr.fees(),
+            chain_impl_mockchain::block::BlockDate::first(),
```
Zeegomo, line 124: mmm, I think there's still some date magic under this and I don't particularly like it. Didn't we fix it already (or was it something different)? IIRC we agreed to not have date manipulation, at least for interfaces which are simply building a transaction like you would on the command line like `Wallet`

testing/jormungandr-integration-tests/src/jormungandr/transactions.rs
```diff
@@ -89,3 +93,124 @@ pub fn accounts_funds_are_updated_after_transaction() {
...
+
+    jcli.fragment_sender(&jormungandr)
+        .send(&valid_transaction)
+        .assert_in_block();
+
+    let expired_transaction = sender
+        .transaction_to(
+            &jormungandr.genesis_block_hash(),
+            &jormungandr.fees(),
+            chain_impl_mockchain::block::BlockDate {
```
Zeegomo, line 203: Same as above, this is also confusing, why would it fail if the validity date is set to the max allowed?

testing/jormungandr-testing-utils/src/testing/fragments/sender.rs
```diff
@@ -142,6 +142,21 @@ impl<'a, S: SyncNode + Send> FragmentSender<'a, S> {
         self.send_fragment(from, fragment, via)
     }
 
+    pub fn send_transaction_with_validity<A: FragmentNode + SyncNode + Sized + Send>(
```
Zeegomo, line 145: This sounds like validity is an optional field, and indeed we have also `send_transaction`. I would pick another less ambiguous name, and also understand why we have these two methods. If `send_transaction` is a convenience method to do some date magic it could just pick the value and reuse `send_transaction_with_validity`

testing/jormungandr-testing-utils/src/testing/fragments/sender.rs
```diff
@@ -142,6 +142,21 @@ impl<'a, S: SyncNode + Send> FragmentSender<'a, S> {
...
+        &self,
+        from: &mut Wallet,
+        to: &Wallet,
+        via: &A,
+        value: Value,
+        valid_until: BlockDate,
+    ) -> Result<MemPoolCheck, FragmentSenderError> {
+        let address = to.address();
+        let fragment =
+            from.transaction_to(&self.block0_hash, &self.fees, valid_until, address, value)?;
```
Zeegomo, line 155: This is misleading, the wallet treats the date as "current date" (poor choice of name there, besides the unwanted manipulation), so this does not set the expiry date for the transaction to `valid_until`, but to the next epoch.
I think this is another evidence that `Wallet` should not do anything with the date it receives, and use it directly to set the `valid_until` transaction field

Zeegomo (changes requested):
The tests themselves are good, but in my opinion the way the test framework is written introduces unnecessary confusion.
I would rather fix that first than keep adding stuff on top of it which we would need to be fixed later

testing/jormungandr-integration-tests/src/jormungandr/transactions.rs
mrzenioszeniou: Yes that is correct. In this case the magic is hidden in `jormungandr_testing_utils::testing::fragments::FragmentBuilder`. We have #3529 which should clean these up, so I didn't bother fixing them again. We can rebase this branch onto it or the other way around.

testing/jormungandr-testing-utils/src/testing/fragments/sender.rs
mrzenioszeniou: Oh my bad, that's leftover code from an attempt to clean up the testing API. I got rid of it. `BlockDate` magic aside, the validity for a transaction is defined upon construction of the `FragmentSender` and all transactions created by it use that `BlockDate`.

testing/jormungandr-integration-tests/src/common/jcli/services/fragment_check.rs
```diff
@@ -156,4 +163,8 @@ impl<'a> FragmentCheck<'a> {
             ),
         }
     }
+
+    pub fn assert_rejected_summary(self) {
```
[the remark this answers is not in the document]
mrzenioszeniou: Often enough the tests end with an assertion, which is why some these assertive functions consume `self`. You're right though, there's no reason why they should. I changed them to `&self`.

testing/jormungandr-integration-tests/src/jormungandr/transactions.rs
Zeegomo: Indeed, I recalled seeing a PR for that. All good then! But I would wait for it to land before merging this PR, as it will require changes to the tests I guess

jormungandr/src/fragment/pool.rs
```diff
@@ -668,6 +669,41 @@ pub(super) mod internal {
             }
             assert!(pool.remove_oldest().is_none());
         }
+
+        #[test]
+        fn expired_transactions_are_removed() {
+            let mut pool = Pool::new(1);
+
+            let tx = Fragment::Transaction(
+                TxBuilder::new()
```
eugene-babichenko, line 678: Unrelated: this is funny, that such no-op transaction can be built.

eugene-babichenko (approved):
All looking good. I merged #3529, so please rebase onto the latest `master`.

Zeegomo (approved):

PR #3547, opened 2021-08-27
metrics: add pending tx count

eugene-babichenko:


eugene-babichenko:
Now that we use only one pool per node this became much simpler.

jormungandr/src/metrics/backends/simple_counter.rs
```diff
@@ -97,6 +104,10 @@ impl MetricsBackend for SimpleCounter {
         self.tx_recv_cnt.fetch_add(count, Ordering::SeqCst);
     }
 
+    fn set_tx_pending_cnt(&self, count: usize) {
+        self.tx_pending_cnt.store(count, Ordering::SeqCst);
```
mzabaluev, line 108: Idly wondering if this needs to be `SeqCst` if you fetch the metrics with `Relaxed`...

This shouldn't matter much in the grand scheme of things, though.

mzabaluev (approved):

jormungandr/src/metrics/backends/simple_counter.rs
eugene-babichenko: I can revisit ordering here. I remember having odd test failures with less strict ordering requirements. Basically, I just copy-pasted these from older implementation.

PR #3548, opened 2021-08-27
Additional TTL Functional Tests

mrzenioszeniou:
Addendum to #3537:

* New test to verify expired transactions do not create entries in the persistent log.
* New test to verify a node will reject transactions that expired after they've entered the mempool.

### DIFF
--- a/testing/jormungandr-integration-tests/src/jormungandr/mempool.rs
+++ b/testing/jormungandr-integration-tests/src/jormungandr/mempool.rs
@@ -1,11 +1,14 @@
+use crate::common::jcli::JCli;
 use crate::common::jormungandr::{starter::Role, Starter};
+use crate::common::transaction_utils::TransactionHash;
 use crate::common::{jormungandr::ConfigurationBuilder, startup};
 use assert_fs::fixture::PathChild;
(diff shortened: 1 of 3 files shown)


Zeegomo (approved):
Looks good, but where we decided that an expired transaction should not be included in persistent logs? It's ok for me, I just wonder if there was any discussion about it, or it was a result of recent changes to the code

mrzenioszeniou:
> Looks good, but where we decided that an expired transaction should not be included in persistent logs? It's ok for me, I just wonder if there was any discussion about it, or it was a result of recent changes to the code

@Zeegomo, I'm not sure what the motivation for this decision is. @eugene-babichenko integrated the TTL changes so maybe he can shed some light.

eugene-babichenko:
> 
> 
> Looks good, but where we decided that an expired transaction should not be included in persistent logs? It's ok for me, I just wonder if there was any discussion about it, or it was a result of recent changes to the code

It is the same case as with any transaction that was rejected upon submission. If a transaction expires while being in the pool it will remain in the fragment log.

eugene-babichenko (approved):

PR #3549, opened 2021-08-27
basic xtask usage

eugene-babichenko:
This commit introduces the basic use of `xtask` approach as described in

https://github.com/matklad/cargo-xtask/

The idea is to bring better development workflow while keeping
everything within the scope of `cargo` and not introducing any
additional scripting like bash/PowerShell/Python.

If we find this approach interesting, I would also suggest porting some
parts of our release pipelines to Rust. This would give us access to
libraries like `cargo-metadata` producing better/easier to understand
scripts.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,0 +1,2 @@
+[alias]
+xtask = "run -p xtask -- "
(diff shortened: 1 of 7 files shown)


minikin:

PR #3556, opened 2021-08-30
allow only one stake pool to operate on a single node

eugene-babichenko:


Zeegomo (commented):
Multiple BFT leaders can be supported without the need to have multiple pools. Not that it's a super common deployment but it does not incur in significant complexity on our side (except a bit of key management).
Also the capability of changing leader keys 'live' could be nice to retain anyway (for example returning an error if more than one praos keys are set at once)

eugene-babichenko:
>  Also the capability of changing leader keys 'live' could be nice to retain anyway (for example returning an error if more than one praos keys are set at once)

Do you refer to the case where we left the possibility to have multiple BFT leaders here?

Zeegomo:
> Do you refer to the case where we left the possibility to have multiple BFT leaders here?

It looks like this capability was removed in this PR.

I'm saying that this PR removes a bit more than just multiple stake pools on a single node like:
* multiple BFT leaders
* key management without having to restart the node (for both praos and bft keys)

And I'm not sure we should remove these

Zeegomo:
Agree these are rather unusual features, but while I know why we introduced the multi-pool thing, I lack context for these other two. Maybe there's something I'm missing?

eugene-babichenko:
The test failures are older ones from `master` and are fixed in https://github.com/input-output-hk/jormungandr/pull/3559

mzabaluev:
Needs rebasing.

mzabaluev:
To reiterate why we want to reduce to one stake pool/ BFT leader per node:

## Reasons to do this

* Saner implementation of mempool logic, fragment log, metrics, other APIs.
* Remove any potential unfairness from running leaders in the same node, leaving a flat network organization (leaving only connectivity/latency issues).

Anything else?

## What are we losing

* The use case of running a shared jormungandr node for multiple stake pools/consensus leaders, potentially segregating their secrets in separate enclave services. This is easily replaced by hosting a node per each pool, for the small price of duplicated storage. In practice, I expect networks would prefer horizontally scaling their leader node instances as much as possible anyway.

doc/api/v0.yaml
```diff
@@ -335,97 +335,6 @@ paths:
...
-          description: Success
-          content:
-            application/json:
-              schema:
-                description: ID of the created leader
-                type: integer
-                minimum: 0
-              example: 5
-
   /api/v0/leaders/logs:
```
mzabaluev, line 338: Looks like this response payload has also changed, no `enclave_leader_id` any more?

jormungandr/src/fragment/pool.rs
```diff
@@ -35,7 +34,7 @@ const DEFAULT_BUF_SIZE: usize = 128 * 1024; // 128 KiB
 
 pub struct Pools {
```
mzabaluev, line 35: Rename to `Pool` for clarity?

jormungandr/src/settings/command_arguments.rs
```diff
@@ -27,7 +27,7 @@ pub struct StartArguments {
     /// Set the secret node config (in YAML format). Can be given
     /// multiple times.
```
mzabaluev, line 28: The doc needs to be updated.

mzabaluev (commented):
Looks good. Minor changes need to be made and the doc updated: it still claims that multiple --secret options can be given.

eugene-babichenko:
> What are we losing

We are also losing the capability to manage leader keys through the API. Although I think change of keys is a rare event and we do not benefit from having such functionality.

mzabaluev (approved):
Good to go after resolving the merge conflict.

PR #3557, opened 2021-08-30
update keys HRP - fund6

Zeegomo:


dkijania (approved):

danielSanchezQ (approved):

mzabaluev:
Hardcoding the curve in the HRP stopped making sense when the backend became selectable.
Can we still extract the curve ID from the backend and use that in the HRP for debug purposes, and making it easy to diagnose when a key generated for a different curve is used?

Zeegomo:
Failure is unrelated

jcli/src/jcli_lib/certificate/mod.rs
```diff
@@ -69,8 +69,8 @@ pub enum Error {
     KeyNotFound { index: usize },
     #[error("Invalid input, expected Signed Certificate or just Certificate")]
     ExpectedSignedOrNotCertificate,
-    #[error("Invalid data")]
-    InvalidBech32(#[from] bech32::Error),
+    #[error("Invalid bech32 data")]
```
mzabaluev, line 72: Maybe "invalid Bech32 address string" would be more descriptive.

jcli/src/jcli_lib/vote/mod.rs
```diff
@@ -17,8 +16,8 @@ pub enum Error {
     Hex(#[from] hex::FromHexError),
     #[error("base64 decode error")]
     Base64(#[from] base64::DecodeError),
-    #[error("bech32 decode error")]
-    Bech32(#[from] bech32::Error),
+    #[error("bech32 error")]
```
mzabaluev, line 19: "Bech32 format error"?

mzabaluev (approved):

jcli/src/jcli_lib/certificate/mod.rs
Zeegomo: It's not used only for keys though, certificates are bech32 encoded as well while they are stored locally

PR #3559, opened 2021-09-01
mempool: BlockDate for fragment expiry based on current time

eugene-babichenko:
A fix for this test failure.

https://github.com/input-output-hk/jormungandr/runs/3473145217

By producing the date from the current system time we are able to always check against the correct time.

### DIFF
--- a/jormungandr/src/fragment/pool.rs
+++ b/jormungandr/src/fragment/pool.rs
@@ -112,7 +112,8 @@ impl Pools {
 
         let mut fragments = fragments.into_iter();
 
-        let block_date = self.get_current_block_date().await;
+        let (block_date, max_expiry_epochs) =
+            self.get_current_block_date_and_max_expiry_epochs().await;
 
         for fragment in fragments.by_ref() {
             let id = fragment.id();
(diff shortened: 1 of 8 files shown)


jormungandr/src/fragment/pool.rs
```diff
@@ -305,6 +307,18 @@ impl Pools {
             block_date.into(),
         );
     }
+
+    async fn get_current_block_date(&self) -> BlockDate {
+        let time = std::time::SystemTime::now();
+        let tip = self.tip.get_ref().await;
```
mzabaluev, line 313: No way we can just use `tip.block_date() + 1` (with some complication as needed to regard for eras)?

mzabaluev (commented):
Looks good. Our block date calculation API is overcomplicated.

eugene-babichenko:
> Looks good. Our block date calculation API is overcomplicated.

Not really, considering the fact that epochs might have different parameters (e.g. slot duration and slot number).

jormungandr/src/fragment/pool.rs
eugene-babichenko: But what about possible missed slots?

xtask/src/install.rs
```diff
@@ -0,0 +1,6 @@
+use xshell::cmd;
+
+pub fn install() {
+    cmd!("cargo install --path jormungandr").run().unwrap();
+    cmd!("cargo install --path jcli").run().unwrap();
+}
```
mrzenioszeniou, line 6: Shouldn't this be in another pull request?

jormungandr/src/fragment/pool.rs
```diff
@@ -305,6 +307,22 @@ impl Pools {
...
+        let era = tip.epoch_leadership_schedule().era();
+        let epoch_position = tip
+            .time_frame()
+            .slot_at(&time)
+            .and_then(|slot| era.from_slot_to_era(slot))
+            .expect("the current time and blockchain state should produce a valid blockchain date");
+        let block_date: BlockDate = epoch_position.into();
+        BlockDate {
+            slot_id: block_date.slot_id + 1,
```
Zeegomo, line 322: I think it could happen that a transaction is received just at the slot start before block creation starts and is incorrectly mislabeled as expired when in theory it could still make its way into the block. In any case, I think this is unlikely and anyway hard to distinguish from practical limits in clocks synchronization among parties

xtask/src/install.rs
eugene-babichenko: uhm, how did it get there -_-
thanks

mrzenioszeniou (approved):

jormungandr/src/fragment/pool.rs
mzabaluev: Right, we can't use that.

mzabaluev (approved):

PR #3562, opened 2021-09-02
TTL Performance Tests

mrzenioszeniou:
* Added a benchmark test to measure the speed at which expired transactions are handled.
* Added a benchmark test to measure the speed at which transactions with long TTL are handled.
* Added a node test to ensure passive nodes remove expired transactions from their fragment pools.

### DIFF
--- a/testing/jormungandr-integration-tests/src/jormungandr/mempool.rs
+++ b/testing/jormungandr-integration-tests/src/jormungandr/mempool.rs
@@ -1,6 +1,7 @@
+use crate::common::jormungandr::StartupVerificationMode;
 use crate::common::jormungandr::{starter::Role, Starter};
 use crate::common::{jormungandr::ConfigurationBuilder, startup};
-use assert_fs::fixture::PathChild;
+use assert_fs::fixture::{PathChild, PathCreateDir};
 use assert_fs::TempDir;
 use chain_impl_mockchain::fee::LinearFee;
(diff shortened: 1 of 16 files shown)


testing/jormungandr-integration-tests/src/jormungandr/mempool.rs
```diff
@@ -438,3 +439,71 @@ pub fn expired_fragment_should_be_rejected() {
...
+#[test]
+/// Verifies that a passive node will reject a fragment that has expired, even after it's been
+/// accepted in its mempool.
+fn expired_fragment_should_be_rejected_by_passive_node() {
+    const N_FRAGMENTS: u32 = 10;
+
+    let receiver = startup::create_new_account_address();
+    let mut sender = startup::create_new_account_address();
+
+    let (leader, _) = startup::start_stake_pool(
```
dkijania, line 452: One remark. I see that you are using GenesisPraos consensus (start_stake_pool). Basically we support also BFT, if you change it to bft consensum mayb we would have better matrxi coverage of passive/leader vs bft/GPraos coverage

dkijania:
Apart from my single comment above looks godd.

testing/jormungandr-integration-tests/src/jormungandr/mempool.rs
mrzenioszeniou: Should we add one test with Praos + Leader and one with BFT + passive? Or 4 tests to cover all pairs?

PR #3569, opened 2021-09-06
Added transaction validity to mjolnir arguments

mrzenioszeniou:
* Added a `BlockDate` argument to transaction-related commands.
* Fixed some `structopt` flag conflicts.
* Did some module cleanup.


testing/mjolnir/src/mjolnir_app/fragment/batch/adversary/all.rs
```diff
@@ -51,6 +51,10 @@ pub struct AdversaryAll {
 
     #[structopt(long = "spending-counter", short = "s")]
     faucet_spending_counter: u32,
+
+    /// Transaction validity deadline (inclusive)
+    #[structopt(long, short, default_value = "1.0")]
```
dkijania, line 56: I have a concern what would happen if our tests last longer than, for example 10 epochs. We would need to provide such block date which will fulfill requirements for valid until date for all transactions (sent from 0.0 date to 10.0), as a consequence we would need to set `tx_max_exipry_epochs` to at least 10 (in our example).  So in this case we would be forced to have some unrealistic parameters in blockchain exclusively for perf/load tests. I wonder if we could use methods from chain-wallet-libs which can calculate what should be expiry date based on blockchain settings  and current time. Knowing that, Client of mjolnir can set expiry duration in slot units (for example --valid-until-duration 5) in cmd input. I'm interested in your opinion. You can check this code for usage of chain-wallet-libs methods*.

```
https://github.com/input-output-hk/vit-testing/blob/updates-to-vitup/iapyx/src/utils/valid_until.rs#L11
```

* - code is still under dev, but you can get general idea. In order to avoid duplication, If you agree that this is the way to go we can put code in jormungandr and use it in vit-testing

mrzenioszeniou: Unless I'm mistaken, what you are suggesting eliminates the possibility to send transactions that are expired or violate `max_expiry_epochs`. So if we don't care about creating such loads, defining the TTL instead of the expiration slot makes more sense.

dkijania: Good point! But after second thought not necessarily will eliminate possibility to send adversary transaction. One can still set some incorrect values, so transaction would be already expired or violate `max_expiry_epochs`. If we set expiry date slot shift to a number which is bigger than `slots_in_epochs`  then all of fragments would be rejected. We can think of adding possibility to provide negative slot shift, so transaction expiry date would be 'less' than current. So we can still generate invalid load (and this is our purpose) by manipulating slot shift and this load will be **alwasy** adversary. If we would use static block date for entire load, then for sure some of the fragments would be valid (for example if we set `valid_until` to 2.0 and run fragments from 0.0 to 10.0 block date, then all fragments sent from `1.0` till `1.{slots_in_epochs}` would be valid) and this is not our goal

mrzenioszeniou: Alright then, how about we use a `--ttl [-]<epoch>.<slot>` argument so we can create `BlockDate`s in the past and the future?

mrzenioszeniou: Okay, so we have two new arguments:
* `valid_until` which will lead to all transactions being sent with a fixed expiration date.
* `ttl` which will lead to all transactions having a fixed time to live depending on the time they're sent. The TTL can be negative if `~` is prepended to the date.

They are mutually exclusive so only one of them can be defined.

testing/jormungandr-integration-tests/src/jormungandr/rest/v1/fail_fast.rs
```diff
@@ -71,7 +71,7 @@ fn world() -> (
     let faulty_tx_builder = FaultyTransactionBuilder::new(
         jormungandr.genesis_block_hash(),
         jormungandr.fees(),
-        BlockDate::first().next_epoch(),
+        BlockDateGenerator::Fixed(BlockDate::first().next_epoch()),
```
mzabaluev, line 74: `BlockDateGenerator` could have a `From<BlockDate>` impl creating the `Fixed` variant, so that this could be written more concisely:

`BlockDate::first().next_epoch().into()`

mzabaluev (approved):
Looks good, but see the note about convenience API below.

testing/jormungandr-integration-tests/src/jormungandr/rest/v1/fail_fast.rs
mrzenioszeniou: Done 👌

dkijania (approved):

dkijania:
Good job!

PR #3578, opened 2021-09-09
Add jcli command to query the proposals a user has voted for in a voteplan

Zeegomo:


jcli/src/jcli_lib/rest/v1/vote/mod.rs
```diff
@@ -0,0 +1,58 @@
...
+    output_format: OutputFormat,
+
+    /// Account address to filter votes
+    #[structopt(short, long)]
+    account: String,
+
+    /// Id of the voteplan for which we want to list proposals
+    /// the account voted for
+    #[structopt(short, long)]
+    voteplan_id: String,
```
mzabaluev, line 21: The name for this option is `--vote-plan-id` [elsewhere](https://github.com/input-output-hk/jormungandr/blob/99fa3c7d8f1120556916cbe55dcc94574fac04ec/jcli/src/jcli_lib/certificate/new_vote_cast.rs#L15-L15), let's make it consistent.

jcli/src/jcli_lib/rest/v1/vote/mod.rs
```diff
@@ -0,0 +1,58 @@
...
+pub struct AccountVotes {
+    #[structopt(flatten)]
+    args: RestArgs,
+
+    #[structopt(flatten)]
+    output_format: OutputFormat,
+
+    /// Account address to filter votes
+    #[structopt(short, long)]
+    account: String,
```
mzabaluev, line 16: This has been a positional parameter in other commands, e.g. `v0 account get`. IDK if this should be a consistent pattern, though.

Zeegomo: I'm not sure, in this case we have two domain-specific arguments and I don't think there's any predefined order. In that sense named parameters gives a bit more flexibility to the user.  In any case, I don't have a strong opinion, I guess either one is fine

mzabaluev (approved):

PR #3579, opened 2021-09-09
Remove CircleCI

Zeegomo:
It stopped working some time ago and a quick look was not
sufficient to identify the root cause.
However, I don't think it has a lot of added value since we run
those tests on GitHub Actions as well and we wait for them anyway.


mzabaluev:
I'll do some investigation in #3580 and then we can resolve this one way or the other.

mzabaluev:
I've fixed the failure in #3580. It was not a waste of time because we have the same issue in GitHub, though it does not yet result in outright failures (probably because a stale index repo is dragged around the cache snapshots).

mzabaluev:
> we have the same issue in GitHub

Correction: we don't. GitHub caching is more crude, but in exchange the steps do not use local repository data.

PR #3590, opened 2021-09-13
Fix dead_code lints by removing unused structs and data members

mzabaluev:
Nightly lints began to catch it.

Some of the lints are overridden because the members are used by the `Debug` impl or potentially useful to jormungandr-testing-utils users.


danielSanchezQ (approved):

mzabaluev:
In fact, what is the purpose of the `StakePoolInfo` struct in the first place? I don't see it used anywhere.

PR #3591, opened 2021-09-13
metrics: less strict atomic ordering requirements

eugene-babichenko:


jormungandr/src/metrics/backends/simple_counter.rs
```diff
@@ -101,45 +101,46 @@ impl Default for SimpleCounter {
...
     }
 
     fn add_peer_available_cnt(&self, count: usize) {
-        self.peers_available_cnt.fetch_add(count, Ordering::SeqCst);
+        self.peers_available_cnt.fetch_add(count, Ordering::Relaxed);
     }
 
     fn sub_peer_available_cnt(&self, count: usize) {
-        self.peers_available_cnt.fetch_sub(count, Ordering::SeqCst);
+        self.peers_available_cnt.fetch_sub(count, Ordering::Relaxed);
```
enzoc4, line 138: I think this could potentially overflow and wrap around, if the `fetch_sub` gets executed before the `fetch_add`, although I guess it would be a strange cornercase...

I'm not that fresh on memory ordering details, though.

jormungandr/src/metrics/backends/simple_counter.rs
```diff
@@ -101,45 +101,46 @@ impl Default for SimpleCounter {
 
 impl MetricsBackend for SimpleCounter {
     fn add_tx_recv_cnt(&self, count: usize) {
-        self.tx_recv_cnt.fetch_add(count, Ordering::SeqCst);
+        self.tx_recv_cnt.fetch_add(count, Ordering::Relaxed);
```
mzabaluev, line 104: Doing this for `load`/`store` (i.e. values updated from another reliably source) is all right, but could `fetch_add` inaccurately increment the count in this mode?

eugene-babichenko: Hm, could you give an example? I don't really understand how this can happen.

jormungandr/src/metrics/backends/simple_counter.rs
eugene-babichenko: Well, I think that is technically possible even with `SeqCst`/`AcqRel` if we somehow manage to try to remove a peer that is not yet counted, there should be more sophisticated synchronization. I think, the Prometheus integration works under the same assumption since under the hood their counters are just atomics with all operations done with the `Relaxed` ordering.

jormungandr/src/metrics/backends/simple_counter.rs
mzabaluev: If two threads do this concurrently, both of them can load the same value and then store the value + 1. This needs to order the load half after the store half of any other concurrent increment. I'd use `Ordering::AcqRel` (we don't care about an absolute ordering of the increments, so `SeqCst` is too heavy indeed). I mean, we're splitting hairs here, but let's split them correctly :trollface:

jormungandr/src/metrics/backends/simple_counter.rs
mzabaluev: These operations are documented to wrap around, but the ordering needs to be `AcqRel` I think.

jormungandr/src/metrics/backends/simple_counter.rs
mzabaluev: I stand corrected by [the nomicon](https://doc.rust-lang.org/nomicon/atomics.html#relaxed).

mzabaluev (approved):

PR #3593, opened 2021-09-13
Concurrent tip update

Zeegomo:
### Problem
Concurrent updates to the tip result in invalid results and data races. See node logs in https://app.circleci.com/pipelines/github/input-output-hk/jormungandr/11546/workflows/1a631406-e672-4344-bdcc-b1dee3962d95/jobs/43478/steps?invite=true#step-110-2026

### Proposed solution
This PR introduces a new component, `TipUpdater`, which is the point of serialization for all tip update requests and is responsible for keeping it up to date (e.g. with periodic checks).
This is done to avoid locking the `Tip` (which is shared and used by pretty much all other modules) or the blockchain process with a mutex for the duration of the selection algorithm.
OTOH, the introduction of a whole new part can be seen as an over complication, but I I think this helps with separating functions (and thus readability) even more.

This has the side effect that it may happen that the tip is not updated right after a new block is received by the blockchain process, but I don't  see this as a problem since there are already no synchronization guarantees when passing messages between tasks.

### Other changes
As a result of the introduction of the `TipUpdater`, and as a way to hide unnecessary information and prevent misuse, the tip cannot be updated or created manually anymore outside of the blockchain module.

Since it needed to be updated anyway following the aforementioned changes, the part of the bootstrap process that was dealing with internals of the blockchain has been moved to the blockchain module from the network one, to better separate functionalities and reuse code. In this regard, the code handling addition of new blocks to the chain from storage and network has been unified, even though doing so made loading from storage a bit less efficient.

### Bug fixes?
I've also fixed some of what I think are bugs in the code:
* `process_network_blocks` always setting the explorer tip to the latest received block https://github.com/input-output-hk/jormungandr/blob/a389bb69b7a97ee27b80e38f18ccfd542d1e7094/jormungandr/src/blockchain/process.rs#L611
* explorer assuming the block applied is the tip https://github.com/input-output-hk/jormungandr/blob/a389bb69b7a97ee27b80e38f18ccfd542d1e7094/jormungandr/src/explorer/mod.rs#L704
* `process_new_ref` not actually creating a new branch for non-optimal forks https://github.com/input-output-hk/jormungandr/blob/a389bb69b7a97ee27b80e38f18ccfd542d1e7094/jormungandr/src/blockchain/process.rs#L343. I traced back this as far as I could but as far as I can tell we never created a branch here and I don't understand why. If `Branches` should contains all of the forks in the chain then I think this step was missing, as this is the only place where we create branches. OTOH, I think the current process will duplicate branches every time we reprocess the tip, as `.apply_or_create` does not check for already existing branches (but can be amended rather easily). I would like some feedback on this one

I've tried to keep this as small as possible, but looks like I've failed :sweat_smile:

### DIFF
--- a/jormungandr/src/blockchain/bootstrap.rs
+++ b/jormungandr/src/blockchain/bootstrap.rs
@@ -1,0 +1,209 @@
+use super::tip::TipUpdater;
+use crate::blockcfg::{Block, HeaderDesc, HeaderHash};
+use crate::blockchain::{self, Blockchain, PreCheckedHeader, Ref, Tip};
+use crate::metrics::Metrics;
+use chain_core::property::{Deserialize, HasHeader};
+use chain_network::data as net_data;
+use chain_network::error::Error as NetworkError;
+use futures::{prelude::*, stream, task::Poll};
+use tokio_util::sync::CancellationToken;
+
+use std::pin::Pin;
+use std::sync::Arc;
+
+#[derive(Debug, thiserror::Error)]
+pub enum Error {
+    #[error(transparent)]
+    BlockchainError(#[from] super::Error),
+    #[error(
+        "received block {0} is already present, but does not descend from any of the checkpoints"
+    )]
+    BlockNotOnBranch(HeaderHash),
+    #[error("received block {0} is not connected to the block chain")]
+    BlockMissingParent(HeaderHash),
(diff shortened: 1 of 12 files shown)


Zeegomo:
superseded by https://github.com/input-output-hk/jormungandr/pull/3596

PR #3595, opened 2021-09-14
use rest for remaining rewards in tests

enzoc4:
Same reasoning than #3322 

This is the more straight forward way of doing this that I could think of, but I'm open to alternatives. 

There are a couple of tests for the reward pot increase that I think we introduced at the end of the ITN. Those use the explorer right now, so this replaces it by a _new_ rest endpoint, as I didn't find a way to do it with the current ones.

### DIFF
--- a/doc/api/v0.yaml
+++ b/doc/api/v0.yaml
@@ -1159,7 +1159,7 @@ paths:
                       },
                     ]
 
-  '/api/v0/rewards/remaining/':
+  '/api/v0/rewards/remaining':
     get:
       description: 'returns the remaining funds to be distributed as rewards'
       operationId: RewardsRemaining
--- a/doc/api/v0.yaml
+++ b/doc/api/v0.yaml
@@ -1159,6 +1159,25 @@ paths:
                       },
                     ]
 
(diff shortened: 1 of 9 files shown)


mzabaluev (commented):
looks good. Document the added API in `doc/api`, please?

enzoc4:
> looks good. Document the added API in `doc/api`, please?

good point, I'll do. I was also wondering if I should add a `jcli` wrapper for it... For completeness at least.

danielSanchezQ:
> > looks good. Document the added API in `doc/api`, please?
> 
> good point, I'll do. I was also wondering if I should add a `jcli` wrapper for it... For completeness at least.

It can be part of another pr imo

enzoc4:
> > I was also wondering if I should add a `jcli` wrapper for it... For completeness at least.
> 
> It can be part of another pr imo

yeah, sure, I don't think it's necessary though, I'm not sure what's our policy with respect to the `jcli rest` subcommand.

Zeegomo (approved):
Good to merge after the conflict is addressed

PR #3599, opened 2021-09-15
[Tests] Remove waits

Zeegomo:
There were a bunch of unnecessary waits in `expired_fragment_should_be_rejected_by_{passive_bft, leader_praos}_node` which, along with a non-ideal implementation and slow runners, were causing the test to fail on CircleCI. This hopefully fixes it


mrzenioszeniou (approved):
Thanks for cleaning up after me :)

PR #3600, opened 2021-09-16
enable all-features flag when building tests

dkijania:
We have many test features (load/soak etc.) which often failed to compile after some refactoring. This PR is an attempt to fix this problem by enabling `all-features` when building tests

PR #3603, opened 2021-09-17
[Tests] increase default slot duration in tests to 2 secs

Zeegomo:
CircleCI is a bit slow and this can cause forking inside a single node.
Let's try to increase the default slot time to 2s and hope this fixes the problem without increasing test times too much


dkijania:
Looking good. Sorry for explicitly setting slot_duration in so many tests...

Zeegomo:
Mmm, this is now too slow for GitHub. Maybe we can indeed put writing to storage in a separate background task

dkijania:
In my PR https://github.com/input-output-hk/jormungandr/pull/3600 I've extended timeout for github action tests. Maybe it's acceptable solution?

mzabaluev (approved):
Looks good to me, especially as it deduplicates TTL setup in most cases.

eugene-babichenko (approved):

PR #3604, opened 2021-09-17
add split limit for explorer load generator

dkijania:
We experienced some issues in `explorer_load_test` in nightly tests . When running test like below:
`cargo test --all-features explorer_load_test --release`
It causes:
```
thread '12' has overflowed its stack
fatal runtime error: stack overflow
```

I've try to debug it using gdb and got another:
```
Thread 65 "58" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7fff77bfd700 (LWP 9734)]
0x0000555555753737 in rayon::iter::plumbing::bridge_unindexed_producer_consumer ()
```

Solution proposed by @Zeegomo was to limit number of splits for Explorer load generator

### DIFF
--- a/testing/jormungandr-testing-utils/src/testing/node/explorer/load/mod.rs
+++ b/testing/jormungandr-testing-utils/src/testing/node/explorer/load/mod.rs
@@ -7,12 +7,15 @@ use rand::RngCore;
 use rand_core::OsRng;
 use std::time::Instant;
 
+const DEFAULT_MAX_SPLITS: usize = 7; // equals to 128 splits, will likely not reach that value but it's there just to prevent a stack overflow
+
 #[derive(Clone)]
 pub struct ExplorerRequestGen {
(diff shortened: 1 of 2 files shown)


testing/jormungandr-testing-utils/src/testing/node/explorer/load/mod.rs
```diff
@@ -153,7 +157,13 @@ impl RequestGenerator for ExplorerRequestGen {
         })
     }
 
-    fn split(self) -> (Self, Option<Self>) {
+    fn split(mut self) -> (Self, Option<Self>) {
+        // Since no transaction will ever be accepted we could split as many times as we want
```
Zeegomo, line 161: This comment is not correct in this context. In this case the explorer queries do not modify the node state and can be run concurrently without problems

Zeegomo (approved):

PR #3606, opened 2021-09-19
Applied fixes for other nightly tests.

dkijania:
Fixed more issues with nightly tests as original connected with explorer load test revealed further issues.

- removed unused method in jcli test service 
- fix soak explorer test by adding expiry date to explorer soak test
- fix rest load tests by introducing split limit for RestRequestGen

### DIFF
--- a/testing/jormungandr-testing-utils/src/testing/node/rest/load.rs
+++ b/testing/jormungandr-testing-utils/src/testing/node/rest/load.rs
@@ -4,12 +4,15 @@ use rand::RngCore;
 use rand_core::OsRng;
 use std::time::Instant;
 
+const DEFAULT_MAX_SPLITS: usize = 7; // equals to 128 splits, will likely not reach that value but it's there just to prevent a stack overflow
+
 #[derive(Clone)]
 pub struct RestRequestGen {
     rest_client: JormungandrRest,
     rand: OsRng,
     addresses: Vec<String>,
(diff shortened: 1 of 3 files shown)


testing/jormungandr-testing-utils/src/testing/node/rest/load.rs
```diff
@@ -112,7 +116,13 @@ impl RequestGenerator for RestRequestGen {
         })
     }
 
-    fn split(self) -> (Self, Option<Self>) {
+    fn split(mut self) -> (Self, Option<Self>) {
+        // Since rest queries do not modify the node state we can split as many times as we want
+        // but that may trigger a bug in rayon so we artificially limit it
```
mzabaluev, line 121: Is there a benefit of using rayon at all?

mzabaluev (approved):
Looks OK.
It's likely very artificial that the load generator gives equal probability to each request type, but it's not the subject of this fix. Consider using a tweakable probability map per request.

testing/jormungandr-testing-utils/src/testing/node/rest/load.rs
dkijania: Maybe @Zeegomo can anwser that. From my perspective it increased load tool transaction volume greatly, but don't know if that was by using framework or fixing my code issues

PR #3607, opened 2021-09-20
Added CI job to check CHANGELOG.md for updates

mrzenioszeniou:


.github/workflows/changelog.yml
```diff
@@ -0,0 +1,16 @@
...
+  pull_request:
+
+jobs:
+  diff_changelog:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - run: git fetch origin master
+      - run: |
+          if [[ ! `git diff --name-only origin/master | grep CHANGELOG.md` ]]
```
eugene-babichenko, line 13:
```suggestion
          if [[ `git diff --exit-code origin/master -- CHANGELOG.md` ]]
```

> --exit-code
>
>    Make the program exit with codes similar to diff(1). That is, it exits with 1 if there were differences and 0 means no differences.

danielSanchezQ:
@input-output-hk/jormungandr are we sure we really want this?
I'm not sure updating the changelog per PR is a good idea. Imo it should change per version or release, even per feature. But sometimes a bunch of PRs are part of a single feature (which could/should be logged into the changelog itself). 
But maybe fixes, some maintenance changes or other stuff will just polluted the file and will make it useless in the end?
It is ok to keep it as a warning and remainder to update it. But I don't think it should drive us to update every change.

eugene-babichenko:
That is why this is not a hard failure but just a warning.

mrzenioszeniou:
@danielSanchezQ

> ​I'm not sure updating the changelog per PR is a good idea.

As @eugene-babichenko mentioned this PR adds an annotation, not a job that fails if the changelog isn't updated. We are all still supposed to use our judgment and update the changelog only when it makes sense.

> Imo it should change per version or release, even per feature.

I'd expect whoever makes a release to go over the items in the changelog and curate them appropriately. Entries may be added/removed/merged/edited to ensure we have neat release notes. Also, I'd trust the author(s) of each feature to do a better job at describing it, over the one preparing the release some weeks down the road.

In any case, this is just to nudge us into actually updating the damn thing once in a while :)

.github/workflows/changelog.yml
mrzenioszeniou: Nice, thanks. Kept the `!` though :)

mzabaluev (approved):
Good, a self-demonstrating pull request!

eugene-babichenko (approved):

PR #3610, opened 2021-09-20
metrics: votesCast

eugene-babichenko:


danielSanchezQ (approved):

jormungandr/src/metrics/backends/prometheus_exporter.rs
```diff
@@ -248,7 +255,10 @@ impl MetricsBackend for Prometheus {
                     Fragment::PoolRetirement(tx) => totals(tx),
                     Fragment::PoolUpdate(tx) => totals(tx),
                     Fragment::VotePlan(tx) => totals(tx),
-                    Fragment::VoteCast(tx) => totals(tx),
+                    Fragment::VoteCast(tx) => {
```
Zeegomo, line 258: What if:
* there's a fork and the node switches branches
* the node is 2 (or more) blocks behind the current tip and receive the updated tip. Those intermediate blocks will never be set as the tip, and their votes will not be counted

PR #3614, opened 2021-09-21
Remove qr code module

dkijania:
Removing qr_code module from jormungandr_testing_utils project since it is not  related to a jormungandr but rather to the catalyst project. Currently we have catalyst-toolbox which has similar function. I think the best idea is to remove qr_code module and enhance kedqr module. Please see related pr:

https://github.com/input-output-hk/catalyst-toolbox/pull/57

Also removing some leftovers like startup::sleep_* methods which now are obsolete.


jormungandr/src/secure/mod.rs
```diff
@@ -29,7 +29,9 @@ pub struct GenesisPraos {
 ///
 #[derive(Clone, Deserialize)]
 pub struct GenesisPraosPublic {
+    #[allow(dead_code)]
```
danielSanchezQ, line 32: Why is this dead code now?

danielSanchezQ (commented):
Lgtm. Just concern about the deadcode tags.

jormungandr/src/secure/mod.rs
dkijania: I was too afraid to remove it :) But then noticed that there is a PR on master which is removing them. so I'll do the same

mzabaluev (changes requested):
Please split the dead_code changes from what this PR is actually supposed to achieve, it's hard to review everything together.

Lint fixes should go in a separate PR like mentioned in the discusssion.

dkijania:
@mzabaluev is it ok now?

mzabaluev (approved):
Looks good. Let's address the lints in a dedicated PR.

PR #3616, opened 2021-09-23
Fix lints on catalyst fund6

dkijania:
Remove unused fields in jormungandr


mzabaluev (approved):
OK, even though it does not remove all lints.

mzabaluev:
Could help being rebased to the latest master. CircleCI failures were fixed some time ago.

mzabaluev:
The CircleCI failure has been fixed in master, but not backported here -> OK to ignore.

PR #3618, opened 2021-09-24
Update chain libs

Zeegomo:


jormungandr-lib/src/crypto/key.rs
```diff
@@ -40,14 +40,34 @@ pub struct Identifier<A: AsymmetricPublicKey>(
...
+    A: AsymmetricKey,
+    SecretKey<A>: Bech32,
+{
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        internal::deserialize_secret(deserializer).map(Self)
+    }
+}
```
danielSanchezQ, line 70: What is the improve of this change?

Zeegomo: Not all `SecretKey<A: AsymmetricKey>` can be serialized with `Bech32` now that this is restricted to fixed size data. I did not want to restrict the usage of `SigningKey` to just the keys that can be serialized / deserialized, so I added the bounds separately.

danielSanchezQ: Aha, I see, didn't notice the bounds though. Nice, thanks!

danielSanchezQ (approved):

PR #3619, opened 2021-09-24
refactor selection function to be tracing-friendly

Zeegomo:
The span instrumentation for the selection algorithm contained an error as it used `span.enter` and held the guard across `.await` points. To make things prettier (and hopefully easier to read) I've refactored the inner part in an function and instrumented that one without manually entering the span.

Also add a debug span which displays the hash of the fragment that is being processed


Zeegomo:
CircleCi failure is unrelated

eugene-babichenko:
Something looks off. This PR has the same set of failures on all platforms in Actions.

eugene-babichenko (approved):

PR #3626, opened 2021-09-28
Check the received block is the requested one

Zeegomo:
During the bootstrap phase, check that the block fetched via gRPC is indeed the one that was requested

### DIFF
--- a/jormungandr/src/network/grpc/client.rs
+++ b/jormungandr/src/network/grpc/client.rs
@@ -28,6 +28,11 @@ pub enum FetchBlockError {
     GetBlocksStream { source: net_error::Error },
     #[error("no blocks received")]
     NoBlocks,
+    #[error("Unexpected block hash: requested {requested} received {received}")]
+    UnexpectedBlock {
+        requested: HeaderHash,
+        received: HeaderHash,
+    },
 }
 
 pub type ConnectError = transport::Error;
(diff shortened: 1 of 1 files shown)


jormungandr/src/network/grpc/client.rs
```diff
@@ -28,6 +28,11 @@ pub enum FetchBlockError {
     GetBlocksStream { source: net_error::Error },
     #[error("no blocks received")]
     NoBlocks,
+    #[error("Invalid block hash: requested {requested} received {received}")]
+    InvalidBlock {
```
eugene-babichenko, line 32: I'd rather call it `UnexpectedBlock`.

jormungandr/src/network/grpc/client.rs
```diff
@@ -71,7 +76,15 @@ pub async fn fetch_block(peer: &Peer, hash: HeaderHash) -> Result<Block, FetchBl
             let block = block
                 .decode()
                 .map_err(|e| FetchBlockError::GetBlocksStream { source: e })?;
-            Ok(block)
+
+            if block.header.id() == hash {
+                Ok(block)
+            } else {
+                Err(FetchBlockError::InvalidBlock {
```
danielSanchezQ, line 83: `UnexpectedBlock`?

Zeegomo: oops, too fast on that one :(

eugene-babichenko (approved):

PR #3662, opened 2021-10-20
watch client implementation

enzoc4:
implements https://github.com/input-output-hk/chain-libs/pull/652

the sync_multiverse algorithm may need some tweaks still.

also, I haven't tested this much, and only on a leader node.

besides that, the places where the `NewBlock` and `NewTip` are propagated basically duplicate the ones that the explorer currently uses. I'd prefer to remove them in a different PR, but I could do it here.

### DIFF
--- a/jormungandr/src/watch_client.rs
+++ b/jormungandr/src/watch_client.rs
@@ -241,11 +241,23 @@ async fn handle_sync_multiverse(
 ) -> Result<(), intercom::Error> {
     let mut checkpoints = checkpoints
         .iter()
-        .map(|id| HeaderHash::deserialize(id.as_bytes()).map_err(intercom::Error::invalid_argument))
+        .map(|id| {
+            let deser = HeaderHash::deserialize(id.as_bytes())
+                .map_err(intercom::Error::invalid_argument)?;
+
+            let chain_length = storage
+                .get_chain_length(deser)
+                .ok_or_else(|| intercom::Error::invalid_argument("unknown checkpoint"))?;
+
+            Ok::<_, intercom::Error>((chain_length, deser))
+        })
         .collect::<Result<Vec<_>, _>>()?;
 
     let block0 = blockchain.block0();
 
+    // we are adding 1 to the lsb later, because it is present then we don't need to send it.
+    // but if the checkpoints are empty then it means the last stable block is the block0, and we
+    // send it here.
     if checkpoints.is_empty() {
         let block = storage
             .get(*block0)
@@ -261,12 +273,15 @@ async fn handle_sync_multiverse(
         .map_err(intercom::Error::failed)?;
     }
 
-    // assume the first checkpoint is the last_stable_block (for now, the docs don't require this)
-    let lsb = checkpoints.pop().unwrap_or(*block0);
+    checkpoints.sort_unstable();
+
+    // TODO: not sure if this is always true
+    const BLOCK0_CHAIN_LENGTH: u32 = 0;
+    let (lsb_chain_length, lsb) = checkpoints.pop().unwrap_or((BLOCK0_CHAIN_LENGTH, *block0));
 
     let mut known_unstable_blocks_by_client = HashSet::new();
 
-    for checkpoint in checkpoints {
+    for (_, checkpoint) in checkpoints {
         if !storage.is_ancestor(lsb, checkpoint) {
             return Err(intercom::Error::invalid_argument(
                 "invalid from/checkpoints",
@@ -285,7 +300,7 @@ async fn handle_sync_multiverse(
         }
     }
 
-    let mut current_length = storage.get_chain_length(lsb).unwrap() + 1;
+    let mut current_length = lsb_chain_length + 1;
 
     loop {
         let blocks = storage
--- a/jormungandr/src/watch_client.rs
+++ b/jormungandr/src/watch_client.rs
@@ -239,23 +239,45 @@ async fn handle_sync_multiverse(
     storage: &Storage,
     sink: &mut intercom::ReplyStreamSink<Block>,
 ) -> Result<(), intercom::Error> {
-    let mut checkpoints = checkpoints
-        .iter()
-        .map(|id| {
-            let deser = HeaderHash::deserialize(id.as_bytes())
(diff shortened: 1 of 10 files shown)


jormungandr/src/watch_client.rs
```diff
@@ -0,0 +1,329 @@
...
+    for (_, checkpoint) in checkpoints {
+        if !storage.is_ancestor(lsb, checkpoint) {
+            return Err(intercom::Error::invalid_argument(
+                "invalid from/checkpoints",
+            ));
+        }
+
+        let mut current = checkpoint;
+
+        while current != lsb {
```
Zeegomo, line 293: We should stop early if two branches converge sooner than the lsb

jormungandr/src/watch_client.rs
```diff
@@ -0,0 +1,329 @@
...
+            block.serialize_as_vec().unwrap(),
+        )))
+        .await
+        .map_err(intercom::Error::failed)?;
+    }
+
+    checkpoints.sort_unstable();
+
+    // TODO: not sure if this is always true
+    const BLOCK0_CHAIN_LENGTH: u32 = 0;
```
Zeegomo, line 279: I think not having this a convention will create a lot of confusion

jormungandr/src/watch_client.rs
```diff
@@ -0,0 +1,329 @@
...
+    let (lsb_chain_length, lsb) = checkpoints.pop().unwrap_or((BLOCK0_CHAIN_LENGTH, *block0));
+
+    let mut known_unstable_blocks_by_client = HashSet::new();
+
+    for (_, checkpoint) in checkpoints {
+        if !storage.is_ancestor(lsb, checkpoint) {
```
Zeegomo, line 285: This actually traverses the whole tree. We can check this in the loop below (by looking at chain length of ancestors and lsb) and save half of database accesses

jormungandr/src/watch_client.rs
```diff
@@ -0,0 +1,329 @@
...
+            let deser = HeaderHash::deserialize(id.as_bytes())
+                .map_err(intercom::Error::invalid_argument)?;
+
+            let chain_length = storage
+                .get_chain_length(deser)
+                .ok_or_else(|| intercom::Error::invalid_argument("unknown checkpoint"))?;
+
+            Ok::<_, intercom::Error>((chain_length, deser))
+        })
+        .collect::<Result<Vec<_>, _>>()?;
```
Zeegomo, line 254: I think this will actually result in a an error if a checkpoint is unknown to the node, which is not what we want. Probably `filter_map` is what we should be using here

jormungandr/src/watch_client.rs
```diff
@@ -0,0 +1,329 @@
...
+        )))
+        .await
+        .map_err(intercom::Error::failed)?;
+    }
+
+    checkpoints.sort_unstable();
+
+    // TODO: not sure if this is always true
+    const BLOCK0_CHAIN_LENGTH: u32 = 0;
+    let (lsb_chain_length, lsb) = checkpoints.pop().unwrap_or((BLOCK0_CHAIN_LENGTH, *block0));
```
Zeegomo, line 280: `pop` works the opposite way as you would like here AFAIU. It will remove the last block, which is the one furthest down the chain, not the lsb

jormungandr/src/watch_client.rs
```diff
@@ -0,0 +1,307 @@
...
+        .map(|id| HeaderHash::deserialize(id.as_bytes()).map_err(intercom::Error::invalid_argument))
+        .collect::<Result<HashSet<_>, _>>()?;
+
+    let branches = blockchain.branches().branches().await;
+    let mut previous_branch = None;
+    let block0 = blockchain.block0();
+    for next_branch in branches {
+        let head_id = next_branch.header().id();
+
+        if let Some(prev) = previous_branch.replace(head_id) {
```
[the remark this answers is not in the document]
Zeegomo: Not too convinced this is the most optimal implementation, but it does not really matter much, as I don't expected this endpoint to be called that often, and I think we've already spent a considerable amount of time here that could be better spent elsewhere.
Anyway, this should be good enough at least for the first implementation

jormungandr/src/watch_client.rs
enzoc4: yeah, sorry, my mind had a glitch.

I had a vague idea that the block0 parent id was configurable somewhere in order for snapshots to refer to the time they were done, but it wouldn't really make sense for the chain length to not be 0.

jormungandr/src/watch_client.rs
enzoc4: Hmm, that's true, I thought it was implemented differently... It will only happen if there are forks, but good point.

jormungandr/src/watch_client.rs
```diff
@@ -0,0 +1,359 @@
...
+                    .for_each(|msg| async {
+                        match msg {
+                            RequestMsg::SyncMultiverse { from, handle } => {
+                                let mut sink = handle.start_sending();
+
+                                if let Err(e) =
+                                    handle_sync_multiverse(from, &blockchain, &storage, &mut sink)
+                                        .await
+                                {
+                                    let _ = sink.feed(Err(e)).await;
```
Zeegomo, line 73: Maybe add some log here if it's not already covered in the funciton

jormungandr/src/watch_client.rs
```diff
@@ -0,0 +1,359 @@
...
+            .and_then(|maybe_block0| {
+                maybe_block0.ok_or_else(|| intercom::Error::failed("block0 not found in storage"))
+            })?;
+
+        sink.send(Ok(chain_network::data::Block::from_bytes(
+            block.serialize_as_vec().unwrap(),
+        )))
+        .await
+        .map_err(intercom::Error::failed)?;
+    }
```
Zeegomo, line 296: nit: you aren't actually accessing `lsb_id`/`lsb_length` here, so some of the logic could be simplified to just look at `Some`/`None`

Zeegomo (approved):
Let's go ahead with this for now, we can iterate from here if needed

PR #3676, opened 2021-11-02
Update chain-libs for parallel spending

mzabaluev:
Initial adaptation for chain-impl-mockchain API changes for independent
spending counters lanes.
Also tonic and prost versions were updated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@
 
 - fix incorrect keys bech32 HRP by always using the ones provided by the library
 - update REST API: add new endpoint AccountVotes (/api/v1/votes/plan/account-votes/{account_id})
+- Support parallel lanes in spending counters on account outputs. This allows
+  submitting transactions that can spend from the same account without
+  requiring any particular order between the transactions.
 
 ## Release 0.13.0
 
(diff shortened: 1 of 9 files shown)


Zeegomo (commented):
I think we should update the changelog here, apart from that LGTM.

There is no way to use lanes other than 0 with jcli, which is not a big issue actually. We should update testing functions at least to be able to stress other lanes, but it can be done in another PR.

Also, I've just realized that this change adds explicit spending counter in the transaction. There were some proposals for improvements if this was known, with the simplest one being better diagnostic errors.

mzabaluev:
> There is no way to use lanes other than 0 with jcli, which is not a big issue actually.

An option could be added to jcli fragment creation commands, but it can be done in a separate PR.

Zeegomo (approved):

Zeegomo:
> There is no way to use lanes other than 0 with jcli, which is not a big issue.
You actually can, albeit with a kind of obscure interface. The spending counter is provided in the witness as a `u32`, and a conversion between `u32` and `SpendingCounter` is well defined and takes lanes into account in the first bits. I think it may be sufficient for now

vincenthz (approved):

Zeegomo:
Will hold this off until after the feature freeze not to bring additional load on the app side

Mr-Leshiy:
@Zeegomo As we have merged it into the chain-libs side, I assume we can merge this PR into master too. But create two new branches related to the fund7 without this feature and deploy them.

PR #3678, opened 2021-11-04
Remove duplicated settings in scenario tests

dkijania:
Removed another duplication between Settings in scenario tests and Settings in testing utils. Thanks for that we can only operate on controller from testing utils and make wrapper around it in scenario tests


testing/jormungandr-scenario-tests/src/scenario/controller.rs
```diff
@@ -88,38 +85,35 @@ impl ControllerBuilder {
...
-        let topology = self.topology.clone().expect("topology not set");
-        let blockchain = self.blockchain.clone().expect("blockchain not defined");
-        self.settings = Some(Settings::prepare(topology, blockchain, context));
-        self.controller_progress.inc(5);
+        self.network_builder = self.network_builder.blockchain_config(blockchain);
+        self
     }
 
     pub fn build(self, context: ContextChaCha) -> Result<Controller> {
+        self.controller_progress.inc(1);
```
mrzenioszeniou, line 107: I don't understand what this progress bar is supposed to represent. We increment by 1, then clear it, then increment it by 5. And why do we increment it when we set the topology and blockchain?

mrzenioszeniou (changes requested):

testing/jormungandr-scenario-tests/src/scenario/controller.rs
```diff
@@ -50,14 +50,12 @@ use std::{
 pub struct ControllerBuilder {
     title: String,
     controller_progress: ProgressBar,
-
-    topology: Option<Topology>,
-    blockchain: Option<Blockchain>,
-    settings: Option<Settings>,
+    network_builder: NetworkBuilder,
+    topology: Topology,
```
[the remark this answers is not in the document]
dkijania: Yes, very good catch. I haven't realized that i can get information about initial topology from node_settings.

testing/jormungandr-scenario-tests/src/scenario/controller.rs
dkijania: Yes, you are right. It doesn't make much of a sense. I've change a bit a mechanism in review fixes

mrzenioszeniou (approved):
Looks good but I'm a little skeptical of the `Observer`/`Observable` implementation. It seems like a lot of code for too little a return. This is a very uneducated take of course, since I haven't worked with the scenario tests __that__ much. But I'll advocate for getting rid of it nonetheless.

Firstly, I don't see why it's so important to give a heads-up status of the __individual__ steps of `NetworkBuilder::build`. Do we actually manually monitor these statuses that often? Also, the caller can simply set the progress bar's message to _"building network"_ or something of the sort, which I think is sufficient.

Secondly, it currently only has two "steps" – _"building topology"_ and _"building block0"_. Of the two, the latter seems to take at least one order longer than the former. This means that _"building topology"_  can barely be observed. So the entire implementation serves to simply set the progress bar's message only once. Which once again begs the same question. Why not just set progress bar's message directly on the calling side and get rid of all the extra hullabaloo?

Lastly, if you choose to keep the current implementation I would suggest moving the `Observer` and `Observable` traits to another module. There is nothing about it that is specific to the `NetworkBuilder` or to the `network` module.

In any case, I trust your judgment more than mine when it comes to the scenario tests, so I am approving.

seal-of-approval

dkijania:
Agree with your comments. I tried to exercise Observer code for future usage. In tests we have a lot of printing and I was thinking of controlling it more rather than just printing on console. But first i need to do proof that there is a working solution. I wasn't aware that only one action is taking 90% of time when creating settings. So in this part it is looking a little bit artificial. In our plans we have reusing the same mechanism to build data for vitup which has more steps, so output will looks better.

Anyway agree that this code should be moved into some more general location

PR #3682, opened 2021-11-10
[Scenario Test Refactor] Move testing directory to utils

dkijania:
Next step in removing scenario tests:
- moved TestingDirectory to jormungandr utils as this is useful struct if we want to preserve all network files for investigation or further usage
- removed further duplications between jormungandr utils and scenario tests


dkijania:
Fix for failing test is here:

https://github.com/input-output-hk/chain-libs/pull/684

PR #3684, opened 2021-11-12
graphql: add total_count to Block::transactions

enzoc4:
for some reason this query doesn't add the total_count field, and
although it is working now, it will cause a panic with async-graphql
2.11 because there are two different `TransactionConnection`


Mr-Leshiy (approved):
LGTM

eugene-babichenko (approved):

PR #3691, opened 2021-11-17
p2p id rework - remove redundant conn

Zeegomo:
### Why
The aim of this PR is mainly to simplify the way we handle p2p connections and remove redundant communications.

Before this, addresses were used as identifiers, which led to a node being present twice in the peers, once as a server and once as a client. In addition, messages were only propagated from a client to a server, even though we have quite some code that is supposed to handle incoming messages for a client (note we use bidirectional streams) which was left unused. Not only was that inefficient (maintain two active connections when that is not needed), it also made understanding flow of information more difficult.

Last but not least, addresses can be forged and we can't use that alone to authenticate messages.

### Changes
I propose with this PR to change the identifier of a node in the p2p layer to the public key used in the topology. From the high level perspective of message dissemination, this hides the fact that a node is a client or a server in the particular protocol used, and avoid having two connections to the same node.

In addition, unify the node id used during the handshake procedure to use the topology key pair. That identifier was only used to avoid propagating the same information to the same node more than once, and we can now reuse the id we already have from topology to do so. 

Furthermore, this can be used to authenticate the peer we're connecting to and is the first step for reliable authentication of messages.

Peer promotion on the server side does not happen right after the handshake but only after the first gossip message is received, to accommodate peers that do not wish to participate in the full dissemination protocol (e.g. private nodes, full wallets, ...).

### Future improvements
At this initial stage we are mainly interested in performance since the network is private and no external users have access to the deployment.
Nonetheless, there is still room for improvements, mainly in the security part like:
 * reliable authentication (ip spoofing works after the initial handshake)
 * proper key handling
 * id-based quarantine whitelist
 * id-based trusted peers

I'm thinking of moving most of the handshake procedure / connection handling to the underlying network layer, something like a tls connection but using tower services. 

Also, we should port this change to chain-network as well, but for that I was waiting for a complete design of the handshake/connections procedure.

### Test

Since this stress code that was previously unused, we should take extra care of testing this PR in detail

### DIFF
--- a/jormungandr-lib/src/crypto/key.rs
+++ b/jormungandr-lib/src/crypto/key.rs
@@ -414,6 +414,12 @@ impl<A: AsymmetricKey> From<SecretKey<A>> for SigningKey<A> {
     }
 }
 
+impl<A: AsymmetricKey> From<SigningKey<A>> for SecretKey<A> {
+    fn from(key: SigningKey<A>) -> Self {
+        key.0
+    }
+}
+
 impl<A: AsymmetricPublicKey> From<PublicKey<A>> for Identifier<A> {
     fn from(key: PublicKey<A>) -> Self {
         Identifier(key)
--- a/jormungandr/src/blockchain/process.rs
+++ b/jormungandr/src/blockchain/process.rs
@@ -243,7 +243,7 @@ impl Process {
                     hash = %header.hash(),
                     parent = %header.parent_id(),
                     date = %header.block_date(),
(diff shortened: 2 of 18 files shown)


eugene-babichenko (commented):
A couple notes regarding the code styling, the logic looks good.

eugene-babichenko (approved):

Zeegomo:
Will coordinate with @dkijania for in-depth testing before merging

PR #3694, opened 2021-11-18
Add ConfigParams as a GraphQl object

Mr-Leshiy:
Adding new ConfigParams GraphQL object and return it as a field of the UpdateProposal GraphQL object

### DIFF
--- a/jormungandr/src/explorer/graphql/config_param.rs
+++ b/jormungandr/src/explorer/graphql/config_param.rs
@@ -21,12 +21,12 @@ use std::num::{NonZeroU32, NonZeroU64};
 
 #[derive(SimpleObject)]
 pub struct Block0Date {
-    block_0_date: u64,
+    block0_date: u64,
 }
 
 impl From<&Block0DateLib> for Block0Date {
     fn from(v: &Block0DateLib) -> Self {
-        Self { block_0_date: v.0 }
+        Self { block0_date: v.0 }
     }
 }
 
--- a/jormungandr/src/explorer/graphql/config_param.rs
+++ b/jormungandr/src/explorer/graphql/config_param.rs
@@ -295,7 +295,7 @@ impl From<&ValueLib> for RewardPot {
 }
 
 #[derive(SimpleObject)]
-pub struct Linear {
+pub struct LinearRewardParams {
     constant: u64,
     ratio: Ratio,
     epoch_start: u32,
@@ -303,7 +303,7 @@ pub struct Linear {
 }
 
 #[derive(SimpleObject)]
-pub struct Halving {
+pub struct HalvingRewardParams {
     constant: u64,
     ratio: Ratio,
     epoch_start: u32,
@@ -312,8 +312,8 @@ pub struct Halving {
 
 #[derive(Union)]
 pub enum RewardParamsUnion {
-    Linear(Linear),
-    Halving(Halving),
+    Linear(LinearRewardParams),
+    Halving(HalvingRewardParams),
 }
 
 #[derive(SimpleObject)]
@@ -330,7 +330,7 @@ impl From<&RewardParamsLib> for RewardParams {
                 epoch_start,
                 epoch_rate,
             } => Self {
-                reward_params: RewardParamsUnion::Linear(Linear {
+                reward_params: RewardParamsUnion::Linear(LinearRewardParams {
                     constant: *constant,
                     ratio: Ratio(*ratio),
                     epoch_start: *epoch_start,
@@ -343,7 +343,7 @@ impl From<&RewardParamsLib> for RewardParams {
                 epoch_start,
                 epoch_rate,
             } => Self {
-                reward_params: RewardParamsUnion::Halving(Halving {
+                reward_params: RewardParamsUnion::Halving(HalvingRewardParams {
                     constant: *constant,
                     ratio: Ratio(*ratio),
                     epoch_start: *epoch_start,
--- a/jormungandr/src/explorer/graphql/config_param.rs
+++ b/jormungandr/src/explorer/graphql/config_param.rs
@@ -1,0 +1,525 @@
+use super::{scalars::Value, BftLeader, Ratio, TaxType};
+use async_graphql::{Enum, SimpleObject, Union};
+use chain_addr::Discrimination as DiscriminationLib;
+use chain_impl_mockchain::{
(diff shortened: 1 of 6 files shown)


jormungandr/src/explorer/graphql/config_param.rs
```diff
@@ -0,0 +1,525 @@
...
+    milli::Milli as MilliLib,
+    rewards::{Ratio as RatioLib, TaxType as TaxTypeLib},
+    value::Value as ValueLib,
+    vote::CommitteeId,
+};
+use std::num::{NonZeroU32, NonZeroU64};
+
+#[derive(SimpleObject)]
+pub struct Block0Date {
+    block_0_date: u64,
```
enzoc4, line 24: I think we usually spell this as `block0_date` for this. Not sure what's the most correct one, but it may be better to keep it consistent?

jormungandr/src/explorer/graphql/config_param.rs
```diff
@@ -0,0 +1,525 @@
...
+impl From<&BftLeaderId> for RemoveBftLeader {
+    fn from(v: &BftLeaderId) -> Self {
+        Self {
+            remove_bft_leader: v.clone().into(),
+        }
+    }
+}
+
+#[derive(SimpleObject)]
+pub struct LinearFee {
```
enzoc4, line 171: Hmmm, technically there is already an object for this in the explorer, but is named `FeeSettings` for one reason or another. It's also slightly different because it flattens the structure, but they could be joined I guess.

I could do it myself later, though.

Mr-Leshiy: What do you think if I will replace the usage of the `FeeSettings` with this `LinearFee` struct ?

enzoc4: I'm ok with it, it was called mostly because I was thinking at the time in the possibility of using something else (non linear?) as fees, but now I don't think it makes much sense tbh.

enzoc4:
I think technically we should also update the `schema.graphql` in the testing directory, it's not that important, but if you can do it the better I guess.

Mr-Leshiy:
@enzoc4 , Got it, will do it 😊

Zeegomo:
We should add an entry in the changelog

testing/jormungandr-testing-utils/resources/explorer/graphql/schema.graphql
```diff
@@ -1,6 +1,146 @@
...
+  discrimination: DiscriminationEnum!
+}
+
+enum ConsensusTypeEnum {
+  BFT,
+  GENESISPRAOS,
+}
+
+type ConsensusType {
+  consensus_type: ConsensusTypeEnum!
```
enzoc4, line 23: I'm a bit curious on what's the reason for this level of indirection. Why not have the enum be called `ConsensusType`?

testing/jormungandr-testing-utils/resources/explorer/graphql/schema.graphql
```diff
@@ -1,6 +1,146 @@
...
+
+type TreasuryParams {
+  treasury_params: TaxType!
+}
+
+type RewardPot {
+  reward_pot: Value!
+}
+
+type Linear {
```
enzoc4, line 93: given that graphql doesn't have modules (as far as I know), I think I would call this `LinearRewardParams`. If you see `Linear` in the schema then it's really not that obvious what it is.

testing/jormungandr-testing-utils/resources/explorer/graphql/schema.graphql
```diff
@@ -1,6 +1,146 @@
...
+type SlotsPerEpoch {
+  slots_per_epoch: Int!
+}
+
+type SlotDuration {
+  slot_duration: Int!
+}
+
+type EpochStabilityDepth {
+  epoch_stability_depth: Int!
```
enzoc4, line 35: Also in general, what do you think is the advantage of these wrappers? It gives you flexibility for adding new fields later, but I'm not really sure it's worth doing it for every thing.

testing/jormungandr-testing-utils/resources/explorer/graphql/schema.graphql
Mr-Leshiy: `assync_graphql` has a problem, it is impossible to make a GraphQl `union` in the following way:
```

#[derive(Enum)]
enum Bar {
   SomeBar,
}


#[derive(Union)]
enum A {
    Foo(Bar)
}

```
Because for the GraphQL union trait it is needed that `Bar` will implement an GraphQL `Object` trait and the only way to workaround it is to wrapper `enum Bar` over other `struct Bar` like I have done here

testing/jormungandr-testing-utils/resources/explorer/graphql/schema.graphql
Mr-Leshiy: The fist reason of adding wrappers for any field for the `enum ConfigParam` type is that for the GraphQL needed that every filed will implement `Object` trait as I described above and all standard types like `u32`, `u16`, `u8` does not implement it, so it is needed to add a wrapper.

enzoc4: This is true, but I rather use a `custom scalar` like all the other numeric types in `scalars.rs`. It's slightly more verbose (although it looks like they added a `scalar!` that wasn't there before, and we could also write a custom macro too), but it just looks nicer in the schema.

testing/jormungandr-testing-utils/resources/explorer/graphql/schema.graphql
enzoc4: Hmmm, apparently the graphql spec doesn't support having nested unions. Interesting.

Maybe we could considering flattening things, but I think this is fine.

testing/jormungandr-testing-utils/resources/explorer/graphql/schema.graphql
Mr-Leshiy: @enzoc4 , unfortunately scalars are also not allowed for the GraphQL `Union` as a field, for the same reasons ((((

enzoc4 (approved):
I still would rename `Linear` and `Halving`, but it looks good overall.

Mr-Leshiy:
@enzoc4 sure. Sorry have missed to update these :)

PR #3700, opened 2021-11-23
Parallel spending counter

Zeegomo:
Add jcli option to specify a spending counter lane in a user-friendly way.

I've also added basic parallel spending counter capability to tests, while still using lane 0 as the default in the test wallet.

I wanted to add tests to ensure the new command is working but I've noticed witness tests are not even compiled right now and have fallen behind. Will fix this and add tests for this command in another PR

### DIFF
--- a/jcli/src/jcli_lib/transaction/mk_witness.rs
+++ b/jcli/src/jcli_lib/transaction/mk_witness.rs
@@ -33,7 +33,7 @@ pub struct MkWitness {
     #[structopt(long = "genesis-block-hash", parse(try_from_str))]
     pub genesis_block_hash: HeaderId,
 
-    /// value is mandatory is `--type=account' It is the counter value for
+    /// value is mandatory if `--type=account`. It is the counter value for
     /// every time the account is being utilized.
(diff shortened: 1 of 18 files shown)


jcli/src/jcli_lib/transaction/mk_witness.rs
```diff
@@ -32,11 +33,17 @@ pub struct MkWitness {
     #[structopt(long = "genesis-block-hash", parse(try_from_str))]
     pub genesis_block_hash: HeaderId,
 
-    /// value is mandatory is `--type=account' It is the counter for
+    /// value is mandatory is `--type=account' It is the counter value for
```
eugene-babichenko, line 36:
```suggestion
    /// value is mandatory if `--type=account`. It is the counter value for
```

eugene-babichenko (commented):
Shall we add tests here, or would it be a separate PR?

Zeegomo:
I'll add tests in a separate PR as witness tests are not compiled and have fallen behind. Need to fix those first

eugene-babichenko (approved):

PR #3701, opened 2021-11-23
Expose account based vote count endpoint

danielSanchezQ:
We need to know the votes per account for other processes (like voters rewards). We already have an endpoint to check per account; but, in this case, we would want to know all of them.


eugene-babichenko (changes requested):
- Update the OpenAPI spec
- Update CHANGELOG.md

And not a request, but a note. This can be a rather heavy endpoint.

danielSanchezQ:
@eugene-babichenko yeah, I know it may be a heavy request but it is needed at some point, and I do not know how to get that information otherwise.

Zeegomo:
AFAIK, this is not actually needed publicly and should probably be hidden behind a proxy for public facing nodes.
It will only be used rarely by "administrators" (I guess with flexibility on execution time) and the impact on resources will be negligible IMHO.

eugene-babichenko (approved):

PR #3702, opened 2021-11-23
Add chain watch grpc tests

enzoc4:
update the proto files to the current chain-network and add some integration tests for the watch api.

### DIFF
--- a/testing/jormungandr-integration-tests/src/jormungandr/grpc/client_tests.rs
+++ b/testing/jormungandr-integration-tests/src/jormungandr/grpc/client_tests.rs
@@ -443,15 +443,15 @@ pub fn test_watch_tip_subscription_is_current_tip() {
     let mut blocks_to_test: usize = 20;
 
     while let Ok(header) = receiver.recv() {
-        blocks_to_test -= 1;
+        let tip = rest.tip().unwrap();
+
+        assert_eq!(header.id(), tip.into_hash());
 
         if blocks_to_test == 0 {
             break;
         }
 
-        let tip = rest.tip().unwrap();
-
-        assert_eq!(header.id(), tip.into_hash());
+        blocks_to_test -= 1;
     }
 }
 
--- a/testing/jormungandr-integration-tests/src/jormungandr/grpc/client_tests.rs
+++ b/testing/jormungandr-integration-tests/src/jormungandr/grpc/client_tests.rs
@@ -476,19 +476,35 @@ pub fn test_watch_sync_multiverse_full() {
         .collect();
 
     let mut block_set: HashSet<Hash> = Default::default();
+    let mut max_length = None;
 
-    let genesis = blocks[0].clone();
-
-    for block in blocks {
+    for block in &blocks {
         if !block_set.is_empty() {
             assert!(block_set.contains(&block.header().block_parent_hash()))
         }
+
+        max_length.replace(block.header().chain_length());
 
         assert!(block_set.insert(block.id()));
     }
 
     // the genesis block is not in the logs
+    let genesis = blocks[0].clone();
     block_set.remove(&genesis.id());
 
-    assert_eq!(block_set, block_hashes_from_logs);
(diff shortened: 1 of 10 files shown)


testing/jormungandr-integration-tests/src/jormungandr/grpc/client_tests.rs
```diff
@@ -396,3 +396,124 @@ pub fn pull_blocks_hashes_wrong_order() {
...
+pub fn test_watch_block_subscription_blocks_are_in_logs() {
+    use std::collections::HashSet;
+
+    let setup = setup::client::default();
+
+    let watch_client = setup.watch_client;
+
+    let (sender, receiver) = std::sync::mpsc::channel();
+
+    watch_client.block_subscription(sender);
```
Zeegomo, line 410: we've already discusses about this but I find it quite confusing 😅

testing/jormungandr-testing-utils/src/testing/node/grpc/client.rs
```diff
@@ -303,3 +308,159 @@ impl JormungandrClient {
...
+            inner_client,
+            rt,
+        }
+    }
+
+    fn client(&self) -> WatchClient<Channel> {
+        self.inner_client.clone()
+    }
+
+    pub fn tip_subscription(&self) -> WatchTipNotifier {
```
Zeegomo, line 363: Old plain mutex are used instead of async stuff because the testing framework is not very async friendly?

Zeegomo (approved):

testing/jormungandr-integration-tests/src/jormungandr/grpc/client_tests.rs
enzoc4: I don't understand, what is it that we discussed?

testing/jormungandr-testing-utils/src/testing/node/grpc/client.rs
enzoc4: I would say is directly not async, all the other grpc functions use `block_on` internally to make the api sync, so I tried to keep it that way.

I wonder if we can use `#[tokio::test]` in the test functions instead and just use streams here.

testing/jormungandr-integration-tests/src/jormungandr/grpc/client_tests.rs
Zeegomo: the naming, I find it a bit confusing that this is a client and is on the sender part of the stream

enzoc4: oh, well... But in this case this is actually a client (the grpc client). It's on the sender part of the stream only because this coe is sync, so this just loops on a separate threads and sends the messages on that channel, but the `watch_client` is a client for the node.

testing/jormungandr-testing-utils/src/testing/node/grpc/client.rs
enzoc4: > I wonder if we can use #[tokio::test] in the test functions instead and just use streams here.

And I think the answer is probably not without doing a big refactor that doesn't belong in this PR. But I think we should remove the runtime from both clients, and just leave it to the callers to deal with the executor. For the client tests we can use `#[tokio::test]`, and the rest of the code can just wrap things in `futures::executor::block_on`.

Or we could rewrite things in terms of `spawn`, instead of `block_on`, maybe with `tokio::runtime::Handle::try_current`.

eugene-babichenko (commented):
I'd say I am OK with the code here given the current paradigm. But I think we should go for `#[tokio::test]`. @enzoc4 if you consider this too big to fit into the current PR, I would ask to save this as a task for the future and we can continue with including this PR into `master`.

eugene-babichenko (approved):

dkijania (approved):

dkijania:
nice work!

PR #3704, opened 2021-11-24
Re enable witness tests

Zeegomo:
Compile and run witness tests.

While re-enabling them I had to remove some of those for which the testing API did not provide an easy access (invalid hashes). If we want to re-add them we will do so in another PR as it will require some work on the test framework.

In addition, I've added an `#[ignore]` tag on top of some tests which seems to target conditions what are not checked anymore, like a utxo witness in an account transaction and vice-versa. Not sure of what we want to do with those, maybe you can help @dkijania? 

This builds on changes from https://github.com/input-output-hk/jormungandr/pull/3700

### DIFF
--- a/testing/jormungandr-integration-tests/src/jcli/transaction/mod.rs
+++ b/testing/jormungandr-integration-tests/src/jcli/transaction/mod.rs
@@ -2,3 +2,4 @@ pub mod e2e;
 pub mod finalize;
 pub mod input;
 pub mod simplified;
+mod witness;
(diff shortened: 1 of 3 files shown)


dkijania:
@Zeegomo what is the reason of lack cross checks (utxo-acccount) ?

Zeegomo:
That combination is now accepted by jcli and does not return an error anymore, not sure of how we want to proceed

dkijania:
Ah, ok i requirements change so should tests. So IMHO test assertion should be alter in order to verify that you can do cross signing

Zeegomo:
The thing is that I'm not sure about requirements. It does not seems a good idea to allow for cross signing, but maybe it's too much of a burden to check this in jcli

dkijania (approved):

PR #3710, opened 2021-11-26
bump tokio to 1.14.0 and resolve data race issue

Zeegomo:


eugene-babichenko:
What was the data race issue though?

Zeegomo:
Something internal in tokio with oneshot channels, see https://rustsec.org/advisories/RUSTSEC-2021-0124

eugene-babichenko (approved):

PR #3721, opened 2021-12-03
Add minting tokens into the initial config

Mr-Leshiy:
Add initial token distribution into the `Block0Configuration`.


jormungandr-lib/src/interfaces/mint_token.rs
```diff
@@ -0,0 +1,93 @@
...
+        Self(val)
+    }
+}
+
+impl Serialize for TokenName {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: serde::Serializer,
+    {
+        hex::encode(self.0.as_ref()).serialize(serializer)
```
Zeegomo, line 24: I would add different behavior depending on whether the serializer/deserializer is human readable or not. There's no reason to dump this in hex format in the file is binary

Mr-Leshiy: Good point !
Have updated it

eugene-babichenko:
Hm, some failing Windows/macOS tests oddity. 🤔

jormungandr-lib/src/interfaces/block0_configuration/BLOCKCHAIN_CONFIGURATION_DOCUMENTED_EXAMPLE.yaml
```diff
@@ -207,6 +207,12 @@ initial:
       - address: {initial_funds_address}
         value: 10000
 
+  # Initial token distribution present in the blockchain
+  - token:
+      name: "7e5d6abc"
+      to: {initial_funds_account_identifier}
+      value: 100
```
eugene-babichenko, line 214: That would hurt readability and bloat the file size. Consider having tens of thousands (that's what we have during funds) of wallets in this form:

```yaml
  - token:
      name: "7e5d6abc"
      to: {initial_funds_account_identifier}
      value: 100
  - token:
      name: "7e5d6abc"
      to: {initial_funds_account_identifier2}
      value: 100
  - token:
      name: "7e5d6abc"
      to: {initial_funds_account_identifier3}
      value: 100
```

Instead you coould follow an example of how are initial fund distributions done but do that per token:

```yaml
initial:
  - token: token_identifier...:
    - address: addr1...
      value: 42
    - address: addr2...
      value: 1337
```

**NB**: Having just a token name is not sufficient considering the general case. Let's go for the agreed human-readable identifier format i.e. `{policy_hash}.{name}`.

eugene-babichenko (changes requested):

Mr-Leshiy:
@eugene-babichenko , yea, this tests also failing even in the master.

jormungandr-lib/src/interfaces/block0_configuration/BLOCKCHAIN_CONFIGURATION_DOCUMENTED_EXAMPLE.yaml
```diff
@@ -215,3 +217,18 @@ initial:
   #    # Legacy Cardano address
   #    - address: 48mDfYyQn21iyEPzCfkATEHTwZBcZJqXhRJezmswfvc6Ne89u1axXsiazmgd7SwT8VbafbVnCvyXhBSMhSkPiCezMkqHC4dmxRahRC86SknFu6JF6hwSg8
   #      value: 123
+
+# Initial token distribution
+initial_tokens:
```
eugene-babichenko, line 222: Why separate section?

Mr-Leshiy: It was added as a additional field inside the `Block0Configuration` for the purpose do not break the current invariant and usage of the `Initial` struct - "one `Initial` == one `Fragment`"

Mr-Leshiy: updated

jormungandr-lib/src/interfaces/block0_configuration/initial_fragment.rs
```diff
@@ -43,68 +51,87 @@ pub struct LegacyUTxO {
...
-                // and that if it changes the compiler will detect it and tell us about the
-                // change so we are reminded of a breaking change
-                let auth = tx.payload_auth().into_payload_auth();
-                let cert = certificate::SignedCertificate::VotePlan(cert, auth);
-                inits.push(Initial::Cert(cert.into()))
-            }
-            _ => return Err(Error::Block0MessageUnexpected),
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
+pub struct Destination {
+    address: Identifier,
```
eugene-babichenko, line 62: Could this be an actual address and not an `Identifier` for consistency with funds? IIRC it should not be hard to do the conversion.

Mr-Leshiy: I am not sure that it is possible, because we will have problems to convert `Address` into `Identifier` in both directions, which is needed there.
`Address` -> `Identifier` problems:
- `Address` type can be `Group` or `Multisig` or `Script` which is hardly to somehow convert or determine a public key of the receiver which is needed for the MintToken fragment.

`Identifier ` -> `Address` problems:
- need to determine `DISCRIMINATION` type of the address which I am not sure that is possible at this place (but maybe I am mistaken, need to investigate)

Mr-Leshiy: For the first issue we can restrict that you have to use only `Account` or `Single` type of `Address`.

Mr-Leshiy: It is possible, I have done 😊

testing/jormungandr-testing-utils/src/testing/network/settings.rs
```diff
@@ -335,9 +335,13 @@ impl Settings {
             self.vote_plans
                 .insert(vote_plan_key.clone(), vote_plan_settings);
         }
-        self.block0
-            .initial
-            .extend(try_initials_vec_from_messages(vote_plans_fragments.iter()).unwrap())
+
+        let disctimination = self.block0.blockchain_configuration.discrimination;
```
eugene-babichenko, line 339:
```suggestion
        let discrimination = self.block0.blockchain_configuration.discrimination;
```

eugene-babichenko (approved):
Good to merge after fixing a typo.

PR #3724, opened 2021-12-06
Add Token REST api

Mr-Leshiy:
Update `AccountState` with the new field `tokens`.
So now it is possible to get an info about the token's balances using the current API just fetching `account state`.

### DIFF
--- a/jormungandr-lib/src/interfaces/mint_token.rs
+++ b/jormungandr-lib/src/interfaces/mint_token.rs
@@ -55,6 +55,21 @@ impl<'de> Deserialize<'de> for TokenName {
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct TokenIdentifier(identifier::TokenIdentifier);
 
+impl PartialOrd for TokenIdentifier {
+    fn partial_cmp(&self, other: &Self) -> Option<std::cmp::Ordering> {
+        self.0
+            .token_name
+            .as_ref()
+            .partial_cmp(other.0.token_name.as_ref())
+    }
+}
+
+impl Ord for TokenIdentifier {
+    fn cmp(&self, other: &Self) -> std::cmp::Ordering {
+        self.0.token_name.as_ref().cmp(other.0.token_name.as_ref())
+    }
+}
+
 impl From<identifier::TokenIdentifier> for TokenIdentifier {
     fn from(val: identifier::TokenIdentifier) -> Self {
         Self(val)
--- a/jormungandr-lib/src/interfaces/mint_token.rs
+++ b/jormungandr-lib/src/interfaces/mint_token.rs
@@ -52,23 +52,8 @@ impl<'de> Deserialize<'de> for TokenName {
     }
 }
 
-#[derive(Debug, Clone, PartialEq, Eq)]
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord)]
 pub struct TokenIdentifier(identifier::TokenIdentifier);
-
-impl PartialOrd for TokenIdentifier {
-    fn partial_cmp(&self, other: &Self) -> Option<std::cmp::Ordering> {
-        self.0
-            .token_name
-            .as_ref()
-            .partial_cmp(other.0.token_name.as_ref())
-    }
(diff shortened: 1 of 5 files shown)


jormungandr-lib/src/interfaces/value.rs
```diff
@@ -59,7 +59,13 @@ impl AsRef<value::Value> for Value {
 
 impl From<value::Value> for Value {
     fn from(v: value::Value) -> Self {
-        Value(v)
+        Self(v)
+    }
+}
+
+impl From<&value::Value> for Value {
```
Zeegomo, line 66: this is redundant and actually inefficient

jormungandr/src/rest/v0/mod.rs
```diff
@@ -222,12 +222,24 @@ pub fn filter(
...
             .and(warp::get())
-            .and(with_context)
+            .and(with_context.clone())
             .and_then(handlers::get_active_vote_plans)
             .boxed();
         root.and(committees.or(vote_plans)).boxed()
     };
 
+    let tokens = {
+        let root = warp::path!("tokens" / ..);
```
Zeegomo, line 232: It looks a bit strange to me that we have a specific route just for tokens where this could be fetched from the account state

Zeegomo (changes requested):
We also need to update the api docs

jormungandr-lib/src/interfaces/value.rs
Mr-Leshiy: I have added to be respectful how it is done with the `AccountState` in this place https://github.com/input-output-hk/jormungandr/blob/9b6db4f32577905f5d14fc067068547fc5dfb490/jormungandr/src/rest/v0/logic.rs#L119 and https://github.com/input-output-hk/jormungandr/blob/33993d0642a3494158297492d75dd0f4ece37da1/jormungandr-lib/src/interfaces/account_state.rs#L162. But if you think that it better to explicitly convert types I can easily change it.

jormungandr/src/rest/v0/mod.rs
Mr-Leshiy: I have added this new route to fetch directly the specific token amount from the specific account, it seems it should be convenient in some cases.

jormungandr-lib/src/interfaces/value.rs
Zeegomo: I mean that we already have a `From<value::Value> for Value` implementation, which makes this not needed. In addition, this involves an unnecessary clone where `value::Value` is `Copy`, which may have performance impacts.

jormungandr/src/rest/v0/mod.rs
eugene-babichenko: > It looks a bit strange to me that we have a specific route just for tokens where this could be fetched from the account state

And that is not how it is specified. This could (and should) have been just a couple of new fields in the account state structure.

eugene-babichenko (changes requested):

jormungandr-lib/src/interfaces/mint_token.rs
```diff
@@ -55,6 +55,21 @@ impl<'de> Deserialize<'de> for TokenName {
...
+            .as_ref()
+            .partial_cmp(other.0.token_name.as_ref())
+    }
+}
+
+impl Ord for TokenIdentifier {
+    fn cmp(&self, other: &Self) -> std::cmp::Ordering {
+        self.0.token_name.as_ref().cmp(other.0.token_name.as_ref())
+    }
+}
```
eugene-babichenko, line 71: This is not entirely correct, because the hash of the minting policy can also be different in the future. Of course it is correct _for now_, so let's change it in a separate PR.

I would suggest you to:

1. Implement `PartialOrd` and `Ord` in chain-libs. You can simply use `derive(PartialOrd, Ord)`, which will compare fields in the order they are declared.
2. Just derive here as well.

eugene-babichenko (approved):
There are things that are not correct strictly speaking, but are fine for the time being.

Zeegomo (approved):

PR #3726, opened 2021-12-06
Add missing cors config

enzoc4:
there are some cases where being able to set these things is necessary.


eugene-babichenko (approved):

eugene-babichenko (changes requested):
Oops, I was too quick to approve this. You need to update the test code as well, otherwise the tests just don't compile.

enzoc4:
I think failure is unrelated now. 

I do wonder why do we have tests for cors that are `#[cfg(windows)]` though. It shouldn't be platform dependent.

eugene-babichenko (approved):

PR #3730, opened 2021-12-10
Update api doc of vote status endpoint

Mr-Leshiy:


doc/api/v0.yaml
```diff
@@ -2215,6 +2215,16 @@ paths:
                   description: Vote Plan
                   type: object
                   properties:
+                    id:
+                      type: string
+                      description: Hex-encoded vote plan ID
+                      pattern: '[0-9a-f]+'
+                      minLength: 64
+                      maxLength: 64
```
eugene-babichenko, line 2223: Could be just `[0-9a-f]{64}`

eugene-babichenko (changes requested):

doc/api/v0.yaml
```diff
@@ -2215,6 +2215,14 @@ paths:
                   description: Vote Plan
                   type: object
                   properties:
+                    id:
+                      type: string
+                      description: Hex-encoded vote plan ID
+                      pattern: '[0-9a-f]+{64}'
```
Zeegomo, line 2221:
```suggestion
                      pattern: '[0-9a-f]{64}'
```

doc/api/v0.yaml
```diff
@@ -2227,23 +2235,28 @@ paths:
...
+                            minimum: 0
+                            maximum: 255
+                          proposal_id:
                             type: string
-                            description: Hash of external id
-                            pattern: '[0-9a-f]+'
-                            minLength: 64
-                            maxLength: 64
+                            description: Hex-encoded proposal id
+                            pattern: '[0-9a-f]+{64}'
```
Zeegomo, line 2259:
```suggestion
                            pattern: '[0-9a-f]{64}'
```

doc/api/v0.yaml
```diff
@@ -2259,36 +2272,130 @@ paths:
...
+                                                        minimum: 0
+                                                        maximum: 15
+                                                      end:
+                                                        type: integer
+                                                        description: upper bound of the available options, maximum being 16 (not included)
+                                                        minimum: 1
+                                                        maximum: 15
+                          votes_cast:
+                            type: integer
+                            description: votes vast
```
Zeegomo, line 2355:
```suggestion
                            description: votes cast
```

doc/api/v0.yaml
```diff
@@ -2259,36 +2272,130 @@ paths:
...
+                                            type: object
+                                            properties:
+                                              start:
+                                                type: integer
+                                                description: starting value of the vote option, starting from 0
+                                                minimum: 0
+                                                maximum: 15
+                                              end:
+                                                type: integer
+                                                description: upper bound of the available options, maximum being 16 (not included)
```
Zeegomo, line 2306: I think the maximum is actually 255

doc/api/v0.yaml
```diff
@@ -2259,36 +2272,130 @@ paths:
...
+                                                      start:
+                                                        type: integer
+                                                        description: starting value of the vote option, starting from 0
+                                                        minimum: 0
+                                                        maximum: 15
+                                                      end:
+                                                        type: integer
+                                                        description: upper bound of the available options, maximum being 16 (not included)
+                                                        minimum: 1
+                                                        maximum: 15
```
Zeegomo, line 2352: this could be abstracted to avoid repetition between public and private tally

doc/api/v0.yaml
Mr-Leshiy: Are you sure ? An upper bound for the `start` is correct ?

Zeegomo: You're right, but I swear I remember there was a limit of 256 regarding choices somewhere.
Maybe it's just a vote that can be constructed for up to 256 options, but it's not currently allowed to create a proposal with that many choices

doc/api/v0.yaml
Mr-Leshiy: Updated

PR #3732, opened 2021-12-10
Catch panics

Zeegomo:
Catch panics from child processes

PS: this `RefCell` thing is getting a bit out of hand

### DIFF
--- a/testing/hersir/src/controller/interactive/args/show.rs
+++ b/testing/hersir/src/controller/interactive/args/show.rs
@@ -218,7 +218,7 @@ fn show_logs_for(
     let logs: Vec<String> = {
         if only_errors {
             logger
-                .get_lines_with_level(LogLevel::ERROR)
+                .get_log_lines_with_level(LogLevel::ERROR)
                 .map(|x| x.to_string())
                 .collect()
         } else if let Some(contains) = &contains {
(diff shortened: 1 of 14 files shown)


testing/jormungandr-testing-utils/src/testing/jormungandr/process.rs
```diff
@@ -72,10 +77,12 @@ impl JormungandrProcess {
...
         let stdout = child.stdout.take().unwrap();
+        let stderr = RefCell::new(OutputCollector::new(child.stderr.take().unwrap()));
 
         Ok(JormungandrProcess {
             child,
             temp_dir,
+            stderr,
             alias,
             grpc_client: JormungandrClient::new(node_config.p2p_listen_address()),
             logger: JormungandrLogger::new(stdout),
```
Mr-Leshiy, line 88: Maybe, is it better to put `stderr` into the `JormungandrLogger` struct, so it will maintain any logging and messaging which is coming from the node ? What do you think ?

Zeegomo: I thought about this at the beginning and then decided to go for adding it to `JormungandrProcess` because those are at a slightly different level but I agree with your point that it is probably simpler

Mr-Leshiy (approved):
LGTM!

dkijania (approved):

dkijania:
please rebase. Looking good

eugene-babichenko (approved):

PR #3733, opened 2021-12-10
Use better measure for network sync

Zeegomo:
Last block height is not a very good measure for network sync, as there could be multiple forks all at the same length and it's the tip that matters rather than the last received block.
Let's use the tip hash for a better measure


dkijania (approved):

dkijania:
Actually it reveal some issue. Particularly in the test: `networking::communication::leader_leader::two_transaction_to_two_leaders`. Maybe it's worth rebasing after tip update improvement?

testing/jormungandr-automation/src/testing/benchmark/sync/measure.rs
```diff
@@ -109,34 +109,29 @@ pub fn measure_and_log_sync_time<A: SyncNode + ?Sized>(
...
             })
-            .collect();
+            .collect::<Vec<_>>();
 
         if report_node_stats_counter >= interval {
-            println!(
-                "Measuring sync time... current block heights: {:?}",
-                block_heights
-            );
+            println!("Measuring sync time... current block tips: {:?}", tips);
```
Mr-Leshiy, line 123: Maybe implement `Display` trait implementation for `tips` as we display it via `println` for the user's convenience, not for the debug purposes.

eugene-babichenko (approved):

eugene-babichenko:
The idea and the code are looking, but we have a consistent test failure on all release builds in GH Actions. I think this is worth looking into, since we never encountered it before.

eugene-babichenko:
UPD: could not reproduce it locally.

Zeegomo:
> UPD: could not reproduce it locally.

`networking::communication::leader_leader::two_transaction_to_two_leaders` fails for me, but you have to enable network tests (`--features network`), which are enabled on ci for release builds

Zeegomo:
I've updated the active slot coeff value used in tests from 1 to 0.5. I think Cardano mainnet uses 0.1, but that caused some tests to fail because transactions were not accepted fast enough into the chain.

PR #3735, opened 2021-12-13
[Test] spending counter increasing update

dkijania:


testing/jormungandr-integration-tests/src/jormungandr/bft/counter.rs
```diff
@@ -0,0 +1,66 @@
...
+/// Ensures that blocks with an incorrect signature are rejected by a BFT leader node
+fn parallel_transaction_using_different_lanes() {
+    let receiver = startup::create_new_account_address();
+    let mut sender = startup::create_new_account_address();
+
+    let (jormungandr, _) = startup::start_stake_pool(
+        &[sender.clone()],
+        &[receiver.clone()],
+        ConfigurationBuilder::new()
+            .with_block0_consensus(ConsensusType::Bft)
```
Zeegomo, line 16: why a stake pool if we use bft consensus?

Cargo.lock
```diff
@@ -14,9 +14,9 @@ dependencies = [
 
 [[package]]
 name = "addr2line"
```
Zeegomo, line 16: Is this a `cargo update` that slipped through?

testing/jormungandr-integration-tests/src/jormungandr/bft/counter.rs
```diff
@@ -0,0 +1,66 @@
...
+    let mut fragment_sender = FragmentSender::new(
+        jormungandr.genesis_block_hash(),
+        jormungandr.fees(),
+        jormungandr.default_block_date_generator(),
+        FragmentSenderSetup::no_verify(),
+    );
+
+    let mut checks = vec![];
+
+    fragment_sender = fragment_sender.witness_mode(WitnessMode::Lane(1));
```
Zeegomo, line 32: Unrelated to this pr, but I see now that `WitnessMode::Lane(..)` is a bit confusing, I would prefer something like an explicit `WitnessMode::Account { lane: 3 }`

testing/jormungandr-integration-tests/src/jormungandr/bft/counter.rs
```diff
@@ -0,0 +1,66 @@
+use chain_impl_mockchain::{chaintypes::ConsensusType, testing::WitnessMode};
+use jormungandr_testing_utils::testing::{
+    jormungandr::ConfigurationBuilder, startup, FragmentSender, FragmentSenderSetup,
+};
+
+#[test]
+/// Ensures that blocks with an incorrect signature are rejected by a BFT leader node
```
Zeegomo, line 7: test comment is unrelated

testing/jormungandr-testing-utils/src/testing/fragments/mod.rs
```diff
@@ -105,14 +99,35 @@ impl FragmentBuilder {
...
+        }
+
+        let value_u64: u64 = value.into();
+        let input_without_fees: Value = (value_u64 * addresses.len() as u64).into();
+        let input_value = self
+            .fragment_factory
+            .fee
+            .calculate(None, 1, addresses.len() as u8)
+            + input_without_fees.into();
+        let input = from.add_input_with_value(input_value.unwrap().into());
```
Zeegomo, line 117: this only works for accounts, not sure if we allowed utxos before / we are interested in providing this service for utxos wallets. If that's not the case I suggest to restrict the input type to `account::Wallet`

dkijania: it was just a copy paste. `add_input_with_value` will panic for utxo. Is that ok, or should i signal it in method name that it is only account based operation?

Zeegomo: if we accept in input a `account::Wallet` instead of a general purpose wallet it's enough, even though signaling this in the name would help

Cargo.lock
dkijania: you mean that i should remove not related to this change update of libs or should not update cargo at all?

Zeegomo: I would rather do this in another pr

Zeegomo (approved):
Good work on the testcases for parallel spending counter

PR #3740, opened 2021-12-17
Improve tip update [catalyst-fund7]

Zeegomo:
Looking at the recent failures in the perf env, it seems like one of the problems is that reprocessing the tip takes a lot of time and the actual tip falls behind.

This happens for two reasons:
* **We maintain a list of branches at the node level in a best effort _non ideal_ way:** 
This can result in a lot of unnecessary branches (as in considering a branch a non-leaf block) in certain edge cases. As a consequence, the reprocessing step will have a lot of useless branches to consider.
In addition, `Branch` currently looks like it's meant to be updated and kept track of, but I argue this is not very useful. 
What does it mean to follow a branch? In the event of a fork, which path should it follow?
The only exception is to follow the tip, as diverging blocks can be discriminated precisely by means of the chain selection rule.

    I'm not sure why we handle it this way (and I'm to blame for this as well, as I did recent work on it), but the storage already keeps track of existing branches in a reliable way and I suggest to use that information.

* **Reprocessing the tip is done in the same loop as the standard tip update**.
To avoid blocking the loop, calculate the best branch in another task and only serialize this single ref as if it was coming from the blockchain

PS: Will need to port this to master as well

### DIFF
--- a/jormungandr/src/blockchain/chain.rs
+++ b/jormungandr/src/blockchain/chain.rs
@@ -49,7 +49,7 @@ See Internal documentation for more details: doc/internal_design.md
 [`Branch`]: ./struct.Branch.html
 */
 #![allow(clippy::large_enum_variant)]
-use super::{branch::Branches, reference_cache::RefCache};
+use super::reference_cache::RefCache;
 use crate::{
     blockcfg::{
         Block, Block0Error, BlockDate, ChainLength, Epoch, EpochRewardsInfo, Header, HeaderDesc,
@@ -59,8 +59,8 @@ use crate::{
 };
 use chain_impl_mockchain::{leadership::Verification, ledger};
 use chain_time::TimeFrame;
+use futures::{StreamExt, TryStreamExt};
 use std::sync::Arc;
-use tokio_stream::StreamExt;
 
 #[derive(Debug, thiserror::Error)]
 pub enum Error {
@@ -167,8 +167,6 @@ pub fn pre_verify_link(
 ///
 #[derive(Clone)]
 pub struct Blockchain {
-    branches: Branches,
-
     ref_cache: RefCache,
 
     ledgers: Multiverse<Ledger>,
@@ -273,7 +271,6 @@ impl Blockchain {
         rewards_report_all: bool,
     ) -> Self {
         Blockchain {
-            branches: Branches::new(),
             ref_cache: RefCache::new(cache_capacity),
             ledgers: Multiverse::new(),
             storage,
@@ -290,12 +287,15 @@ impl Blockchain {
         &self.storage
     }
 
-    pub fn branches(&self) -> &Branches {
-        &self.branches
-    }
-
-    pub fn branches_mut(&mut self) -> &mut Branches {
-        &mut self.branches
+    pub async fn branches(&self) -> Result<Vec<Branch>> {
+        futures::stream::iter(self.storage().get_branches()?)
+            // FIXME: this should always return a valid ref, as the branches
+            // are directly fetched from the node storage, but account for
+            // misses in the ref cache which are not handled currently by the node
+            .filter_map(|branch| async move { self.get_ref(branch).await.transpose() })
+            .map(|try_ref| try_ref.map(Branch::new))
+            .try_collect()
+            .await
     }
 
     pub async fn gc(&self, tip: Arc<Ref>) -> Result<()> {
@@ -631,8 +631,6 @@ impl Blockchain {
         let block0_id = block0.header().hash();
         let block0_date = block0.header().block_date();
 
-        let mut branches = self.branches.clone();
-
         let time_frame = {
             use crate::blockcfg::Block0DataSource as _;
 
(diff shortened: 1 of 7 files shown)


jormungandr/src/blockchain/tip.rs
```diff
@@ -202,52 +202,69 @@ impl TipUpdater {
...
+    async fn reprocess_tip(
+        blockchain: Blockchain,
+        tip: Arc<Ref>,
+        mut mbox: MessageBox<Arc<Ref>>,
+    ) -> Result<(), Error> {
+        use std::cmp::Ordering;
+        let branches = blockchain.branches().await?;
+        let storage = blockchain.storage();
+
+        let best_branch = branches.into_iter().map(Branch::into_ref).max_by(|a, b| {
```
enzoc4, line 218: This may add some bias to hashes with lower leading bits, I think? At least I think that's the order of iteration.

Not sure if it matters, though. And it may be more a thing with the `compare_against` function.

jormungandr/src/blockchain/tip.rs
```diff
@@ -202,52 +202,69 @@ impl TipUpdater {
...
+            match chain_selection::compare_against(storage, &a, &b) {
+                ComparisonResult::PreferCurrent => Ordering::Greater,
+                ComparisonResult::PreferCandidate => Ordering::Less,
+            }
+        });
 
-        for other in others {
-            self.process_new_ref(Arc::clone(other)).await?
+        if let Some(new_tip) = best_branch {
+            if !Arc::ptr_eq(&tip, &new_tip) {
```
enzoc4, line 226: I wonder if this is truly necessary. But I can't imagine a case where it should case problems, so I guess it's fine (except maybe some `RefCache` refactors?)

jormungandr/src/blockchain/chain.rs
```diff
@@ -290,12 +287,15 @@ impl Blockchain {
...
 
-    pub fn branches(&self) -> &Branches {
-        &self.branches
-    }
-
-    pub fn branches_mut(&mut self) -> &mut Branches {
-        &mut self.branches
+    pub async fn branches(&self) -> Result<Vec<Branch>> {
+        futures::stream::iter(self.storage().get_branches()?)
+            // FIXME: this should always return a valid ref, as the branches
```
enzoc4, line 292: can this happen? wouldn't a `Vec<Result<_, _>>` make more sense? Although for the fund7 branch I guess it doesn't really matter.

enzoc4 (commented):
I actually always wondered if it is really that necessary to reprocess tips periodically. If there is a branch that becomes more desirable I feel like it should get selected anyway when we receive (or produce) a block for it.

jormungandr/src/blockchain/tip.rs
Zeegomo: It's mostly to provide better clarity in the logs on what's happening

jormungandr/src/blockchain/tip.rs
Zeegomo: > This may add some bias to hashes with lower leading bits, I think

Why should it matter? the max function (in this case tip selection) does not really depend on the order (apart from equal score chains, but in this case I guess it's arbitrary the one you choose)

jormungandr/src/blockchain/chain.rs
Zeegomo: > wouldn't a Vec<Result<_, _>> make more sense?

Partly yes, but I think it will be just more cumbersome to use. Also because I expect errors not to depend on the specific branch but rather on external factors (e.g. spurious storage failures)

jormungandr/src/blockchain/tip.rs
enzoc4: Well... It may not matter. I meant for equal score chains. The storage is a btree, so I think it's sorted, and this will pick the first in that order. 

It doesn't matter for fund7 though, so I'm not proposing to change it. What I was wondering is whether it could incentivize pools to produce blocks with small hashes.

jormungandr/src/blockchain/tip.rs
enzoc4: I meant the `ptr_eq` instead of using normal equality, but I thought it was a the tip id and not a `Ref`, and I guess that's not comparable, so please ignore this.

enzoc4 (approved):

Zeegomo:
> If there is a branch that becomes more desirable I feel like it should get selected anyway when we receive (or produce) a block for it.

I'm a bit battled as well as I was trying to understand the rational behind this. I can find two reasons for it:
* Time de-sync. If a block looks to be in the future we do not consider if for the tip selection, but it will be a valid one when that time comes. For example, if we lag behind the longest chain with respect to time (which is ofc non-desiderable) without branch reprocessing we will never select that as the tip, which with reprocessing we will select it with some delay. 
* In general, a bit of resilience around one-off conditions and occasional failures

But I agree that for a live chain receiving a new block from it should trigger the right selection rule.

Mr-Leshiy (approved):
LGTM

jormungandr/src/blockchain/tip.rs
Zeegomo: In a standard flow, tips would go first through the normal processing function, where time is the biggest advantage, so I guess the additional work to produce blocks with small hashes would not be beneficial. 
Good call out though, didn't think of that

eugene-babichenko (approved):

PR #3741, opened 2021-12-17
remove apparently unused dependencies

enzoc4:
Maybe I'm missing something, but these dependencies don't seem to be used anymore. At least everything seems to compile just fine. For the QR code ones I think it is because it was moved to `catalyst-toolbox`, I guess @dkijania can confirm. 

The other one I don't know.

### DIFF
--- a/jormungandr-lib/Cargo.toml
+++ b/jormungandr-lib/Cargo.toml
@@ -29,7 +29,6 @@ hex = "0.4"
 bech32 = "0.8"
 base64 = "0.13.0"
 http = "0.2.2"
-eccoxide = { git = "https://github.com/eugene-babichenko/eccoxide.git", branch = "fast-u64-scalar-mul", features = ["fast-u64-scalar-mul"], optional = true }
 
 [dev-dependencies]
 quickcheck = "0.9"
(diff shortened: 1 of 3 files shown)


jormungandr-lib/Cargo.toml
```diff
@@ -29,7 +29,6 @@ hex = "0.4"
 bech32 = "0.8"
 base64 = "0.13.0"
 http = "0.2.2"
-eccoxide = { git = "https://github.com/eugene-babichenko/eccoxide.git", branch = "fast-u64-scalar-mul", features = ["fast-u64-scalar-mul"], optional = true }
```
Zeegomo, line 32: `eccoxide` was never a direct dependency of this crate, but rather a hack to work around the current build process https://github.com/input-output-hk/jormungandr/pull/3184. We should switch away from that `eccoxide` version though, as I moved that improvement directly in chain-crypto IIRC.

Zeegomo:
We did make the switch from the custom `eccoxide` version already in https://github.com/input-output-hk/chain-libs/pull/587, somehow it got overridden.  I've opened https://github.com/input-output-hk/chain-libs/pull/714 to fix this

jormungandr-lib/Cargo.toml
enzoc4: oh, I remember now. As far as I understand, we still need to remove it from here then, right?

Zeegomo: yes, and actually it was re-inserted in chain-libs somehow, https://github.com/input-output-hk/chain-libs/pull/714 will fix it

dkijania:
Yes, i can confirm, we can get rid of them

Mr-Leshiy (approved):

eugene-babichenko (approved):

PR #3742, opened 2021-12-21
Improve tip update

Zeegomo:
Port of https://github.com/input-output-hk/jormungandr/pull/3740 to master

Looking at the recent failures in the perf env, it seems like one of the problems is that reprocessing the tip takes a lot of time and the actual tip falls behind.

This happens for two reasons:
* We maintain a list of branches at the node level in a best effort non ideal way:
    This can result in a lot of unnecessary branches (as in considering a branch a non-leaf block) in certain edge cases. As a consequence, the reprocessing step will have a lot of useless branches to consider.
    In addition, Branch currently looks like it's meant to be updated and kept track of, but I argue this is not very useful.
    What does it mean to follow a branch? In the event of a fork, which path should it follow?
    The only exception is to follow the tip, as diverging blocks can be discriminated precisely by means of the chain selection rule.
    I'm not sure why we handle it this way (and I'm to blame for this as well, as I did recent work on it), but the storage already keeps track of existing branches in a reliable way and I suggest to use that information.
* Reprocessing the tip is done in the same loop as the standard tip update.
    To avoid blocking the loop, calculate the best branch in another task and only serialize this single ref as if it was coming from the blockchain

### DIFF
--- a/jormungandr/src/blockchain/branch.rs
+++ b/jormungandr/src/blockchain/branch.rs
@@ -1,160 +1,25 @@
 use crate::blockchain::Ref;
-use futures::stream::{FuturesUnordered, StreamExt};
 use std::sync::Arc;
-use tokio::sync::RwLock;
-
-#[derive(Clone)]
-pub struct Branches {
-    inner: Arc<RwLock<BranchesData>>,
-}
-
-struct BranchesData {
-    branches: Vec<Branch>,
-}
-
-#[derive(Clone)]
-pub struct Branch {
-    inner: Arc<RwLock<BranchData>>,
-}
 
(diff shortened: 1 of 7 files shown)


dkijania (approved):

dkijania:
Please make clippy happy and merge ! :)

dkijania:
Ah-a, this failure again: 

```
jormungandr::grpc::client_tests::test_watch_tip_subscription_is_current_tip
```

I have separate story for fixing this issue as already talked with Enzo. i think we can merge this one

Mr-Leshiy (approved):

PR #3745, opened 2021-12-29
Update with the latest chain-libs

Mr-Leshiy:


testing/jormungandr-testing-utils/src/testing/vit/builder.rs
```diff
@@ -99,6 +112,11 @@ impl VotePlanBuilder {
         self
     }
 
+    pub fn voting_token(&mut self, voting_token: TokenIdentifier) -> &mut Self {
```
enzoc4, line 115: isn't this the same as `with_voting_token`?

testing/jormungandr-testing-utils/src/wallet/account.rs
```diff
@@ -92,6 +92,10 @@ impl Wallet {
         ))
     }
 
+    pub fn internal_counters_increasing(&self) -> &SpendingCounterIncreasing {
```
enzoc4, line 95: @dkijania did something different in #3735, so I think it may be better to do the same.

enzoc4 (commented):
what about the errors in `prometheus_exporter.rs` which are behind the `prometheus-metrics` feature?

Thank you for taking care of the voting token part.

testing/jormungandr-testing-utils/src/testing/vit/builder.rs
Mr-Leshiy: Yes it is the same, I have done it with the respect of other fields. But yes, I agree with you it is better to leave only one method.

enzoc4 (approved):
LGTM

PR #3751, opened 2022-01-05
Update 'transaction' jcli command, add posibility to create evm transaction fragment

Mr-Leshiy:
Firstly need to merge https://github.com/input-output-hk/chain-libs/pull/726

Updated `Transaction` jcli command, add new subcommand `AddEvmTransaction` for the `Transaction` command.
Add `EvmTransaction` interface type for the library's `EvmTransaction` type


enzoc4 (commented):
why can't the `EvmTransaction` be a certificate? Maybe it was discussed already, but doesn't it follow the same pattern?

Mr-Leshiy:
@enzoc4 it is a good question. Indeed currently we do not have any cryptographic validation of the evm transactions, which is a vulnerability.
I am not sure but assume that it is not implemented currently  because we also need to determine how to sing such transactions and how to match our `jormungandr` address format with the `evm` address format and now we are trying to add some MVP part of the evm.

@eugene-babichenko maybe can elaborate about it.

Mr-Leshiy:
Also needs to merge first this https://github.com/input-output-hk/jormungandr/pull/3757

jcli/src/jcli_lib/transaction/mod.rs
```diff
@@ -45,9 +46,15 @@ pub enum Transaction {
     /// set a transaction expiration date
     SetExpiryDate(set_expiry_date::SetExpiryDate),
     /// set a certificate to the Transaction. If there is already
-    /// an extra certificate in the transaction it will be replaced
-    /// with the new one.
+    /// an evm transaction in the transaction it will be reseted.
```
eugene-babichenko, line 49:
```suggestion
    /// an evm transaction in the transaction it will be reset.
```

jcli/src/jcli_lib/transaction/mod.rs
```diff
@@ -45,9 +46,15 @@ pub enum Transaction {
...
+    /// If there is already an extra certificate in the transaction
+    /// it will be replaced with the new one.
     AddCertificate(add_certificate::AddCertificate),
+    /// set a evm transaction to the Transaction. If there is already
+    /// an extra certificate in the transaction it will be reseted.
```
eugene-babichenko, line 54:
```suggestion
    /// an extra certificate in the transaction it will be reset.
```

jcli/src/jcli_lib/transaction/staging.rs
```diff
@@ -358,10 +375,19 @@ impl Staging {
...
+            None => match &self.evm_transaction {
+                None => self.finalize_payload(
+                    &chain::transaction::NoExtra,
+                    fee_algorithm,
+                    output_policy,
+                ),
+                Some(_tx) => {
+                    // let tx: chain::evm::EvmTransaction = tx.clone().into();
+                    // self.finalize_payload(&tx, fee_algorithm, output_policy)
+                    unreachable!()
```
eugene-babichenko, line 387: what's the reason for this change?

jcli/src/jcli_lib/transaction/staging.rs
```diff
@@ -573,8 +606,15 @@ impl Staging {
...
         let res = match &self.extra {
-            None => self.transaction_sign_data_hash_on(TxBuilder::new().set_nopayload()),
-            Some(ref c) => match c.clone().into() {
+            None => match &self.evm_transaction {
+                None => self.transaction_sign_data_hash_on(TxBuilder::new().set_nopayload()),
+                Some(_tx) => {
+                    // let tx: chain::evm::EvmTransaction = tx.clone().into();
+                    // self.transaction_sign_data_hash_on(TxBuilder::new().set_payload(&tx))
+                    unreachable!()
+                }
```
eugene-babichenko, line 615: Here's another one

jcli/src/jcli_lib/transaction/staging.rs
```diff
@@ -358,10 +375,21 @@ impl Staging {
...
+                    output_policy,
+                ),
+                Some(tx) => {
+                    #[allow(unreachable_code)]
+                    #[allow(unused_variables)]
+                    let tx: chain::evm::EvmTransaction = tx.clone().into();
+                    #[allow(unreachable_code)]
+                    self.finalize_payload(&tx, fee_algorithm, output_policy)
+                }
+            },
```
eugene-babichenko, line 391: Are these really unreachable/unused now?

Mr-Leshiy: We have the following situation at this place. When we are compiling with the `feature = "evm"` so it works fine and compiles with any problem, but when we are compiling without  `feature = "evm"` so `chain::evm::EvmTransaction` become a `uninhabited` type, so it is not way how to initialize a value with this type so according to this this code will be unreachable.
```
pub enum EvmTransaction {}
```

Maybe it is better than I will do it in the following way
```
#[cfg(feature(evm)]
{
     let tx: chain::evm::EvmTransaction = tx.clone().into();
     self.finalize_payload(&tx, fee_algorithm, output_policy)
}
#[cfg(feature(evm)]
unreachable!()
```

eugene-babichenko: That would be the correct approach IMO. These unreachables would hit us in the back later.

eugene-babichenko (approved):

PR #3757, opened 2022-01-07
Update VotePlanStatus with the token identifier

Mr-Leshiy:
Relates to the following update https://github.com/input-output-hk/chain-libs/pull/725 of the VotePlanStatus

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -454,7 +454,7 @@ dependencies = [
 [[package]]
 name = "cardano-legacy-address"
 version = "0.1.1"
-source = "git+https://github.com/input-output-hk/chain-libs.git?branch=master#c30099ab1c44eeb766ca98354f4fceca1199c793"
+source = "git+https://github.com/input-output-hk/chain-libs.git?branch=master#4d2c52ace7dc3d76e9c6bfc777e1d3fa6cd68053"
 dependencies = [
  "cbor_event",
  "chain-ser",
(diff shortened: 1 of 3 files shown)


Cargo.lock
```diff
@@ -619,9 +619,9 @@ dependencies = [
 [[package]]
 name = "chain-vote"
 version = "0.1.0"
-source = "git+https://github.com/input-output-hk/chain-libs.git?branch=master#c30099ab1c44eeb766ca98354f4fceca1199c793"
+source = "git+https://github.com/input-output-hk/chain-libs.git?branch=master#4d2c52ace7dc3d76e9c6bfc777e1d3fa6cd68053"
 dependencies = [
- "cfg-if 1.0.0",
+ "cfg-if 0.1.10",
```
enzoc4, line 624: I find this a bit weird, but I guess it's not important

enzoc4 (approved):
I was going to do this together with #3749, but I think it is fine to do it here too, thank you.

Cargo.lock
Zeegomo: I don't understand it, but it's changing pretty much every other update of Cargo.lock 🤷🏼‍♂️

enzoc4: I wonder if it's really ok to use a `*` [here](https://github.com/input-output-hk/chain-libs/blob/43663b1a0661abcf1dab1ac1bb9cf4053727891d/chain-vote/Cargo.toml#L23). Not sure if it's related though. And as I said, I don't think it really makes any difference.

Zeegomo: > git blame

> ahhh that was me



probably just hacked it while trying out stuff because I didn't bother to look at what was the current version and then forgot to update it

Mr-Leshiy: Hmm, need to double check it

PR #3779, opened 2022-01-18
increase stack size on windows

Zeegomo:


.github/workflows/test.yml
```diff
@@ -126,6 +126,10 @@ jobs:
           esac
           echo "JORMUNGANDR_FEATURES=$features" >> $GITHUB_ENV
 
+          if [ '${{ matrix.profile }}' == debug ] && [ '${{ matrix.os }}' == windows-latest ]; then
+            echo "RUSTFLAGS=$RUSTFLAGS -C link-args=/STACK:2000000" >> $GITHUB_ENV
```
enzoc4, line 130: is this twice the current stack size?

enzoc4 (approved):
not sure what's this for, are we really overflowing the stack in tests?

But anyway, I think this is probably ok, it's a small number anyway.

.github/workflows/test.yml
Zeegomo: So, yes, we were overflowing the stack size on windows for debug builds in certain edge cases (https://github.com/input-output-hk/jormungandr/runs/4852642457?check_suite_focus=true).

Looking at the binary with `editbin` it looks like the current value was 100k, which is quite small, especially since the default should be 1 MB (it's 8 MB on linux for comparison). Not sure of why it was that small though

Mr-Leshiy (approved):

PR #3798, opened 2022-01-25
Port catalyst fund7 to master

dkijania:


Zeegomo:
I think it's better to port each change individually, there shouldn't be a lot of them, and if there are we should get better at opening PRs for both branches at the same time, otherwise it's hard to track why some change didn't land on master and what are the reasons for that

PR #3799, opened 2022-01-25
Remove chrono

Zeegomo:
Switch from `chrono` to `time-rs` to resolve long-standing security advisory about possible unsafe memory access.
The last library we use that depends on `chrono` is `jortestkit`, but it's only used in tests in jormungandr and a patch is available in https://github.com/input-output-hk/jortestkit/pull/108


jormungandr-lib/src/time.rs
```diff
@@ -532,11 +529,9 @@ mod test {
 
     impl Arbitrary for LocalDateTime {
         fn arbitrary<G: Gen>(g: &mut G) -> Self {
-            // no nanoseconds for the local date time as this is not displayed
-            const NSECS: u32 = 0;
-
-            let secs = i64::arbitrary(g) % 0xFF_FFFF_FFFF;
-            LocalDateTime(Local.timestamp(secs, NSECS))
+            const MAX: i64 = ::time::Date::MAX.midnight().assume_utc().unix_timestamp();
```
enzoc4, line 532: isn't this actually the beginning of the last day? (not that it matters though)

jormungandr/src/fragment/process.rs
```diff
@@ -59,11 +60,11 @@ impl Process {
...
-                let next_hour = current_hour + Duration::hours(1);
-                let sleep_duration = (next_hour - now).to_std().unwrap();
-                tokio::time::sleep(sleep_duration).await
+                let now = OffsetDateTime::now_utc();
+                // truncate date to hour so that rotation always happens at the hour mark
+                let current_hour = now.replace_time(Time::from_hms(now.hour(), 0, 0).unwrap());
+                let next_hour = current_hour + Duration::HOUR;
+                tokio::time::sleep((next_hour - now).try_into().unwrap()).await
             } else {
                 future::pending().await
```
enzoc4, line 69: unrelated, but if the decision to do something comes as an argument I wonder why not just avoid calling the function in the first place

enzoc4 (approved):

jormungandr-lib/src/time.rs
Zeegomo: yes, I was a bit lazy and didn't add the remaining hours. I won't be around to blame if something goes wrong in year 9999 anyway :)

jormungandr/src/fragment/process.rs
Zeegomo: I guess it's to avoid loads of branching points in the same place and improve readability

eugene-babichenko (approved):

PR #3809, opened 2022-01-27
update Cargo.lock for newest chain libs changes

dkijania:
Updating Cargo.lock to consume changes in tests (especially FragmentFactory method adjustmens)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -14,9 +14,9 @@ dependencies = [
 
 [[package]]
 name = "addr2line"
-version = "0.16.0"
+version = "0.17.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "3e61f2b7f93d2c7d2b08263acaa4a363b3e276806c68af6134c44f523bf1aacd"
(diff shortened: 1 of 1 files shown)


enzoc4:
> I think it was enough to update chain-libs deps only, but if you used cargo update those should only be patch-compatible changes anyway

By the way, I think it's usually possible to run `cargo update -p chain-impl-mockchain` to just update those.

Zeegomo:
> By the way, I think it's usually possible to run cargo update -p chain-impl-mockchain to just update those.

Mmm, I'm pretty sure I've done that myself multiple times. It updates all crates in the github repo at the same time, but that's usually what we want to do.

enzoc4:
> Mmm, I'm pretty sure I've done that myself multiple times. It updates all crates in the github repo at the same time, but that's usually what we want to do.

Yeah, I was just pointing it out because maybe @dkijania was not aware you could do that.

dkijania:
apologies for my misbehavior.  I knew that i should update only chain-libs i just forgot :(

Zeegomo:
> apologies for my misbehavior. I knew that i should update only chain-libs i just forgot :(

No problem at all, we just saved ourselves 100s of PRs from dependabot :)

PR #3817, opened 2022-01-27
Discard test cases with equal fragments

Zeegomo:


enzoc4 (approved):
I wouldn't expect that to happen, but this seems to make sense anyway.

dkijania (approved):
Thanks for fixing this!

PR #3834, opened 2022-02-07
fix Cargo.lock: duplicated lru

eugene-babichenko:


Cargo.lock
```diff
@@ -2602,15 +2602,6 @@ dependencies = [
  "hashbrown",
 ]
 
-[[package]]
-name = "lru"
-version = "0.7.2"
```
Mr-Leshiy, line 2607: But why we are removing the newer version `lru 0.7.2` instead of `lru 0.6.6`

eugene-babichenko: The entry for 0.7.2 was duplicated. Cargo is fine with having multiple version of the same package.

PR #3848, opened 2022-02-14
refactor RemoteJormungandrBuilder move from &mut Self -> Self

dkijania:
Small refactoring for RemoteJormungandrBuilder. Migrated for not very convenient usage of builder methods:

`fn ___(&mut self, ...) -> &mut Self `

into 

`fn __(mut self) -> Self`


Mr-Leshiy (approved):
LGTM !

PR #3862, opened 2022-02-21
add tarpaulin-coverall integration

dkijania:


.github/workflows/test.yml
```diff
@@ -263,3 +263,93 @@ jobs:
...
+          args: '--ignore-tests --out Lcov'
+          timeout: 600
+
+      - name: upload to Coveralls
+        uses: coverallsapp/github-action@master
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          path-to-lcov: './lcov.info'
+
+  test_coverage:
```
Zeegomo, line 312: looks like it's duplicated?

Zeegomo (changes requested):

.github/workflows/test.yml
dkijania: right...

Zeegomo (approved):

PR #3864, opened 2022-02-22
update chain-impl-mockchain

dkijania:
Fix compilation issues after update of chain-impl-mockchain

### DIFF
--- a/jcli/src/jcli_lib/vote/tally/decrypt_tally.rs
+++ b/jcli/src/jcli_lib/vote/tally/decrypt_tally.rs
@@ -3,14 +3,14 @@ use crate::jcli_lib::utils::{
     vote::{self, SharesError},
     OutputFormat,
 };
-use chain_vote::EncryptedTally;
+use chain_vote::tally::{batch_decrypt, EncryptedTally};
 use jormungandr_lib::{
     crypto::hash::Hash,
     interfaces::{PrivateTallyState, Tally},
 };
 use rayon::prelude::*;
 use serde::Serialize;
-use std::{convert::TryInto, num::NonZeroU64, path::PathBuf};
+use std::{convert::TryInto, path::PathBuf};
 use structopt::StructOpt;
 
 #[derive(StructOpt)]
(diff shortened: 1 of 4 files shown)


jcli/src/jcli_lib/vote/tally/decrypt_tally.rs
```diff
@@ -76,7 +79,9 @@ impl TallyVotePlanWithAllShares {
                 }
             }
         }
-        let table = chain_vote::TallyOptimizationTable::generate(max_stake);
+        let table = chain_vote::TallyOptimizationTable::generate(
+            NonZeroU64::new(max_stake).ok_or(Error::ZeroStakeWhileDecryptingTally)?,
+        );
```
cameron1024, line 84: The aim with the chain-libs change is that consumers of the API don't need to worry about setting up a table, instead there's a function `batch_decrypt(&[ValidatedTally]) -> Result<Vec<Tally>>` which sets up and reuses a table for all the tallies. It also handles the zero case, so that error variant is probably not needed.

`chain-libs/chain-impl-mockchain/src/testing/builders/vote.rs` has a `decrypt_tally` function like this that's changed pretty significantly, should be able to just copy/paste from there

eugene-babichenko: I'm going to take on that.

dkijania:
looking good. should we merge it?

testing/thor/src/wallet/committee.rs
```diff
@@ -222,7 +226,7 @@ impl PrivateVoteCommitteeDataManager {
                         &decrypt_shares,
                     )
                     .unwrap()
-                    .decrypt_tally(max_votes, &table)
```
Zeegomo, line 225: we should use the `batch_decrypt` function as used in jcli

Zeegomo (changes requested):

dkijania:
@enzoc4 @cameron1024 can you guys give a look to Giacomo changes?

cameron1024 (approved):
LGTM

enzoc4 (approved):

dkijania:
ok, merging then, to unblock rest of projects

PR #3868, opened 2022-02-24
test - vote with invalid choice would be rejected

dkijania:
Currently as there are some discussion if it should be rejected during votes casting or not, I'm testing current implementation


Zeegomo:
I think those votes should be rejected right away, and this should be enforced by the incremental tally implementation, so I guess this needs to be rebased and updated

dkijania:
@Zeegomo rebased and looks like test is passing

PR #3872, opened 2022-02-28
add testing parameter to mjolnir in order to control tool discrimination

dkijania:


testing/jormungandr-integration-tests/src/startup.rs
```diff
@@ -15,11 +15,11 @@ pub fn create_new_utxo_address() -> Wallet {
 }
 
 pub fn create_new_account_address() -> Wallet {
-    Wallet::new_account(&mut rand::rngs::OsRng)
+    Default::default()
 }
 
 pub fn create_new_delegation_address() -> Wallet {
-    let account = Wallet::new_account(&mut rand::rngs::OsRng);
+    let account: Wallet = Default::default();
```
cameron1024, line 22: minor nit: `Wallet::default()` here?

testing/mjolnir/src/mjolnir_lib/mod.rs
```diff
@@ -78,3 +79,17 @@ pub fn build_monitor(progress_bar_mode: &ProgressBarMode) -> Monitor {
         ProgressBarMode::None => Monitor::Disabled(10),
     }
 }
+
+pub trait DiscriminationExtensions {
+    fn from_bool(testing: bool) -> Self;
```
cameron1024, line 84: I'd be tempted to call this something like `from_testing_bool`, or something that describes what `true` means. If I saw `Discrimination::from_bool(true)`, I'm not sure it would be obvious whether it was prod or test 🤷

cameron1024 (approved):
Looks good, both suggestions are pretty minor

dkijania:
I will stabilize failing test in other PR

PR #3885, opened 2022-03-10
Split explorer in bin and lib

enzoc4:
This is slightly similar to #3876 (in spirit)

This changes the `explorer` crate to a library with a binary in `bin/jexplorer.rs` (yes, also changes the name of the executable, I think this way is more consistent with the rest). 

The motivation is mostly so that `jormungandr-automation` can provide access to an explorer without needing the binary compiled and available (in path), which I think is useful if using it from outside projects, like vit-testing. This way, there is no need to deal with paths to the binary, and there is no risk of using an outdated/incompatible version of the explorer in tests (doesn't matter for development, but may happen in local environment).


.github/workflows/test.yml
```diff
@@ -195,13 +195,6 @@ jobs:
           command: build
           args: --tests --all-features ${{ env.CARGO_FLAGS }}
 
-      - name: Build explorer
```
Zeegomo, line 198: The explorer should be built anyway with cargo test right?

enzoc4: The lib should, since it's going to be used.

PR #3887, opened 2022-03-15
Add jcli support for EVM mapping certificates

saibatizoku:
Need to add handling of the new Certificate/Fragment EvmMapping in jormungandr, and also update jcli to generate and sign this new type of Certificate/Fragment.

### DIFF
--- a/jormungandr-lib/src/interfaces/certificate.rs
+++ b/jormungandr-lib/src/interfaces/certificate.rs
@@ -56,6 +56,9 @@ impl SignedCertificate {
             certificate::SignedCertificate::UpdateVote(c, _) => {
                 Certificate(certificate::Certificate::UpdateVote(c))
             }
+            certificate::SignedCertificate::EvmMapping(c, _) => {
+                Certificate(certificate::Certificate::EvmMapping(c))
+            }
         }
     }
 }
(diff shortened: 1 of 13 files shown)


Mr-Leshiy:
need to update with this update https://github.com/input-output-hk/chain-libs/pull/774#pullrequestreview-911328609

jormungandr-lib/src/interfaces/certificate.rs
```diff
@@ -217,6 +224,11 @@ impl property::Serialize for SignedCertificate {
                 writer.write_all(c.serialize().as_slice())?;
                 writer.write_all(a.serialize_in(ByteBuilder::new()).finalize().as_slice())?;
             }
+            certificate::SignedCertificate::EvmMapping(c, a) => {
+                writer.write_all(&[15])?;
```
Zeegomo, line 228: why 15?

Zeegomo (commented):
I've also pushed some changes, but they are not appearing for some GitHub quirk

jormungandr-lib/src/interfaces/certificate.rs
saibatizoku: From what I can tell, the serialization of these certificates use the same numbering scheme as the tag enum for fragments in chain-libs

https://github.com/input-output-hk/chain-libs/blob/master/chain-impl-mockchain/src/fragment/mod.rs#L53-L70

Zeegomo: Do they? most of `Certificate` and `SignedCertificate` overlap anyway and those are at two different levels

Zeegomo (approved):

PR #3889, opened 2022-03-16
_ (underscore) is not needed since it is added by the prometheus lib

garbas:


garbas:
Confirming that this fixes the issue in monitoring (tested with catalyst-signoff namespace).

Zeegomo (approved):

PR #3935, opened 2022-04-13
Initialize a new RPC service

Mr-Leshiy:
Added initial RPC service initialization which will be used for the Ethereum JSON-API implementation.
Added two more dependencies:
- `jsonrpc`, [as for the server instance](https://github.com/paritytech/jsonrpc.git)

### DIFF
--- a/jormungandr/src/lib.rs
+++ b/jormungandr/src/lib.rs
@@ -618,7 +618,7 @@ fn initialize_node() -> Result<InitializedNode, start_up::Error> {
     init_os_signal_watchers(&mut services, cancellation_token.clone());
 
     let rest_context = match settings.rest.clone() {
-        Some(rest) => {
+        Some(rest_config) => {
             use tokio::sync::RwLock;
 
             let mut context = rest::Context::new();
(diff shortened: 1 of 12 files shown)


jormungandr/src/settings/start/mod.rs
```diff
@@ -153,13 +154,22 @@ impl RawSettings {
         }
     }
 
+    fn rpc_config(&self) -> Option<Rpc> {
+        let cmd_listen_opt = self.command_line.rpc_arguments.listen;
+        match cmd_listen_opt {
+            Some(cmd_listen) => Some(Rpc { listen: cmd_listen }),
+            None => None,
+        }
```
cameron1024, line 162: Small style thing: could this become: `cmd_listen_output.map(|listen| Rpc { listen })`?

jormungandr/src/lib.rs
```diff
@@ -645,6 +646,15 @@ fn initialize_node() -> Result<InitializedNode, start_up::Error> {
         None => None,
     };
 
+    match settings.rpc.clone() {
```
cameron1024, line 649: Could this be `if let Some(rpc)`?

cameron1024 (approved):

enzoc4 (commented):
I'm out of the loop on this, but did you consider something like https://github.com/paritytech/jsonrpc ? Seems to be MIT, and uses `hyper` underneath anyway. I don't see much benefit on using warp if we are not even serving it with the already existing server.

> hyper, as for the server instance

Well, not sure if that counts as adding a dependency, since warp uses hyper underneath anyway.

Mr-Leshiy:
@enzoc4 , yes I have take a look into this crate also. But initially I though how to use the same dependencies as we have currently, so that is why I have chosen `warp` json rpc extension over the original warp. At least if it will be needed to establish the usage of `promotes` or `cors` it can be easily done in the same way how it was done with the REST apis.
But if it is not needed and it is ok to use some other crates as https://github.com/paritytech/jsonrpc I can take that one.

Mr-Leshiy:
Changed dependency crate to the https://github.com/paritytech/jsonrpc.git

saibatizoku (approved):
Seems fine to me, even though I don't know the rpc crate, no cargo audit warnings + hyper under the hood is good enough for me.

Are you going to put the RPC server behind the `#[cfg(feature = "evm")]` gate?

Mr-Leshiy:
@saibatizoku , 
> Are you going to put the RPC server behind the #[cfg(feature = "evm")] gate?
generally yes, but I am considering to add just a specific roc functions under this gate, not the whole service, maybe it could be useful for the jormungandr itself

PR #3973, opened 2022-05-03
Add get chain info RPC endpoints

Mr-Leshiy:
Add a dummy implementation for the bunch of the RPC endpoints.
The main goal to provide a correct signature of the RPC endpoints and correct data types for the arguments and return value.
Functionality for the endpoints will be added later.

### DIFF
--- a/jormungandr/src/jrpc/mod.rs
+++ b/jormungandr/src/jrpc/mod.rs
@@ -23,14 +23,15 @@ pub async fn start_jrpc_server(config: Config, _context: ContextLock) {
     let mut modules = RpcModule::new(());
 
     #[cfg(feature = "evm")]
-    modules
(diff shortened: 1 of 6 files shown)


jormungandr/src/jrpc/mod.rs
```diff
@@ -22,7 +24,12 @@ pub async fn start_jrpc_server(config: Config, _context: ContextLock) {
 
     #[cfg(feature = "evm")]
     modules
-        .merge(eth_block_info::eth_get_blocks_info_module(_context))
+        .merge(eth_block_info::eth_get_blocks_info_module(_context.clone()))
+        .unwrap();
+
+    #[cfg(feature = "evm")]
+    modules
```
Zeegomo, line 31: nit: I think you can have a code block and use only one conditional compilation directive

Zeegomo (approved):

PR #3981, opened 2022-05-06
try cargo nextest in CI

cameron1024:


cameron1024:
Fails due to ports being in use, non trivial to parallelize

PR #3987, opened 2022-05-11
Add rustfmt.toml and fmt specifications

nicopado:
Added rustfmt.toml file to leverage fmt command features.
Added option "imports_granularity="Crate"" which sort use statement alphabetically and merges them when necessary.
Being this an unstable feature, it needs to run on nightly channel
`cargo +nightly fmt`


nicopado:
Regarding actions / CI:

My concern was that, given this is a breaking change for anyone using plain "cargo fmt", it would disrupt some CI step.
It looks like it doesn't, since we run the "fmt" command only with the "-- --check" option in the lint step.
This step will now fail according to what specified in the rustfmt.toml file, as can be seen [here](https://github.com/input-output-hk/jormungandr/runs/6385998075?check_suite_focus=true)

Mr-Leshiy (approved):
Seems it is ok it we can merge it

Mr-Leshiy:
@nicopado are we ready to merge it ? Or some additional actions needed ?
(Fix clips CI fist of course but it is not so complicated)

Mr-Leshiy:
@nicopado , what do you think also to add `group_imports="One"` config ? seems it will also add great structuring. https://rust-lang.github.io/rustfmt/?version=master&search=#One%5C%3A

nicopado:
@Mr-Leshiy LOVE IT!
Another good one we could add is 
`reorder_modules=true`
this one sorts the modules alphabetically.

Adding those two right now, but let's close it after this so is not parked here for eternity :)

nicopado:
image

PR #3988, opened 2022-05-12
add jcli merge voteplan results command

enzoc4:
This probably needs more tests, and an integration test, but I'm not sure if that should be done here (and I'm still not sure how).

### DIFF
--- a/jcli/src/jcli_lib/vote/tally/merge_results.rs
+++ b/jcli/src/jcli_lib/vote/tally/merge_results.rs
@@ -1,0 +1,274 @@
+use crate::jcli_lib::utils::io;
+use crate::jcli_lib::utils::OutputFormat;
+use jormungandr_lib::crypto::hash::Hash;
+use jormungandr_lib::interfaces::VotePlanId;
+use jormungandr_lib::interfaces::{PrivateTallyState, Tally, VotePlanStatus, VoteProposalStatus};
+use serde::Deserialize;
+use serde::Serialize;
+use std::collections::BTreeMap;
+use std::path::PathBuf;
+use structopt::StructOpt;
+
+#[derive(Debug, thiserror::Error)]
+pub enum Error {
+    #[error("voteplan should be already decrypted before merging")]
+    VotePlanEncrypted,
+    #[error("voteplans have different privacy type")]
+    PrivacyMismatch,
+}
+
+#[derive(StructOpt)]
+#[structopt(rename_all = "kebab-case")]
+pub struct MergeVotePlan {
+    /// The path to json-encoded list of voteplans to merge. If this parameter is not specified, it
+    /// will be read from the standard input. Voteplans must be already decrypted before merging.
+    /// Two voteplans in the list will be merged if they have ALL the same proposals according to
+    /// the proposal (external) id.
+    #[structopt(long)]
+    vote_plans: Option<PathBuf>,
+    #[structopt(flatten)]
+    output_format: OutputFormat,
+}
+
+impl MergeVotePlan {
+    pub fn exec(&self) -> Result<(), super::Error> {
+        let voteplans: Vec<VotePlanStatus> =
+            serde_json::from_reader(io::open_file_read(&self.vote_plans)?)?;
+
+        let results = merge_voteplans(voteplans)?;
+
+        let output = self
+            .output_format
+            .format_json(serde_json::to_value(results)?)?;
+        println!("{}", output);
+
+        Ok(())
+    }
+}
+
+#[derive(Serialize, Deserialize, Debug, Clone, Eq, PartialEq)]
+pub struct MergedVotePlan {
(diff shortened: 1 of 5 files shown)


jcli/src/jcli_lib/vote/tally/mod.rs
```diff
@@ -20,6 +21,11 @@ pub enum Tally {
     /// The decrypted tally data will be printed in hexadecimal encoding
     /// on standard output.
     DecryptResults(decrypt_tally::TallyVotePlanWithAllShares),
+    /// Decrypt all proposals in a vote plan.
```
Zeegomo, line 24: copy paste?

jcli/src/jcli_lib/vote/tally/merge_results.rs
```diff
@@ -0,0 +1,274 @@
...
+    let mut results = Vec::with_capacity(voteplans.len() / 2);
+
+    for (curr, vp1) in voteplans.iter().enumerate() {
+        let p1: BTreeMap<Hash, VoteProposalStatus> = vp1
+            .proposals
+            .iter()
+            .map(|vps| (vps.proposal_id, vps.clone()))
+            .collect();
+
+        if let Some(vp2) = voteplans.iter().skip(curr + 1).find(|vp2| {
```
Zeegomo, line 65: why do we merge two voteplans only? Why not merging all of those with the same proposal set?

Zeegomo (commented):
Wondering if this should happen by default when decrypting instead of requiring another command, but not really crucial (and would require some changes to the output format). 
Also, please update the changelog

jcli/src/jcli_lib/vote/tally/merge_results.rs
```diff
@@ -0,0 +1,344 @@
...
+/// indexes with the current implementation.
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Eq)]
+pub struct MergedVoteProposalStatus {
+    pub proposal_id: Hash,
+    pub options: Range<u8>,
+    pub tally: Tally,
+    pub votes_cast: usize,
+}
+
+fn merge_voteplans(mut voteplans: Vec<VotePlanStatus>) -> Result<Vec<MergedVotePlan>, Error> {
```
Zeegomo, line 67: this is more complicated than I would have expected, especially dealing with raw indexes. 
Can we partition the voteplans based on proposals right at the start?
Smth like:
```
struct WrappedVotePlan(VotePlansStatus);

impl PartialEq for WrappedVotePlanStatus {
    fn eq(&self, other: &Self) -> bool {
        let p1 = self.proposals.sort() // it's in-place in practice, but whatever...
        let p2 = other.proposals.sort()
        p1 == p2
    }
}

impl Hash for WrapperVotePlan {
    fn hash(&self, hasher: &mut Hasher) {
        hasher.input(self.proposals.sort())
    }
}

fn merge_voteplans(mut voteplans: Vec<VotePlanStatus>) -> ... {
    // HashMap<WrappedVotePlan, Vec<WrapperVotePlan>>
    let vps = voteplans.into_iter().map(WrapperVotePlan).fold(|mut acc, vp| acc.entry(vp.clone()).or_default().push(vp) );
    for (_, vps) in vps.into_iter() {
        do_merge(vps)
    }
}
```

jcli/src/jcli_lib/vote/tally/merge_results.rs
```diff
@@ -0,0 +1,344 @@
...
+        matches.reverse();
+        for i in matches {
+            voteplans.swap_remove(i);
+        }
+
+        // FIXME: I'm not actually sure if this should be the correct behavior.
+        //
+        // On one hand, a voteplan without matches can't be merged.
+        //
+        // On the other, it can be considered as is trivially merged.
```
Zeegomo, line 162: Given the current use case, I would at least emit a warning (if not an error) if a voteplan can't be merged

Zeegomo (changes requested):

jcli/src/jcli_lib/vote/tally/merge_results.rs
```diff
@@ -64,111 +64,84 @@ pub struct MergedVoteProposalStatus {
...
-
-            for (mut ps1, ps2) in p1.values_mut().zip(p2.into_iter().map(|(_, v)| v)) {
-                ps1.votes_cast += ps2.votes_cast;
-                ps1.tally = match (&ps1.tally, &ps2.tally) {
-                    (Tally::Public { result: result1 }, Tally::Public { result: result2 }) => {
-                        Tally::Public {
-                            result: result1.merge(result2),
+        group_by_proposals
+            .entry(ids)
+            .or_insert_with(Vec::new)
```
Zeegomo, line 83: very nit:  `or_default`

Zeegomo (approved):

PR #3997, opened 2022-05-24
EVM address mapping endpoints

Mr-Leshiy:
Add a new `v2` REST API endpoints for the evm purposes: `address_mapping/jormungandr_address/{id}`,  `address_mapping/evm_address/{id}`, for getting info about address mapping.


Zeegomo:
I don't think we need to create another version, or better, AFAIK we are not committing to any compatibility scheme and anyway `v1` was never declared complete anyway

saibatizoku (approved):
LGTM, I would only argue against bumping to `v2` at this point.

Mr-Leshiy:
@saibatizoku I am considered all these `v2` API to be an ethereum specific REST endpoints. Or do you think it would be better to rename it to `evm` ?

Mr-Leshiy:
@Zeegomo , pls take a look I have moved all these endpoints into the separate path `evm`. Or it would be better to extend existing `v1`/`v2` ?

Zeegomo:
I would just add them to the existing `v1` endpoint, but I'm also not a expert here. @enzoc4 maybe knows better about common convetions

enzoc4:
> I would just add them to the existing `v1` endpoint, but I'm also not a expert here. @enzoc4 maybe knows better about common convetions

I don't think I'm an expert neither, but I would also add them to `v1`. It can be `v1/evm` anyway if we want to group them.

As a side note, I would also document what's the response if the node was compiled without the `evm` feature.

PR #4002, opened 2022-05-30
Add new /v1/account-votes-all endpoint 

Zeegomo:
Remove /v1/account-votes-count endpoint and add new /v1/account-votes-all
to return the full list of proposals a user has voted for instead
of just the count.

This is going to be needed for a few rewards distribution but it's
likely going to result in a lot of data and should thus be
restricted and/or not made publicly available.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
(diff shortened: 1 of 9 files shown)


Zeegomo:
Failing fmt check is on unrelated piece of code

nicopado (approved):

saibatizoku (approved):

PR #4018, opened 2022-06-14
Mock all JSON RPC miners module

Mr-Leshiy:
This miner's module in our case, does not actually needs and it is not so vital functionality in our case so we can easily mock it.

### DIFF
--- a/jormungandr/src/jrpc/mod.rs
+++ b/jormungandr/src/jrpc/mod.rs
@@ -43,8 +43,8 @@ pub enum Error {
     Fragment(FragmentsProcessingSummary),
     #[error("Cound not decode Ethereum transaction bytes, erorr: {0}")]
     TransactionDecodedErorr(String),
-    #[error("We are not supporting mining functionality")]
-    MiningAreNotAllowed,
+    #[error("Mining is not currently supported")]
+    MiningIsNotAllowed,
 }
 
 pub async fn start_jrpc_server(config: Config, _context: ContextLock) {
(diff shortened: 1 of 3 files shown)


jormungandr/src/jrpc/mod.rs
```diff
@@ -43,6 +43,8 @@ pub enum Error {
     Fragment(FragmentsProcessingSummary),
     #[error("Cound not decode Ethereum transaction bytes, erorr: {0}")]
     TransactionDecodedErorr(String),
+    #[error("We are not supporting mining functionality")]
+    MinningAreNotAllowed,
```
Zeegomo, line 47: typo

Mr-Leshiy: fixed

jormungandr/src/jrpc/mod.rs
```diff
@@ -43,6 +43,8 @@ pub enum Error {
     Fragment(FragmentsProcessingSummary),
     #[error("Cound not decode Ethereum transaction bytes, erorr: {0}")]
     TransactionDecodedErorr(String),
+    #[error("We are not supporting mining functionality")]
+    MiningAreNotAllowed,
```
enzoc4, line 47: why not _is_ instead of _are_? 

I also think something like:

> "Mining is not currently supported" 

Would be a more standard error message, but it's mostly nitpicking.

enzoc4 (approved):

PR #4045, opened 2022-07-08
add limits to qraphql query depth and complexity

saibatizoku:
Adding global limits for queries

### DIFF
--- a/explorer/src/api/mod.rs
+++ b/explorer/src/api/mod.rs
@@ -8,9 +8,6 @@ use jormungandr_lib::interfaces::{Cors, Tls};
 use std::{net::SocketAddr, time::Duration};
 use warp::{http::Response as HttpResponse, Filter, Rejection, Reply};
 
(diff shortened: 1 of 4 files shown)


saibatizoku:
Values were chosen so that the current tests could pass.

saibatizoku:
Limit values were chosen so that current tests could pass.

explorer/src/api/mod.rs
```diff
@@ -8,6 +8,9 @@ use jormungandr_lib::interfaces::{Cors, Tls};
 use std::{net::SocketAddr, time::Duration};
 use warp::{http::Response as HttpResponse, Filter, Rejection, Reply};
 
+const QUERY_DEPTH_LIMIT: usize = 15;
```
Mr-Leshiy, line 11: Is it possible to put it somewhere in config struct or something similar, if it is presented of course 😊?

Mr-Leshiy (approved):

kukkok3 (approved):

PR #4077, opened 2022-08-23
Added new command for thor which is sending a vote plan.

dkijania:
Along with send command added two helpful utility method (decode and example), so user can easily create vote plan. For example: Let say that we want to create a vote plan from single proposal. This can be done in 3 commands:

thor utils examples proposal > proposal.json // will save example proposal.json file which hash will be used in vote plan

thor utils decode proposal // will print out hash which will should be used in vote plan

thor utils examples vote-plan > vote_plan.json // will save example vote_plan.json file which hash will be our vote plan metadata

thor send vote-plan --json vote_plan.json --pin 1111


minikin:
@dkijania That looks great. 
Can we document what you describe in the PR description somewhere?

PR #4084, opened 2022-09-12
Npg 3331 explorer testing infra

cong-or:
Initial groundwork for expanding explorer tests. PR also addresses NPG-2899.


explorer/tests/explorer_tests.rs
```diff
@@ -0,0 +1,55 @@
...
+
+        let mut config = ConfigurationBuilder::new();
+        config.with_consensus_genesis_praos_active_slot_coeff(ActiveSlotCoefficient::MAXIMUM);
+
+        let (jormungandr, _initial_stake_pools) =
+            startup::start_stake_pool(&[faucet], &[], &mut config).unwrap();
+
+        let params = ExplorerParams::new(query_complexity_limit, None, None);
+        let explorer_process = jormungandr.explorer(params);
+        let explorer = explorer_process.client();
```
cameron1024, line 50: This seems like something we might want to pull into a function. Maybe something like:
```rust
let config = ExplorerTestConfig {
  query_complexity_limit: 123,
  ..Default::default()
};
let (jormungandr, explorer) = explorer_test(config);

// actual test assertions
```

explorer/tests/explorer_tests.rs
```diff
@@ -0,0 +1,55 @@
...
+                    error.message.to_string(),
+                    "internal error (this shouldn't happen) Couldn't find block in the explorer"
+                        .to_string()
+                );
+            }
+        }
+    }
+
+    pub fn get_valid_block(explorer: &Explorer, genesis_block: Hash) {
+        if let Some(data) = explorer.block(genesis_block).unwrap().data {
```
cameron1024, line 31: Similarly, if `data` is `None` here, this function doesn't panic. Maybe worth a second `unwrap()`

explorer/tests/explorer_tests.rs
```diff
@@ -0,0 +1,55 @@
...
+    use jormungandr_lib::crypto::hash::Hash;
+    use jormungandr_lib::interfaces::ActiveSlotCoefficient;
+
+    pub fn get_invalid_block(explorer: &Explorer) {
+        let hash =
+            Hash::from_str("000000000019d6689c085ae165831e934ff763ae46a2a6c172b3f1b60a8ce26f")
+                .unwrap();
+
+        if let Some(errors) = explorer.block(hash).unwrap().errors {
+            for error in &errors {
```
cameron1024, line 20: This doesn't panic if `errors` is empty, is that expected? I wonder if we could just compare `errors` to another `Vec` to make sure they're exactly the same

explorer/src/tests.rs
```diff
@@ -0,0 +1,76 @@
...
+use jormungandr_automation::jormungandr::JormungandrProcess;
+use jormungandr_automation::jormungandr::{
+    explorer::configuration::ExplorerParams, ConfigurationBuilder, Explorer,
+};
+
+use jormungandr_integration_tests::startup;
+use jormungandr_lib::crypto::hash::Hash;
+use jormungandr_lib::interfaces::ActiveSlotCoefficient;
+
+pub struct ExplorerTestConfig {
```
cameron1024, line 12: My guess is that as this struct grows, we probably don't want to have to always specify the values. A `Default` implementation is probably handy so we can just override the values we care about for a specific test.

Not sure what a sensible default is, but I'm sure there's some other tests to copy from.

cong-or:
@cameron1024 suggestions addressed and added to the PR. 

CI build still failing due to `cargo audit` and `clippy` errors.

I have created a separate ticket `3361` for fixing cargo audit `async-graphql` . API has changed since version `3.0` - vulnerability is fixed from versions  `>=4.0.6`.. So not a trivial change.

Last `clippy` errors seem to be legacy and not related to this change. Can we ignore for now? 

Thanks :)

cameron1024:
The security issue seems like something we want to address IMO, thanks for making a ticket. Though doesn't have to be this PR.

As for clippy, the lint that is triggering is actually one that we want to allow, so I'd probably suggest adding `#![allow(clippy::clippy::derive_partial_eq_without_eq)]` to the failing crates. That one I think probably could be done here

cameron1024 (approved):
LGTM 🎉

cong-or:
> LGTM 🎉

thanks for the help :)

I will pick up ticket `3361` (_re cargo audit dependency vulnerability `(async-graphql)`_)  to get it out of the way..

PR #4098, opened 2022-09-19
Npg 1941 last block test

kukkok3:


kukkok3:
merged in catalyst-core

PR #4142, opened 2023-11-01
feat: Update the node's prometheus metrics

saibatizoku:
Updates jormungandr node to initialize mempool and peer gauges to 0.

Closes https://github.com/input-output-hk/catalyst-core/issues/600


kukkok3:
Would not be better to wait until we have a stable jor node version and merge this change into that?

cong-or:
> Would not be better to wait until we have a stable jor node version and merge this change into that?

yea but we want to compare the "new" nodes performance vs the old node. 

[Old:](golden commit: https://github.com/input-output-hk/jormungandr/commit/47e874c5792bccd2eb073ef65ea89ac5d0525c04)

New: Branch from gold. Silver?

saibatizoku:
> > Would not be better to wait until we have a stable jor node version and merge this change into that?
> 
> yea but we want to compare the "new" nodes performance vs the old node.
> 
> [Old:](golden commit: [47e874c](https://github.com/input-output-hk/jormungandr/commit/47e874c5792bccd2eb073ef65ea89ac5d0525c04))
> 
> New: Branch from gold. Silver?

I think that we can rename to whichever name is better suited to do A/B testing (e.g. branch-A, branch-B), and create a tag when we have reached a conclusion.

